PR #221, opened 2014-12-31
relaxing type constraint for THEN and THENL

jeremydaw:
to permit
val THEN           : list_tactic * tactic -> list_tactic
val THENL          : list_tactic * tactic list -> list_tactic
as well as the usual types

also EVERY_LT for list of list-tactics


xrchz:
Cool. I'd like to see `CONJ_ASMx_TAC` implemented more simply in terms of `USE_SG_THEN` as part of this pull request.

jeremydaw:
Done - also the help file for USE_SG_THEN now seems OK

xrchz:
Thanks. Only problem now is it seems to break the build on expk (https://travis-ci.org/HOL-Theorem-Prover/HOL/jobs/45588681).

jeremydaw:
Added infix THEN_LT in Tactic.sml.  Not sure why needed, since builds OK here without it.  Also apparently not needed for THEN in Tactic.sml.  Hoping this works now.

jeremydaw:
Hi Ramana,

You've been pretty prompt about looking at and considering my 
contributions - thanks for that.

Except for https://github.com/HOL-Theorem-Prover/HOL/pull/218
on that one, no action, no critical comments, no queries, either.
What about that one?

Cheers,

Jeremy Dawson

PR #222, opened 2015-01-11
infix THEN_LT - hopefully in the right places now

jeremydaw:
see comments on https://github.com/HOL-Theorem-Prover/HOL/pull/221

now infix THEN_LT in _all_ instances of src/*/Overlay.sml,
having tested the build here on _all_ the kernels


mn200:
Incidentally, you might like to invent a symbolic abbreviation for this along the lines of `>>` for `THEN`, as defined in `src/boss/lcsymtacs`.  You'd need to make the new symbol infix in `Overlay`, and then define it in `lcsymtacs`.  I reckon `>>>` would probably be good.

PR #228, opened 2015-01-28
selftests of goalStack.flatn, and some list-tactics

jeremydaw:
selftest.sml in src/proofman is new, and my consequent changes to the Holmakefile were the product of trial and error, since the Holmakefile seems to do things rather differently from the others I looked at.  Ie, if there's supposed to be a system for constructing these files, then I'm afraid I haven't adhered to it.

### DIFF
--- a/src/1/selftest.sml
+++ b/src/1/selftest.sml
@@ -507,6 +507,18 @@ val _ = List.app test badnames
 val _ = print "OK\n"
 
 val _ = let 
+  val _ = tprint "Testing structural list-tactics"
+  val tac = REPEAT DISCH_TAC THEN REPEAT CONJ_TAC THEN_LT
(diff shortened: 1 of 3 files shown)


mn200:
The issue is that your `Holmakefile` in `src/proofman` now only builds `sefltest.exe`, and not the other target in that makefile.  See the bottom of the [latest release notes](http://hol.sourceforge.net/kananaskis-10.release.html#incompatibilities) for a detailed discussion.  In this case, you should probably just add an `all` target:


    all: selftest.exe $(TARGET)
    .PHONY: all

so that both things get built.  The `$(TARGET)` in this case is a Poly/ML heap (called `hol.builder0`) that is necessary for subsequent parts of the build.

PR #233, opened 2015-02-10
Cleanups in .sig files, doc and file mode bits

ptroja:
* extra semicolons removed from `.sig` files (they do not help when exploring the source code)
* some minor documentation fixes
* executable mode bits removed from non-executable files in `examples/`

NOTE: I think that .class files should be also removed from `examples/` but I decided to leave this decision to core developers.


mn200:
I agree about those class files.  The alternative might be to entirely delete that example (we don't distribute it in formal releases, and no-one has touched it in years).

mn200:
Thanks for the work on this!

PR #236, opened 2015-02-17
HolyHammer

barakeel:


xrchz:
Can we get it pass the Travis build? It looks like it causes problems with the -expk kernel.

barakeel:
Ok I'll delete this fork and make a new one. Hopefully it will make Travis succeed, as it did for the first commits.

barakeel:
And maybe i should exclude holyhammer to the sequence if a flag (like -expk) appears.

mn200:
You can run shell-scripts easily enough by putting them into the relevant `Holmakefile`.  If it's only going to work on Unix, you might restrict the directory to be just poly/ml in the sequence file (though this is a hacky way of achieving that restriction)

barakeel:
Currently, I am using four linux commands make, ocaml and perl and sh, for creating and using the holyhammer binaries.  Would it be a good idea to install mac and windows to see if it works on them? I saw you have an installer for windows. Does it mean the sources are not usable?

mn200:
I don't think I mind the changes you have ended up making to the kernel, but I wonder if you couldn't just create another kernel, similarly to the way Ramana created the logging kernel.   People wanting to use Holyhammer would build with that kernel, generating the prover calls they need.  Then, those calls could be used in the original kernels.  I think I prefer that approach, because it reassures the really paranoid people.

mn200:
2nd question: could you adjust the build sequence so that the Holyhammer code is tested by Travis please?  When I try to build on my local machine, I get the following error from `g++`

    cc1plus: error: unrecognised command line option ‘-std=c++11’

You will have to adjust the environment that Travis gets told to use so that `ocaml` is present.  See for example the code in `developers/travis/before_install.sh`.  You should be able to use `apt-get` to install the OCaml packages rather than having to build from source.

xrchz:
I am against the option of making another kernel. I think it imposes too much overhead on the use of HolyHammer, so people will just end up not using it. The idea of HolyHammer, as I understand it, is to be useful for many people much of the time.

I think the separate kernel for OpenTheory logging was also a mistake, but at least in that case there is some merit in the argument that you only want to use it rarely (exporting proofs is less common than constructing them).

mn200:
I take Ramana's point about separate kernels being a pain.  We can leave that.  For the moment, I'd like to see the rest of it building successfully on a Travis machine.  If it's going to get used, it needs to be built as much as possible.  That means putting `src/holyhammer` into the standard build sequence.  Now, of course, there will be people who *don't* have OCaml and the rest available, so when the build gets to `src/holyhammer`, it should also be able to fail gracefully.

mn200:
Testing it on a Mac and Windows would be nice too, yes.

konrad-slind:
What is HolyHammer?         --Konrad.

barakeel:
To Konrad:

HolyHammer is a tool that given a conjecture returns a theorem or an error if the proof fails.

It contains two parts:
- First, it searches the library for relevant theorems: either theorems similar to the conjecture 
  (with some similarity measure) or theorems that are used for proving theorems 
  similar to the conjecture (that's why recording dependencies is necessary).
- Second it translate the 120 most relevant theorems together with the conjecture to an external provers such as eprover, vampire or z3. The provers then returns the theorems used in the proof about (5 to 30). These theorems can be given to Metis to reprove the conjecture.

barakeel:
To Michael: (first question)
   I agree with Ramana that few people will use it if it's not included in the default build. As I view it,  HolyHammer automatizes the search of relevant lemmas for Metis, which should be a useful feature.

barakeel:
To Michael: (building the src/holyhammer binaries)
  It seems that you don't have g++ (4.7) with support for C++11. 
  To fix this, is it a good idea to install g++-4.7 (in before_install.sh) 
  and renamed the g++ call to g++-4.7 ?
  I am sorry to ask this again but I translated the building bash script to a Makefile. 
  Now, I have a  Makefile in src/holyhammer, 
  how do I translate or include it into  a  Holmakefile?

All the best

konrad-slind:
Sounds very good. I would love to use such a tool.

Here's my question: why does it need kernel alterations?

Thanks,
Konrad.

xrchz:
The kernel alterations are for the "recording dependencies" that barakeel mentioned. HolyHammer adds extra metadata (like the current tags) to theorems to record what theorems they depend on.

konrad-slind:
OK, I see. One of the goals of the tagging system was to allow such
extralogical work (and data) to get passed along and altered as inference
proceeds.

  I was never very happy with the tag system since it is too inflexible.
Having to modify the kernel for such things as OpenTheory and HolyHammer
shows that.

I would favour a solution that could apply to all these kinds of primitive
inference
add-ons. Probably having a universal datatype for tags would work, although
there
is a requirement that tags need to be writable/readable from disk.

Konrad.

mn200:
As it happens, there is a universal datatype interface in `Theory.sml` (see the `LoadableThyData` sub-structure there), that does require readability/writability from/to disk.  Perhaps this could be used for tags like this.  At the moment, this stuff is mostly used to sets of theorems being stored (supporting the stageful rewriter, storing monotonicity theorems, storing congruence rules *etc*).

mn200:
@barakeel The format of a `Holmakefile` is a subset of normal make-files.  The big omission is that you can't use pattern rules.  Stuff like

    .o.c: ...

and

    %.o: %.c
        $(CC) -c $<

won't work.  Lots else should work.  There's documentation in the Description manual. 

If the version of g++ you need is installed by default on the Travis machines, then I'm happy to let it through.  If it isn't, then I'll take that as an indication that you're asking for the bleeding edge.

barakeel:
I don't think you can avoid modifying the datatype of theorems and inference rules. Nevertheless, I agree that this modification should be generic and usable by OpenTheory and the proof recording kernel as well as HolyHammer.

konrad-slind:
Hi Barakeel,

  Can you explain why theorems and inference rules need changing?

The ideal tagging system for an LCF style system should leave all primitive
inferences
to operate completely independently of the computation happening on tags.

Konrad.

barakeel:
Hello Konrad,

Sorry for the imprecision. I agree that the computation of tags should be
parallel to the inference rules.

What i meant is that an extension of the theorems' tag are necessary. 
This extension will have to modify the source code inside Thm.sml so the tags can be modified depending on which rules is applied. This is currently what is done for the oracles' and axioms' tags.
However, the function that is applied at different point inside the inference rules (Tag.merge) 
is specific and too restrictive.

I think a common approach could be found for all the kernels that are using the tagging method.

Thibault

konrad-slind:
Hi Thibault:

  Agreed. The merge operation on tags should be aware of the inference rule
being used,
and should be given all the arguments the inference rule is given.

Konrad.

mn200:
@barakeel will resubmit this in polished form, so closing this version of it.

PR #246, opened 2015-03-23
Minor fixes for help and description

ptroja:


mn200:
Thanks for these.  

With respect to the second commit, I think that, In theory, we should probably be encouraging tildes everywhere in `\texttt` style blocks.  If the resulting boxes get too large, and cause bad line-breaks, then the blocks should probably be displayed rather than inline within paragraphs.

PR #251, opened 2015-04-02
Trial

jeremydaw:
tactic RULE_L_ASSUM_TAC, like RULE_ASSUM_TAC but f produces list of theorems

example -  RULE_L_ASSUM_TAC CONJUNCTS  splits all conjuncts in assumptions of goal


mn200:
Could you adjust the documentation for `VAR_EQ_TAC` to say that it won't do a substitution if the variable occurs in what is to be substituted please?

mn200:
Also, in the updated proofs in `containerTheory`, there are a few places where you have excessive parentheses.  For example

    EVERY [(tac1), (tacfn arg), (foo bar baz)]

is unnecessary and could be 

    EVERY [tac1, tacfn arg, foo bar baz]

Likewise, a few places where you have parentheses in `THEN` chains (though these can sometimes be necessary, as *per* FAQ).

jeremydaw:
Note that Thm.INST_TT_HYPS is potentially security critical, 
commit 3120ad1455f4f2d6123dcbfa759f35c80bf7bc18 includes a test of it whenever 
Drule.INST_TY_TERM is called (and this was tested by fully building HOL);
commit 5b2ce89be9d0b6a3ac11b21bb2eda78dc08f28c0 then removes this test code

mn200:
Unless you think it's a *big* efficiency loss to do so, I'd strongly prefer to have `INST_TT_HYPS` outside the kernel.  It certainly seems to me that this should be possible.  (All the code that calculates the eventual list of terms corresponding to the hypotheses could be calculated outside the kernel, and the other part of the result is just the result that `INST_TY_TERM` would return (right?)

jeremydaw:
Well, a couple of possibilities:

(1) substitute the theorem (with its hypotheses in a set) and the 
hypotheses separately: this means substituting the hypothesis terms 
twice, which was the loss of efficiency in my mind that motivated doing 
it the way I did

(2) Discharge all the hypotheses, do the substitution using 
Drule.INST_TY_TERM, then undischarge them again, thus giving the 
substituted hypotheses in a list

I've just thought of this second one - maybe that would be best?

mn200:
Yes: the second option looks best.

PR #261, opened 2015-06-08
Trial

jeremydaw:
shorter proofs of theorems in sortingTheory


mn200:
Could you reformat to make the `THENL` continuations nest deeper than what comes before please?
Currently you have things like

    EVERY [tac1, tac2,
      tac3, tac4] THENL [
      tac1, 
      tac2,
      tac3
    ]

I'd prefer something like

    EVERY [tac1, tac2, 
      tac3, tac4]
    THENL [
      tac1,
      tac2,
      tac3
    ] 

I'm not sold on `EVERY` btw; you can't select a subsequence of the list and apply it easily with cut-and-paste (or editor mode help).  If you join them up with `>>` or `THEN`, any subsequence is easy to apply.  (I can see that `THEN` is tedious to have to type; but `>>` (from `lcsymtacs`) is very easy.

mn200:
Thanks for this!

PR #263, opened 2015-06-11
Trial

jeremydaw:
DROP and TAKE are total but some of the theorems don't seem to reflect this


mn200:
Thanks for these.  These constants became total with 71f60b68 (a while ago now), so it's nice to have some of the consequences properly in place.

PR #264, opened 2015-06-18
Trial

jeremydaw:
both new proofs clearer (partly due to setting out intermediate lemmas) and, even so, shorter.
Also new proof of perm_cons_append reduces build time noticeably (due to not using METIS_TAC)


mn200:
It strikes me that the `PERM_SINGLE_SWAP` constant is not brilliantly named.  I'd expect such a constant to swap just a single element of a list, rather than a segment of the list. Moreover, as it stands, it necessarily swaps to the end of one of the lists.

jeremydaw:
True enough.  

Another thing about that file, they go to quite some length to avoid defining a new constant called perm, which seem like it's the only reason why they don't just use Hol_reln

mn200:
Right.  "They" in that part of the file was me.  As it says in the comments, I didn't want to create another constant (which people would then find with `DB` tools and the like). 

Other people are responsible for other bits of the script, just as you are yourself now. :smile:

PR #279, opened 2015-08-20
Definition of coinductive relations

acondolu:
This update adds the possibility to define coinductive predicates and have induction principles automatically derived (`Hol_coreln` function). The code originates from the `IndDef` library: the corresponding proofs were "dualized" in order to derive co-induction, rules & cases theorems.


xrchz:
This is an exciting looking pull request! Can you get the Travis check to pass?

acondolu:
Thank you! Yep, this should make it.

PR #296, opened 2015-09-18
Trial

jeremydaw:
needing to do some stuff with rationals, found existing proofs seemed more complicated than they need be


xrchz:
This looks great! Document the new functions?

jeremydaw:
OK I'll do that but maybe not for a couple of weeks, since I'm 
travelling much of next week and have the following week off

Jeremy

jeremydaw:
OK, I've done doc pages for the new functions PART_MATCH_A and REWR_CONV_A
Would it make sense to do the same things for HO_PART_MATCH and HO_REWR_CONV ?

xrchz:
Yes certainly, if you're happy to.

jeremydaw:
OK, will do, but it will have to wait a week or more

xrchz:
I think you can just open another pull request later :)

PR #303, opened 2015-10-05
theorems re ABS_DIFF also proved more easily by induction on all vars

jeremydaw:


mn200:
Please don't add the monad stuff to this pull-request; it's not related.  I'd also object to some of the changes in a7d7cb8.  In particular, it adds too many constants.  For example, there's no point in defining `OPTION_UNIT` when it's equal to `SOME`.  At most, you could define `OPTION_UNIT` as an overload for `SOME` (but prefer `SOME` when printing).  I'm also dubious about defining `OPTION_EXT` as a new constant; I'd probably go for an overload there also (to `\x y. OPTION_BIND y x`)

mn200:
I merged in the non-monad commits *via* 1d9d23d72

PR #308, opened 2015-10-13
approach to state_transformer monad using Kleisli category

jeremydaw:
results related to categories/functors,
other monad axioms follow fairly easily
also errorStateMonad, indicating how to similarly
compose any monad with the state transformer monad


jeremydaw:
changed from earlier version by doing without constants OPTION_EXT and OPTION_UNIT,
comment saying UNIT of option monad is SOME

PR #309, opened 2015-10-28
Monad

jeremydaw:
can do definition (rather than specification) of LUNFOLD using option monad functions; 
define LTL_HD as it's "inverse" to LUNFOLD

### DIFF
--- a/src/llist/llistScript.sml
+++ b/src/llist/llistScript.sml
@@ -184,7 +184,7 @@ val llist_CASES = store_thm(
     SRW_TAC [][llist_repabs']
   ]);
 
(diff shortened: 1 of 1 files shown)


xrchz:
This is pretty cool! Can you get it to work on -expk?

jeremydaw:
Well, I think I've got a change to make it work in both.

It's a pain that both kernels behave differently, I wonder if it is easy 
to fix this.

xrchz:
I agree that the kernels ought to behave the same. While they don't, though, the instruction to the user is to not depend on generated names (e.g., rename variables first if you have to refer to them).

jeremydaw:
Have renamed variables, in essence, using X_GEN_TAC, which I didn't find earlier.  (A much better solution, I agree).

PR #317, opened 2015-11-17
Trial

jeremydaw:
modified my original implementation of irule to behave similarly to Michael's: this is a bit complicated and it may be simpler to stick with Michael's approach.  

But I suggest HYP_CONV_RULE and UNDISCH_SPLIT are likely to be worth having regardless.


jeremydaw:
Changed this pull request to include HYP_CONV_RULE, UNDISCH_SPLIT, omitting the stuff relating to irule (reflecting email exchange with Michael - his implementation of irule seems more natural for the changed functionality)

PR #334, opened 2016-03-13
Make COND_CASES_TAC handle repeated use of guard properly

bacam:
If a conditional's guard appears in one of the arms then `COND_CASES_TAC` fails to substitute the conditional:

    > COND_CASES_TAC ([], ``if x then f x else z``);
    <<HOL message: inventing new type variable names: 'a>>
    val it =
       ([([``x``],
          ``if T then f T else z``),
         ([``¬x``],
          ``if F then f F else z``)],
        fn): goal list * validation

This fixes it so that you end up with `f T` and `z`, respectively.  I also tried sticking strictly to the documentation (which would produce `f x`), but that breaks existing proofs.

### DIFF
--- a/help/Docfiles/Tactic.COND_CASES_TAC.doc
+++ b/help/Docfiles/Tactic.COND_CASES_TAC.doc
@@ -14,8 +14,8 @@ i.e. a sub-term of the form {p=>u|v}, choosing one by its own criteria if there
 is more than one. It then induces a case split over {p} as follows:
 {
                              A ?- t
-    =======================================================  COND_CASES_TAC
-     A u {p} ?- t[u/(p=>u|v)]   A u {~p} ?- t[v/(p=>u|v)]]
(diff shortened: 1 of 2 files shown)


xrchz:
Are the documentation and behaviour in sync after these commits? Could we fix the documentation if not? Fixing the behaviour and the resulting broken proofs is also an option in my opinion. Which behaviour is better?

bacam:
Yes, it brings them into sync.  In fact, I probably ought to have put them in separate commits because the documentation change reflects existing behaviour.  I think this is the best behaviour because it's closer to other cases tactics, which replace the term throughout the goal.

xrchz:
I'm happy to merge once the Travis check succeeds.

mn200:
The failure on Travis's side is due to the git version of Poly failing to build.  Manifestly not our problem.  I'll merge the pull request.  Thanks Brian!

PR #344, opened 2016-05-17
Fix the ConseqConv documentation so that it matches the actual types

Armael:
It might also be good to update the documentation of `ConseqConv.EXT_CONSEQ_REWRITE_CONV` to explain what is the `thm list` that is passed as an argument to the "conversions" (I do not know the answer).

### DIFF
--- a/help/Docfiles/ConseqConv.EXT_CONSEQ_REWRITE_CONV.doc
+++ b/help/Docfiles/ConseqConv.EXT_CONSEQ_REWRITE_CONV.doc
@@ -1,7 +1,7 @@
 \DOC
 
 \BLTYPE
-EXT_CONSEQ_REWRITE_CONV : conv list -> thm list ->
+EXT_CONSEQ_REWRITE_CONV : (thm list -> conv) list -> thm list ->
                           (thm list * thm list * thm list) ->
                           directed_conseq_conv
 \ELTYPE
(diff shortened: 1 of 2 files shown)


xrchz:
@thtuerk might know the answer to your question about `EXT_CONSEQ_REWRITE_CONV`.

thtuerk:
This theorem list is the context that can be used. Since `EXT_CONSEQ_REWRITE_CONV`, traverses a term, at sub-positions knowledge about the context might have accumulated. `EXT_CONSEQ_REWRITE_CONV` exploits knowledge about the implication operator, other functions use conjunctions and disjunctions as well.

@Armael Thanks for updating the documentation. What do you prefer? Should I update the documentation according to what you noticed is missing or should I do it?

Armael:
@thtuerk well, if you don't mind updating the documentation that would be teh best, I think :-).

Btw, slightly related question: I noticed that `CONSEQ_TOP_REWRITE_CONV` has a [documentation page](https://hol-theorem-prover.org/kananaskis-10-helpdocs/help/Docfiles/HTML/ConseqConv.CONSEQ_TOP_REWRITE_CONV.html) but is not exposed in the API (it seems that commit fa0f4beb594734 removes it). It looked like an useful function, plus documentation of `CONSEQ_REWRITE_CONV` explains the tactic refering to `CONSEQ_TOP_REWRITE_CONV`.  
Is there now another function that implements the same features? Should the documentation be updated?

thtuerk:
@Armael no, there is no new feature. I will update the documentation. However, since I'm quite busy with other stuff, it might take till the weekend, till I find the time. I have not looked at the consequence conversions for years. So, I need to refresh my mind a bit and see what changed in HOL meanwhile anyhow. By the way, what are you using it for. I'm just curious. P.S. Perhaps we should continue this chat via private email. My email address is in my profile.

PR #373, opened 2016-11-18
tweak case expression pretty printing

thtuerk:
- when PMATCH-case expression parsing is enabled, print tree-based case- expressions as "dtcase"
- if more than the variables used in the patter are bound, print all variables
- add parenthesis around PMATCH case expressions if necessary

I finally found the time to look at the pattern matches again. I ran into a few issues with pretty-printing / parsing. Since Michael Norrish is the expert for this code and since I also modify term_pp, I did not want to push directly to master. Thus, I create a pull request and would appreciate if anyone interested, esp. Michael could have a look.

@mn200 Could you please have a look, whether my changes make sense and either merge the pull request, tell me what to change or (if it is simpler) change it yourself?

### DIFF
--- a/src/parse/term_pp.sml
+++ b/src/parse/term_pp.sml
@@ -79,7 +79,13 @@ fun convert_case tm =
       end
 
 val prettyprint_cases = ref true;
+val prettyprint_cases_dt = ref false;
 val _ = register_btrace ("pp_cases", prettyprint_cases)
+val _ = register_btrace ("pp_cases_dt", prettyprint_cases_dt)
+
+fun prettyprint_cases_name () = if !prettyprint_cases_dt then "dtcase" else "case";
+
+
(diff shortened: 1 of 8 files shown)


mn200:
The general gist of the change seems fine, though I think you should introduce regression tests for the “case as `dtcase`” behaviour.  The regression failures can't be ignored also.  I guess they have to do with the logic around 

> if more than the variables used in the patter are bound, print all variables

but printing `case n of 0 => 3 | ...` as `case n of () .| 0 => 3 | ...` is not good.

thtuerk:
:-( Yes, I was not careful with some special cases when fixing the pretty printing bug. This should hopefully do it.

mn200:
Very nice!  Thanks for the work on this.

PR #374, opened 2016-11-19
Any multiword

xrchz:
Multiword arithmetic theory that works over 'a word (generalised from one for word64).


xrchz:
@myreen does this need anything more in order to be merged?

myreen:
I'll close it when I've finished editing it. There's a number of things that needs to be done on it still, e.g. the stack needs to be removed, and a clock needs to be added to each recursive function.

Issue #345, opened 2016-05-21
Provide easy way to include case-equality theorems


mn200:
There are “case equality” theorems for all algebraic types. For example

```
(option_CASE opt n f = v) <=> (opt = NONE /\ n = v \/ ?u. opt = SOME u /\ v = f u)
```

There should be an easy way to include all of these in a simplifier invocation, perhaps like the way abbreviations and AC-theorems are included, such as 

```
simp[use_case_eqs]
```

where `use_case_eqs` is a theorem value with a special form that prompts the simplifier to do the right thing. 
##

thtuerk:
I hope full request #380 addresses this issue.

PR #380, opened 2016-11-30
Provide easy way to include case-equality theorems / issue #345

thtuerk:
Issue #345 reminded me of some work I had done in 2013 (see commit 3ecc5da6a9385cf854831db64a3999880c63bc6c). The library `DatatypeSimps` provides theorems and simpsets similar to the ones requested by issue #345. However, it did not implement the ones needed for #345. This pull requests extends `DatatypeSimps` accordingly.

This pull request adds `DatatypeSimps.elim_case_eq_or_stateful_ss` and `DatatypeSimps.elim_case_eq_and_stateful_ss` as well as several similar spirited functions. When eliminating case splits, one can introduce disjunctions or conjunctions

````
> SIMP_CONV (std_ss ++ DatatypeSimps.elim_case_eq_or_stateful_ss ()) [] 
     ``(option_CASE opt n f = v)``

val it =
   |- (option_CASE opt n f = v) <=>
   (opt = NONE) /\ (n = v) \/ ?x. (opt = SOME x) /\ (f x = v):
   thm
````

````
> SIMP_CONV (std_ss ++ DatatypeSimps.elim_case_eq_and_stateful_ss ()) [] 
     ``(option_CASE opt n f = v)``

val it =
   |- (option_CASE opt n f = v) <=>
   ((opt = NONE) ==> (n = v)) /\ !x. (opt = SOME x) ==> (f x = v):
   thm
````

I hope this addresses issue #345 .

### DIFF
--- a/src/datatype/DatatypeSimps.sig
+++ b/src/datatype/DatatypeSimps.sig
@@ -55,6 +55,33 @@ val mk_type_exists_thm_tyinfo : tyinfo -> thm
 val mk_type_forall_thm_tyinfo : tyinfo -> thm
 val mk_type_quant_thms_tyinfo : tyinfo -> thm * thm
 
+
+(* mk_case_pred_elim_thm_tyinfo_or and
+   mk_case_pred_elim_thm_tyinfo_and generate an elimination theorem
+   for case-splits that are applied to a predicate.  There is a choice
+   to expand to either a disjunction or a conjunction. The most common
+   case is an equality used as this predicate. Therefore, the
+   specialised versions mk_case_eq_elim_thm_tyinfo_or and
+   mk_case_eq_elim_thm_tyinfo_and are provided as well.
+
+   |- !P M r1 r2. (P (case M of C1 => r1 | C2 x0 => r2 x0)) <=>
+      ((M = C1) ==> P r1) /\ (!x0. (M = C2 x0) ==> P (r2 x0))
+
+   |- !P M r1 r2. (P (case M of C1 => r1 | C2 x0 => r2 x0)) <=>
+      ((M = C1) /\ P r1) \/ (?x0. (M = C2 x0) /\ P (r2 x0))
+
+   |- !r M r1 r2. (case M of C1 => r1 | C2 x0 => r2 x0) = r <=>
+      ((M = C1) ==> r1 = r) /\ (!x0. (M = C2 x0) ==> (r2 x0 = r))
+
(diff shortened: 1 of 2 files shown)


mn200:
A few things:
- it's not efficient: every time you use the invocation in your example you recompute all such theorems for all datatypes in the `TypeBase`
- it's not “easy” yet: there are many characters to type in the example invocation. As *per* the request, I'd like something as easy as `simp[alleqsplits]`, and perhaps something that lets a *per*-type specification be made: `simp[eqsplit “:'a option”]`
- it should reuse the existing code for doing much of this stuff in `Prim_rec` (which is where this sort of code has been historically). In particular, there is code there for proving the “or” theorems already.

thtuerk:
I did not realise the overlap with `Prim_rec`. This needs checking. For me it seems however that you duplicated some functionality already present in `DatatypeSimps` in `Prim_rec`. All this seems comparably new anyhow (2014 and 2013).

About efficiency: it depends how you use it. If you use it a lot like in this toy example, it is not very efficient. If you precompute it, as I would do in such cases, it is efficient. Alternatives to ensure that it is efficient, if used very naively are much more work, obfuscate what is going on and are in my opinion not worth it. 

About ease: I believe it is easy. I use a simpsets instead of a special form of theorem and modifying the simplifier to recognize this special theorem. It's a bit more work typing it, but much easier to implement, less error prone and easier for users to understand what is going on. So, I believe my approach is easier, but requires perhaps to type a few more chars. The per type versions are implemented by the way (`elim_case_eq_and_ss` and ` elim_case_eq_or_ss`)

Anyhow, I just wanted to quickly add this functionality to an existing, related lib in the style of this lib. Since you seem to want something completely different, I will stop working on it now and let you sort it out in good time.

mn200:
I think that making it properly efficient would involve a reference cell that was updated when the TypeBase updates.  This is by analogy with how the value behind `srw_ss` updates. Another approach would be to simply prove and store the relevant theorems in the TypeBase from the start. (I'm tempted by this.)

I don't want to blame anyone for the `Prim_rec` situation.  Apart from the newer case_eq theorems that we're talking about, the entry-points in that module have been there for an extremely long time.

konrad-slind:
The TypeBase was conceived just for this: to provide a central database of
datatype theorems.
Prove the theorems once, when the type is declared, and stuff them in
there. I can help if needed.
TypeBasePure is the structure where all that functionality is implemented.

Having even more _ss thingies to remember is not good user interface
design, so I would
prefer something along Michael's suggestion of having a token like
"eq_splits" in the list
of rewrites.

Konrad.

PR #388, opened 2017-01-02
fixed parse_Z3_version() for 64-bit Z3

binghe:
Z3 4.x still works with HolSmtLib. However, 64-bit Z3 now has version strings like "Z3 version 4.5.0 - 64 bit". To correctly get the part "4.5.0" from both 32-bit and 64-bit Z3, I think we parse_Z3_version() should always get the 3rd space-separated token (instead of the last one).


mn200:
Thanks for the patch!

PR #389, opened 2017-01-05
Isabelle

jeremydaw:
for the purpose of writing code to get full unification (of terms and types), 
I want access to the internal structure of terms (involving de bruijn bound vars).
So this gives access to terms and types in the "original" (before Overlay.sml)
format.  This should be a safe change to the correctness-critical Term/Type/Thm.sml
because it doesn't allow recreating standard terms (etc) (type Term.term)  from terms 
(etc) of type KernelTypes.term (which my code will have produced).

the second kernel change is variant_ty - which just copies code in variant,
just uses mk_vartype for a chosen argument


mn200:
The problem with this idea is that the internal representation is allowed to vary across kernels and does (see the experimental kernel).  Exposing this function ties the code you write on top of it to a particular kernel, which seems like a bad idea (and no doubt explains your regressions).

I suggest you write unification against the public API (actually this has been done already for terms, see `src/simp/src/Unify` for term unification). If you think there is strong evidence that this isn't efficient enough, then I wouldn't be too set against the idea of providing unification in the kernel API.

PR #390, opened 2017-01-16
Fixed building of fmap_encodeScript and signedintScript in HOL-ACL2 bridge

binghe:
I tried to build the HOL-ACL2 bridge, it doesn't work in latest HOL4 because the theorem IN_LIST_TO_SET (listTheory) and definition LSB_def (bitTheory) are not defined any more. However, removing these non-existing things results into successful building of HOL-ACL2 bridge (at least the ML part).

In Holmake, the option "--no_holmakefile" should be "--no_hmakefile". The name must have changed in recent versions of HOL4.

What's left unsolved is the following definition in "examples/acl2/ml/testFunctionsScript.sml":

val SPLIT_def = tDefine "SPLIT"
    `(split1 [] (X,Y) = (X,Y)) /\
     (split1 (x::ys) (X,Y) = split2 ys (x::X,Y)) /\
     (split2 [] (X,Y) = (X,Y)) /\
     (split2 (x::ys) (X,Y) = split1 ys (X,x::Y))`
    (WF_REL_TAC `measure (sum_case (LENGTH o FST) (LENGTH o FST))` THEN
     RW_TAC arith_ss [listTheory.LENGTH]);

I think "sum_case" (sumTheory) has been removed after Kananaskis 8. But I don't know how to fix above definition ...


mn200:
The best way to fix the sum case issue would be to write a lambda expression as the argument to `measure` and to use the `case ... of ...` syntax.  Something like: 

```
(\s. case s of INL s => LENGTH (FST s) | INR s => LENGTH (FST s))
```

That way you don't need to worry about what the underlying constant is called.

At a higher level, if you want to use this code, and to be sure that it stays maintained, can you please adjust the build sequence file `tools/sequences/final-examples`, and make sure that the directories you put there really do build.

mn200:
Also, please separate out the fix to `HINT_EXISTS_TAC.doc`.

binghe:
Hi, could you please explain the leading "!"s in sequence files? Should I also put "!" (or "!!", "!!!") before the directory name "examples/acl2/ml"?

binghe:
Not good ... this time Holmake command hangs and I see two "poly" processes consuming 100% CPU:

binghe@binghe-mac5:~/ML/HOL/examples/acl2/ml$ Holmake 
*** Switching to execute /Users/binghe/ML/HOL/bin/Holmake
*** (As we are in/under its HOLDIR)
defaxiomsTheory                                                              OK
defaxiomsTheory                                                              OK
defaxiomsTheory                                                              OK
.. defaxiomsTh.....                                                    OK !!! 

It's been 10 minutes like this.

binghe:
Hi again, please hold for this PR (or should I just cancel it?).  I found that, actually fmap_encodeTheory and signedintTheory can't finish building - automatic proof searching seems cannot terminate. (that's why Holmake hangs)  I'll figure out the broken theorems soon.

binghe:
More work needed, let me close it for now.

mn200:
Thanks for looking into this.  The leading exclamation marks are the self-test level you want the directory to build at.  Use one for things that build quickly, and three for things that take multiple hours.

PR #394, opened 2017-01-20
holmode: add support for hol.bare

thtuerk:
I got tired of having to set the HOL executable to hol.bare manually in
the emacs HOL mode when working on theories that are build before
bossLib. This commit changes holmode to add an option for appending
".bare" to the hol-executable to the HOL-process menu.

Is this OK for other users or too much clutter?


mn200:
This looks good to me.  Thanks for the update!

PR #395, opened 2017-01-23
patternMatchesLib: add chapter to Description

thtuerk:
This commit adds a chapter about the pattern matches lib to the HOL Description.

It would be great if someone (esp. @mn200) could check, whether this addition is acceptable.


mn200:
Thanks for your work on this!

PR #398, opened 2017-02-03
Location information in the PEG parser

hferee:


xrchz:
A comment: opening `lcsymtacs` should not be necessary (all the tactics that were re-exported after rename by `lcsymtacs` have been moved back to their original libraries with the new names).

Apart from that, and if you can get the Travis check to pass, this look good to merge.

xrchz:
(oh I also saw there's a few stray comment like `(* HERE *)` and `TODO: ...` that you might want to fix)

hferee:
lcsymtacs was in fact already opened, I just moved the line up with bothering if it was useful or not.

Sorry for the unclean TODOs, but I reckoned it was ok since there was already some in other files. I'll fix them later, I was hoping to get a quick merge so that Magnus can start using it.
But this is not going to be the case: there are a lot of theorems to fix in simpleSexpParse, some of which may be hard to reformulate so that they still hold.

Issue #341, opened 2016-05-02
Make Unwind work with some


mn200:
There's a theorem roughly like

   (some x. (x = v) /\ P(x)) = if P(v) then SOME v else NONE

that should be used in the simplifier's unwind technology.

thtuerk:
I believe the new simple quantifier heuristics, in particular `SIMPLE_SOME_INSTANTIATE_CONV` implement this.

PR #399, opened 2017-02-06
implement simple quantifier heuristics

thtuerk:
The quantifier heuristics library is powerful, but often slow. The unwind conversions are fast, but have serious restrictions. This commits adds simple quantifier heuristics. They are based on the techniques
developed for quantifier heuristics library, but only search gap guesses that don't contain free variables. Some very limited experiments suggest that it is similarly fast as unwind. It can handle all the cases unwind can handle. However, it supports also a much richer syntax. Moreover, it can instantiate in addition to universal and existential quantifiers also unique existential quantifiers, eliminate select and some. Moreover, it knows about pairs and lists.

Perhaps this is useful for #341. This pull request also adds a new simpset fragment `SQI_ss` to bossLib.

I'm a bit uncertain how to do this. I need support for instantiating quantifiers fast while using information about pairs for the PMATCH pattern matching automation. I was looking into adding this on top of existing unwind implementations. However, `simpLib/Unwind` is in the build sequence before pairs. Moreover, I found it fiddly to adapt and it results in a less powerful tool than the much easier to implement it based on quantifier heuristics. However, this now adds some kind of duplication, since the resulting tool is rather similar to unwind. 

@mn200 and whoever else is interested in this: do you think this duplication is an issue?

### DIFF
--- a/Manual/Description/QuantHeuristics.tex
+++ b/Manual/Description/QuantHeuristics.tex
@@ -239,6 +239,16 @@ problematic. Therefore, the simplifier and Metis are used to prove
 the validity of the explicitly given instantiations. This succeeds
 only for simple examples.
 
+\subsection{Simple Quantifier Heuristics}
+
+The full quantifier heuristics described above are powerful and very flexible.
+However, they are sometimes slow.
+The unwind library\footnote{see \texttt{src/simp/src/Unwind.sml}} on the other hand is limited, but fast.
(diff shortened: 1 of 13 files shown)


xrchz:
I don't think duplication is too bad, however, maybe it would be nice to replace the duplicated unwind functionality at some point. I'll leave it to @mn200 to comment further.

PR #402, opened 2017-02-10
Make parser prefer charset64 alphabets in regexp example

IlmariReissumies:
...or else some proofs break in regexp_compilerScript.sml


xrchz:
I'm surprised this is necessary - where is the dependence on charsetTheory picked up from? Does it work without these changes if you cleanAll first?

konrad-slind:
Yes, I was wondering the same thing. When I converted to word64x4 there was
no need to change
anything. Curious.

Konrad.

IlmariReissumies:
Honestly it didn't make much sense to me either. I'll try with cleanAll per Ramana's suggestion, and if that helps I'll withdraw thte pull request...

IlmariReissumies:
Thanks guys! It didn't cross my mind that problem might be cruft in .HOLMK...

Issue #159, opened 2014-04-04
Emacs mode: function to de-Unicode-ify buffer


mn200:
##

mn200:
Done in 97e64bf57546

PR #404, opened 2017-02-16
issue #159 : provide support for de-Unicode-ify in emacs mode

thtuerk:
I could not sleep. So I spend some time to compile a list of the common unicode symbols used by HOL and added a replacement function to the emacs mode. I hope this adresses issue #159. 
@mn200 please check whether this is what you had in mind 

There is a new command `hol-remove-unicode` now that replaces commonly
used unicode symbols with their ASCII counterpart. It can be found in
the menu as "HOL / Misc / Replace common HOL unicode symbols". If a
region is selected, it operates only on this region, otherwise on the
whole buffer.

Currently the most common unicode symbols are replaced. Probably a few
more need adding in time. This should be straightforward by updating the
list "hol-unicode-replacements".

### DIFF
--- a/tools/hol-mode.src
+++ b/tools/hol-mode.src
@@ -1458,6 +1458,98 @@ second is to show the new term.
    (hol-remove-tailing-empty-lines)
    (message "Buffer cleaned up!"))
 
+;;replace common unicode chars with ascii version
+(setq hol-unicode-replacements '(
+  (nil "¬" "~")
+  (nil "∧" "/\\\\")
+  (nil "∨" "\\\\/")
+  (nil "⇒" "==>")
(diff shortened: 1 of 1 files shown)


mn200:
Looks great to me.  I never got 'round to doing this because I was wary of the repeated passes over the buffer finding the replacement strings. I felt as if there should be a way to make one pass.  Equally, I think it's clear that we shouldn't let the perfect be the enemy of the good, so thank you very much for actually going ahead and putting in the necessary work!

thtuerk:
Thanks. It's far from perfect but hopefully handy as it is. And it is a starting point for improvement after we figured out which ones are really needed.

PR #411, opened 2017-04-08
Fixed misleading messages and provided a README file

vo1stv:
As per request of @mn200


vo1stv:
Actually, please defer this pull until the next commit.
Going to add routines to the effect that "not proving is not disproving".
Investigating whether prove(~th,tac) is sufficient to disprove.

vo1stv:
This is going to have to wait until later this month.  Happy Easter!

Issue #160, opened 2014-04-07
Consider implicational rewriting tactics


xrchz:
https://github.com/aravantv/HOL4-impconv

Also, the current implementation of `HINT_EXISTS_TAC` in HOL4 is somewhat broken, and I think fixed up in the above.
##

PR #412, opened 2017-04-10
BackchainingLib

thtuerk:
I started working for the PROSPER project now. There one commonly has to apply lemmata of the form 

``! x y. R x y ==> !x' y'. C x y x' y' ==> R x' y'``

to prove that `R a b` holds for some `a`, `b`, while a term `R a' b'` is explicitly present in the list of assumptions. This is a bit fiddly with the HOL tools I know (see longer example below). Therefore I implemented a small library for handling this. Essentially it combines matching against assumptions (as done by e.g. IMP_RES_TAC) with conditional rewriting done by the simplifier. It is a tiny lib but handy for our purposes (detailed description below).

I would like to add it somewhere to HOL, but am uncertain where to put it. Since it is closely linked with the simplifier, since  and since I can imagine it becoming even more closely linked if it is used a lot, I added the new lib to the simplifier directory in this pull request. I'm very open to other suggestions, though.

### DIFF
--- a/src/simp/src/BackchainingLib.sig
+++ b/src/simp/src/BackchainingLib.sig
@@ -1,0 +1,60 @@
+signature BackchainingLib =
+sig
+ include Abbrev
+
+  (* Applying theorems of the form
+
+     !x y. R x y ==> !x' y'. some complicated condition ==> R x' y'
+
+     is a bit tricky. However, such theorems occurs frequently when e.g. R is a
+     relation between states and one wants to show that after certain steps
+     the resulting states are still in this relation.
+
+     Conditional rewriting (as implemented by the simplifier) requires
+     finding instances for the variables "x" and "y". Such instances
+     usually cannot be found by the simplifier, even if a term of the
+     form "R a b" is present in the list of assumptions. Applying
+     MATCH_MP_TAC is of course possible, but often tedious. Similarly,
+     consequence conversions work, but easily loop or might - in the
+     presence of multiple rules - pick up the wrong rule. IMP_RES_TAC
+     and similar tools, easily produce too many instances and clutter
+     the goal-state.
+
+     This library tries to address that. Let "bc_thms" be a
+     list of theorems of the form
+     (!x1 ... xn y1 ... yn. P x1 ... xn ==> Q x1 ... xn y1 ... yn).
+     Then "BACKCHAIN_THEN (fn thms => tac thms) bc_thms" is a tactic
+     that tries to instantiate the variables x1 ... xn and discharge
+     the precondition P. The resulting theorems are handed to a
+     function to produce a tactic. This is done via matching against
+     assumptions. BACKCHAIN_THEN is similar to IMP_RES_THEN. However,
+     IMP_RES_THEN normalises P and Q, introduces many implications and
+     resorts them. It applies a tactic to all theorems resulting from
+     discharging one of these preconditions. In contrast
+     BACKCHAIN_THEN only normalises the precondition P and returns all
+     theorems resulting from discharging all preconditions produced by P.
+     Moreover, BACKCHAIN_THEN applies a tactic to all resulting theorems instead
+     of calling a theorem-tactic for each of them.
+
+     BACKCHAIN_THEN is the basis of this library. By providing multiple
+     instantiations for the tactic argument, the other tactics are derived.
+  *)
+  val BACKCHAIN_THEN : (thm list -> tactic) -> thm list -> tactic
+
+
+  (* apply a theorem tactic to every resulting theorem *)
+  val BACKCHAIN_EVERY_THEN : (thm -> tactic) -> thm list -> tactic
+
+  (* Assume each resulting theorem or add it as a precondition to the goal. *)
+  val BACKCHAIN_ASSUME_TAC : thm list -> tactic
+  val BACKCHAIN_MP_TAC : thm list -> tactic
+
+  (* Use the resulting theorems for backchaining via consequence conversions *)
+  val BACKCHAIN_TAC : thm list -> tactic
+
+  (* Use resulting theorems together with the simplifier *)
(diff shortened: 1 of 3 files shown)


xrchz:
Incidentally, have you seen the `mp_then` library? It might have some similar functionality. But probably not as integrated with the simplifier as yours.

Could you add some `.doc` files for this? I'd be happy to merge it if so. (I defer to @mn200 either way.)

thtuerk:
## Usage Example

Let `R` be simply the greater relation for natural numbers. Then we can easily prove some lemma like the one above.

````
val R_def = Define `R x y = (x:num) > y`

val R1 = prove (``R x y ==> x <= x' /\ y' <= y ==> R x' y'``,
  SIMP_TAC arith_ss [R_def])
````

If the then want to prove a simple consequence of `R1` like ``R x y ==> x < x' /\ x <= 10 /\ y' <= y ==> R x' y'``, it is already a bit fiddly.

#### Simplifier
The simplifier can't solve this goal easily. While applying conditional rewriting instances for `x` and `y` in the precondition need to be found. Even when adding `SATISFY_ss` or `QI_ss` these can't be found automatically.

#### IMP_RES_TAC
`IMP_RES_TAC R1` does too much. It only adds the assumption `R 10 y'`, which is useless for us. First, as we want it instantiates `x` and `y` as we want. However, then it instantiates `x'` with 10 while getting rid of `x <= x'`. Therefore, the result is useless for us. In other situations, `IMP_RES_TAC` might produce a lot of additional assumptions.

#### MATCH_MP_TAC / CONSEQ_REWRITE_TAC
One can of course manually do backchaining. However, this is often painful. In our cases, a lot of such goals need solving. `HINT_EXISTS_TAC` does a good job, but if one wants to automate common proof patterns, it is easy to pick the wrong lemmata for CONSEQ_REWRITE_TAC. Moreover, if one is not careful, the consequence rewrite might either loop or not modify the parts of goal that need processing. So, while this is a feasible option, the resulting proofs are in our case lengthy and a bit fiddly.

#### METIS_TAC 
Since the precondition needs processing, using something like PROVE_TAC or METIS_TAC is only possible in trivial instances.

### New Backchaining Lib

With the new backchaining one can combine the benefits of `IMP_RES_TAC` and conditional rewriting. With methods similar to `IMP_RES_TAC` exactly the first precondition is discharged. Then conditional rewriting picks in a save way one of the resulting theorems that can solve part of the goal.

thtuerk:
@xrchz No I have not seen the `mp_then` library. I will have a look. I'm creating this pull request to have a discussion whether this is sensible or not and in case it is sensible then where to put it. So please don't merge without such a discussion. Otherwise, I could just merge it myself or not create a pull request at all. Of course I will add documentation (not just doc files), but also some selftests etc. However, before doing so, I would like to have this discussion here.

xrchz:
Got it. Another point of comparison would be "implicational rewriting" (#160 and [this paper](https://link.springer.com/chapter/10.1007/978-3-319-08970-6_4))

konrad-slind:
And Peter Homeier's "dependent rewriting", which has been integrated
somewhere in the system.

Konrad.

thtuerk:
I mentioned impliciational rewriting (as of #160) in the form of `HINT_EXISTS_TAC`. For the other tactics, I believe that the consequence conversions I mention do the same job.

@konrad-slind Thanks. I expect you mean "src/1/dep_rewrite.sml". I did not know about it. I will have a look.

xrchz:
In my understanding `HINT_EXISTS_TAC` is superseded by `part_match_exists_tac`. And both are potentially superseded by `mp_then`.

thtuerk:
I had a closer look meanwhile. As I understand it, `part_match_exists_tac` and `mp_then` take arguments in order to find instantiations for quantifiers using the assumptions. However, this is not what I want. For my purposes HINT_EXISTS_TAC is more appropriate and even this one has issues (see above). I want to write automation that is robust and can be applied without getting the proof into dead ends. Using these existential quantification heuristics, one needs to do something like `MATCH_MP_TAC` before. This is tricky (see above for MATCH_MP_TAC and CONSEQ_REWRITE_TAC), because I want a robust automation that gets many theorems and picks only ones whose precondition can be discarded. Moreover, I got the impression that for these tactics I need to encode in the tactic call on the ML side, which part of the theorem to match assumptions against. In my approach, this is encoded in the structure of the theorem, which is more flexible for my purposes. So, I don't see how these either `part_match_exists_tac` or `mp_then` can be used to tackle my problem or implement something like `BACKCHAIN_THEN`.

I had a closer look at Peter's dependent rewrites and the tactics discussed in #160. Both can be used to replace MATCH_MP_TAC or CONSEQ_REWRITE_TAC. They are more general in the sense that they provide really dependent rewriting for arbitrary equations and introduce new preconditions to be discharged. However, here I want to reason about relations and I got the impression that for this purpose the infrastructure provided by ConseqConv is more flexible. Anyhow, I don't want to use even this infrastructure because of the issue of picking the right theorem for backchaining. I believe it might for other purposes (not my current ones) be nice to combine `BACKCHAIN_THEN` with dependent rewrites, though. 

Many thanks for pointing me to these libraries. I did not know about `mp_then` and had forgotten about dependent rewrites (which I noticed I used before though). Anyhow, having looked at these libraries, I got more convinced that what I want cannot be mimicked by them easily. So I got more convinced that adding it might be worthwhile.

thtuerk:
I noticed that I can simplify my code by using IMP_CANON instead of reimplementing this functionality. Since the code is in an early stage in an unused branch, I pushed the change via a forced commit to keep history a bit cleaner.

thtuerk:
Many thanks for your comments. After having looked at the mentioned libraries, I still believe this functionality cannot be mimicked easily with the level of automation I want using the other tools.  Since noone objected about the functionality itself or the location I'm putting it in, I'm merging now. I will add more documentation over the next few days

PR #419, opened 2017-05-15
[example] a formalization of process algebra CCS (ported from HOL88)

binghe:
Hi,

This is a formalization of CCS (Milner's Calculus of Communicating Systems) by Prof. Monica Nesi during 1992-1995, ported by Chun Tian (me) from old HOL88 proof scripts with some improvements.   Original proof scripts, project report (draft) and the original paper can be found here: https://github.com/binghe/informatica-public/tree/master/CCS

Among other features, I think this work is a classical example to demonstrate how Hol_coreln can be used to define "bisimulation equivalence" - the most well-studied co-induction relation.

The code is confirmed working with all HOL kernels (stdknl, expk, otknl), also in Kananaskis-11 final release.  (It doesn't build in Kanasnaskis-10 because Hol_coreln is not available)

This is only partial of the entire work. I'm receiving more old scripts from Monica Nesi, and in the future I may request to submit more code into this directory. The final aim is to have a model checking tools running on top of HOL4.

Thanks,

Chun Tian
University of Bologna

### DIFF
--- a/examples/CCS/CCSLib.sig
+++ b/examples/CCS/CCSLib.sig
@@ -1,0 +1,47 @@
+(*
+ * Copyright 1991-1995  University of Cambridge (Author: Monica Nesi)
+ * Copyright 2016-2017  University of Bologna   (Author: Chun Tian)
+ *)
+
+signature CCSLib =
+sig
+  include Abbrev
+
+  val PAT_X_ASSUM		: term -> thm_tactic -> tactic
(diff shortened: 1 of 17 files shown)


xrchz:
Can you add this to the build sequence so that it gets tested by the regression test? Probably that means add a line like `!examples/CCS` to `tools/sequences/more-theories`.

binghe:
Done.

binghe:
Thanks!

Issue #416, opened 2017-05-08
Hol replacing a quoted, bound variable by another


rmonat:
Hi everyone,

I stumbled upon a weird behavior of HOL. I reduced my example to something smaller, that does not really make any sense now. When defining the _test_ function below, I wanted to enforce m' = m in the case of the Downcast constructor.
```sml
open HolKernel bossLib;

val _ = Datatype `exp = Const num num | Downcast num exp`;

val test_def = Define `
  test (e:exp) (map: exp -> num option) : bool =
    case e of
     | Const m n => (case map e of
                      | SOME m' => m = m'
                      | NONE => F)
     | Downcast m e1 => (case map e, map e1 of
                           | SOME m', SOME m1 => (m' = m) /\ test e1 map
                           | _, _ => F)
     | _ => F`;
```

However, when giving this to HOL, this constraint is not forced anymore, and has been replaced by m'' = m''.

```
val test_def =
   |- ∀map e.
     test e map ⇔
     case e of
       Const m n => (case map e of NONE => F | SOME m' => m = m')
     | Downcast m' e1 =>
         case (map e,map e1) of
           (NONE,v1) => F
         | (SOME m'',NONE) => F
         | (SOME m'',SOME m1) => (m'' = m'') ∧ test e1 map:
   thm
```

Have I missed something or is this a bug? I am using HOL4 Kananaskis 11.

konrad-slind:
It's a bug. Not a soundness bug, but the pattern-match expansion is
capturing a variable
it shouldn't.

Konrad.

thtuerk:
I believe pull request #426 fixes this issue

thtuerk:
closing after #426 got merged

PR #426, opened 2017-06-13
fix issue #416

thtuerk:
I debugged issue #416 and hope this will fix it.

Issue #416 describes a bug in the case expression parsing. During case-parsing two bound variables could be identified with each other while they should really be separate ones. I tracked this problem down to `Pmatch.mk_functional`. After doing the real work and building `a` and `case_tm`, variable renamings should be applied to make the variable names similar to the ones used in the input. The bug is in this renaming code (see https://github.com/HOL-Theorem-Prover/HOL/blob/dea93fe0a3b84f048a4f7e82caf2dd5a9a014fa6/src/1/Pmatch.sml#L770-L781).

`pat_match3` computes the necessary variable renaming as a substitution and applies it via `rename_case` (see https://github.com/HOL-Theorem-Prover/HOL/blob/dea93fe0a3b84f048a4f7e82caf2dd5a9a014fa6/src/1/Pmatch.sml#L774-L775). The implementation of `rename_case` strips away all outer occurrences of case-expressions applies the substitution and builds the case-expressions again with substituted variable names (see https://github.com/HOL-Theorem-Prover/HOL/blob/dea93fe0a3b84f048a4f7e82caf2dd5a9a014fa6/src/1/Pmatch.sml#L708-L719). If a clause of the input case-expression has another case-expression at top-level, then this stripping is doing too much. The substitution is applied under a lambda abstraction which has not been taken in consideration while building the substitution. This is when variable capture might occur as demonstrated by issue #416.

I was thinking about how to best fix this bug. One option is to use markers to prevent the sustitution being applied too low. However, this would be not make the intention of variable renaming clear, be rather hackish and moreover cause dependency issues (or duplicate definitions) since marker and combinator theories are later in the build-chain. Instead I implemented a comparably large change to the code, which is however cleaner in my opinion.

I looked what is really supposed to happen. The substitution computed by `pat_match3` is really a renaming of bound variables. I made this explicit by changing the return type and adding checks that fail and produce an error message if this is not guaranteed. Moreover, by changing how the auxiliary variable names are generated by `vary`, I ensure that the bound variables renamed are only occurring once, i.e. not in the input. This allows me to implement `rename_case` completely differently as a renaming of bound variables via `rename_bvar`.

The works well for my tests. However, it is a potentially troublesome change. Moreover, I tested only with the standard kernel and after my last change, I only ran selftest level 1. So, it would be good, to check the results of the automatic tests carefully and also review my changes in case I implemented some nonsense.

### DIFF
--- a/src/1/Pmatch.sml
+++ b/src/1/Pmatch.sml
@@ -507,18 +507,21 @@ fun no_repeat_vars pat =
 (*---------------------------------------------------------------------------
      Routines to repair the bound variable names found in cases
  ---------------------------------------------------------------------------*)
+fun pat_match1 fvs pat given_pat =
+ let val (sub_tm, sub_ty) = Term.match_term pat given_pat
+     val _ = if null sub_ty then () else (raise ERR "pat_match1" "no type substitution expected");
 
-fun subst_inst (term_sub,type_sub) tm =
-    Term.subst term_sub (Term.inst type_sub tm);
-
-fun pat_match1 (pat,exp) given_pat =
- let val sub = Term.match_term pat given_pat
- in (subst_inst sub pat, subst_inst sub exp);
-    sub
+     fun is_valid_bound_var v = (is_var v andalso not (List.exists (fn tm => aconv tm v) fvs))
+     val _ = if List.all (fn m => is_valid_bound_var (#residue m)) sub_tm then () else
+           (raise ERR "pat_match1" "expected a bound variable renaming");
+ in sub_tm
  end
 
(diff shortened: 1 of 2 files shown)


xrchz:
Travis suggests this works fine on `-expk`. Also, I didn't see any problems when looking at the changes. (I have not run a selftest at level 3 though.)

mn200:
I'm running a `-t2` standard kernel build now (against 775d634).  After this, I'll try experimental kernel, and then assuming that passes (which seems likely), I'll merge with master and let my regression machines try the `-t3` builds overnight.

PR #432, opened 2017-07-05
Move 40 small theorems (+ Schroeder-Bernstein theorem and FUNSET) from probability/util_prob to pred_setTheory

binghe:
I have moved some small set-theoretic theorems from probability/util_prob to pred_setTheory. I think they're very common theorems, and people who needs pred_setTheory may not know the existence of these ready-to-use theorems:

BIGINTER_SUBSET
BIGUNION_IMAGE_UNIV
BIJ_SYM
BIJ_SYM_IMP
BIJ_TRANS
DIFF_BIGINTER
DIFF_INTER
DIFF_INTER2
DISJOINT_ALT
DISJOINT_COUNT
DISJOINT_DIFF
DISJOINT_DIFFS
EQ_SUBSET_SUBSET
FINITE_REST_EQ
INFINITE_INJ
INJ_IMAGE_BIJ
IN_BIGINTER_IMAGE
IN_BIGUNION_IMAGE
IN_EQ_UNIV_IMP
SUBSET_ADD
SUBSET_INTER1
SUBSET_INTER2
SURJ_IMP_INJ

Tested HOL builds in both --stdknl and --expk modes.  I think my goal is to make "probability/util_prob" smaller and smaller, and finally remove it.

### DIFF
--- a/src/pred_set/src/pred_setScript.sml
+++ b/src/pred_set/src/pred_setScript.sml
@@ -384,6 +384,11 @@ val _ = TeX_notation {hol = "SUBSET", TeX = ("\\HOLTokenSubset{}", 1)}
 val _ = TeX_notation {hol = UChar.subset, TeX = ("\\HOLTokenSubset{}", 1)}
 val _ = ot0 "SUBSET" "subset"
 
+val SUBSET_THM = store_thm (* from util_prob *)
+  ("SUBSET_THM",
+   ``!(P : 'a -> bool) Q. P SUBSET Q ==> (!x. x IN P ==> x IN Q)``,
+    RW_TAC std_ss [SUBSET_DEF]);
+
 val SUBSET_TRANS = store_thm
     ("SUBSET_TRANS",
      (--`!(s:'a set) t u. s SUBSET t /\ t SUBSET u ==> s SUBSET u`--),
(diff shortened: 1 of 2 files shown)


binghe:
Further moved Schroeder-Bernstein theorem (and relative definitions) from util_probTheory. It's an important result about INJ / SURJ / BIJ.

util_probTheory has R-associative THEN and THENL, I have to understand the proof first before migrating the proof into pred_setTheory with modifications.

mn200:
Thanks very much for this!  I totally appreciate the work.  Incidentally, if you're looking for more results in similar vein, there are some in `examples/set-theory/hol_sets/cardinalScript.sml`, including another (now redundant) proof of S-B.

binghe:
Thanks!

P. S. I see your cardinalScript.sml is a complete theory about the cardinality of infinite sets (while pred_setTheory has only cardinality results for finite set). I'm learning the ordinalScript in that folder and using it in my own (CCS) theory.

binghe:
Hi ... I took a deeper look at cardinalScript.sml and surprisingly found that the proof of cardleq_ANTISYM actually took 143 lines (over 10% of the entire file!), while the core proof of this theorem is quite simpler in latest pred_setTheory (about 50 lines).

Do you mind if we remove the big proof in cardinalScript and use the following version instead?

(* Schroeder-Bernstein theorem *)
val cardleq_ANTISYM = store_thm (
   "cardleq_ANTISYM",
  ``∀s t. s ≼ t ∧ t ≼ s ⇒ s ≈ t``,
    REWRITE_TAC [cardleq_def, cardeq_def]
 >> REWRITE_TAC [SCHROEDER_BERNSTEIN]); (* in pred_setTheory *)

I think in this way, maybe people could focus more on the other useful stuff in cardinalTheory ... what do you think?

mn200:
Of course, that would be great.  Notice that the proof in `pred_set` does depend on the otherwise uninteresting constant `schroeder_close` and its properties so that the result there is more like 80 lines long.  Nonetheless, it‘s silly to have two proofs now that it’s available so early in the core distribution, so please do make the change you suggest.

binghe:
OK, great, I will submit another pull request for the cleanups in cardinalScript. (the other version of Schroeder-Bernstein theorem has already been removed in "examples/miller")

PR #433, opened 2017-07-10
Further cleanup of util_probTheory with set theorems moved back to pred_setTheory

binghe:
Hi,

This is a further work of PR #432:

1. In util_probTheory, there's a duplicated (simpler) definition for countable sets (countable_def), which is not the same but equivalent with the one in current pred_setTheory. I have proved the equivalence and then merged the two work together into pred_setTheory.  The alternative definition is now called COUNTABLE_ALT in pred_setTheory as a theorem, while the original alternative definition of countable sets (based on bijections), is now called COUNTABLE_ALT_BIJ in pred_setTheory. Some other beautiful theorems on countable sets were also moved:

COUNTABLE_SUBSET
COUNTABLE_ENUM

2. More theorems about bijections are moved to pred_setTheory:

BIJ_ALT
BIJ_INSERT_IMP
BIJ_FINITE_SUBSET
FINITE_BIJ
FINITE_BIJ_COUNT

3. the concept and theorems for PREIMAGE of sets, are moved from util_prob to pred_setTheory. There's no deep results, all basic definition and theorems.

Now there's no set theory theorems left in util_probTheory.  My changes didn't change the name of any existing theorem in pred_setTheory, thus I believe the potential compatibility issues (if there is) must be subtle. Now pred_setTheory looks rich: each new concept has enough supporting theorems.

### DIFF
--- a/examples/miller/formalize/extra_pred_setScript.sml
+++ b/examples/miller/formalize/extra_pred_setScript.sml
@@ -51,27 +51,6 @@ val Cond =
 (* Definitions.                                                              *)
 (* ------------------------------------------------------------------------- *)
 
-val countable_def = Define
-  `countable s = ?f. !x : 'a. x IN s ==> ?n : num. f n = x`;
-
-val pscountable_countable = store_thm(
-  "pscountable_countable",
-  ``pred_set$countable = extra_pred_set$countable``,
-  SRW_TAC [][countable_def, pred_setTheory.countable_def, FUN_EQ_THM] THEN
-  EQ_TAC THEN STRIP_TAC THEN1
-    (IMP_RES_TAC inj_surj THEN SRW_TAC [][] THEN METIS_TAC [SURJ_DEF]) THEN
-  Q.EXISTS_TAC `\x. @n. f n = x` THEN ASM_SIMP_TAC (srw_ss())[INJ_DEF] THEN
-  MAP_EVERY Q.X_GEN_TAC [`e1`, `e2`] THEN STRIP_TAC THEN
-  SELECT_ELIM_TAC THEN CONJ_TAC THEN1 METIS_TAC [] THEN
-  Q.X_GEN_TAC `u` THEN STRIP_TAC THEN
-  SELECT_ELIM_TAC THEN CONJ_TAC THEN1 METIS_TAC [] THEN
-  Q.X_GEN_TAC `v` THEN REPEAT STRIP_TAC THEN SRW_TAC [][]);
-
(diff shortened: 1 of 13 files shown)


mn200:
I think the `pred_set` changes have broken `examples/miller/formalize/extra_realScript.sml`.  Do you think you have time to look into this?  I will do so otherwise.

binghe:
Yes, I'll check and fix it today and submit the solution as additional commits here.

mn200:
That would be great - thanks.

binghe:
Fixed. This example folder seems containing an early version of probability theory and an "extra_pred_setTheory", so there're duplicated definitions with new additions in pred_setTheory.   After I removed many of these duplicated definitions, the issue is resolved.

I'll take a look at "extra_pred_setTheory" and see if there're useful things which can be moved to pred_setTheory.

binghe:
There're still small issues in "examples/miller/miller" folder, I'm fixing it!

binghe:
All fixed. Now every sub-folders under "examples/miller" builds correctly.

P. S. Actually there's already a FINITE_BIJ_COUNT in pred_setTheory, I didn't notice it. But the version in extra_pred_setTheory has different bound variables, and it turns out that at least one theorem (PROB_BERN_BIND_COUNTABLE) in "examples/miller/prob/probScript.sml" is very sensitive to the variable names in FINITE_BIJ_COUNT, so at the end I chose to keep the old FINITE_BIJ_COUNT in that extra_pred_setTheory. (the one in src/probability/util_probTheory was safely removed, though)

mn200:
Brilliant; thank you.  You’re right that the material is related historically.  In fact, I’d like to have all of the duplication removed—it’d be nice to have the Miller-Rabin algorithm verified with respect to the newer probability development in `src`. Anyway, that’s probably a deal more work.

binghe:
That's doable ... but I'm afraid that I cannot commit any determine deadline for this work.   Now it's obvious that the Miller-Rabin algorithm only depends on Elementary Probability theory (which has no links to Lebesgue integral theory), which is supposed to be well supported in the newer probability development in src.  But right now I'm focusing on Advanced Probability: I'm going to prove the Strong Law of Large Numbers in HOL4 in near future.  If you want, I'll pick up the job for maintaining the probability related stuff from now on and clean up them increasingly.

mn200:
As the original authors of the material in `src/probability` seem to have lost interest, I don't have a problem with that.  I'm sure you can maintain some level of backwards compatibility as well, so I don't think there will be any issue with you taking “ownership”.

binghe:
Great, thank you, let's see how this plan goes!

PR #434, opened 2017-07-10
Simplified proof of cardleq_ANTISYM in cardinalTheory

binghe:
After the merge of PR #432, now HOL's pred_setTheory contains a version of Schroeder-Bernstein theorem (`SCHROEDER_BERNSTEIN`) expressed in basic INJ/SURJ/BIJ languages. This obsoleted the complicated proof of  `cardleq_ANTISYM` in `examples/set-theory/hol_sets/cardinalScript.sml`.

With the author of cardinalScript (Michael Norrish) agreed, I create this pull request to replace the proof of `cardleq_ANTISYM` with a trivial version based on `SCHROEDER_BERNSTEIN` in pred_setTheory. The goal is to make cardinalTheory more clear for learners and users, and remove duplicated theorems in HOL's code base as much as possible ...

Another small theorem (`INJ_BIJ_SUBSET`) is also moved to pred_setTheory as it's a common result and depends only on theorems in pred_setTheory, after removing unicode chars and changed to use tacticals available in pred_setTheory.


mn200:
It looks as if the Travis failure is spurious so I'm happy to merge.  Thanks for the work!

binghe:
Yes, when I manually ran the test sequence (--seq=developers/travis/more_examples_seq -t1) in my Linux workstation, there's no failure. Thanks.

PR #438, opened 2017-07-17
Removed right-associative THEN and THEN1 in util_prob, extreal and lebesgue theories

binghe:
Some probability related scripts has the following historical setup (right-associative THEN and THEN1):

```
infixr >> >|
infix 1 >-
```

Now they're removed with minor changes in the proof of several theorems.

I also added a few new theorems in probabilityTheory, they're ported from version 2 of probability theory (examples/miller/formalize/probabilityScript.sml). they may be needed in the future (when robin-miller algorithm were ported to latest probabilityTheory)


mn200:
Thanks for this!

PR #442, opened 2017-07-21
Joe Hurd's Miller-Rabin probabilistic primality test ported to latest probabilityTheory

binghe:
Hi,

in response to the recent "hope" of Michael Norrish (#433), I have successfully ported the Miller-Rabin probabilistic primality test (`examples/miller`, Joe Hurd's ph.D thesis) to latest probabilityTheory (`src/probability`).

Code changes are *huge*. Actually almost every line is changed, because I also renamed all legacy tactics ("++" for THEN, "<<" for THENL, ">>" for THEN1) used by Joe Hurd to their standard symbols in HOL4. 

But you can verify my work though the following visible stuff:

 1.  The "version 2 of probability theory" files (`examples/miller/formalize/measureScript.sml`, and `examples/miller/formalize/probabilityScript.sml`) are removed.

 2. The building process finishes without error, and there's no cheated theorems.

 3. No theorem in `examples/miller/miller` (the core scripts of the Miller-Rabin probabilistic primality test) is removed.

 4. There're little changes in HOL's core probabilityTheory: just I removed those recently added theorems and put them into `examples/miller/prob/prob_algebraScript.sml`.

It remains to clean up the folder `examples/miller/formalize` with duplicated theorems removed and useful theorems moved to core theories. I'll send new PRs occasionally.   I hope this work has proved my ability to maintain HOL's probabilityTheory stuff:)

Regards,

Chun Tian (binghe)
University of Bologna, Italy


gilith:
Great work Chun Tian, thank you for maintaining and modernizing the Miller-Rabin example.

binghe:
Hello Joe Hurd,

You're welcome!  I found your work actually contains a rather complete formalization of discrete probability theory based on Bernoulli infinite trails model. The theorems related to independent functions are also excellent designed and proved.  I think the Bernoulli model has no other formalizations outside of HOL4, even in Isabelle (core and AFP) there's no such work. Compared to these  the Miller-Rabin test itself ... actually I know little and I'm not interested in it:)

On the major difficulty in my work: in the latest probabilityTheory in HOL4 (and real Math), the concept of "measurable" functions and the "measure_preserving" between two measure spaces, the related algebra must be sigma-algebra.   However, your "prob_algebra" defined in "prob_algebraScript.sml" is not sigma-algebra (yet). To preserve all your existing theorems, I have to define a new concept  "premeasurable" and a weakened "prob_preserving" (same definition as in version 2, but in latest probabilityTheory we have `prob_preserving = measure_preserving`), which both require only "algebras".  But these concepts seem to be only needed in the middle of proving later theorems. In the future I should take a deeper look in your probScript.sml, and see if some theorems on "bern" should use "measure_preserving" instead.

Any way, I've learnt so much from your work.

mn200:
Brilliant - thanks!

binghe:
Yesterday I found another duplication of probability theories in `example/dinningcryptos` directory (also by Joe Hurd), this directory looks more interesting to me: it depends on advanced probability theory with lebesgueTheory included, but the "measureScript.sml" file there seems belong to Joe Hurd - there're several important theorems (e.g. Beppo Levi's monotone convergence) not cover by HOL's core measure&lebesgue theories done by Tarek Mhamdi, Osman Hasan and Sofiene Tahar, although the two versions share the same root (maybe Aaron Coble).

So I'm going to merge the two versions and remove "measure, lebesgue and probability" 3 theories from `examples/diningcryptos`, because I may also need those extra theorems proved by Joe Hurd in my future probability project.

mn200:
Yes, you're right.  Merging this stuff would be great. There's also a `measureScript` under `examples/pgcl/src`.

PR #443, opened 2017-07-25
Try to get diningcryptos example to build

binghe:
I want to get diningcryptos example to build first, then try to merge the measure/lebesgue/probability theories into core HOL files.

I think example/diningcryptos was broken since last year: the PAT_ASSUM -> PAT_X_ASSUM changes haven't been fully applied to all scripts under `example/diningcryptos`. In last commit c73e58d (on 17 Aug 2016), Michael added some new theorems which used `IN_LIST_TO_SET`, but this theorem has been renamed to `LIST_TO_SET` later.

In this PR, so far I have fixed all scripts before the core file `dining_cryptoScript.sml`. But I'm blocked at the following definition in `dining_cryptoScript.sml`, which seems fine:

```
val dc_random_states_def = Define
  `(dc_random_states (0:num) = {(\s:string. s = (STRCAT "coin" (toString 0))); (\s:string. F)}) /\
   (dc_random_states (SUC n) = (IMAGE (\(s :bool state).
					(\x:string. if x = (STRCAT "coin" (toString (SUC n))) then
							T
						    else (s x)))
				       (dc_random_states n))
				    UNION
			       (IMAGE (\(s :bool state).
					(\x:string. if x = (STRCAT "coin" (toString (SUC n))) then
							F
						    else (s x)))
				       (dc_random_states n)))`;
```

but HOL4 throws the following strange type error:

```
Exception raised at Preterm.type-analysis:
in unknown location:
Attempt to unify different type operators: min$fun and list$list
Exception-
   HOL_ERR
     {message =
      "in unknown location:\nAttempt to unify different type operators: min$fun and list$list",
      origin_function = "type-analysis", origin_structure = "Preterm"} raised
```

Can you (e.g. Michael) please help checking if above definition has triggered a HOL bug, or the definition itself is invalid somehow?


mn200:
That bug (thanks for the report!) should be fixed as of 7493f6afe1.

binghe:
You're welcome. Please hold this PR until I push more commits to get the diningcryptos example to build (but without merging the probability).

mn200:
I won't move on this until you say you think it is ready.

binghe:
Hi Michael,

I'm afraid the current status is all what I can do in this direction. Currently the building process is blocked at `dc3_leakage_result` (line 365 of `dining_cryptoScript.sml`), in which a very complicated `CONV_TAC` seems trapping into infinite loops ...  I tried to analysis it, but the goal statement is super complicated, I couldn't make any progress.

I think it's safe to merge this PR at its current state, because I do have fixed everything except for the last file. Later, someone else, more experienced than me, may be able to push the progress further ...

On the other side, I'll still analyze the measure & probability theory code and try to merge useful parts into HOL's core probability stuff, but this work won't touch the current example until everything needed is merged, thus it will happen in separated PRs, and not related to the rest fixes of `dining_cryptoScript.sml`, not blocked at all.

mn200:
Thanks for looking into this!  I'll merge what you have, and as you say, someone may have the time and patience to extract all of the useful things.  If you are getting anything out of it all, that's good in itself.

PR #444, opened 2017-07-25
New version of the CCS example

binghe:
Hello HOL4 maintainers,

First of all, thanks for incorporating the CCS example into HOL in commit 0fe375e (May 15). After that initial work, I received more old scripts from the original author (Prof. Monica Nesi) and I did a further project to port the rest old work, with my own additions (several deep lemmas in CCS).

The "last" part in which I have used new_axiom() (c.f. recent discussions in hol-info mailing list), is NOT included in this PR, because I'm still trying new approaches.   With this PR merged, now HOL4 contains a rather complete implementation of the process algebra CCS, slightly better than similar work found in user contribute codes of Coq [1] and Isabelle AFP [2].

A project report can be found in my own GitHub repository [3], which contains also the original proof scripts in Hol88.

Regards,

Chun Tian
University of Bologna

[1] https://github.com/coq-contribs/ccs
[2] https://www.isa-afp.org/entries/CCS.shtml
[3] https://github.com/binghe/informatica-public/tree/master/CCS2


mn200:
In your `CCSSyntax` file you call the parser directly in various places.  This is bad style as it makes a module dependent on the global grammar.  If the user hides constants like `par`, your code will stop behaving correctly.  You should use the idiom as at the top of files like `src/1/Prim_rec.sml` to fix a grammar that the embedded calls can use.

binghe:
Hi Michael,

Thanks for reviewing the code!  You mean the use of `mk_const` may fail with the related names were hidden?   I will add the following code on top of `CCSSyntax.sml`:

```
structure Parse = struct
  open Parse
  val (Type, Term) = parse_from_grammars CCSTheory.CCS_grammars
end
open Parse
```

But I still have one related question: in the original code, there's no use of `mk_const`:

```
fun args_par tm = let
    val (opn, [P1, P2]) = strip_comb tm
in
    if opn = ``par`` (* HERE *)
    then (P1, P2)
    else failwith "term not CCS parallel"
end;
```

But code like `opn = ``par`` ` doesn't work if the CCS datatype contains two type variables, because the variable `opn` may contains a term in which the type variables are instantiated, like ``:(string, string) CCS``.  In my code, I try to get the actual type from the variable, and then call `mk_const` to "build" a new ` ``par`` ` const with the same type:

```
fun args_par tm = let
    val (opn, [P1, P2]) = strip_comb tm
in
    if opn = mk_const ("par", type_of opn)
    then (P1, P2)
    else failwith "term not CCS parallel"
end;
```

I don't know how HOL handle `mk_const` internally, but it looks like a performance & GC issue: I keep creating new constants for one-time use and throw them away immediately after leaving the function.  Is there a better way to do this equality check without respect to type variables?

Regards,

Chun

mn200:
If I understand your application, I think you probably want to use the `same_const` function.  (Alternatively, use dest_thy_const to pull the user's term apart and then check that it has the right `Name` and `Thy` fields.)

acjf3:
Hi Chun,

My preferred code for something like this would be:

fun args_par tm = 
  case Lib.total boolSyntax.dest_strip_comb tm of
     SOME ("CCS$par", [p1, p2]) => (p1, p2)
   | _ => failwith "term not CCS parallel”

Alternatively you could use:

val (par_tm, mk_par, dest_par, is_par) = HolKernel.syntax_fns2 "CCS" “par"

or

val (_, _, dest_par, _) = HolKernel.syntax_fns2 "CCS" “par"

Here dest_par is essentially the same as args_par but with a different error message for the failure case.

Regards,
Anthony

binghe:
Hi Anthony (and Michael),

Thanks for the coding suggestions. Amazing .. so I don't even need to write these functions, they are already there! (the difference in error messages doesn't matter)

Please temporarily hold this PR until I get these things fixed.

Regards,
Chun

konrad-slind:
Hi Chun,

  The function "same_const" tests whether two terms are the same constant,
ignoring
how their types may have been instantiated. So you could do

  val par_tm = ``par`` ;

     (or without the parsing invocation)

  val par_tm = prim_mk_const{Name="par", Thy="CCS"};

and then

 fun args_par tm =
  case strip_comb tm
    of (opn,[P1,P2]) =>
         if same_const par_tm opn then (P1,P2)
         else failwith ...
     | otherwise => failwith ...

Konrad.

konrad-slind:
Oops! Sorry for the redundancy.

binghe:
Thanks everyone, what a honor to receive help from 3 HOL maintainers together:)

I have fixed `CCSSyntax.sml` and `CCSConv.sml` adding a head piece of code to fix the CCS grammar, as suggested by Michael; And whenever possible, I use directly `HolKernel.syntax_fns1` and `HolKernel.syntax_fns2` to prevent defining new functions, as suggested by Anthony; finally, I have used `prim_mk_const` (in comb. with `same_const`) in place of `mk_const` whenever the purpose is just to test the constant equality, as suggested by Konrad and Michael.

mn200:
Can you confirm that you now think this is ready for merging?  Thanks!

binghe:
Yes I think so.

PR #447, opened 2017-07-28
The chapter of conversions in HOL manual

binghe:
Hi,

recently I've learnt a lot (from example code) the so-called _conversion functions_ (of type `:conv` or `:term -> thm`) in HOL4, a large group of functions including `REPEATC`, `THENC`, and `ORELSEC` for combining conversions into larger one. This is like a whole new world to me, because they were not mentioned systematically in HOL's _DESCRIPTION_.

But then I found that, actually there's indeed a chapter talking about this important facility, just it's not included when building the manual.  And from the several commented lines in existing chapter 2 (drules), the hidden chapter for conversions seems known to chapter 2.  Any way, they're 20 pages of valuable guides to me, and it also contains the answer to a question I ever asked in the mailing list months ago, i.e. the use of QCONV to handle unchanged conversions.

I tried to include that chapter back, then LaTeX shows several errors, all easily fixable. And I think it's better to insert this chapter before the chapter for "Goal Directed Proof", and after the "Derived Inference Rules", so it's the new Chapter 3.

I don't know the story, the author, and the reason why the chapter is not used so far. But this PR could fix the PDF building issue. Please feel free to merge it, or not.

Regards,

Chun Tian


mn200:
I'd like to see the sessions that are in this section of the manual redone with the "polyscripter" tool so that REPL exchanges are actually correct.  You can see how this is done by looking at the files with `.stex` extensions and the build commands in `Manual/Description/Holmakefile`.  The documentation for what is possible with `polyscripter` is in `Manual/Tools/README`.

If you don't have time for this, I will probably merge the chapter and then make the fixes myself at some stage.

binghe:
I didn't know `polyscripter` before. I'll make the changes and convert this chapter into .stex. Please hold this PR for now.

binghe:
The failed checks (`developers/travis/before_install.sh` failed) should be accidents, not related to TeX code changes in this PR).

PR #448, opened 2017-07-28
Fixed HolBddLib (for mosml only)

binghe:
The HolBddLib must have been broken since last year (due to the PAT_X_ASSUM related changes).

Also, to build the muddy.so dynamic library, the compiler parameters for setting the mosml includes doesn't fit into the directory structure of modern installations of mosml in both Linux and Mac OS X: whenever mosml is detected at somewhere, say `/usr/local`, the directory holding includes should be `/usr/local/include/mosml` instead of `/usr/local/include`. But instead of changing Makefiles, I think a standard way is to change the sources and to include files like `<mosml/mlvariables.h>` instead.

This PR fixed above two issues. After adding `Holmakefile`s for the two examples, the Solitaire example hangs, while the other (Katipuzzle) builds successfully.

P. S. A sad thing to see that `HolBddLib` doesn't support Poly/ML.  The `muddy.so` built by mosml, shouldn't work when loading into Poly/ML, and the ML interfaces of the `DynLib` between mosml and poly are only similar but not the same.


binghe:
My CCS example could have a "third" part in which HolBddLib is used to decide the bisimilarity of CCS processes, that's why I'm interested in HolBddLib.  Any comments in (the potential value of) making HolBddLib work in Poly/ML?

mn200:
Thanks for the work on this.  I will have to try to figure out how to make sure we properly test this stuff so that regressions can't linger like this. 

The linkage between the host ML system and the BDD library requires a bit of care because you want the ML system to be able to signal that BDDs have become unreachable so that the BDD library can do its own garbage collection.  (I believe this is done on the BDD side through destructors being called.) But I'd be more than happy to see it worked out for Poly/ML if it's possible.

binghe:
I think Poly/ML makes loading dynamic libraries a little easier: there's no need to compile with any ML headers, Poly/ML simply can load (and use) any dynamic library (.so or .dll). Muddy authors have taken care the garbage collection issues, by preventing each side (Mosml and Muddy) collecting garbage of the other side.  I feel there's a possibility to skip Muddy and compile Buddy directly into a dynamic library and then let Poly/ML use it.

On the other side, can you clarify one thing: when HOL has done something with help of external BDD programs, does HOL blindly trust BDD as an oracle, or the BDD actually returns enough information for  HOL to re-construct the proof, so that the resulting theorems can also be dumped into OpenTheory .art files?

PR #449, opened 2017-07-31
Move FINITE_BIJ_COUNT_EQ to pred_setTheory to simplify FINITE_BIJ_COUNT

binghe:
There's a theorem called `FINITE_BIJ_COUNT` in pred_setTheory since long time ago. It says for any countable set, there exists a bijection from {0, 1, 2, ..., b} to that set:
```
|- !s. FINITE s ==> ?f b. BIJ f (count b) s
```

But actually the stronger equation version also holds: (with more reasonable var names, found in miller and diningcryptos examples:
```
|- !s. FINITE s = ?c n. BIJ c (count n) s
```

I moved this equation theorem into `pred_setTheory` as `FINITE_BIJ_COUNT_EQ`, then the existing proof of `FINITE_BIJ_COUNT` becomes trivial. Total length of `pred_setTheory` decreased slightly, while one more useful theorem added.  The version in miller and diningcryptos examples can be deleted then, with other duplications identified and cleaned up too.

Some uses of `FINITE_BIJ_COUNT` in `lebesgueTheory` should actually be `FINITE_BIJ_COUNT_EQ`, because `GSYM` was called there, and it's only meaningful when calling with an equation theorem.

### DIFF
--- a/examples/diningcryptos/extra_pred_setScript.sml
+++ b/examples/diningcryptos/extra_pred_setScript.sml
@@ -386,35 +386,6 @@ val BIJ_INSERT = store_thm
    >> REPEAT STRIP_TAC
    >> METIS_TAC [IN_INSERT]);
 
-val FINITE_BIJ = store_thm
-  ("FINITE_BIJ",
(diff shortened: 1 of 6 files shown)


mn200:
Looks great to me - thanks!

PR #450, opened 2017-08-03
Added TeX notation for O (relation composition)

binghe:
This trivial PR extends previous PR #436, in which the TeX notations for RUNION, RINTER and RSUBSET in relationTheory were added. The TeX notation for big "O" (relation combination) is still missing, as a result, any theorem containing the big "O" cannot be output into TeX files and pass the compiling process.

I added new command `\HOLTokenRCompose`, and of course user can re-define it if they want.


mn200:
Thanks for this!

PR #454, opened 2017-08-21
Use _Theory.dat for loading theory data into _Theory.sml

barakeel:


xrchz:
Note, there is some `loadPath` hackery in `load_thydata` in `TheoryReader.sml` - if there's a better way to do this that works, it would be good to know. @mn200 maybe you have an idea?

barakeel:
Somehow, the cakeml build failed because of the loadPath hackery.

binghe:
I think after this work the `--otknl` build on `master` has been broken...

mn200:
I certainly agree that `--otknl` is broken; fixing it is on my todo list.

PR #455, opened 2017-08-23
Change behavior of MATCH_MP_TAC to fail only when being applied to a goal state

HeikoBecker:
Beforehand the following tactic would fail, even if ACCEPT_TAC succeeds:

    fun simple_apply thm = (ACCEPT_TAC thm) ORELSE (match_mp_tac thm);

Now the tactic only fails, if ACCEPT_TAC failed and match_mp_tac either is not given an implication or cannot match the theorem with the conclusion being matched on.

For more details, see https://sourceforge.net/p/hol/mailman/message/36008112/


mn200:
The documentation is vague about exactly when this error must occur, so I'm happy that this would not break any promises we have made.  The fact that the regression tests pass is also encouraging.  Thanks for the prompt to fix this.

PR #456, opened 2017-08-25
Adding formalisation of a translation from LTL formulae into generali…

simonjantsch:
…zed buechi automata via

weak alternating automata.
Includes theories for ltl, alternating automata, generalized buechi automata,
infinite words and correctness proofs for the translation.


mn200:
Can you add your directory name to an appropriate file under `tools/sequences` (perhaps `more-theories` or `final-examples`) please?  You should put in an appropriate number of exclamation marks.  If the directory builds in a couple of minutes or less, one `!` is fine.  If it takes less than 30 minutes, use `!!`, and if it takes longer still use `!!!`.  (These times all assume you are using a reasonably powered modern machine.)

simonjantsch:
I've renamed the directory and added it to tools/sequences/final-examples

mn200:
Many thanks for the example!

PR #469, opened 2017-10-05
Various documentation fixes

AndreasLoow:
I noticed that the type for Cases_on was incorrect in the documentation, and made the mistake of writing a short program that checked for more incorrect cases. Here is the result of this mistake.

I mainly tried to fix the obvious cases I found. I initially wanted to “synchronize” all type annotations in the source code and the documentation, including things like “thm -> tactic” vs. “thm_tactic”. But I gave up on the latter detail after a while (but there are not ridiculously many such cases left, most of them in Tactic/Term/Thm/Type if I recall correctly).

This PR removes some documentation for removed functions, moves documentation for functions that has been moved previously, and includes some minor various fixes (such as correcting type annotations).

Here are some comments and notes about things I’m not sure how to handle, or that I did not have time to do right now, in case anyone’s interested (this PR can be merged without doing them, I just wrote them down for a potential future):

* Docfiles/HolKernel.sort_vars.doc - not part of signature
* HolSatLib.SAT_PROVE.doc - not visible in rendered documentation, should maybe be moved to minisatProve.SAT_PROVE.doc?
* holyHammer.hh.doc - type and documentation (at least example) outdated I think
* listLib.LIST_CONV.doc and listLib.PURE_LIST_CONV.doc - functions removed?
* pairSyntax.is_prod - no such function any more? there’s no function to check if a type is a prod?
* Docfiles/Psyntax.doc - not visible anywhere I think
* wordsLib.WORDS_EMIT_RULE.doc - moved to basis_emitTheory, so we cannot move the documentation with the function?
* Drule.HALF_MK_ABS (potentially) moved to jrhUtils, so I updated some pointers to new location, not sure this is what we want to do
* Cond_rewrite.COND_REWR_CONV.doc + Cond_rewrite.COND_REWR_CONV.doc + Cond_rewrite.COND_REWR_TAC.doc types not same as in signature file
* dep_rewrite.DEP_PURE_LIST_ASM_REWRITE_TAC wrong type
* patriciaLib.PTREE_EVERY_LEAF_CONV.doc + patriciaLib.PTREE_EXISTS_LEAF_CONV.doc - outdated
* Many things in Prim_rec have outdated docs (types)
* proofManagerLib.expand_listf.doc + proofManagerLib.expandf.doc + proofManagerLib.flatn.doc + proofManagerLib.r.doc + proofManagerLib.set_goal.doc - outdated docs, now returns things
* Definition.\*.doc and holCheckLib.\*.doc - no corresponding source files
* res_quanTools pair syntax incorrect in types for functions (minor/not important)
* simpLib.type_ssfrag.doc - outdated
* Tactic.GSUBST_TAC.doc - outdated
* Thm.BETA_CONV.doc - type conv in documentation, type term -> thm in code, code does not include Abbrev. documentation should probably say term -> thm also? Same for Thm.REFL.doc.


xrchz:
Cool! Maybe you could also add the program you used (to `developers`)?

AndreasLoow:
The program is undocumented and not robust in any way. I think the documentation generation program already does most of the things a proper checker would have to do (parse signature files, proper parsing of .doc files, etc.). I think it would be better to modify the documentation generation program to check what my script checks currently, than to publish my script in its current form. I will try to look into this at some point.

xrchz:
OK fair enough. Why are some files deleted, for example, for `recInduct`? That function certainly exists...

AndreasLoow:
There's no recInduct in BasicProvers, so I removed that doc file. The doc file for bossLib.recInduct is still there. The other deleted files are deleted for similar reasons, I hope (i.e., I did not find a corresponding function in the source code).

PR #476, opened 2017-10-13
Add simple snippets to emacs mode

HeikoBecker:
I have been collecting some snippet for the emacs yasnippet plugin that I find helpful when writing down definitions, lemmas or proofs in HOL4 and I thought these may be helpful to some other HOL4 users. I also added a binding for "\M-h Tab" to expand snippets and give suggestions with company mode.

I tried to keep everything such that nothing breaks for any user, if yasnippet or company mode is not installed.

### DIFF
--- a/tools/hol-mode.src
+++ b/tools/hol-mode.src
@@ -2308,4 +2308,9 @@ second is to show the new term.
    '("Start HOL - horizontal split" . hol-horizontal))
 
 (define-key global-map [menu-bar hol-menu hol-process hol-start]
-   '("Start HOL" . hol))
+  '("Start HOL" . hol))
+
+;; Support for snippet expansion, if yasnippet is installed.
+(when (require 'yasnippet nil 'noerror)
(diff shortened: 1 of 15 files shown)


mn200:
This looks cool!  Is it obvious to a `yasnippet` user how to use these?  Thanks.

HeikoBecker:
It should be obvious after looking at the keywords defined in the files once.
Usually, I tried to stick to obvious abbreviations, e.g.  `once_rewrite_tac<tab>`  expands to `once_rewrite_tac[ <cursor> ] <cursor after pressing tab again>`.

For some larger constructs, I defined abbreviations that are maybe not so obvious and could be changed, as desired, e.g. `sg<tab>` expands to \` `<cursor>` \` `by (<cursor after tab>)` or `Lemma` expands to
```
val <cursorPos1>_def = store_thm (
  "<repeat what is typed at cursorPos1>",
  ` <cursor after tab>`);
```

I am happy to change any snippet definitions, if desired to make them more memorable/ easy to find.

mn200:
Sounds great.  Many thanks.  I will have to try these out.

PR #479, opened 2017-10-15
Fixed building of Manual/Description.pdf due to removal of DISJ_ss

binghe:
I found HOL document "Description.pdf" can't build in latest GitHub.  It seems that `DISJ_ss` has been recently removed from HOL (?), as a result when manual tries to use it, error returned:

```
libraries.tex                                                       FAILED! <1>
 Exception raised on line 2618: Value or constructor (DISJ_ss) has not been declared
 Fail "Static Errors"
```

solution is to define it in the document and eval more ML code before using `DISJ_ss`.  The syntax of `SSFRAG` seems also changed, and previous definition in document must be modified according to be evaluated correctly in latest HOL.

### DIFF
--- a/Manual/Description/libraries.stex
+++ b/Manual/Description/libraries.stex
@@ -2596,9 +2596,10 @@ For example, the following \simpset{} fragment will cause
 AC-normalisation of disjunctions
 \begin{hol}
 \begin{verbatim}
-   SSFRAG {ac = [(DISJ_ASSOC, DISJ_COMM)],
-           rewrs = [], filter = NONE, convs = [],
-           dprocs = [], congs = []}
+   SSFRAG {name = NONE,
+           convs = [], rewrs = [], congs = [],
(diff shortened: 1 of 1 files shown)


mn200:
I don't know about `DISJ_ss` being removed recently, but I'm sure I saw this code work recently when I made the relevant change (`9c873eec8`).  Very strange.  In order to avoid the repetition of `DISJ_ss`'s definition, can you adjust your fix to use the `##eval` directive please?  (See `Manual/Tools/README`.)

binghe:
Then I think it's never defined in HOL4 (I didn't find it in `boolSimps` lib). The manual has always "assumed" its existence, and when you started to eval it, the command fails.  Any way, I've made the changes using `##eval`.

mn200:
Thanks for making the fix(es)!

I thought I'd managed to build the Description manual successfully after the change I made, but I must have been mistaken.

PR #482, opened 2017-10-19
Added TeX_notation for "++" in listTheory

binghe:
Hi,

The "++" (APPEND) operator has no TeX notation, as a result the theorem containing "++" outputs `UnicodeChars.doubleplus` directly into TeX files and it cannot be correct handled.  There's no other special symbols without TeX notations in `listTheory`.

This PR fixed the issue by adding `\HOLTokenDoublePlus` with a valid definition that I found on Internet [https://tex.stackexchange.com/questions/4194/how-to-typeset-haskell-operator-and-friends].

Chun Tian


mn200:
Thanks for this!

PR #487, opened 2017-10-25
[CCS] some late developments with new theorems; fixed with K11; selftest added

binghe:
Hi,

this PR contains some "late" developments of the CCS theory. Changes include:

 1. I have used "optionTheory" to define the `Action` type which was based on `Datatype`;
 2. Added new CCS theorems and theoretical concepts (Observational contraction)
 3. Fixed the compatibility of proof scripts with HOL K11 final release. (it's not easy to get the same code working on both K11 and latest HOL)
 4. Renamed some CONV functions according to "modern" names of related concepts.
 5. Added a `selftest.sml` file to cover the most important decision procedure in the project, and converted manual tests to auto tests.

Now the code (and all names in it) are aligned with the final version of my MSc thesis, I won't commit new PRs for CCS in this year.  Future updates (in next year) should be considered as "post-thesis" work or based on new papers.

The code can be correctly built with latest HOL in all configurations, plus K11 final release (test with STD kernel, built by PolyML 5.6)

Regards,

Chun Tian


binghe:
Thank you!

mn200:
It turns out that this doesn't build from an fresh HOL checkout with `build -t` because of the dependency on `realTheory` (listed in the dependencies for the heap in the `Holmakefile`).  This could be fixed by making sure that `examples/CCS` was later than `src/real` in the sequence file, but I am trying a fix that removes the dependency on reals (which seems unnecessary).

binghe:
Yes, `realTheory` can be removed from the heap.  It was needed by my `graphTheory`, but finally I decided to not include it into the project, because I want to keep a pure CCS datatype which is the same as classic textbooks.

PR #490, opened 2017-10-30
Add some identity properties about rounding with ties to even

HeikoBecker:
All properties are proven in src/floating-point/lift_ieeeScript.sml.

- finite_float_within_threshold is a helper lemma, showing that any finite floating-point value must always be within its threshold when translated to a real number

- round_finite_normal_float_id shows translating any finite, normal floating-point number to a real number and rounding with tiesToEven is the identity function

- real_to_float_finite_normal_id is similar to round_finite_normal_float_id, lifting the theorem to  the function real_to_float

- float_to_real_real_to_float_zero_id shows that translating the real number 0 to a floating-point number and translating back to real must be 0

- non_representable_float_is_zero shows that any floating-point number that cannot be represented because its value is too small will flush to zero when rounding with ties to even

### DIFF
--- a/src/floating-point/lift_ieeeScript.sml
+++ b/src/floating-point/lift_ieeeScript.sml
@@ -1577,4 +1577,85 @@ val float_mul_sub_relative = Q.store_thm ("float_mul_sub_relative",
 
 (* ------------------------------------------------------------------------- *)
 
+val finite_float_within_threshold = Q.store_thm (
+  "finite_float_within_threshold",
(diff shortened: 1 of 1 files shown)


mn200:
I'll let @acjf3 approve this when it's ready.

FWIW, it looks as if bc0406b002 still has Unicode comparison operators on line 1584 (and maybe other places).

acjf3:
I'll merge and then remove the Unicode.

PR #493, opened 2017-11-06
hol-mode: Fix cleanup of leading connectives in "Apply tactic"

abxy:
The bug was caused by a location pragma being prepended before
cleaning up the tactic string. Reversing the order of those operations seems to fix it.

### DIFF
--- a/tools/hol-mode.src
+++ b/tools/hol-mode.src
@@ -467,9 +467,9 @@ With a prefix ARG, uses `expandf' rather than `e'."
   "Send selected region to HOL process as tactic."
   (interactive "r\nP")
   (let*
-      ((region-string0 (with-hol-locpragma start (buffer-substring start end)))
+      ((region-string0 (tactic-cleanup (buffer-substring start end)))
+       (region-string1 (with-hol-locpragma start region-string0))
        (ste "\"show_typecheck_errors\"")
-       (region-string1 (tactic-cleanup region-string0))
        (region-string (concat "let val old = Feedback.current_trace "
                               ste
                               " val _ = Feedback.set_trace "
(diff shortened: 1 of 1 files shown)


abxy:
I believe there is still some more work to do on this. The location pragma is generated based on the start position of the uncleaned input. It needs to be adjusted by adding the number of characters removed from the start of the string by tactic-cleanup.

I'm willing to do those changes, but I need some advice on how to test them. I'm assuming that the location pragma is added so that HOL can include location information in its error reports, but I haven't been able to produce such an output. Maybe the location pragma is not supported by Poly/ML?

mn200:
The location information is added so that parsing of terms can report errors.  If you want to see this sort of thing in action, you should try tactics such as 

    `P x` by tactic

To get error messages, try things that don't parse in place of `P x` (*e.g.*, `x +`).  Poly/ML can't use the location pragmas unfortunately.

abxy:
Thanks! I was able to verify my suspicion and fix the location problem, but I'm not really happy with my solution so far. This is my first go at lisp and I feel like it could be done in a cleaner way.

I have two alternative ideas for how to refactor it: (1) Replace `tactic-cleanup` with a function that takes start/end and returns trimmed start/end. (2) Replace `tactic-cleanup` with 2 functions that trim leading and trailing connectives respectively. Call them from `copy-region-as-hol-tactic` and diff the lengths before and after trimming leading connectives to find the start offset.

I think I prefer (2). Any other ideas?

mn200:
I agree that the second option looks better.

abxy:
I updated and force pushed my branch. I'm happy with the code now. Unlike my previous attempt it doesn't branch or use mutiple return values.

mn200:
Thanks for your work on this!

PR #497, opened 2017-11-26
Add Holmake support for the 'shell' function

oskarabrahamsson:
Adds Holmake support for the 'shell' command (see [GNU make manual](https://www.gnu.org/software/make/manual/make.html#Shell-Function)). 

Writing
```
VAR=$(shell ls -lah)
```
executes `/bin/sh -c "ls -lah"` and assigns the output of the command to `VAR`.


mn200:
Thanks!

PR #500, opened 2017-12-01
Added TeX notation for relation inversion (relinv)

binghe:
This small PR added TeX notation for relation inversion operator ($^T$).


mn200:
Thanks!

PR #501, opened 2017-12-12
HOL(y) Hammer OCaml 4.06.0 support

AndreasLoow:
Some simple Makefile fixes for hh to compile with OCaml 4.06.0.

Not sure how far back the -unsafe-string flag goes, so might break compatibility with very old OCaml versions. We also depend on ocamlfind being installed now, which was not the case previously.

### DIFF
--- a/src/holyhammer/README
+++ b/src/holyhammer/README
@@ -9,11 +9,7 @@ Requires:
 
     opam install num
 
-* g++ >= 4.8 (recent version with C++11 support)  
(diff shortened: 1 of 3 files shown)


AndreasLoow:
@oskarabrahamsson: Seems that the shell command is executed multiple times, do you know why?

(Also, I never see the ``@echo`` output for some reason.)

barakeel:
Great, thanks for the fix.

PR #502, opened 2017-12-18
Use the Unicode turnstile (user-customizable) when printing theorems

marioxcc:
When the trace “Unicode” is changed to 1 or 0 the theorem prefix is set to “⊢”
or the ASCII approximation “|-”, respectively. The prefix and suffix (empty by
default) can be customized via Globals.{thm_pp_prefix,thm_pp_suffix}.

### DIFF
--- a/Manual/Tools/expected1
+++ b/Manual/Tools/expected1
@@ -18,14 +18,14 @@ val it =
 > open arithmeticTheory;\quad\textit{\small\dots{}output elided\dots{}}
 > ADD_CLAUSES;
 val it =
-   |- (0 + m = m) \(\land\) (m + 0 = m) \(\land\) (SUC m + n = SUC (m + n)) \(\land\)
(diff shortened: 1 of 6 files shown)


marioxcc:
I can not tell what “Manual/Tools/expected1” _should be_. Replacing “|- ” by “\(\turn\) ” makes the comparison with “testoutput1” fail because of a difference in white space (indentation on the left). Since I do not know what the correct white space should be, I did not change this file.

marioxcc:
Some experimentation indicates that the pretty-printer does not take into account the 3 spaces in the first line (printed automatically by Poly/ML?) and thus when there is nothing to show before the turnstile, it indents as if the turnstile was the first character in the column, even though it is the third. I updated “expected1” with the correct indentation according to this reasoning.

mn200:
Thanks for the update!

marioxcc:
Thanks.

PR #504, opened 2017-12-20
Include Abbrev in dep_rewrite

AndreasLoow:
Just a micro fix, before this fix this happened (not sure why exactly):

```
> rw [];
val it = fn: tactic

> open dep_rewrite;
[...]

> rw [];
val it = fn: bossLib.tactic
```


mn200:
Thanks for this.

PR #505, opened 2017-12-29
[CCS] Fixed TeX_notation of STRONG_EQ; added ExampleScript

binghe:
Hi, 

this is a late update of the CCS formalization. There's a bug in the TeX_notation of STRONG_EQ (strong bisimulation equivalence), now it's fixed. Then I added ExampleScript.sml, which contains some important historical (now obsoleted) work from Monica Nesi's original code, plus a few examples that I used in the final version of my thesis (available at arXiv [1]).

Regards,

Chun Tian

[1] https://arxiv.org/abs/1712.09402


mn200:
Thanks for the update!

PR #513, opened 2018-01-31
vector theory

LiLiming:
All files are translated from the corresponding code files of Harrision in Hol-Light. They are all developed in Kananaskis-6. The order is permutation-iterate-vectors-determinant. If you run in the latest version, there may be some problems. Welcome to amend.

### DIFF
--- a/src/vectors and matrix/Determinant.sml
+++ b/src/vectors and matrix/Determinant.sml
@@ -1,0 +1,875 @@
+(* ========================================================================= *)
+(* Determinant and Matrix *)
+(*      
+It is translated from the corresponding code file of Harrision in Hol-Light.	            
+AUTHORS  : (Copyright) Liming Li, Yong Guan and Zhiping Shi
+             Beijing Engineering Research Center of High Reliable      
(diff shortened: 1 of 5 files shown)


mn200:
We'd greatly appreciate it if you could turn these into proper `Script.sml` files and make sure they work with the current tip of the `master` branch.

mn200:
Closing until the necessary preliminary work has been done.

binghe:
These code are now in a branch [1] of my HOL4 fork. I plan to first get them buildable in latest HOL4, under the folder `examples/vector`, with conflicted terms renamed and broken proof fixed, then create a new PR and think how to further clean it up.

The authors said "It is translated from the corresponding code file of Harrision in Hol-Light." and the  `iterateTheory` should be essentially the same thing (older or newer) as `src/probability/iterateTheory` in latest HOL4, thus they can be merged together.

So I'll take care of these code, once my current work on real-valued Probability Theory is done (with duplications in `miller` and `diningcryptos` examples merged).

[1] https://github.com/binghe/HOL/tree/vectorTheory/examples/vector

PR #519, opened 2018-03-07
Words theory tex fixes

AndreasLoow:
This PR is an unsatisfactory workaround for a LaTeX problem occuring in the documentation for words theory.

The package stix enable T1 font encoding, which apparently enables ligatures inside ``holtxt``s. I think the proper fix to this is to use XeTeX instead of standard LaTeX because XeTeX is able to handle ligatures correctly (i.e., they are enabled outside ``holtxt``s but not inside). (But using XeTeX requires some porting work, mainly how fonts are selected/loaded I think.) One alternative workaround is to not use T1 font encoding (e.g., by saying notext to the stix package), another workaround is to use the microtype package and tell it to disable ligatures inside ``holtxt``s (but this has other consequences). (One should note that ``'``s are seemingly always pretty-printed inside ``holtxt``s regardless of standard-LaTeX-based workaround -- this is correctly handled by XeTeX also.) Because of all this complexity I did the simplest workaround possible...

### DIFF
--- a/Manual/Description/description.tex
+++ b/Manual/Description/description.tex
@@ -7,7 +7,6 @@
 \usepackage{latexsym}
 %\usepackage{amsmath}
 \usepackage{amssymb}
-\usepackage{stix}
 %\usepackage{amsbsy}
 %\usepackage{amsthm}
 \usepackage{makeidx}
--- a/Manual/LaTeX/layout.sty
+++ b/Manual/LaTeX/layout.sty
@@ -36,8 +36,12 @@
 % FONT
 % ---------------------------------------------------------------------
 
+\usepackage{stix}
 \renewcommand{\rmdefault}{bch}
 \def\@pnumwidth{8mm}
+
+\usepackage{microtype}
+\DisableLigatures{encoding = T1, family = tt*}
 
 % ---------------------------------------------------------------------
 % MATH INDENTATION.  = \tabcolsep + three small verbatim spaces  (!)


mn200:
If it’s only `stix` that is causing the problem we could dispense with that and find some other way to get LaTeX to render “not iff”.  (It looks from 004eebd300 as if this is why `stix` is there.)

AndreasLoow:
`stix` is only the problem in the sense that it enables T1 font encoding (i.e., `\usepackage[T1]{fontenc}`), I think. Furthermore, whether one wants to include `stix` also depends on which math font one wants: 

With stix:
stix

Without stix:
no-stix

(I think the package also sets the main font, but that is overridden somewhere after the `stix` package is loaded.)

(Currently we also depend on `stix` for `\cupleftarrow`.)

mn200:
I am happy to merge whatever you end up settling on. 

Incidentally, some googling revealed [this stack exchange page](https://tex.stackexchange.com/questions/57718/ligatures-disabling), which suggests:

    \DisableLigatures[f]{encoding = *, family = tt* }

Have you tried that?

AndreasLoow:
Yes, that's what I meant by the microtype fix. I didn't take that approach because loading the microtype package affects more things than just ligatures. But if we load the microtype package in the common style file, along with the stix package, all documentation should be rendered consistently. So this is what I did now instead. I.e., now all documentation (or at least the documentation based on the common style file) uses the stix math font (and the microtype package).

mn200:
Thanks for all the work on this!

PR #520, opened 2018-04-06
Fix HOL mode emacs snippets not being loaded correctly, if the HOLDIR environment variable is not set

HeikoBecker:
I found out that I missed that on some platforms `HOLDIR` might not be set when starting emacs. 
Consequently snippets could not be used on these. This should fix this issue.


mn200:
Thanks for this!

PR #525, opened 2018-04-24
Render theory graph as SVG instead of JPEG

AndreasLoow:
I was initially planning to just render the theory graph as a PNG file instead of a JPEG file, as PNG is a much more suitable format for images containing text, but then I noticed that dot can output to SVG also, so I went with that alternative.

Pros: Looks much better

Cons: Irritating hover titels (that cannot be disabled in dot, apparently), and no cursor hover effect for the clickable links, at least not in my browser (it's possible to add `curser:pointer` to `ellipse` elements etc. (using CSS), but I didn't work very well in my browser)

**Edit**: Oh, I just tested it in Firefox, and it works much better than in Chrome (which was the browser I used initially). The curser is changed into "pointer" style when hovering links. Hopefully Chrome will improve in the future.

### DIFF
--- a/help/src-sml/DOT
+++ b/help/src-sml/DOT
@@ -9,10 +9,8 @@
 
       1. Generate the dependencies from the current theory hierarchy
       2. Write them to a file ready for processing by "dot"
-      3. Invoke dot to generate the JPEG-format image, the image map
-         information, and a PDF file while we're at it.
-      4. Generate a HTML file to hold the reference to the image, and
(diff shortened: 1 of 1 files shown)


AndreasLoow:
Before:
screen shot 2018-04-24 at 22 06 10

After:
screen shot 2018-04-24 at 22 05 58

mn200:
Looks nice!  Thanks for this.

PR #527, opened 2018-04-26
Emacs HOL mode fix, need subr-x for string-remove-suffix

AndreasLoow:
PR #520 calls `string-remove-suffix` without requiring `subr-x`. My Emacs crashes saying that "I do not know what that function is" with the current code. I don't know anything about Emacs Lisp but this PR seems to fix it.

### DIFF
--- a/tools/hol-mode.src
+++ b/tools/hol-mode.src
@@ -10,7 +10,7 @@
 
 (require 'thingatpt)
 (require 'cl)
-
+(require 'subr-x)
 
 (defgroup hol nil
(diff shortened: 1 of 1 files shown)


AndreasLoow:
Ping @HeikoBecker: Is the fix ok?

HeikoBecker:
@AndreasLoow: Thank you for fixing this. I was not aware of this issue. I have tested it locally on my machine and the snippets are still working, so all good from my side.

mn200:
Thanks for this!

PR #534, opened 2018-05-06
Ensure escape sequences generated by QuoteFilter are 3 digits long

bacam:
Otherwise they break holdep.  For example, if you have a newline after the `:` introducing a type it was changed into `\10` instead of `\010`.


mn200:
Thanks for the fix!  I like the extra paranoia in the code emitting the escapes, but really the `\10` shouldn't be being emitted at all: I think the rule that does this should be changed to handle newlines as they are handled elsewhere inside quotations.

PR #538, opened 2018-05-23
Some further developments on CCS formalization

binghe:
Hi,

recently my supervisor (Davide Sangiorgi) and I am working on a paper for EXPRESS/SOS 2018 workshop, in which we're going to formally introduce the new contraction relation variant that I found during my thesis work. And some discussions have led to further developments of CCS formalization, include two new theorems:

- COARSEST_PRECONGR_FINITE: the "coarsest (pre)congruence contained in `contraction` for finite-state CCS (van Glabbeek scenario)
- UNIQUE_SOLUTION_OF_ROOTED_CONTRACTION: a stronger version of UNIQUE_SOLUTION_OF_OBS_CONTRACTION with a rooted bisimulation conclusion.

I hope these work can be merged into HOL4 CCS example, so that we could directly cite HOL4 code repository in our paper. Thanks.

Regards,

Chun Tian

### DIFF
--- a/examples/CCS/ContractionScript.sml
+++ b/examples/CCS/ContractionScript.sml
@@ -4,8 +4,9 @@
 (*                                                                            *)
 (* THESIS        : A Formalization of Unique Solutions of Equations in        *)
 (*                 Process Algebra                                            *)
-(* AUTHOR        : (c) Chun Tian, University of Bologna                       *)
-(* DATE          : 2017                                                       *)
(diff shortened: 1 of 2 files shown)


mn200:
Thanks for the update.  Do you think your release notes paragraph needs updating too?  If so, please add your change to that file as part of this PR.

binghe:
Hi, I think the current paragraph is still precise enough, and I don't want make it longer (or into the abstract of my thesis & future paper), after all it's just an example theory living inside HOL.

mn200:
OK - thanks.

PR #541, opened 2018-05-24
Fixed building of HOL Reference under TeXlive 2018

binghe:
Hi,

I got the following errors when trying to build HOL manual PDFs under newly released TeXlive 2018:

```
[471]

! Package inputenc Error: Unicode character ∀ (U+2200)
(inputenc)                not set up for use with LaTeX.

See the inputenc package documentation for explanation.
Type  H <return>  for immediate help.
 ...                                              
                                                  
l.21922    [|- 2 = SUC 1, |- ∀n. SUC n = 1 + n]:
```

The Unicode character (seems encoded in UCS, not even UTF-8) appears inside `verbatim` block, it's not working by TeX default configurations unless `inputenc` package is loaded with Unicode encoding enabled.

Beside this fix, I changed to use new Latin Modern font, and added support of copy&paste 8bit characters from PDF.  Also added the Manual title into PDF metadata.

--Chun

### DIFF
--- a/Manual/Reference/reference.tex
+++ b/Manual/Reference/reference.tex
@@ -3,13 +3,21 @@
 % =====================================================================
 
 \documentclass[12pt,fleqn]{book}
+
+%  Font related packages
+\usepackage[utf8x]{inputenc}
+\usepackage[T1]{fontenc}        % This is necessary for the correct copy-paste of 8-bit characters
+\usepackage{lmodern}            % Latin Modern fonts (better than CM-Super)
+
 %\usepackage{amsmath}
-\usepackage{alltt}
 \usepackage{../LaTeX/layout}
 \usepackage{makeidx}
 \usepackage{alltt}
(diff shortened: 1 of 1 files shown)


AndreasLoow:
Setting inputenc to utf-8 might be fine, but I don't think the other changes should be merged. The font used by all manuals is set in `Manual/LaTeX/layout.sty` (`\usepackage{stix}`). Loading this font also enables T1 fontenc. Lastly, I think loading hyperref the way you do is redundant because it's already loaded, and the title seems to already end up in the metadata of rendered PDFs with the current setup.

mn200:
The issue is that we shouldn’t have that character in the file at all, and it’s good that TL 2018 is being better about reporting the problem.

I will fix the relevant doc file.

Michael

binghe:
OK, fair enough, as long as the PDF can be built I’m fine.

Chun

Inviato da iPhone

PR #542, opened 2018-05-25
Fixed src/temporal/examples.sml with HOL4_SMV_EXECUTABLE=NuSMV

binghe:
Hi,

I believe the SML script `src/temporal/examples.sml` is a dead one for long time. Now I've fixed it, tested with NuSMV [1] - a reimplementation and extension of SMV, the first model checker based on BDDs. I don't think anyone is still using the original SMV program by McMillan, but my code changes didn't break the compatibility if the original SMV program were used.

Following the way of calling external programs like Z3, Yices, etc. by HolSmtLib, now I defined a new environment variable called `HOL4_SMV_EXECUTABLE` for user to setup an SMV-compatible program for the model checking of LTL (actually it's CTL in this case).

Now the LTL_CONV command works again:

```
> LTL_CONV ``(a SUNTIL b) 0
              = (EVENTUAL (\t. b t /\ PNEXT (PALWAYS a) t)) 0``;
val it = ⊢ ((a SUNTIL b) 0 ⇔ EVENTUAL (λt. b t ∧ PREV (PALWAYS a) t) 0) ⇔ T:
   thm
```

and it can produce a countermodel if the  proof fails:

```
> LTL_CONV ``(a UNTIL b) 0
                = (EVENTUAL (\t. b t /\ PNEXT (PALWAYS a) t)) 0``;
# SMV computes the following countermodel:
===============================================
Formula is not true! Consider the countermodel:
===============================================
===============================================
Trace Description: CTL Counterexample 
Trace Type: Counterexample 
  -- Loop starts here
  -> State: 1.1 <-
    a = TRUE
    b = FALSE
    ell0 = TRUE
    ell1 = TRUE
    ell2 = TRUE
    ell3 = FALSE
  -> State: 1.2 <-
===============================================
Exception-
   HOL_ERR
     {message = "", origin_function = "NO_CONV", origin_structure = "Conv"}
   raised
```

The only thing which still confuses me is the last theorem in this script:

```
val SEPARATE_EVENTUAL_ALWAYS_THM = store_thm (
   "SEPARATE_EVENTUAL_ALWAYS_THM", ``
                (EVENTUAL(ALWAYS (\t. a t \/ PNEXT b t))
                 =  EVENTUAL(ALWAYS (\t. NEXT a t \/ b t))
                ) /\
                (EVENTUAL(ALWAYS (\t. a t \/ PSNEXT b t))
                 =  EVENTUAL(ALWAYS (\t. NEXT a t \/ b t))
                ) /\
                (EVENTUAL(ALWAYS (\t. a t \/ (b PSUNTIL c) t)) 0
                =
                if ALWAYS (\t.~c t) 0
                  then EVENTUAL(ALWAYS a) 0
                  else if ALWAYS (EVENTUAL c) 0
                         then EVENTUAL(ALWAYS
                                (\t. b t \/ c t \/ ((NEXT c) BEFORE (\t. ~a t)) t)) 0
                         else if EVENTUAL(ALWAYS a) 0
                                then EVENTUAL(ALWAYS a) 0
                                else EVENTUAL(\t. c t /\ ALWAYS (NEXT b) t) 0

                )
                ``,
        REPEAT STRIP_TAC THEN CONV_TAC LTL_CONV);
```

Here the formula is well-formed, but HOL reports the following strange internal error:

```
Don't expect to find a | in this position after a <beginning of input>
at line 210, character 38 and in compiler-generated text.
Proof of 

(EVENTUAL (ALWAYS (λt. a t ∨ PREV b t)) =
 EVENTUAL (ALWAYS (λt. NEXT a t ∨ b t))) ∧
(EVENTUAL (ALWAYS (λt. a t ∨ PSNEXT b t)) =
 EVENTUAL (ALWAYS (λt. NEXT a t ∨ b t))) ∧
(EVENTUAL (ALWAYS (λt. a t ∨ (b PSUNTIL c) t)) 0 ⇔
 if ALWAYS (λt. ¬c t) 0 then EVENTUAL (ALWAYS a) 0
 else if ALWAYS (EVENTUAL c) 0 then
   EVENTUAL (ALWAYS (λt. b t ∨ c t ∨ (NEXT c BEFORE (λt. ¬a t)) t)) 0
 else if EVENTUAL (ALWAYS a) 0 then EVENTUAL (ALWAYS a) 0
 else EVENTUAL (λt. c t ∧ ALWAYS (NEXT b) t) 0)

failed.
Failed to prove theorem SEPARATE_EVENTUAL_ALWAYS_THM.

Exception raised at Absyn.Absyn:
in compiler-generated text:
Don't expect to find a | in this position after a <beginning of input>
at line 210, character 38 and in compiler-generated text.

Exception-
   HOL_ERR
     {message =
      "in compiler-generated text:\nDon't expect to find a | in this position after a <beginning of input>\nat line 210, character 38 and in compiler-generated text.\n",
      origin_function = "Absyn", origin_structure = "Absyn"} raised
```

The "|" in that position must be a logical "OR" from the SMV language. I guess it's not well handled by some LTL-specific code when trying to build a oracle theorem by calling MK_THM. But currently I didn't find the related code yet. Beside this issue, all other parts in this example scripts work well.

Besides, in `Past_Temporal_LogicScript.sml`, I have defined some overload names for past temporal operators, following their modern names found in major LTL literatures, e.g. instead of `PNEXT`, a better name is `PREV` (or `PREVIOUS` if you prefer). `PEVENTUAL` is `ONCE`, `PUNTIL` is `SINCE` now.  Feel free to cancel this part if you don't like it.

Regards,

Chun Tian

[1] https://es.fbk.eu/technologies/nusmv-symbolic-model-checker

### DIFF
--- a/src/temporal/examples.sml
+++ b/src/temporal/examples.sml
@@ -1,5 +1,9 @@
-load "temporalLib"; open temporalLib;
+(* ---------------------------------------------------------------------------- *)
+(* Please set HOL4_SMV_EXECUTABLE env variable before evaluating this file,     *)
+(* Tested with NuSMV 2.6.0 by Chun Tian <ctian@fbk.eu> on May 25, 2018.         *)
+(* ---------------------------------------------------------------------------- *)
 
+load "temporalLib"; open temporalLib;
 
 (* ---------------------------------------------------------------------------- *)
 (* First, we prove that SUNTIL can be expressed by unary temporal operators     *)
@@ -8,12 +12,20 @@ load "temporalLib"; open temporalLib;
 (* and EVENTUAL, we could not define the binary temporal future operators.      *)
 (* ---------------------------------------------------------------------------- *)
 
-LTL_CONV (--`(a SUNTIL b) 0
-                = (EVENTUAL(\t. b t /\ PNEXT(PALWAYS a) t)) 0`--);
-LTL_CONV (--`(a SWHEN b) 0
-                = (EVENTUAL(\t. a t /\ b t /\ PNEXT(PALWAYS (\t.~b t)) t)) 0`--);
-LTL_CONV (--`(a SBEFORE b) 0
-                = (EVENTUAL(\t. a t /\ ~b t /\ PNEXT(PALWAYS (\t.~b t)) t)) 0`--);
+val SUNTIL_BY_UNARY_OPERATORS = save_thm (
+   "SUNTIL_BY_UNARY_OPERATORS",
+    LTL_CONV ``(a SUNTIL b) 0
+              = (EVENTUAL (\t. b t /\ PNEXT (PALWAYS a) t)) 0``);
+
+val SWHEN_BY_UNARY_OPERATORS = save_thm (
+   "SWHEN_BY_UNARY_OPERATORS",
+    LTL_CONV ``(a SWHEN b) 0
+              = (EVENTUAL (\t. a t /\ b t /\ PNEXT (PALWAYS (\t. ~b t)) t)) 0``);
+
+val SBEFORE_BY_UNARY_OPERATORS = save_thm (
+   "SBEFORE_BY_UNARY_OPERATORS",
+    LTL_CONV ``(a SBEFORE b) 0
+              = (EVENTUAL (\t. a t /\ ~b t /\ PNEXT (PALWAYS (\t. ~b t)) t)) 0``);
 
 (* ---------------------------------------------------------------------------- *)
 (* Manna and Pnueli consider several normal forms for temporal logic formulas.  *)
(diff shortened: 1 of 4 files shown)


thtuerk:
Nice fixes. Thanks. Perhaps I can find the time soon to go through my old code in `examples/temporal_deep` and see how much bit-rot occurred there.

mn200:
Thank you for the work!  Does NuSMV have a sufficiently open-source licence that we might distribute it with HOL?  And how large are the sources?

I'm keen to make sure that contributions like this don't rot again, so I'm looking for a way to ensure that the regression test machinery can test it along with everything else that is tested.

mn200:
The message about `|` is generated by the HOL parser, and I'd be very surprised if the code was passing SMV formulas to HOL.  Could it be that the code is trying to use the old syntax for conditional expressions (`guard => truebranch | falsebranch`)?

binghe:
Hi Michael,

according to [1], "New versions of NuSMV are distributed under the LGPL v2.1 license." but its source tarball has 20MB+ (if you try to download it from [1]), so I don't think it's a good idea to ship it with HOL. Advanced users may choose to use "nuXmv" instead (another compatible model checker made by FBK with more features), you should treat them as the same thing as SML tool "Yices" (close source).  Thus, to setup a test environment, it's better to manually install latest version of NuSMV on that test machine.

Hope this helps,

Chun

[1] http://nusmv.fbk.eu/index.html

binghe:
Thanks for your hints, I'll check if any ML code (in `src/temporal`) is trying to use the old syntax for conditional expressions.

mn200:
Thanks for the information about the NuSMV sources.  You're right that it seems excessive to distribute this.

PR #543, opened 2018-05-25
Lambek calculus

binghe:
Hi,

this was my first HOL project done in 2016-2017 on the formalization of Lambek Calculus (a version of Categorial Grammar for formal/natural languages). Basically I ported the same work from Coq's `contribs` library.   For students who is learning the same thing in NLP courses, now HOL4 can be used for proving the grammatical correctness of simple sentences  (as shown in the `ExampleTheory`).

The whole formalization (I mean the original one) is quite a beautiful work: it's a deep embedding, but it has a flexible relation definition (based on `Hol_reln`) with ability to accept extensions. Thus various version of Lambek Calculus (associative, non-associative, commutative or non-commutative) can be supported together, with possibility of other extensions. I believe, with some extra efforts we can turn HOL4 into a language parser based on Lambek grammars.

In `CutFreeTheory` I've created a deep embedding for "derivation tree (proof)", which is first-class object in Coq but not in HOL. This idea was learnt from prof. Jeremy Dawson (c.f. "Difficulties when migrating proof scripts from Coq" in HOL mailing list, on Jan 12, 2017).   I ported and proved many necessary lemmas towards the "Cut elimination theorem" for Lambek Calculus (L and NL), but I didn't finish the final proof (so is the original work in Coq!)  I hope I could find some time to finally finish it (and draft a paper).

I have to fix some `irule` issues to be able to build it in latest HOL4, maybe that's another good reason to put it inside HOL code base, because it may not run in old HOL versions any more.

Regards, 

Chun Tian

### DIFF
--- a/doc/next-release.md
+++ b/doc/next-release.md
@@ -14,7 +14,7 @@ Contents
 -   [Bugs fixed](#bugs-fixed)
 -   [New theories](#new-theories)
 -   [New tools](#new-tools)
--   [Examples](#examples)
+-   [New Examples](#new-examples)
 -   [Incompatibilities](#incompatibilities)
 
 New features:
--- a/doc/next-release.md
+++ b/doc/next-release.md
@@ -66,7 +66,7 @@ New tools:
 New examples:
 ---------
 
-*   We have resurrected Monica Nesi’s CCS example (from the days of HOL88), ported and extended by Chun Tian (based on HOL4’s co-induction package `Hol_coreln`).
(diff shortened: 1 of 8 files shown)


mn200:
Thank you for this!  Could you please add it to a build-sequence file as well?  Choose the number of exclamation marks in the testing annotation to reflect the length of time the directory takes to build.  If it's less than 10 minutes, one is fine.

binghe:
Thanks for accepting this example.

mn200:
For future reference, [this is a link to Jeremy Dawson's post](https://www.mail-archive.com/hol-info@lists.sourceforge.net/msg04167.html), as referred to above.

PR #484, opened 2017-10-20
Add reswites and the corresponding ssfrags to “res_quan”

marioxcc:

PR #550, opened 2018-06-14
Add “without loss of generality” tactic

marioxcc:
This is an adaption of tactics from [my personal repository](https://puszcza.gnu.org.ua/projects/hol-proofs/).


marioxcc:
I get this from traivs-ci.com: 

> We couldn't find the repository
> HOL-Theorem-Prover/HOL

help/Docfiles/Conv.SPLICE_CONJ_CONV.doc
```diff
@@ -0,0 +1,23 @@
+\DOC
+
+\TYPE {conv}
+
+\SYNOPSIS
+Partially normalize a conjunction.
+
+\DESCRIBE
+Normalize to right associativity a conjunctiong without recursing in the right
```
mn200: Typo on `conjunction`

mn200:
I've no idea what Travis's problem is.  I will see if I can manually get it to test the code.

mn200:
Shouldn't the various entry-points be `...wlog...`?  I don't understand why you have `...wloc...` everywhere.

marioxcc:
Thanks for the reports. You are correct that it should be “wlog”, not “wloc”. I will correct the problems you reported tomorrow (10 June UTC+05).

marioxcc:
Done. Let me know if there is any problem left.

mn200:
The Travis errors are to do with the inclusion of non-ASCII characters in the source.  It's a little annoying that we don't currently allow such characters even in comments, so I'd be happy to see a separate fix for the `testsrc` test to allow non-ASCII in comments (this would force that tool to lex SML comments).

Alternatively, you can put the string `UOK` onto lines that have non-ASCII characters and that tester will let them pass.

Finally you might just adjust the original sources.

marioxcc:
All non-ASCII characters in the SML source have the “UOK” token. The non-ASCII characters are in the documentation. If I placed “UOK” It would show in the generated files, no? Anyway, what is the point in forbidding non-ASCII character? HOL4 itself uses them in the pretty printer.

mn200:
Also, you can't use non-Unicode in `.doc` files because these are consumed by LaTeX, and our infrastructure is not yet smart enough to deal with this.  (The `Doc2Tex` tool in `help/src-sml` would need adjusting. This could probably be as simple as a string-replace to generate the right LaTeX, pretty well as encoded by the `umap` file in `Manual/Tools`.)

mn200:
One last thing: the `\TYPE` field in `.doc` files should mention the name of the SML value as well as the type. See, for example, `help/Docfiles/Tactic.CONJ_TAC.doc`.

marioxcc:
I will fix those issues tomorrow, because this is the time of day I have to go to sleep.

> For the moment, the policy is that we keep src ASCII only to make sure that it is easy for people to edit these files.
But this is easy. I edited my keyboard layout to include the common Unicode symbols of HOL and the `.XCompose` file for some rarely-used symbols. I will share the file on the mailing list, but probably it is only useful as an example because it is designed around a custom phonetic Russian keyboard layout, which is in turn loosely based on the Latin American keyboard layout.

Anyway, eliminating non-ASCII characters is a major usage of my time when contributing changes to HOL4 and I do not think it makes things any easier for anyone. If anything, it makes things harder for those who have customized keyboard layout. I can type `∧` much more easily than `/\`

mn200:
I'm totally sympathetic: however, if people cannot even see the characters properly because of a font-failure or similar, that makes files very hard to edit safely.

thtuerk:
A year ago, I was giving a HOL 4 course for master students. So plenety of people using HOL for the first time on their own maschines with plenty of different systems and setups. I believe that even nowadays having good uinicode fonts that display certain characters paperlyis unluckily an issue. So sticking with ASCII is definitely a good thing in my opinion.

@marioxcc The emacs mode has even a key-combination for replacing common unicode symbols. If you are not using emacs, a tiny scripte (e.g. sed) does the trick. So, I don't see why removing unicode is a major drain on your time. This is especially true since the policy does not affect your own theories and libs.

thtuerk:
Talking about teaching HOL: The new without-loss-of-generality tactics seems like something needed only in special, rather rare situations. Polluting central files like `Conv` or `BasicProvers` with something special makes learning HOL harder. How about moving this stuff to a less prominent place, e.g. into a new lib instead of `BasicProvers`? I had to read the doc very carefully and multiple times to hopefully understand what is going on. How about an example and some motivation when these tactics are useful?

marioxcc:
@thtuerk: The tactic is simply an incarnation of the idiom “without loss of generality, assume _P_”. It is very general and so I think it belongs to BasicProvers which already has the somewhat-specialized induction tactics. The description is complicated because I tried to make it _define_ a unique tactic (up to the appropriate equivalence relation for tactics). Free feel to suggest a simpler description that meets the same goal. You can see examples in my repository. About non-ASCII characters: Most if not all general purpose GNU/Linux distribution include DejaVu fonts which have the required symbols and they are used as fallback for characters not found in the user-requested font. A system that lacks support for the basic symbols in the Unicode BMP is broken and the burden is onto the administrator to fix it, not us. Replacing the symbols with sequence of characters can not be done mechanically because one must _selectively_ reflow text to keep the margins and simple heuristics are bound to be inaccurate. I always do line wrapping of formulas manually.

thtuerk:
I disagree that it is something very general. For me it is in contrast to `measureInduct_on` something not core that should not live in such a prominent place. You might use it a lot, but I believe not many others do.

With the idiom "without loss of generality" I can image many slightly different things. That's why I asked for documentation, which I believe is missing. Your text is very technical, just a tiny bit above the implementation level. It is not real documentation since it lacks the higher level of abstraction, the additional information you should give in a documentation. I could just as well read the source code. 

Not all people use a modern, general purpose GNU&Linux distribution and as I said, I know of plenty of subtle problems especially for new-bees. Setting up emacs on certain Windows machines has for example unicode issues. Judging from your strange comments about replacing symbols you yourself seem to have serious issues with using a text editor. So you probably know what I'm talking about.

help/Docfiles/BasicProvers.var_wlog_tac.doc
```diff
@@ -0,0 +1,25 @@
+\DOC
+
+\TYPE {term quotation -> term quotation list -> tactic}
+
+\SYNOPSIS
+Assume a proposition without loss of generality.
+
+\DESCRIBE
+Given {var_wlog_then q vars_q} let {asm ⊢? c} be the the goal. {q} is parsed
```
xrchz: "the" is duplicated

help/Docfiles/BasicProvers.var_wlog_tac.doc
```diff
@@ -0,0 +1,25 @@
...
+in the goal context to a proposition {P}. {vars_q} are parsed to variables in
+the goal context. Let {egv} (effectively free variables) be the free
```
xrchz: Why does "egv" stand for "effectively free variables"? Should it be "efg"?

help/Docfiles/BasicProvers.var_wlog_tac.doc
```diff
@@ -0,0 +1,25 @@
...
+variables of {P} and {c} that are not free in the assumptions and are not in
+{vars}, from left to right and first {P}, then {c}. Let {gen_vars} be {vars @
+egv}. Let {asm'} be the elements of {asm} in which any of {vars} is a free
+variable. Let {ant} be the result of splicing {p :: asm'}. The first subgoal
+is {asm, (∀(gen_vars). ant ⇒ c), ¬P ⊢? c}. The proposition {P} is added to the
+assumptios with {strip_assume_tac}. If this generates subgoals (as is usually
+the case), then those subgoals follow.
+
```
xrchz: By the end of this, I still have no idea what this tactic does. The description is too low-level. And some examples would help in the documentation.

help/Docfiles/BasicProvers.var_wlog_tac.doc
```diff
@@ -0,0 +1,25 @@
...
+\DESCRIBE
+Given {var_wlog_then q vars_q} let {asm ⊢? c} be the the goal. {q} is parsed
+in the goal context to a proposition {P}. {vars_q} are parsed to variables in
+the goal context. Let {egv} (effectively free variables) be the free
+variables of {P} and {c} that are not free in the assumptions and are not in
+{vars}, from left to right and first {P}, then {c}. Let {gen_vars} be {vars @
+egv}. Let {asm'} be the elements of {asm} in which any of {vars} is a free
+variable. Let {ant} be the result of splicing {p :: asm'}. The first subgoal
+is {asm, (∀(gen_vars). ant ⇒ c), ¬P ⊢? c}. The proposition {P} is added to the
+assumptios with {strip_assume_tac}. If this generates subgoals (as is usually
```
xrchz: typo: assumptios

help/Docfiles/BasicProvers.var_wlog_then.doc
```diff
@@ -0,0 +1,18 @@
+\DOC
+
+\TYPE {term quotation -> term quotation list -> thm_tactic -> tactic}
+
+\SYNOPSIS
+Apply a theorem-tactic using a proposition that can be assumed without loss of
+generality.
+
+\DESCRIBE
+Like {var_wlog_then}, but the theorem {P ⊢ P} is passed to the user-provided
```
xrchz: I thought this _was_ `var_wlog_then` (not just "like" it).

help/Docfiles/BasicProvers.wlog_then.doc
```diff
@@ -0,0 +1,19 @@
...
+\TYPE {term quotation -> thm_tactic -> tactic}
+
+\SYNOPSIS
+Apply a theorem-tactic using a proposition that can be assumed without loss of
+generality.
+
+\DESCRIBE
+Like {var_wlog_tac}, but the theorem {P ⊢ P} is passed to the user-provided
+theorem-tactic instead of {strip_assume_tac} and does not handle existing
+hypotheses especially.
```
xrchz: "especially" should probably be "specially"?

marioxcc:
> So you probably know what I'm talking about
@thtuerk: I understand the rest of your message, but about this I have no idea what you are talking about.

@xrchz: Thanks. I will address those problems.

mn200:
There is no great rush - thanks for your work!  I think Ramana and Thomas are right that some examples of the tactics being used would be a great addition to the documentation.

marioxcc:
@mn200: Please consider merging this change; let me know if there is any remaining problem.

src/basicProof/BasicProvers.sml
```diff
@@ -4,6 +4,16 @@
...
+"COPYRIGHT" in the HOL4 distribution (note added by Mario Castelán         UOK
+Castro).
+
+For the avoidance of legal uncertainty, I (Mario Castelán Castro) hereby   UOK
+place my modifications to this file in the public domain per the Creative
+Commons CC0 1.0 public domain dedication <https://creativecommons.org/publ
+icdomain/zero/1.0/legalcode>. This should not be interpreted as a personal
+endorsement of permissive (non-Copyleft) licenses. *)
+
 structure BasicProvers :> BasicProvers =
```
thtuerk: Such copyright notices are not used normally in HOL. I know that  @marioxcc believes they are necessary and tries to push people towards using them. I believe otherwise. If someone adds them to their own files, thats their own business. But before adding them to something central like BasicProvers, some general discussion on how to do it for HOL is needed, I believe. If we do it, it should be done in the same fashion for all files.

src/basicProof/BasicProvers.sml
```diff
@@ -285,6 +295,188 @@ fun Induct (g as (_,w)) =
...
+      end
+  end
+  handle HOL_ERR _ => raise ERR "UNDISCH_CONJ" ""
+
+(* Proves
+  ⊢ (¬P ⇒ (∀(vars). P_hyp ⇒ t) ⇒ t) ⇒   (* “wlog_thm”. *)                  UOK
+     (∀(vars). P ∧ hyp ⇒ t) ⇒           (* “new_sg_thm”. *)                UOK
+     hyp ⇒                                                                 UOK
+     t»                                                                    UOK
+except that when "hyp" is "NONE", it is omitted the obvious way. If "hyp" is
```
thtuerk: Using UOK here I consider an abuse. I believe this unicode should be removed here.

help/Docfiles/BasicProvers.wlog_tac.doc
```diff
@@ -0,0 +1,88 @@
...
+parsed to variables in the goal context. Let {efv} (effectively free
+variables) be the free variables of {P} and {c} that are not free in the
+assumptions and are not in {vars} from left to right and first {P}, then
+{c}. Let {gen_vars} be {vars @ efv}. Let {asm'} be the elements of {asm} in
+which any of {vars} is a free variable. Let {ant} be the result of splicing
+{p :: asm'}. The first subgoal is
+{asm, (!(gen_vars). ant ==> c), ~P ?- c}. The proposition {P} is added to the
+assumptions with {strip_assume_tac}. If this generates subgoals (as is usually
+the case), then those subgoals follow.
+
```
thtuerk: This is a very technical description. It is nearly at the level of the source code and therefore not suitable as documentation in my opinion. I would opt for either deleting it completely since it has no useful information or (preferably) rewrite it at a higher level of abstraction.

help/Docfiles/BasicProvers.wlog_tac.doc
```diff
@@ -0,0 +1,88 @@
...
+\USES
+A typical use case is to continue the proof assuming one case where all cases
+are symmetric. The first subgoal is a good candidate to be solved by a first
+order prover like {PROVE_TAC} or {METIS_TAC} providing to it the appropriate
```
thtuerk: What is "symmetric"? Grammar on "where".

help/Docfiles/BasicProvers.wlog_then.doc
```diff
@@ -0,0 +1,20 @@
...
+\TYPE {wlog_then : term quotation -> term quotation list -> thm_tactic -> tactic}
+
+\SYNOPSIS
+Apply a theorem-tactic using a proposition that can be assumed without loss of
+generality.
+
+\DESCRIBE
+Like {wlog_then}, but the theorem {P |- P} is passed to the user-provided
+theorem-tactic instead of {strip_assume_tac}.
+
```
thtuerk: Example. Why is this useful?

src/basicProof/BasicProvers.sml
```diff
@@ -285,6 +295,188 @@ fun Induct (g as (_,w)) =
  end
  handle e => raise wrap_exn "BasicProvers" "Induct" e
 
+(*---------------------------------------------------------------------------*)
+(* No loss of generality                                                     *)
+(*---------------------------------------------------------------------------*)
```
thtuerk: No -> without ?

src/basicProof/BasicProvers.sml
```diff
@@ -285,6 +295,188 @@ fun Induct (g as (_,w)) =
...
+     t»                                                                    UOK
+except that when "hyp" is "NONE", it is omitted the obvious way. If "hyp" is
+none then "p_hyp" is just "hyp" else it is the splice of "p" with "hyp". The
+tuple returned is the aforementioned theorem and "P_hyp". *)
+fun wlog_rule vars p hyp t =
+  let
+    val the = liteLib.the
+    fun forall vars t = list_mk_forall (vars, t)
+    fun implies t1 t2 = mk_imp (t1, t2)
+    val new_sg_t = forall vars (implies (case hyp of
```
thtuerk: non standard functions `forall`, `implies`, `the` used without significant benefit. Please remove for clearer, better maintainable code.

src/basicProof/BasicProvers.sml
```diff
@@ -285,6 +295,188 @@ fun Induct (g as (_,w)) =
...
+                                           SOME hyp_t => mk_conj (p, hyp_t)
+                                         | NONE       => p)
+                                        t)
+    val new_sg_thm = ASSUME new_sg_t
+    val thm_p = MP (SPECL vars new_sg_thm)
+                   (case hyp of
+                      SOME hyp_t => CONJ (ASSUME p) (ASSUME hyp_t)
+                    | NONE       => (ASSUME p))
+    val wlog_t = ref NONE
+    val wlog_ant = ref NONE
```
thtuerk: unused reference `ẁlog_t`

src/basicProof/BasicProvers.sml
```diff
@@ -285,6 +295,188 @@ fun Induct (g as (_,w)) =
...
+    val splice_thm = ref NONE
+    val _ =
```
thtuerk: I see no reason to use references here. Please use pure code for deriving `wlog_and` and `splice_thm`.

help/Docfiles/BasicProvers.wlog_tac.doc
marioxcc: I do not see the grammar error. The sentence is of the form “A typical use case _is_ ... _where_ ...”.

marioxcc:
@thtuerk:
- About the Copyright notices: I see no better way to _selectively_ place my changes in the public domain. You should note that such notices already appears in the Kananaskis-12 release; therefore, I hold that including these notices in some files is the status quo and if you want to forbid them, the burden of justifying such a decision as well as removing the existing notices in the master is upon you.
- The limitation that non-ASCII characters are not allowed without the “UOK” token is already of dubious utility and lowers readability of the code.
- There is no way to write purely functional tactics in HOL4 because primitive inference steps are not purely functional. As for the local references in the tactic itself: Before I submitted the pull request, I wrote a version without local references (that I have since discarded) but found it unnecessarily verbose and decided for the version with local references.
- I hold that there should be a specification of the code _somewhere_. I will not remove details without which, the specification would be incomplete.
Overall, I find your comments to be unproductive. I will happily fix any errata (“bug”) and follow any _project-wide_ consensus as I have already done by making the files pass `unicode-grep`. However, I am unwilling to use my time on accommodating your _personal taste_.

help/Docfiles/BasicProvers.wlog_tac.doc
marioxcc: `strip_assume_tac` has many special cases. It does not simply strips and adds the proposition to the assumption. It can solve the goal outright, for example. I see no concise way to accurately describe what happens with the proposition `P` other than saying it is passed to `strip_assume_tac` or describing the complete behavior of `strip_assume_tac` here.

help/Docfiles/BasicProvers.wlog_then.doc
marioxcc: That is explained for the general case in the Description manual. There is no need to repeat it for every tactical that takes a theorem-tactic as an argument.

src/basicProof/BasicProvers.sml
marioxcc: This is a valid observation. I will fix it.

binghe:
@marioxcc, if you think your new tactics are really that useful, you should first put them in an example theory and let others see the power. I think HOL maintainers would like to merge more examples, as long as they're meaningful.   So far I don't see how your new tactics should belong to "BasicProvers", because that will add the things that new HOL learners think they must learn.

Second, there's absolutely no practice in other open source projects, that, the person who just add some copyright notes need also put his name together saying things like "I, xxx, added these texts".  If I as newbie go to Linux kernel project and find an (unlucky) code file and do the same as you did, Linus Torvalds will use his most malicious words to attack me.  Here you are just too lucky with HOL's kind maintainers. Plus, all issues about Unicode characters come from the accent of your name!

So as another HOL user I suggest this whole PR should be rejected.

binghe:
I also suggest, for PRs like this, all HOL maintainers (those who has direct write access, basically) should vote in private and make a group decision based on majority opinions. In this way, it is for the real benefits of major HOL users.

xrchz:
Please keep the discussion polite and friendly :) It's great to have many voluntary contributors, including those passionate about our code and documentation quality! I believe there's some good work here on a useful tactic and would like to see it merged in some form. I found the MAX example particularly compelling.

binghe:
I see no better way to express the same message in a more “polite” way.

Inviato da iPhone

marioxcc:
I want to address some of the comments made by @binghe:

> If I as newbie go to Linux kernel project and find an (unlucky) code file and do the same as you did, Linus Torvalds will use his most malicious words to attack me.

Linux uses a Copyleft license any modification thereof is automatically under the GNU GPL 2 (otherwise it would be a Copyright violation). Moreover, **Linux has a [mechanism with the same intent as my file headers](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst?id=HEAD),** except that they use commit messages instead.

HOL4 uses a permissive license and therefore modifications are not automatically free software. My notices make it clear beyond all doubt that my contributions (so far) are public domain, and I use them because I consider it my duty to make sure that my changes are free software. Moreover, I repeat that **my headers are already part of HOL4 and are in the Kananaskis-12 release; they are not new.**

> Plus, all issues about Unicode characters come from the accent of your name!

No, that was only the case with a previous pull request, but it was solved back then.

> If you think your new tactics are really that useful, you should first put them in an example theory and let others see the power.

In the body of the pull request I gave a link to my personal repository where they are used several times in a real-world case (Note: the identifiers of the tactics in my personal repository are slightly different, but the behavior is the same).

As for the usefulness of these tacticals in informal terms: I believe teaching this is beyond the scope of HOL4. Writing a non-trivial theory exclusively with the intent of showing how to use `wlog_tac` and `wlog_then` would be useful, but I think it would be a double standard to _require_ it.

- We have induction and case split tactics, and we assume users already know how to use induction in an informal proof. The first example theory in the tutorial (about Euclid’s theorem) addresses only how to translate an informal proof by induction into a formal proof, not how to use induction to find an informal proof.
- We have a theory of naturals, integers, reals and complex numbers, but we do not teach analysis to the user in our documentation. We do not have theories written expressly as an example of how to use these theories because we assume he has the necessary knowledge.
- It is the same here: Now we have a “no loss of generality” tactic, but I assume users already know how to invoke this concept in informal proofs. Please bear in mind I did not invent the concept of “Assume _P_ with no loss of generality”.

binghe:
Let me do this in a polite way: I don’t want to see your name with copyright notes, even they’re in fact added by you. The fact that they were already in some files of K12 release doesn’t mean you can do that again, and previous additions (the part of your name) need to be removed when those files were modified (due to other reasons) again. But keep in mind that I’m not HOL maintainer at all, so my these words do not count at all.

Inviato da iPad

thtuerk:
For me, one huge issue here is the license. I agree fully with @binghe. I believe we need a common policy for the core files of HOL. If a certain example or external tools are under a different license that is fine. However, having mulitple licenses on the core files of HOL and even multiple licenses for a file is not sensible for me. Moreover, it is not @marioxcc decision which license core HOL 4 files should be distributed under. If @marioxcc does not like the license of HOL 4 he should not contribute to it. 

Such copyright noticed by @marioxcc indeed made it already into the release. I don't know whether this happened deliberately. However, I have neither heard about a decision by one of the main maintainers nor of a discussion in the community about it. For me these headers should be removed as well. Having these in some places can't be seen as a permission to add them to additional files! 
Either we should put certain headers in all files or none. Similar to @binghe I don't want to see a single author in such a copyright notice, but the full list of contributors.

This is an issue not really related to the rest of this pull request and for me a quite important one. That's why I created issue #550 about it.

mn200:
I am on long service leave at the moment and don't really have time to engage with this long discussion.  I am happy to delegate responsibility for resolving this to @xrchz and @konrad-slind.

marioxcc:
@thtuerk: The code does what it is supposed to do and does it well. No errata is known to me. Saying it is _poor quality_ is unfounded.

> In most cases this tactic can be used, I strongly advice a declarative approach in order to have maintainable proofs.

src/basicProof/BasicProvers.sml
marioxcc: I added these functions to make the subsequent code more clear by reducing the need for breaking long lines (especially noticeable in “forall” against “list_mk_forall”).

src/basicProof/BasicProvers.sml
xrchz: Thanks for creating the separate issue #559. Let's resolve this there, ok?

src/basicProof/BasicProvers.sml
xrchz: I am in favour of allowing Unicode in comments because I think it aids readability, but I am sympathetic to the argument that it hinders readability too. I suggest we discuss Unicode in a separate issue and/or mailing list thread and focus on the wlog tactics in this pull request.

help/Docfiles/BasicProvers.wlog_tac.doc
xrchz: I think @marioxcc has written good high-level documentation above the low-level description, and the low-level part is clearly marked as such ("Detailed description"). As such, I suggest we keep all the documentation, and retain the option, as always, to revise and improve it in subsequent work.

help/Docfiles/BasicProvers.wlog_tac.doc
xrchz: This reads fine to me.

help/Docfiles/BasicProvers.wlog_then.doc
xrchz: I agree with @marioxcc here. If more explanation or examples are desired, let's have them in future commits - I don't think they are a blocking requirement for this PR.

src/basicProof/BasicProvers.sml
xrchz: I agree with @thtuerk here: these abbreviations don't add much and break (informal) convention. However, I wouldn't require their removal for this PR if @marioxcc is strongly opposed to removing them.

xrchz:
@marioxcc there are a few requested changes outstanding in @thtuerk's review, which I would be grateful for you to submit. @thtuerk I've replied to each of the points in your review with a view to moving the review process forward. I would be grateful if, after @marioxcc addresses the remaining requests, you could approve the changes (or reply with any blocking items that are both unresolved and not going to be discussed in a separate thread). Thanks very much for your detailed work on reviewing these changes so far!

thtuerk:
After I noticed the copyright notice, it became the biggest issue with this pull request for me. I agree with @binghe that with these notices in, the pull request can't be merged. I would demand this anyhow, but the situation is even made worse by @marioxcc  strange interpretation what permissions an approved pull request grants. Following his interpretation, pull request #558 allows to add typos and broken files to HOL.

Apart from the license issue, my main concern is that `BasicProvers` is a very central file, containing core functionality of HOL 4. Therefore it is important for new users just learning about HOL 4. I asked for example my students during the master course at KTH last summer to look at it. For me, the wlog-tactics are sometimes handy, but do not belong in such a prominent place. If they are added to such a place, they need to be higher quality than if being added to some less prominent place. And I still believe that both documentation and implementation are poor quality.

I agree with Ramana that the documentation is no show stopper. I stand with my opinion though that the documentation is poorly written. However, the other points are important for me. So in order to approve this pull request, I would demand one of the following options. I would prefer the first one.

**Option 1: remove wlog from BasicProvers**
Move it to a less prominent place. An example might be a new lib in `src/1` called something like `wlogLib`. This would of course create some clutter by creating a new file. However, depending on the outcome of the license issue, having a separate file might even be benefitial. If wlog is moved out of `BasicProvers`, I would not press the points about code quality. 

**Option 2: code remains in BasicProvers**
In this case it needs cleaning up. It is read by many people and therefore needs to be easily understandable and maintainable. It needs to follow common ways of doing things. Therefore, I would insist in this case on the points made above including the removal of unicode.

help/Docfiles/BasicProvers.wlog_tac.doc
thtuerk: For me, there is confusion, to what _where_ refers. One can read it as _... assuming one case where ..._ This is a minor point though.

help/Docfiles/BasicProvers.wlog_then.doc
thtuerk: My question was: why is low level interface exposed. Is there a useful application for this low level interface. Or to word it differently, why is `wlog_then` added to the signature?

help/Docfiles/BasicProvers.wlog_tac.doc
thtuerk: I just spotted a wrong name. This is the documentation for `wlog_tac` not `wlog_then`.

help/Docfiles/BasicProvers.wlog_then.doc
thtuerk: `Like {wlog_then}` should be `Like {wlog_tac}` I believe.

xrchz:
> the situation is even made worse by @marioxcc strange interpretation what permissions an approved pull request grants. Following his interpretation, pull request #558 allows to add typos and broken files to HOL.

This seems unfair: as far as I can tell, @marioxcc had nothing to do with #558 and has not mentioned it before.

src/basicProof/BasicProvers.sml
thtuerk: Yes, let's resolve it there. However, having this copyright notice in is a blocking issue for me at the moment.

konrad-slind:
Separate note: SPLICE_CONJ_CONV has an incorrect example: (variable a3
appears on the rhs of the equivalence but not the left.

+> SPLICE_CONJ_CONV “(a1 ∧ a2) ∧ (b1 ∧ b2) ∧ c”;
+val it = ⊢ (a1 ∧ a2) ∧ (b1 ∧ b2) ∧ c ⇔ a1 ∧ a2 ∧ a3 ∧ (b1 ∧ b2) ∧ c: thm

thtuerk:
@xrchz I did not want to state that @marioxcc had anything to do with #588. I apologize for confusion. It was for me just a simple example of mistakes happening while approving pull requests. In pull request, a typo was approved that stopped HOL from getting compiled. This happens easily. Mistakes do happen. @marioxcc argues above that he has the right to add copyright notices, because noone complained about them being added in previous pull requests. I wanted to demonstrate on an as simple as possible example that this kind of argumentation leads to all kinds of problems. I wanted to give an example why it is not a valid argument for me.

xrchz:
@marioxcc would you be OK to move your code into a separate library, like `src/1/wlog.sml`?

thtuerk:
Moving wlog to a separate library and importing this lib into `bossLib` is a compromise I find OK. As written above, I will not press the quality and unicode issues for this new lib.

This leaves the main point: the license.

marioxcc:
@xrchz: Yes, that is fine for me. I moved it. Can you please merge or tell me if there is any issue remaining?

@thtuerk: I already made a compromise when I decided for the common good to make some of my modifications public domain and contribute it upstream instead of keeping it under the GNU GPL and in my personal repository or as a private patch. **I will not remove my license headers. _This decision is final and not open to further compromises;_ I already explained the reasons the headers are there.**  Although there is no legal obligation to keep my file headers, I will take issue if you unilaterally remove the existing ones because that would be arbitrarily overturning a decision by @mn200 . My headers have been there since #484 and from the dialog you can see clearly that Michael noticed them and choose to approve the changes. You mentioned an unrelated pull request with a typographic error that most probably _passed unnoticed_; this is unrelated and just a straw man fallacy.

konrad-slind:
I have been discussing this with Ramana. As the person who originally put
BasicProvers.sml together (since augmented by Michael and other) I have no
problem with adding wlog to that file.

Re: copyright and license issues. The current approach (do as little as
possible) has worked well enough. If we decide to pay more attention to IP
issues---and it's not clear we should---my main desire is that it be easy.
HOL4 is research software and we want to spend as much time as possible
doing fun research with it.

Konrad.

konrad-slind:
I am also fine with putting wlog in some other file. Not clear it needs to
be a library of its own.

Konrad.

thtuerk:
@konrad-slind In principle I have no issues with putting wlog in BasicProvers. I argued that it needs to be well readable in such a prominent place and follow common HOL naming and coding practices.

About the license. I would like to keep it as simple as possible as well. And exactly that is a problem of adding a separate license to parts of files. Even if this license is more permissive than the one used for HOL. I believe @jeremydaw summons it up very well in issue #559. Perhaps let's discuss it in issue #559.

marioxcc:
> About the license. I would like to keep it as simple as possible as well. And exactly that is a problem of adding a separate license to parts of files.

There is no more complexity. The user can simply ignore the fact that some code is in the public domain. The user already has to comply with the HOL4-wide license and having some strictly more permissive code does not alter that.

thtuerk:
@marioxcc let's discuss license issues on #559 please. I replied there.

marioxcc:
@xrchz: Thanks for your kind mediation.

Issue #561, opened 2018-07-03
Strange tactic behavior HOL-k10 vs HOL-k12


ahmedwaqar:
I observed a strange behavior of the same tactic in HOL-k10 vs HOL-k12

Consider a proof goal

``?!m. ((m = n) \/ m < n) /\ ((if m < n then f m else a) = x)``

e (CONV_TAC(ONCE_DEPTH_CONV COND_ELIM_CONV));

(* In HOL-k10, the tactic results a desired behavior*)

``?!m. ((m = n) \/ m < n) /\ (~(m < n) \/ (f m = x)) /\ (m < n \/ (a = x))``

(* In HOL-k12, the tactic results *) 

``?!m. ((m = n) \/ m < n) /\ (if m < n then $= (f m) else $= a) x``

barakeel:
After analysing HOL-k10 and HOL-k12 code I figured that the unwanted behaviour comes from the use of TB in the COND_ELIM_CONV which only applies a conversion if the term is a boolean. I don't see any reason to do that.

So I propose we use this code instead with TB removed.
 I needed to add PURE_REWRITE_CONV [] to rewrite
(x \/ F to x) and (x /\ T to x)

 val COND_ELIM_CONV =
   TOP_DEPTH_CONV
     (COND_RATOR_CONV ORELSEC
      (fn tm => if same_const (op_of_app tm) COND_t then failwith "fail"
                else COND_RAND_CONV tm) ORELSEC
      COND_ABS_CONV) THENC
   REPEATC ELIM_COND1 THENC
   PURE_REWRITE_CONV []

@ahmedwaqar I let you do the pull request to see if this change breaks any proofs.

ahmedwaqar:
Thanks! It worked perfectly. I also created the pull request. Please, let me know if you want me to do anything else.

xrchz:
I guess a regression test would be nice for this too, which you could put in `src/1/selftest.sml`

xrchz:
I committed a selftest for this but it still fails.

ahmedwaqar:
Now, the behavior is bit changed ....

 ``?!m. ((m = n) \/ g m n) /\ ((if g m n then f m else a) = x)``

 e (CONV_TAC(ONCE_DEPTH_CONV COND_ELIM_CONV));

``?!m.
       (g m n ==> ((m = n) \/ T) /\ (f m = x)) /\
       (~g m n ==> ((m = n) \/ F) /\ (a = x))``

[...]

PR #562, opened 2018-07-04
Strange tactic behavior HOL-k10 vs HOL-k12 (#561)-Solved

ahmedwaqar:
Modified COND_ELIM_CONV as suggested by T. Gauthier. This modification corrected the strange behavior in HOL-k12. From regression testing, all the proofs are building correctly locally. Please, merge this change. Thanks!

### DIFF
--- a/src/num/arith/src/Sub_and_cond.sml
+++ b/src/num/arith/src/Sub_and_cond.sml
@@ -270,12 +270,24 @@ val SUB_AND_COND_ELIM_CONV =
 (*                                                                           *)
 (*---------------------------------------------------------------------------*)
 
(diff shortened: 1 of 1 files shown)


mn200:
There was an efficiency related change to `Arith.sml` earlier in its history (normalisation for arithmetic terms shouldn’t be causing blowups on non-arithmetic terms) — please check that this change hasn’t undone this.

xrchz:
@mn200 do you have any relevant commit? The history of `Arith.sml` (even with --follow) doesn't show anything obvious.

mn200:
a5b09774c9070e is what I was thinking of.

Issue #559, opened 2018-06-30
licence of HOL


thtuerk:
Recently the subject of the license of HOL 4 and files distributed with it was brought up by @marioxcc several times. As I understand it, he believes the current license statement is not sufficient and a header in each file is needed (see #485). As a result, he placed copyright notices in `Conv.sml`, `res_quanLib.sml` and `res_quanScript.sml`. Currently the issue is becoming hot as part of pull request #550, which contains such a copyright notice for `BasicProvers.sml`.

I believe @marioxcc is wrong and the current practice of having a central copyright notice is sufficient. However, even if he is right, it is not his decision to change the copyright of HOL 4 and add copyright notices like the following to central files of HOL:

````
(* This file as a whole is assumed to be under the license in the file
"COPYRIGHT" in the HOL4 distribution (note added by Mario Castelán         UOK
Castro).

For the avoidance of legal uncertainty, I (Mario Castelán Castro) hereby   UOK
place my modifications to this file in the public domain per the Creative
Commons CC0 1.0 public domain dedication <https://creativecommons.org/publ
icdomain/zero/1.0/legalcode>. This should not be interpreted as a personal
endorsement of permissive (non-Copyleft) licenses. *)
````

If he is right (which I daub), we need to add a proper header to all files. I don't like multiple licenses
per file. If the license of HOL 4 is not permissive enough for Mario, I have no problem with him placing his own files in the example directory under a different license. However changing the license on parts of the core files in unacceptable for me. 

What to do about the license should in my opinion not be a decision by a single contributor, but be taken by the main developers foremost @mn200 and @konrad-slind and preferably after a discussion in the HOL 4 community. It have so far not heard of such a decision or discussion (please tell me if I missed something). That's why I start this issue.

I see the following options;

- we keep the copyright as it is
- we add a short copyright header to __all__ files

In any case, I would remove the copyright notices by @marioxcc. He should of course be listed as an author, but not put his own contributions to certain files under a different license. If this means removing this contributions, this would be regrettable, but for me a price worth paying.

I would appreciate, if the main authors @mn200, @konrad-slind and perhaps some of the very active HOL developers would comment on this issue and tell their opinion about it.

jeremydaw:
Hi,

I'm not at present an active developer, but here's my view.

(1) If you contribute to HOL, you accept that your contributions may be 
copied as per the general HOL COPYRIGHT notice

(2) If you contribute to HOL, you accept that your contributions may be
[...]

PR #563, opened 2018-07-04
Remove copyright notes wrongly touched by Mario Castelán Castro

binghe:
In case #559 concluded with the decision of removing copyright notes wrongly touched by Mario Castelán Castro, the current PR quickly gets the job done.

### DIFF
--- a/src/1/Conv.sml
+++ b/src/1/Conv.sml
@@ -13,17 +13,6 @@
 (* TRANSLATOR    : Konrad Slind, University of Calgary                   *)
 (* DATE          : September 11, 1991                                    *)
 (* Many micro-optimizations added, February 24, 1992, KLS                *)
-
-(* This file as a whole is assumed to be under the license in the file
-"COPYRIGHT" in the HOL4 distribution (note added by Mario Castelán         UOK
-Castro).
-
-For the avoidance of legal uncertainty, I (Mario Castelán Castro) hereby   UOK
-place my modifications to this file in the public domain per the Creative
-Commons CC0 1.0 public domain dedication <https://creativecommons.org/publ
-icdomain/zero/1.0/legalcode>. This should not be interpreted as a personal
-endorsement of permissive (non-Copyleft) licenses. *)
-
 (* ===================================================================== *)
 
 structure Conv :> Conv =
--- a/src/res_quan/src/res_quanLib.sig
+++ b/src/res_quan/src/res_quanLib.sig
@@ -5,19 +5,9 @@ DESCRIPTION   : Signature for restricted quantification inference rules
 AUTHOR        : P Curzon
 DATE          : May 1993
 UPDATED       : Joe Hurd, October 2001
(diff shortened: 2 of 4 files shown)


jeremydaw:
There is more reason to do this, beyond the points I made in the discussion on PR #559:
a copyright notice referring to some licence means that someone in future wanting to copy the file would need to look up that licence (unless it were made clear that the notice has the effect of only granting additional permissions).
In this particular case, well, I personally understand "public domain" to mean anyone can do anything they like with it, but that means I understand "public domain" to make any further reference to any licence quite redundant.  So I don't know what it is intended to mean.  (Maybe it would be clear what it means if I looked up the licence in question, but my point is that we want people to be able to use the file without having to do that).
And the discussion on PR #559 shows the importance of the precedent, of leaving these notices in or not.

marioxcc:
I will continue use my time to work on tactics and formalizing theories (for my personal repository and for contributing upstream) while @binghe and @jeremydaw continue use theirs for discuss among themselves the best way to thwart my contributions and making pull requests that merely remove comments.

thtuerk:
@marioxcc Repeating what has been said on #559: Your "comments" seriously harm HOL 4.

marioxcc:
@thtuerk: **My notices themselves make it clear that the license of both HOL4 as a whole and the files is unaffected.** You intentionally misinterpreted my words to claim a non-existent “harm” to HOL4. As long as you continue acting with malice in ignoring or twisting my words do not expect further responses from me on this non-issue. Again, I will be working on doing actual work with HOL4 while you discuss among yourselves how to thwart my work instead of doing an actual contribution.

mn200:
@thtuerk @marioxcc Please refrain from further discussion of this (here or on the other PR): I'm not sure any value is being added anymore.  I think it's clear that the historical HOL approach of basically ignoring the issue is not really tenable. (This latest manifestation is not the first time this has arisen.)  On the other hand, it's not yet clear what the best way forward is. 

FWLIW, my ideal solution would involve 
- credit being given to contributors; 
- a clear legal ownership story (allowing also the "no ownership" solution of public domain; note incidentally that if something is in the public domain, no-one can claim any copyright on it)
- as little legalistic boilerplate in our source files as possible

I do not know if my ideal solution is possible.

thtuerk:
@mn200 Agreed. Let's discuss it in issue #559. Perhaps you could copy your ideal solution there. I agree with the goals and for me the question is how to achieve these with as little effort as possible.

jeremydaw:
@mn200 " (allowing also the "no ownership" solution of public domain; note incidentally that if something is in the public domain, no-one can claim any copyright on it);"

"public domain" is not a good term to use when you're getting legally technical: in practice it seems to be used in two quite different ways (though the practical effect is similar)

(1) material which is not subject to copyright law, for whatever reason (eg, that it's too old)
(2) material for which the copyright owner allows it to be dealt with in any way, by anyone

PR #565, opened 2018-07-20
Add sort_vars to HolKernelDoc

AndreasLoow:
The function sort_vars was missing from HolKernelDoc, so the documentation available in help/Docfiles/HolKernel.sort_vars.doc was not visible when browsing the HTML documentation for HolKernel.

(Maybe HolKernel should have a signature file so it's easier to compare HolKernelDoc with the actual HolKernel?)


AndreasLoow:
Oh, I see. Maybe something like this would work then: AndreasLoow@2d221f9d8c8e488061a622e3284209e3db3de62b? I.e. moving all functions from HolKernel to a new structure HolKernelUtils (that can use the old HolKernelDoc as signature). There might be better names than HolKernelUtils however...

mn200:
What about moving `HolKernelDoc.sig` to be `HolKernel.sig`?  What breaks then?

mn200:
You're right that this is the issue.  The "code duplication" is unfortunate (if you update `Term` you will not see the new entry-point in most contexts without also updating the proposed `HolKernel.sig`), but I think I could cope with that.

Perhaps there is some funky solution involving functors...

PR #571, opened 2018-08-13
Moved some real_seq theorems from util_prob (and extreal) to seqTheory

binghe:
Hi,

the purpose of this PR is to move some useful results from `src/probability` to their best places.

I'm currently working on improving HOL's probability-related script by merging from upstream [1]. Currently HOL's seqTheory is missing some important results, for example, the fundamental theorem about the convergence of non-negative real sequences, saying "if all its partial sums have upper bound, then the infinite sum is summable" (and has the same upper bounded)

```
   [POS_SUMMABLE]  Theorem
      
      ⊢ ∀f. (∀n. 0 ≤ f n) ∧ (∃x. ∀n. sum (0,n) f ≤ x) ⇒ summable f

   [SUMMABLE_LE]  Theorem
      
      ⊢ ∀f x. summable f ∧ (∀n. sum (0,n) f ≤ x) ⇒ suminf f ≤ x
```

These results were proved in `util_probTheory` for theirs uses in `measureTheory`, but I think they should go to their correct places, so that users of `seqTheory` could benefit from them. (I also proved the extreal version of `POS_SUMMABLE`, it will be part of future updates in `extrealTheory`)

Another deep result about `sums` and `summable` is `SUMINF_2D`, whose proofs rely on a number of other lemmas:

```
   [SUMINF_2D]  Theorem
      
      ⊢ ∀f g h.
            (∀m n. 0 ≤ f m n) ∧ (∀n. f n sums g n) ∧ summable g ∧
            BIJ h 𝕌(:num) (𝕌(:num) × 𝕌(:num)) ⇒
            UNCURRY f ∘ h sums suminf g
```

The monotonicity of real sequence is captured by the definition `mono`, however, sometimes it's useful to discuss if the sequence is mono-increasing or descreasing, and there're theorems concerning about one side. The definitions `mono_increasing_def` and `mono_decreasing_def` were defined in `extrealTheory` with supporting theorems. Now I also move them back to `seqTheory`.

The following two elegant theorems could be further moved to `pred_setTheory`, however, their proofs need `RESQ_STRIP_TAC` (`res_quanTools`) in a way that I couldn't understand, so I have to leave them in `seqTheory`, as they're needed in the proof of `SUMINF_2D`:

```
   [NUM_2D_BIJ_BIG_SQUARE]  Theorem
      
      ⊢ ∀f N.
            BIJ f 𝕌(:num) (𝕌(:num) × 𝕌(:num)) ⇒
            ∃k. IMAGE f (count N) ⊆ count k × count k
   
   [NUM_2D_BIJ_SMALL_SQUARE]  Theorem
      
      ⊢ ∀f k.
            BIJ f 𝕌(:num) (𝕌(:num) × 𝕌(:num)) ⇒
            ∃N. count k × count k ⊆ IMAGE f (count N)
```

P. S. some special theorems on real numbers were moved from `extrealTheory` to `util_probTheory`. The main purpose is to make `extrealTheory` cleaner, beacuse my next move will be fixing `extrealTheory` with correct definition on `add`, `sub` and `inv`, but I can't submit them right now, as it will break the existing measure and probability theories in HOL4.

I've tested the changes, they're build-able, and didn't beak one major application in example folder (`example/miller`).

--Chun

[1] http://hvg.ece.concordia.ca/code/hol/DFT/index.php


binghe:
Let me cancel this PR for now, and submit another one once the whole work is done.

mn200:
> Let me cancel this PR for now, and submit another one once the whole work is done.

Looking forward to it!

PR #574, opened 2018-08-23
hol-light's topology theory and dependents (up to 2015)

binghe:
Hi,

currently HOL4 is quite weak on Topology (the topologyScript.sml has only 467 lines). This PR contains a rather complete elementary topology in Euclidean space ported from HOL-light (22,761 lines, 1760 theorems), up to the year 2015, by Muhammad Qasim and Osman Hasan [1].  I tested the proof scripts in latest HOL4, fixed most broken proofs (left two unfixed) and confirmed building in latest HOL (both std and expk), with Poly/ML and MosML. Some common lemmas were moved to early scripts.

It depends on hol-light's cardinal theory (called `cardTheory`) and others (iterateTheory and productTheory), all scripts were currently put into "src/probability".  They cannot go to `src/real/topologyScript.sml` because HOL4's `topologyTheory` is loaded too early (before `transcTheory` and `seqTheory`).

For the future work, the dependence on hol-light's `cardTheory` should be replaced by HOL4's `cardinalTheory` and then removed. Also, more theorems from latest hol-light (its `Multivariate/topology.ml' has 36780 lines and about 2400 theorems) could be ported into HOL4, but if the rest are not used by Probability, I leave this job to future contributors who are interested in topology (and real analysis) itself.

HOL4's Probability Theory scripts are not touched for now, but the new version of Probability Theory will depend on theorems here.

--Chun

[1] http://hvg.ece.concordia.ca/code/hol/DFT/index.php

(This PR contains all code changes in previous closed #571)


binghe:
sorry, building in mosml is not really tested. (I forgot to switch to current branch before making the builds ...)  There's currently a small issue in `wlogLib.sml` which blocks the mosml build, I've fixed it and am waiting for the building process. I'll commit once more.

binghe:
Fixed building in mosml.

mn200:
Thanks for this work!

PR #577, opened 2018-08-27
Merged HOL4's original topologyTheory with HOL-light's version

binghe:
Hi,

this PR is a further work of previous #574. I successfully merged HOL4's original topologyTheory with HOL-light's version.

HOL4's original `topologyTheory` mainly defined a "metric" topology (`mtop`) on R1 (1-dimension real line) to support `netsTheory` and `seqTheory`. The basic topology concepts, i.e. open and closed sets, are not used at all.  My idea is to replace them with HOL-light's version while keep all the rest concepts and theorems as is.

In HOL-light's "topology" definition, J. Harrison starts with the slightly more general notion of one set being open in another, then use this "localized" notion to further define the ‘global’ version [1]. In another words, the defnition of `istopology` doesn't require that `UNIV` (universal set) is also in `L`:
```
   [istopology]  Definition
      ⊢ ∀L.
            istopology L ⇔
            ∅ ∈ L ∧ (∀s t. s ∈ L ∧ t ∈ L ⇒ s ∩ t ∈ L) ∧
            ∀k. k ⊆ L ⇒ BIGUNION k ∈ L
```
then another "global" topology is defined with `UNIV` included:
```
   [open_DEF]  Definition      
      ⊢ ∀s. open s ⇔ open_in s 𝕌(:α)
```

As a consequence, the old HOL4 term "open" has become the new term "open_in".  This creates minor incompatibility issues as I can see, because the only place (outside of `topologyTheory`) in which the term "open" is used, is just one place in `netsTheory`, in the body of a proof.

As you can see, the beginning part of `toplogyScript.sml` is completely changed with definitions and theorems moved from `rich_topologyScripts.sml`, but the latter part (limit points, metric topology) is not touched at all, all theorems and concepts are still the same. As a result, HOL4's realTheory and related scripts were not touched at all (except for just one word in a proof in `netsTheory`). 

So I believe this is the right direction, now we have again just one version of "topology" type (based on open sets) in HOL4's official theory library, and HOL4's remaining original concepts in `topologyTheory` are now based on HOL-light's topology definitions!

A small number of common and interesting set-theoretic theorems (`FORALL_IN_*`, `EXISTS_IN_*` and others) were also moved from various places to `pred_setTheory`. Please let me know if they're appropriate, and I'm fine to put them back if you think they're not general enough.

Regards,

Chun Tian

[1] J. Harrison, A HOL Theory of Euclidean Space, TPHOLs. 3603 (2005) 114–129. doi:10.1007/11541868_8.


mn200:
All sounds great to me.  If the history isn't clear, incidentally, the original topology material was also due to John Harrison but is much older of course, and he has clearly improved it a deal since then. 

I'm also keen on the idea of having the abstract topology stuff move to somewhere other than `src/real` so that one can define other topologies without ending with up a dependency on the real numbers as well.

binghe:
Hi Michael,

thanks for quick merging and background information (I didn't know the original topology material was also John Harrison's work).

Speaking about abstract topology stuff, I think maybe we can do this:

1. Rename the current `topologyTheory` in `src/real` to `metricTheory`, as this is the actually part used by `netsTheory` and `seqTheory`. Also, in the definition of metric space (`ismet`), the `real` type is inevitable (zero distance and triangular property).

2. Move the abstract topology stuff (`topology` type, all theorems about open and closed sets) from renamed `metricTheory` to somewhere outside of `src/real`, as the new `topologyTheory`, in a new folder `src/topology`.  These abstract topology stuff depends only on pred_setTheory and combinatorial theories, so it can be built quite early in the build sequence.

3. Later, I (or others) could move more abstract topology stuff from `rich_topologyTheory` (currently in `src/probability`) to the new `topologyTheory`, to further enrich it.   And if someone ported more advanced topology stuff from hol-light, e.g. the "homologyTheory", it can be put in parallel with the new "topologyTheory" in the same folder, as long as they do not use real numbers.

What do you think about this plan?

--Chun

PR #578, opened 2018-08-28
Abstract (general) topology independent of realTheory

binghe:
Hi,

this PR is a further work of previous #577, in which Michael wishes to have "the abstract topology stuff move to somewhere other than src/real so that one can define other topologies without ending with up a dependency on the real numbers as well."

My idea is to move these abstract stuff (`topology` type, open and closed sets, neighbor, and "hull" from `rich_topologyTheory`) to `topology/topologyScript.sml` whose build sequence is just after `integerTheory` (mostly because `jrhUtils` is under `src/integer`), while the remaining stuff in `src/real/topologyScript.sml` now becomes `src/real/metricScript.sml`.  Then I have to update some real theories to make sure they open `metricTheory` instead.

I think this is the right direction, just I'm not sure if a new folder `src/topology` is too *heavy*, after all there's currently only one script inside it, and by design it cannot hold any further topologies depending on real numbers.  However, as I said in #577, this folder can be used to hold some Algebraic Topology stuff in the future, e.g. HOL-light's homology theory.

--Chun


mn200:
Brilliant!  Thank you.

Issue #238, opened 2015-02-24
Link to Definition.new_type_definition is missing from HTML index of symbolic identifiers


ptroja:
The link to documentation of `Definition.new_type_definition` is not included in (automatically generated) `help/src-sml/htmlsigs/idIndex.html`.

mn200:
Hmm.  This is because there is no longer a separate `Definition` module.  Rather, it's a sub-structure of `Theory`.   There is good documentation of `new_type_definition` though, and it's a shame to lose the link to it.  We'll have to think about that one.

PR #582, opened 2018-09-01
Hack to make Definition visible in the HTML documentation

AndreasLoow:
Workaround for issue #238

### DIFF
--- a/help/src-sml/Htmlsigs.sml
+++ b/help/src-sml/Htmlsigs.sml
@@ -179,7 +179,8 @@ fun processSig db version bgcolor HOLpath SRCFILES sigfile htmlfile =
            end
 
         val aliasStrName =
-           fn "FinalType" => "Type"
+           fn "DefinitionDoc" => "Definition"
+            | "FinalType" => "Type"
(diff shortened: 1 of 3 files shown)


AndreasLoow:
Hm, bossLib.new_specification is actually defined as pairLib.new_specification, but maybe the documentation still should say Theory.Definition.new_specification?

xrchz:
I think whatever documentation is given for the (presumably non-existent) `Theory.Definition.new_specification` should instead be given for `pairLib.new_specification`.

AndreasLoow:
I think saying something short about the relation between `Theory.Definition.new_specification`, `Theory.Definition.gen_new_specification` and `pairLib.new_specification` would be helpful. More concretely, if `pairLib.new_specification` is meant to supersede `Theory.Definition.new_specification`, then saying that as the documentation for the latter function would be helpful in my opinion (in combination with moving the documentation as suggested).

mn200:
I think this is probably an improvement on what we had, so I don't want to let the perfect be the enemy of the good.

PR #583, opened 2018-09-01
Make sure all HTML pages are interpreted as utf-8 encoded

AndreasLoow:
Yet another minor fix:

E.g. computeLib.CBV_CONV didn't display properly before this commit

Also use the "html5 doctype" in affected files


mn200:
Thanks!

PR #585, opened 2018-09-08
Merged hol-light's cardTheory (rich_cardinalTheory) with HOL4's cardinalTheory

binghe:
Hi,

hol-light's Cardinal Theory ([`Library/card.ml`](https://github.com/jrh13/hol-light/blob/master/Library/card.ml), now at `src/probability/cardScript.sml`) contains a overlapped  set of cardinal theorems as HOL's `cardinalTheory`, with some cardinal arithmetics on + (`add_c`) and × (`mul_c` or `CROSS`):

```
[add_c]  Definition
      ⊢ ∀s t. s + t = {INL x | x ∈ s} ∪ {INR y | y ∈ t}
[mul_c]  Theorem      
      ⊢ ∀s t. s × t = {(x,y) | x ∈ s ∧ y ∈ t}
```

and algebraic laws like the following:

```
   [CARD_ADD2_ABSORB_LE]  Theorem
      ⊢ ∀s t u. INFINITE u ∧ s ≼ u ∧ t ≼ u ⇒ s + t ≼ u
   
   [CARD_ADD2_ABSORB_LT]  Theorem
      ⊢ ∀s t u. INFINITE u ∧ s ≺ u ∧ t ≺ u ⇒ s + t ≺ u
   
   [CARD_ADD_ABSORB]  Theorem
      ⊢ ∀s t. INFINITE t ∧ s ≼ t ⇒ s + t ≈ t
   
   [CARD_ADD_ABSORB_LE]  Theorem
      ⊢ ∀s t. INFINITE t ∧ s ≼ t ⇒ s + t ≼ t
   
   [CARD_ADD_ASSOC]  Theorem
      ⊢ ∀s t u. s + (t + u) ≈ s + t + u
   
   [CARD_ADD_C]  Theorem
      ⊢ ∀s t. FINITE s ∧ FINITE t ⇒ (CARD (s + t) = CARD s + CARD t)

  [CARD_MUL2_ABSORB_LE]  Theorem
      ⊢ ∀s t u. INFINITE u ∧ s ≼ u ∧ t ≼ u ⇒ (s × t) ≼ u
   
   [CARD_MUL_ABSORB]  Theorem
      ⊢ ∀s t. INFINITE t ∧ s ≠ ∅ ∧ s ≼ t ⇒ (s × t) ≈ t
   
   [CARD_MUL_ABSORB_LE]  Theorem
      ⊢ ∀s t. INFINITE t ∧ s ≼ t ⇒ (s × t) ≼ t
   
   [CARD_MUL_ASSOC]  Theorem
      ⊢ ∀s t u. (s × (t × u)) ≈ ((s × t) × u)
```

Plus, it contains basic cardinality results about real numbers and sets:

```
   [CARD_EQ_REAL]  Theorem
      ⊢ 𝕌(:real) ≈ 𝕌(:num -> bool)
   
   [CARD_EQ_REAL_IMP_UNCOUNTABLE]  Theorem
      ⊢ ∀s. s ≈ 𝕌(:real) ⇒ ¬COUNTABLE s

   [UNCOUNTABLE_REAL]  Theorem      
      ⊢ ¬COUNTABLE 𝕌(:real)
```

but the whole theory was based on its own definitions on cardinality comparison operators and others, rendering the whole theory incompatible with HOL4's existing `cardinalTheory`.

In this PR, I've merged the two cardinal theories by forcing hol-light's version to overload cardinality operators defined HOL4's `cardinalTheory`. As a result, theorems in both theories now use the same set of operators, thus completely compatible.   My idea and principles are: 1) use `overload_on` to let `cardTheory` directly use operators from `cardinalTheory`, 2) Convert definitions in `cardTheory` into equivalent theorem, 3) Keep the old ASCII symbols in `cardTheory` and not touch any statement in existing theorems. Finally, I renamed `cardTheory` to `rich_cardinalTheory` indicating it is an extension of HOL4's `cardinalTheory`.

In old `cardinalTheory`, the `<</=` operator was overloading the one defined in `listTheory` with Unicode and TeX representations. But if user loads only `cardinalTheory` without loading `listTheory` at all, there will be no Unicode representation at all.  To solve this issue, I have to move the shared symbol representation and infix definitions to upstream, now at `pred_setTheory`.

Also, in `cardinalTheory`, the cardinal equality (`cardeq`) operator is overloaded with ASCII symbol `=~`. This is not a very good choice, because a term like `a=~b` (without spaces) could be supposed to mean `a = ~b` (`~b` is either a minus number or negated Boolean term). Putting `=~` into `pred_setTheory` has broken several proof scripts in such a way, and I have to fix them by putting some extra spaces, then I moved it back to `cardinalTheory` with my spaces fixes left in the changes. (I would choose "~=" instead, however, it is too late, as `cardinalTheory` is already in K12 release, I don't want break existing user code after K12 release.)

One theorem (`SUBSET_ANTISYM_EQ`) recently added into `pred_setTheory` now has a better name: `SUBSET_SUBSET_EQ`.

--Chun


xrchz:
I appreciate the careful work and detailed PR message!

mn200:
Which results have to be in a later "rich" theory, and which could just be pushed straight into `cardinalTheory`?  We have used the `rich_` prefix in a few places already, and I am willing to bear it if there is no better alternative, but I wonder if we couldn't just have one theory about cardinalities.  

If I understand your description correctly, you cannot pick up `rich_cardinal` without also picking up the reals (so that theorems about `𝕌(:real)` can be stated).  This doesn't seem ideal.  I'd prefer to have the real-specific stuff in a "real" theory, and all of the non-real stuff back in `cardinal`.  

Is this possible?

binghe:
Hi Michael,

good suggestion. I think it's totally possible to append hol-light's cardinal theorems into your `cardinalScript.sml`.  The only exceptions, as you expected, are those related to real sets, which are just a few. I don't think they deserve to go into dedicated new theory file under `src/real/`. Do you expect adding them into `realScript.sml` with your `cardinalTheory` depended? or somewhere else?

On those "rich_*" theories, I think `rich_topology` should be renamed to `real_topology`; the current `rich_cardinal` will disappear; the remaining `rich_listTheory`... I don't know, don't care. There's no others.

Also, we should keep in mind that, the purpose of this whole "merge" work are twofold:

1. Extend HOL4's existing `cardinalTheory`.
2. Provide a compatibility layer for porting more proof scripts (using its `Library/card.ml`) from hol-light.

Thus, even I succeed in merging almost everything into HOL4's `cardinalTheory`, I still want to keep all hol-light's cardinal theorem names as is, even some of them have same statements with existing theorems in `cardinalTheory`. (of course, duplicated proofs can be eliminated, by using `save_thm` instead of `store_thm`)

If you approve above plan, please hold this PR and I'll start working on it. (or please provide your further amendments)

mn200:
Totally happy with this plan, including the renames.  Thanks!

I would put the cardinality results that are real-specific into the script file that first uses them.  If they're just proved because they're nice results and aren't actually used in the material you're looking at, I'd put them in `real`.  (The reason to delay is just to keep the dependency graph comparatively uncontaminated until the last possible minute.)

binghe:
Great, the work is in progress now.

It turns out that the cardinality results of real sets are needed by `real_topologyTheory`, thus I'm moving all real-related results from the old `cardTheory` into new `real_topologyTheory`, with the rests merged into your `cardinalTheory`.

binghe:
Done, now `cardTheory` is removed with contents moved to `wellorderTheory` and `cardinalTheory` and the renamed `real_topologyTheory`.

Also, the term `wellfounded` defined in `wellorderTheory` is already defined in `prim_recTheory` but with different types (curried and uncurried relations). I try to resolve the conflicts by letting both term to overload `Wellfounded`. It seems good.

PR #593, opened 2018-09-21
Add Dockerfile

lorenzleutgeb:
This adds nothing to HOL itself. However, it allows those interested to build a Docker image around HOL easily. Apart from being hip one may also see a Dockerfile as a common way of expressing build instructions.

```
$ docker run --tty --interactive lorenzleutgeb/hol:latest
```

will spawn a HOL session.


mn200:
Thanks!

PR #595, opened 2018-10-04
Add a faster PERM conv, add a conv for ALL_DISTINCT.

talsewell:
Improves the performance of PERM_ELIM_DUPLICATES_CONV from O(n^2)
to O(nlogn) for large lists with many duplicates. This is done by
adding another phase that splits large problems with binary splits.

Also, add necessary helper lemmas about PERM to sortingTheory.

Adds a new conv ALL_DISTINCT_CONV which proves ALL_DISTINCT goals
by permuting to a strictly ordered list.

### DIFF
--- a/src/enumfset/comparisonScript.sml
+++ b/src/enumfset/comparisonScript.sml
@@ -279,6 +279,18 @@ val good_cmp_trans = Q.store_thm ("good_cmp_trans",
  fs [] >>
  metis_tac [cmp_thms]);
 
+val good_cmp_Less_trans = Q.store_thm ("good_cmp_Less_trans",
+`!cmp. good_cmp cmp ==> transitive (\k k'. cmp k k' = Less)`,
+ rw [relationTheory.transitive_def] >>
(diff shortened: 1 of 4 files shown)


mn200:
Can you fix the CI error please?  It's existing policy that Unicode shouldn't be used under `src`, which is checked for by the `testsrc` tool.  You can use the Unicode quotes rather than backticks though.

talsewell:
@mn200 Oops, sorry, didn't see this response, will fix up right away.

mn200:
Thanks for this!

talsewell:
No problem. Obviously I had a potential use for the ALL_DISTINCT mechanism, though I don't know if I'm going to end up following that up.

mn200:
I see from the code that you have some tests for the code; you might also move those into a `selftest.sml` script so that we get a regression check.  See the `Holmakefile` and the bottom of `selftest.sml` (use of `testutils.convtest`) in `src/list/src` for how you might set this up.

PR #596, opened 2018-10-09
cleaning example temporal_deep

thtuerk:
The example temporal_deep got severe bit-rot. @binghe wants to use this example. He started getting it to work again, I (@thtuerk) continued since I created the example more than 10 years ago. 

This pull request should make the example build with both the experimental and standard kernel again. Moreover, some concepts from standard libs are now used instead of specialised ones. For example a own definition of power-sets was replaced with the standard one. 

The proofs are ugly and hard to maintain. I would never write it in that style again. However, cleaning up all the proofs is a lot of work and this is for now just getting them to work again, was already plenty of work. 

Besides the proofs also the libraries got fixed. However, this excludes "model_checkLib", an interface to model checkers at the moment. This needs fixing. I currently don't have the time to do it. Hopefully I will be able to do it in the near future.


binghe:
Hi Thomas, thank you so much for fixing the "temporal_deep" example!!!  I'll take a look at "model_checkLib", to see if my previous work on integrating NuSMV model checker (#542) is also useful here.

thtuerk:
Great! Then I will leave fixing the model_checkLib to you. Let me know, if you want me to help.

mn200:
My cursory check through the GitHub web interface suggests you haven't enabled `temporal_deep` somewhere in the standard build sequence.  Please find it a home so that future bitrot can be avoided!  I'm sure you can judge what selftest level it should be at.  (If you have actually already put it somewhere, I am happy to merge straightaway.)

binghe:
I found that `model_checkLib` depends on `examples/muddy/bdd.{sml|sig}`, thus can only work with HOL built by Moscow ML. On the other side, it doesn't directly call the model checker, instead it calls API provided by `src/temporal/src/temporalLib.{sml|sig}` which is already fixed and is confirmed working with FBK's NuSMV model checker.

So I'm now building HOL in Therk's branch by Moscow ML and see if some sample test code work or now. In case it works, I'll move them into a separated `selftest.sml` file.  The latest HOL `master` is broken under Moscow ML due to a small issue in `src/postkernel/DefinitionDoc.sig`, I have a trivial fix for this.

I'll also enable `temporal_deep` in the standard build sequence by adding one line into `tools/sequences/more-theories`, just after the PSL examples.

So maybe it's better to hold this PR temporarily until I send another PR to @thtuerk 's current branch.

thtuerk:
Chun (@binghe) added it to the build sequence and worked on the model checking lib. I hope it is ready for merging now.

binghe:
Here is an example from the model checking lib:

```
val ltl1 = ``LTL_SUNTIL (LTL_PROP (P_PROP a), LTL_PROP (P_PROP b))``;

val ltl2 = ``LTL_EVENTUAL (LTL_AND (LTL_PROP (P_PROP b),
				    LTL_PNEXT (LTL_PALWAYS (LTL_PROP (P_PROP a)))))``;

val test1 = model_check___ltl_equivalent_initial ltl1 ltl2;
```

Since the two LTL formulas stored in `ltl1` and `ltl2` are equivalent (under initial semantics), the call of `model_check___ltl_equivalent_initial` returns a theorem (of type `:thm option`) indicating their equivalence:

```
- valOf test1;
> val it =
    [oracles: SMV] [axioms: ] []
    ⊢ LTL_EQUIVALENT_INITIAL
        (LTL_SUNTIL (LTL_PROP (P_PROP a),LTL_PROP (P_PROP b)))
        (LTL_EVENTUAL
           (LTL_AND
              (LTL_PROP (P_PROP b),
               LTL_PNEXT (LTL_PALWAYS (LTL_PROP (P_PROP a))))))
     : thm
```

Noticed that, the oracle is SMV for this theorem.

mn200:
Thanks for your work on this!

mn200:
One last comment: is there any prospect of merging the definitions of LTL in this example and that in `examples/logic/ltl`?

thtuerk:
I don't like my old example much. From what little I have seen, I would much prefer the stuff in `examples/logic/ltl`. I have not compared carefully, but my impression is that `temporal_deep` has more on automata, and stuff like CTL and the connection to PSL. However, it is not anything interesting in my opinion. Moreover, I believe the algorithm for translating ltl to omega automata in `examples/logic/ltl` is superior to the one in `examples/temporal_deep`. Moreover, the proofs in `examples/temporal_deep` are awkward and the design crude. Without looking too deeply, my feeling was that `examples/temporal_deep` it is redundant and could be removed.

@mn200 What exactly did you mean with merging?

@binghe You compared several LTL formalisations you said. To my surprise you wanted to use `examples/temporal_deep`. Can you explain, which parts might be worth merging into `examples/logic/ltl`?

mn200:
Both examples need and use a basic definition of LTL syntax and semantics.  I'm happy for both collections of material on top of that to be kept, but it seems a shame to duplicate the foundations.  So, by *“merging”* I just mean to have one canonical base LTL theory for subsequent developments to build on top of.

Unfortunately, there are still some aesthetic decisions to be made about this.  I know that in `examples/logic/ltl` there is a core set of operators chosen, and that others are made to be derived.  There are always choices to be made: if anyone cares enough about the material, they’re probably the right person to make those choices.

thtuerk:
@mn200 The basic definitions are in both examples tiny compared to the size of the whole development (very roughly a couple of dozen lines vs 20 000 or 30 000 lines). The definition of LTL is straightforward. However, `temporal_deep` has past operators in its LTL while `tools/ltl` does not. This is a big difference. While the expressiveness is the same, some properties can be expressed exponentially more succinct in the presence of past operators. So, I don't think it is worth merging these basic definitions. One could provide a translation between the future fragment of the logic in `temporal_deep` and the one in `logic/ltl`. This should be very simple and just a few lines of code. I could add it to the `temporal_deep` translations, if you (@mn200) wants it.

binghe:
@thtuerk 's work has provided formalizations of LTL with past operators, CTL* and a deep embedding of propositional logic formulae (plus the "next" version, `xprop_logicTheory`). More importantly, it has provided a complete formalization of Symbolic Kripke Structure and automata.  The formalization of (explicit-state) alternating automata is based on an universal datatype (`alternating_automaton`) plus various acceptance conditions, and I think this choice saved a lot of proof duplications comparing with the work in `examples/logic/ltl`:

```
val alternating_acceptance_condition_def =
Hol_datatype
`alternating_acceptance_condition =
    TRUE |
    FALSE |
    INITIAL of 'state set |
    CO_INITIAL of 'state set |
    BUECHI of 'state set |
    CO_BUECHI of 'state set |
    PARITY of 'state->num |
    WEAK_PARITY of 'state->num |
    WEAK_BUECHI of 'state set |
    WEAK_CO_BUECHI of 'state set`;
```

My work in Runtime Verification is based on Symbolic Kripke Structure and "LTL with past operators,"  thus @thtuerk 's work would be my only choice to base on.

binghe:
> @binghe You compared several LTL formalisations you said. To my surprise you wanted to use `examples/temporal_deep`. Can you explain, which parts might be worth merging into `examples/logic/ltl`?

for this question, I would say it's much easier if we merge `examples/logic/ltl` into your `temporal_deep` framework, because all one need to do is to go ahead to formalize the LTL-to-VWAA (very weak alternating automata) translation algorithm, because all needed supporting data structures are already formalized in `temporal_deep`.

I did have compared all three LTL formalisations in HOL [1,2,3] and another one in Isabelle/HOL AFP [4, which depends on 5].   [1] is a deep embedding, it's very hard to work with it since there's no explicit semantics definition, so I gave up after spending many efforts.  All others are shallow embeddings, with datatype definitions representing LTL formulae.  But as I told in preceding comment, I need a formalization from LTL to symbolic transition system (i.e. Kripke structure), thus the only choice is [2], but I need to point out that, [2] also provides explicit-state automata structures, thus I believe [3] can be re-done easily in framework of [2].   The one in Isabelle/HOL doesn't have any new features comparing with the three formalizations in HOL4.

[1] https://github.com/HOL-Theorem-Prover/HOL/tree/master/src/temporal
[2] https://github.com/HOL-Theorem-Prover/HOL/tree/master/examples/temporal_deep
[3] https://github.com/HOL-Theorem-Prover/HOL/tree/master/examples/logic/ltl
[4] https://www.isa-afp.org/entries/LTL_to_GBA.html
[5] https://www.isa-afp.org/entries/LTL.html

mn200:
Thanks for the discussion on this.  I will leave it to you guys to figure out what to do.  

PS: @binghe you have the deep / shallow terminology the 'wrong' way round above.  If you define a type to represent the syntax, and then an explicit semantics function to interpret that syntax, you are performing a deep embedding.  (This would be why the `temporal_deep` example has that name.)

binghe:
sorry ... sometimes I think one word and write the other:)

binghe:
There's one more issue: on Mac (case-insensitive file system), the name of `src/coalgebras/pathScript.sml` and `examples/PSL/path/PathScript.sml` are conflicting. This is not an issue for building any theory file if the relevant `Holmakefile`s put `${HOLDIR}/examples/PSL/path` into `PRE_INCLUDES` variable, however, the following `examples/temporal_deep/src/model_check/selftest.exe` cannot build correctly:

```
selftest.exe: selftest.uo modelCheckLib.uo
	$(HOLMOSMLC) -o $@ $<
```

even when I put `-I ${PRE_INCLUDES}` into the command line, simply because `$(HOLMOSMLC)` always search theory files from `${HOLDIR}/sigobj` first.  How can we resolve this issue? Change `examples/PSL/path/PathScript.sml` to another name?

mn200:
I think perhaps the PSL path theory should have its name changed.  I think `PSL_path` is probably a good name for it.

thtuerk:
Renaming the PSL Path Theory sounds like a good suggestion.

A similar problem exists for 'ltlTheory'. One exists in 'temporal_deep' and one in 'logic/ltl'. I think the one in 'temporal_deep' should be renamed to 'pltlTheory' to reflect that it has past temporal operators in.

I will prepare a pull request renaming the path and ltl theories.

binghe:
I like the new name `pltlTheory`, but I would rather interpret it as "Propositional LTL" (to distinguish with "Predicate LTL" or other LTL variants with quantifiers).

There exists LTL variants with ONLY past operators, and they are called something very close to "Past LTL". On the other side, LTL with both past and future operators were usually called "Full LTL" or "LTL with Past Operators" in recent literatures.

thtuerk:
@binghe OK, I used `full_ltl` now

PR #598, opened 2018-10-11
Simplify proofs in lift_ieeeScript.sml by removing is_normal assumption

HeikoBecker:


mn200:
@acjf3 Any comment?  I'm generally a bit loth to change theorem names.  Could/should we keep the old names as well, even if the content has changed?

HeikoBecker:
@mn200 Sorry, I forgot to add the information: I added these theorems a year ago and did not change any of the work done by @acjf3. I am however fine with changing the names back to the old ones I used back then,

mn200:
OK, let's go with the PR as it stands.

PR #602, opened 2018-10-15
rename temporal theories

thtuerk:
in order to avoid name clashes between theories rename

/examples/PSL/path/Path -> PSLPath
/examples/PSL/path/FinitePath -> FinitePSLPath
/examples/temporal_deep/src/deep_embeddings/ltl -> full_ltl

This change has been discussed as part of the ongoing discussion of pull request #596


mn200:
Thanks for your work on this!

PR #603, opened 2018-10-15
fix bit-rot in pattern_match_demo

thtuerk:


binghe:
I guess @mn200 will suggest putting also this example into a build sequence ...

mn200:
Absolutely.  If it's not regression-tested, it is the sort of thing that will ultimately get culled.

I'd suggest merging into the existing self-test for `src/pattern_matches`.  If it needs to be in a theory file (or if it is particularly slow), put it into a new `src/pattern_matches/theory_tests` directory.

thtuerk:
I fixed a very minor problem because I spotted it coincidentally. I'm not willing to put in the time and effort to restructure the whole thing. If a fix is only acceptable, if it is redesigned, please just reject this pull request. By the way, the file is not related to `src/pattern_matches`.

mn200:
Thanks for this.  I will probably merge it into `src/tfl/src/test/pattern_matchingScript.sml` now that I better understand what it's doing.

PR #605, opened 2018-10-17
example ltl-translations added

thtuerk:
This example relates the 3 existing ltl formalisations. This requered adding some minor lemmata to the formalisations.


mn200:
Could you move it be under examples/logic please?  Maybe even examples/logic/ltl.  Ideally, temporal_deep should be in the same sub-directory, but we can think about that as a separate issue.

thtuerk:
I moved it now.

binghe:
> Ideally, temporal_deep should be in the same sub-directory, but we can think about that as a separate issue.

I ... think it's not suitable to move complex examples (old, large, or with multiple sub-directories) into `examples/logic`, even they're kind of "logic".  Examples are: `examples/PSL`, `examples/temporal_deep`, `examples/separationLogic`, and `examples/decidable_separationLogic`, etc.

New logic formalization additions in the future, especially when all files are in a flat directory, should be put into `examples/logic`, to minimize the number of directories under top-level of the `examples` directory.

mn200:
I agree that `ltl-translations` is not ideal.  

> I ... think it's not suitable to move complex examples (old, large, or with multiple sub-directories) into examples/logic, even they're kind of "logic". Examples are: examples/PSL, examples/temporal_deep, examples/separationLogic, and examples/decidable_separationLogic, etc.

Why?

konrad-slind:
I think examples/logic is not a bad place for meta-theory of various logics
to live

Instead of ltl-translations, how about ltl-transformations?

Konrad.

binghe:
> Why?

well, I just thought the directory tree may be too deep ... but never mind.

binghe:
> I'm sorry, but can we think a better name other than `ltl-translations`? The current name looks like a formal theory of translating LTL to something else (e.g. omega automata) but actually it's just a "translation" of LTL terms between different LTL definitions current available in HOL...

I retreat my above (inappropriate) comments, as it seems to have blocked the merging process of this PR. Now I suggest the PR should be merged as is.

mn200:
Thanks for your work on this!

PR #612, opened 2018-11-09
A rudimentary RL environment for proving in HOL

xrchz:
(I thought this had been added to `master` already but apparently it's still just on this branch.)


mn200:
It needs inclusion in a build sequence file at a suitable test level and a self-test mechanism to ensure it doesn't bit-rot.

PR #619, opened 2018-11-29
Cumulative fixes for HOL4 building in Moscow ML

binghe:
Hi, this trivial patch fixes all issues raised when building latest HOL4 in Moscow ML.

### DIFF
--- a/src/TeX/holindex.sml
+++ b/src/TeX/holindex.sml
@@ -190,7 +190,7 @@ let
    val _ = if (isSome content) then () else Feedback.fail();
    val empty_posn = (0,0);
    val opts = mungeTools.parseOpts empty_posn ("alltt,"^options);
-   val _ = if command = mungeTools.Theorem then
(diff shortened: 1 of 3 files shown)


mn200:
I think my recent changes mean that the patches to do with the `Theorem` “keyword” aren't required anymore. I think it is just the dependency information in `src/metis` that needs fixing.

binghe:
What recent changes? I was building on 8649983c4924097c8a62e531247200c9bf24067c (the latest commit on `master` at this moment), and without those changes related to `Theorem` the mosml build was still broken...

mn200:
Changes I am seconds away from pushing ...

binghe:
Very good, then this PR is not needed any more.

binghe:
Sorry, I didn't realized that you still need the changes in `src/metis/Holmakefile` (because I just found the mosml build after your these changes is still broken at that part), but I guess you can make the changes by yourself.

PR #624, opened 2018-12-04
Reconcile store_thms with diverging names in db and val binding

IlmariReissumies:
In order to avoid breaking too much, I have tried to be conservative with changing db names in `src/` even when they are somewhat strange, with the exception of `SEG_FIRSTN_BUTFISTN` which is just bizarre.


mn200:
Thanks!

PR #625, opened 2018-12-06
Boyer-Moore Substring Search Algorithm

JamesShaker:
Added Boyer Moore substring search algorithm implementation (on lists and strings) to examples


mn200:
You need to update a build sequence file to make sure this example is built and tested as part of our regression test machinery.  Put it into one of the files in `tools/sequences` and put an appropriate number of `!` characters in front of the directory name.  (One if the directory builds in less than a few minutes, two for longer times, and three for super-long times.)

You might also add a `selftest.sml` file that runs and shows that the algorithm is doing the right thing.

PR #637, opened 2019-01-08
Local Program Variables?

immler:
I know that dealing with references and synchronization is a very subtle business, so this pull request is mostly meant as a starting point for discussion.

Some of the refs in HOL4 track global state, whereas some are private program variables.
The refs affected in this commit are the ones that are heavily allocated while building HOL and looked (upon superficial inspection) reasonably local

Would it make sense to mark them explicitly as private program variables?

Best regards,
Fabian

### DIFF
--- a/src/0/Term.sml
+++ b/src/0/Term.sml
@@ -717,6 +717,7 @@ fun strip_binder opt =
                                     end handle HOL_ERR _ => NONE)
  in fn tm =>
    let
+     open Unsynchronized
      val (prefixl,body) = peel f tm []
      val AV = ref (Redblackmap.mkDict String.compare) : ((string,occtype)Redblackmap.dict) ref
      fun peekInsert (key,data) =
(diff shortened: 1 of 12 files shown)


mn200:
It's not clear to me why in some circumstances you've used qualified names, and in some cases you've used `open`.  I'd prefer the latter in more situations.

mn200:
I believe the GitHub UI allows you to adjust the target branch for PRs (I can't see it because it's not my PR).  If so, can you target `develop` rather than `master` please?  (If not, I can manually merge on the command-line locally and push that.)

immler:
Okay, I managed to target `develop` with the GitHub UI.

mn200:
Thanks - will wait for Travis to be done.

mn200:
I realise that we should have an `Unsynchronized` module for Moscow ML.

immler:
I see...

It seems that in mosml, something like
`structure Unsynchronized = struct datatype foo = datatype ref end`
does not export the constructor `ref`.
I.e., `Unsynchronized.ref` yields an `Unbound value component: Unsynchronized.ref`.

I therefore introduced a new signature and structure Uref that provides its own constructor.

PR #643, opened 2019-01-22
 Remove eqtype declaration for term type

immler:
@Jonas-R and I started to work on something like this branch when, to our luck, we found out it already existed.

Have you ever considered merging this into develop?
Would there be any problems with that?


mn200:
It's tempting, though also quite a radical change.  Our experience with this branch was fairly positive so I think it's doable.  I will consult on the developers' mailing list.

PR #645, opened 2019-01-30
Fixed hol-unicode-replacements (hol-mode.el) for UNIV, added "o"; ...

binghe:
Hi,

currently in hol-mode.el, the feature "Replace common HOL Unicode symbols" replaces universal set symbol 𝕌 into `UNIV`, but this is wrong, it should be `univ`, because 𝕌 always appear with the underlying set element type, e.g. 𝕌(:α) or 𝕌(:num), with the corresponding ASCII term `univ(:'a)` or `univ(:num)`. There should be no space after `univ` before the type part.

Beside, the functional composition symbol ∘ isn't handled. I added this one with replacement `o`.

Also, I don't understand why the subtraction symbol `-` has also a Unicode char `UTF8.chr 0x2212` (which looks almost the same as `-`). This was done in commit eb606344 but the commit message looks irrelevant.  In my proof scripts, sometimes I copied some term from HOL's output back to script file, then all the subtraction symbols being copied are the Unicode version, and of course it's not handled by "Replace common HOL Unicode symbols". I was trying hard to figure out those `UTF8.chr 0x2212` chars.   If there's no particular reason, I hope it can be removed as the current PR does.

--Chun

### DIFF
--- a/src/num/theories/arithmeticScript.sml
+++ b/src/num/theories/arithmeticScript.sml
@@ -105,8 +105,7 @@ val _ = ot0 "BIT2" "bit2"
 val _ = add_numeral_form (#"n", NONE);
 
 val _ = set_fixity "-" (Infixl 500);
-val _ = Unicode.unicode_version {u = UTF8.chr 0x2212, tmnm = "-"};
-val _ = TeX_notation {hol = UTF8.chr 0x2212, TeX = ("\\ensuremath{-}", 1)}
+val _ = TeX_notation {hol = "-", TeX = ("\\ensuremath{-}", 1)}
 
 val SUB = new_recursive_definition
    {name = "SUB",
--- a/tools/hol-mode.src
+++ b/tools/hol-mode.src
@@ -1595,8 +1595,8 @@ second is to show the new term.
   (t "⊆ᵣ" "RSUBSET")
   (t "∪ᵣ" "RUNION")
   (t "∩ᵣ" "RINTER")
-  (t "∅" "EMPTY")
-  (t "𝕌" "UNIV")
+  (t "∅" "{}")
+  (nil "𝕌" "univ")
   (t "⊆" "SUBSET")
   (t "∪" "UNION")
   (t "∩" "INTER")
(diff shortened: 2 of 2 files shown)


mn200:
I think the theory is that the unicode subtraction symbol will look better than a simple hyphen in some fonts. But I agree that the effect is so minor (at least with the fonts I use) that there's no real point to doing this, and it does just cause confusion. 

Note, incidentally, that it's fine to input `univ(:'a)` with a space after the `univ` if you desire. 

Thanks for the fixes!

binghe:
Thanks. I know that it's allowed to have a space after `univ`, but to eliminate that space, I had to set the first list element in your replacement table to `nil`.

binghe:
Hi Michael,

I just found a related issue in LaTeX manual generation:

In `Manual/Description/libraries.stex`, line 3131, we have the following code

```
>> EVAL ``fact 4``;
```

Its corresponding expansion in `Manual/Description/libraries.tex` contains some Unicode version of "minus" symbol:

```
fact 4 = if 4 = 0 then 1 else 4 * fact (4 − 1)

fact 3 = if 3 = 0 then 1 else 3 * fact (3 − 1)

fact 2 = if 2 = 0 then 1 else 2 * fact (2 − 1)

fact 1 = if 1 = 0 then 1 else 1 * fact (1 − 1)

fact 0 = if 0 = 0 then 1 else 0 * fact (0 − 1)
```

And these Unicode chars has stopped TeXlive 2018 for building HOL Description.

From `Manual/Description/Holmakefile` I see you have used a ML-based tool `../Tools/polyscripter ../Tools/umap` to convert all Unicode chars in `*.stex` files to its ASCII version, but this conversion seems happening before the expansion of the embedded HOL sessions in `*.stex` files. As a result, it doesn't convert above "minus" (−) symbos at all.

I don't know how to fix this.

--Chun

mn200:
Fixed in e057d7623

mn200:
The issue was the code in `computeLib`, which was not emitting its output through the streams controlled by the `Feedback` module.  Now that it does so, the `polyscripter` “magic” can capture it appropriately.

binghe:
Thanks for the fixes and the explanation.

PR #648, opened 2019-01-31
Add toSortedAList to sptree.

talsewell:
Adds an O(n) traversal of sptree's that extracts an
association list sorted in index order. The absence
of such a traversal was noted as a major drawback
in the existing documentation.

The algorithm is adapted from the one for Braun trees
(sptree's are Braun trees with gaps/partiality added).
It won't handle very sparse arrays well, that is, it's
O(n) in the largest index, not the number of elements.


mn200:
Please fix the CI failure (from theorem `sum_spt_size_less`).  It stems from the fact that the `save_thm` Theorem syntax has changed.  Instead of
```
   Theorem foo (expression)
```
you have to write
```
   Theorem foo = expression
```

mn200:
Thanks!

PR #650, opened 2019-02-03
Fixed polyscripter.sml according to API changes of

binghe:
.. of `eReadHMF.extend_patch_with_includes`; Fixed one Unicode char in Docfiles.


mn200:
Yes please.

PR #656, opened 2019-02-11
Fix examples/machine-code/lisp for remove-term-eqtype

Jonas-R:


mn200:
I think I fixed the same files myself already: sorry to have caused needless work.  I will check to see that what I pushed does indeed build, and if so close this.

mn200:
I have merged my version of your changes into `develop`.

Thanks for looking at this!

PR #658, opened 2019-02-14
remove-term-eqtype in EquivType

immler:
Apparently bin/build does not complain when "Building directory src/datatype/equiv", but I do not understand why. I believe the attached change should be necessary...

Is there something else going on?


mn200:
Thanks for spotting these.  The issue is that under Poly/ML, the build process only actually compiles code if it is used.  EquivType is old technology that the distribution no longer uses (the machinery in the quotient directory is better).  

I assume the same problem applies to the other files you had to touch.  I will add selftest files to check that this code actually does compile.  With luck, the same selftests can check for functionality...

PR #662, opened 2019-02-19
Removed duplicate theorem unibag_IN

lxndrcx:


mn200:
This is fine in itself, but it'd be even better if you instead proved:

   BAG_IN e (unibag b) <=> BAG_IN e b

and made that the theorem called `BAG_IN_unibag`.  And made it a simp.

Also, `unibag_FINITE` should be oriented the other way, and should also be a simp.

lxndrcx:
Done 👍

PR #667, opened 2019-02-25
Add IntuitionisticProof to logic examples.

lxndrcx:
This is the project I have been working on over the summer. The intuitionistic version is complete, so here it is as an example. Feedback welcome.


mn200:
The Travis error is my fault and the commit looks basically fine.  However, I'm wondering if you can link your mechanisation to the example directory's existing PropLogic script.  Actually maybe you can't. You can't define the intuitionistic connectives in terms of implication and negation (see https://www.ps.uni-saarland.de/~dang/ri-lab/indep/independence.pdf), so if anything the stuff in PropLogic should be explicitly labelled as classical.  (Or we should create classical and intuitionistic directories to keep things apart.)

PR #668, opened 2019-02-27
Corrected a typo thence --> hence in Description/tactics.stex

ahmedwaqar:


mn200:
Actually, I think “thence” may be correct. The intended meaning is that the theory is the place from which those lemmas can be taken, and “thence” is thus OK.  To my mind, “hence” would be to imply a reason or cause.  But, to be honest, I think avoiding both words would be preferable.  They are both borderline archaic and I think we should try to use simpler phrasings where they are possible.

ahmedwaqar:
Thanks for clarifying the confusion. I think there is no need to merge this change now. I'm going to close this PR.

PR #671, opened 2019-03-02
Add Hol_list_reln, which defines a relation from a list of quotations.

lxndrcx:
When defining relations which build upon each other, at the moment a lot of redundancy is introduced. With this function, you make a list of quotations, each of which is a rule for the relation, but with a universally quantified relation name. You call the function with a new name which is used in place of the old one, plus the list of rules, and get a relation just like when using Hol_reln.

You could write, for example:
```sml
val base_rule_list = [`!R. <rule1 def>`,`!R. <rule2 def>`]
val ext_rule_list = `!R. <rule3 def>` :: base_rules

val (base_rules, base_ind, base_cases) = Hol_list_reln "base" base_rule_list
val (ext_rules, ext_ind, ext_cases) = Hol_list_reln "ext" ext_rule_list
```

I would have found this useful when I was defining proof calculi for different logics, i.e., minimal logic, intuitionistic logic, classical logic, which build on each other for some calculi like natural deduction.

Let me know what you think, and if you think it would be useful for anyone other than me.

### DIFF
--- a/src/IndDef/IndDefLib.sig
+++ b/src/IndDef/IndDefLib.sig
@@ -10,6 +10,7 @@ sig
 
   val Hol_reln      : term quotation -> thm * thm * thm
   val xHol_reln     : string -> term quotation -> thm * thm * thm
+  val Hol_list_reln : string -> term quotation list -> thm * thm * thm
   val Hol_mono_reln : string -> monoset ->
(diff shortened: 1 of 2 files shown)


lxndrcx:
This isn't working as expected. Don't merge.

PR #675, opened 2019-03-11
Add type unification and mk_ucomb to boolSyntax

hrutvik:
Expose four new functions in `boolSyntax`:
 - `type_unify : hol_type -> hol_type -> (hol_type,hol_type) subst` - classical type unification.
 - `sep_type_unify : hol_type -> hol_type -> (hol_type,hol_type) subst * (hol_type,hol_type) subst` - type "unification" which returns two separate substitutions (one for each argument), avoiding problems with occurs checks and so on. In other words, gives separate substitutions for each argument, to "unify" them to some common type - particularly useful when the types have type variables in common.
 - `mk_ucomb : term * term -> term` - a version of `mk_comb` which first uses `sep_type_unify` to allow type instantiation of both function and argument types, reducing the need for manual type instantiations.
 - `list_mk_ucomb : term * term list -> term` - folds `mk_ucomb` over a list of arguments.

Help docfiles are included for these four new functions. The functions `type_unify` and `sep_type_unify` are adapted from `examples/miller`.

### DIFF
--- a/help/Docfiles/boolSyntax.list_mk_ucomb.doc
+++ b/help/Docfiles/boolSyntax.list_mk_ucomb.doc
@@ -1,0 +1,28 @@
+\DOC list_mk_ucomb
+
+\TYPE {list_mk_ucomb : term * term list -> term}
+
+\SYNOPSIS
+Folds {mk_ucomb} over a series of arguments.
+
+\DESCRIBE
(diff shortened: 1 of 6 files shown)


mn200:
Thanks for your work on this!

PR #676, opened 2019-03-14
Add RISC-V support to graph decompiler

myreen:
It is worth noting that this PR includes the gcc output for the seL4 kernel when compiling to RISC-V with the RVC extension. Since seL4 is under a GPLv2 licence, I have been instructed by Matthew Brecknell and Gerwin Klein to add a line to `COPYING` pointing to text in `examples/machine-code/graph/seL4-kernel/README`. This new `README` file describes the licensing of the files under `examples/machine-code/graph/seL4-kernel`.


mn200:
How is it tested?  It can't be merged without an automatically running regression test as part of a standard build sequence.  (I can see a theory called _something_`_testScript.sml`, but I don't see a change to `tools/sequences`...)

myreen:
@mn200, the new files are tested on a regression test, since I added the following to the `INCLUDES` in `examples/machine-code/Holmakefile`:

           $(HOLDIR)/examples/machine-code/graph

That was part of commit 9b3661d40324002521683ff67e6529add6c1ce03.

My understanding is that a level 3 regression test runs `Holmake` in `$(HOLDIR)/examples/machine-code` due to the following line in `tools/sequences/final-examples`:

    [poly]!!!examples/machine-code

mn200:
Ok - thanks for that.

PR #696, opened 2019-04-23
Add lookup_union to balanced_mapTheory

AndreasLoow:
The obvious thm for lookup + union in balanced_mapTheory was missing for some reason. (There are thms such as lookup_insert etc. already included in the theory, and this new thm fits this family of thms.)


mn200:
Thanks for this!

PR #697, opened 2019-04-26
-all-p -> every

AndreasLoow:
Seems that -all-p is not from the packages required by hol-mode.


mn200:
Thanks for this.  I didn’t check that my own environment was not contaminating my development of this code!

PR #698, opened 2019-05-02
Move some theorems from CakeML

xrchz:


mn200:
Content up to 989b1c4 looks great.

mn200:
Thanks for this!

PR #702, opened 2019-05-12
Reduce diff between otknl-thm and std-thm

xrchz:
Also set kernelid to otknl (rather than logknl) since that is the name
used everywhere else.


mn200:
Thanks for your work on this.  I am sorry I've put off merging it in for so long!

Issue #693, opened 2019-04-17
Document/suggest methods for Unicode input


mn200:
The manuals are rather quiet on the issue of how users might input Unicode characters.  This varies with editor mode of course, but we could describe those we know about.

mn200:
This is mostly done now in the Interaction guide (for the emacs mode), as of 9a5b66138b2

PR #705, opened 2019-05-21
Add an alternative input method for emacs

minchaowu:
Usage:
Add the following two lines to `.emacs`.
```
(load "~/HOL/tools/hol-input.el")

(add-hook 'sml-mode-hook
	  (lambda () (set-input-method "Hol")))
```

#693. This is the input method for Emacs used by both the Agda and the Lean theorem prover.  Existing HOL4 users might prefer the current Unicode input defined in `tools/hol-unicode.el`, but new comers might take quite some time to remember all the key bindings.  This PR provides a tex-like syntax for typing Unicode symbols. Emacs users can choose between this one and the existing method.

Examples: `\forall`  -> ∀, `\le` -> ≤, ...

### DIFF
--- a/tools/hol-input.el
+++ b/tools/hol-input.el
@@ -1,0 +1,1141 @@
+;;; hol-input.el --- The Hol input method (based/copied from Lean)
+;;;
+;;; DISCLAIMER: This file is based on lean-input.el provided with the Lean theorem prover.
+;;; We did minor modifications
+;;
+;;; Commentary:
+;;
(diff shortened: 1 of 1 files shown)


mn200:
Thanks -- this looks very nice!

mn200:
However, please change so that the PR targets the `develop` branch.  Merges into master only happen once `develop` has been properly tested.

minchaowu:
Done!

PR #707, opened 2019-05-22
Holmake: remove `/bin`

qoelet:


mn200:
- There is no need to use `env`; the Holmake variables you are changing are expanded in lines that are passed to a shell for execution, so `$PATH` will naturally be consulted if bare executable names are used (*e.g.*, `cp`). 

- Please target the `develop` branch with your PR

- Please convince me that this is a good idea.

qoelet:
Build was failing for me with the complaint that `/bin/cp` not found. I'm not familiar with the internals of Holmake, this was more of me following the error messages and finding this. If `$PATH` is consulted then why is the `/bin` required?

If it helps I'm on NixOS.

mn200:
I take it NixOS doesn't have `cp` in `/bin/cp`?  Where is it? `/usr/bin`?

The reason to use an absolute path is to prevent weird behaviours where users have put a `cp` in their PATH that hides the "actual" `cp` that Holmake wants to use.  I guess your argument is that we should trust the user and run that `cp` anyway. 

Alternatively, I can change the code so that it looks in "reasonable" places for a `cp` program.

qoelet:
Yes, `cp` is in some other absolute path in my case (i.e., `/run/current-system/sw/bin/cp`). I'm just confused because if `$PATH` is consulted by Holmake, then we shouldn't have an absolute path like `/bin/cp`. But for portability's sake maybe this should be presented as an option to the user? 

`smart-configure.sml` seems like a good place for that? It's already showing me a bunch of information and prompting me for checks. Maybe something like

```shell
toolsdir: /bin
DOT_PATH: SOME ...
MLTON: NONE
```

mn200:
I have created an issue documenting that we should address this (#708).

PR #709, opened 2019-05-27
Fix mismatched parentheses

nbos:


mn200:
Thanks for this!  Clearly I need to figure out a way to check for elisp regressions automatically.

PR #715, opened 2019-07-06
fixing =+ typo

GergelyBuday:
Reading this section it came through that the operators here should be =+ the function override symbol, and not += the addition operator in imperative languages.


mn200:
Thanks!  Well spotted.

PR #716, opened 2019-07-12
Added Bisimulation to relationTheory

JamesShaker:
Just added a simple generic definition of what makes a relation a bisimulation and also the notion of bisimulation relation (which says two states are related if a bisimulation exists between them). Proved the bisimulation relation is an equivalence.

PR #719, opened 2019-08-02
Some supporting theorems of relation$BISIM

binghe:
Hi,

the recent bisimulation addition in `relationTheory` by @JamesShaker (#716) is very inspiring to me. In this PR, I further enriched it with the following supporting theorems:

```
   [BISIM_ID]  Theorem
      
      ⊢ ∀ts. BISIM ts $=
   
   [BISIM_INV]  Theorem
      
      ⊢ ∀ts R. BISIM ts R ⇒ BISIM ts Rᵀ
   
   [BISIM_O]  Theorem
      
      ⊢ ∀ts R R'. BISIM ts R ∧ BISIM ts R' ⇒ BISIM ts (R' ∘ᵣ R)
   
   [BISIM_RUNION]  Theorem
      
      ⊢ ∀ts R R'. BISIM ts R ∧ BISIM ts R' ⇒ BISIM ts (R ∪ᵣ R')
```

Besides, the existing proof of `BISIM_REL_IS_EQUIV_REL` is slightly simplified.

Then I re-defined `STRONG_BISIM` (bisimulation of CCS) in `examples/CCS/StrongEQScript.sml` using the new `relation$BISIM`. Using ALL above theorems, some theorems in CCS's `StrongEQTheory` now become trivial, the whole file is shortened by 100+ lines.

P. S. A general definition of "weak bisimulation" (and weak bisimilarity) is also possible, as I partially tried: (some transitions should be "weak" transitions)

```
   [WBISIM_REL_def]  Definition
      
      ⊢ ∀ts p q. WBISIM_REL ts p q ⇔ ∃R. WBISIM ts R ∧ R p q
   
   [WBISIM_def]  Definition
      
      ⊢ ∀ts R.
            WBISIM ts R ⇔
            ∀p q l.
                R p q ⇒
                (∀p'. ts p (SOME l) p' ⇒ ∃q'. ts q (SOME l) q' ∧ R p' q') ∧
                (∀q'. ts q (SOME l) q' ⇒ ∃p'. ts p (SOME l) p' ∧ R p' q') ∧
                (∀p'.
                     ts p NONE p' ⇒
                     ∃q'. (λx y. ts x NONE y)⃰ q q' ∧ R p' q') ∧
                ∀q'.
                    ts q NONE q' ⇒ ∃p'. (λx y. ts x NONE y)⃰ p p' ∧ R p' q'
```

However, it's much more complex and depends on `optionTheory` (`NONE` is the invisible transition label) which is built after `relationTheory`.  I can't see a good place to put it. Going further in this direction, I think it will become a formalization of (general) labelled transition system (of type `'a -> 'b option -> 'a -> bool`) like [https://github.com/coq-contribs/ccs/blob/master/Trans_Sys.v] and serve as a basis of process algebra formalization including CCS.

Chun Tian

### DIFF
--- a/examples/CCS/StrongEQScript.sml
+++ b/examples/CCS/StrongEQScript.sml
@@ -1,6 +1,7 @@
 (*
  * Copyright 1991-1995  University of Cambridge (Author: Monica Nesi)
- * Copyright 2016-2017  University of Bologna   (Author: Chun Tian)
+ * Copyright 2016-2017  University of Bologna, Italy (Author: Chun Tian)
+ * Copyright 2018-2019  Fondazione Bruno Kessler, Italy (Author: Chun Tian)
  *)
 
 open HolKernel Parse boolLib bossLib;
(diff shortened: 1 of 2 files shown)


mn200:
I agree about how it might be nice to have a general "labelled" theory.  Incidentally, the `path` theory is designed to support things in this direction.  Also, @JamesShaker is due to provide the “obvious” coinduction principle for bisimilarity sometime soon.  Anyway, many thanks for this update!

IlmariReissumies:
Thanks @binghe for your work on bisimulation!

Just a minor usability suggestions. Most formalisations of weak bisimulation for labelled transition systems that I've seen do not use `NONE` for the unobservable label, but rather define datatypes like

`datatype 'a label = Input 'a | Output 'a | Tau`

One way to make your notion of weak bisimulation compatible with this usage, while also getting rid of the dependency on `optionTheory`, would be to parameterise your definition of weak bisimilarity on an auxiliary predicate `WBISIM (obs: 'label -> bool) ts R` where `obs` is a predicate that specifies whether a label should be treated as an observable or as a tau.

binghe:
> Thanks @binghe for your work on bisimulation!
> 
> Just a minor usability suggestions. Most formalisations of weak bisimulation for labelled transition systems that I've seen do not use `NONE` for the unobservable label, but rather define datatypes like
> 
> `datatype 'a label = Input 'a | Output 'a | Tau`
> 
> One way to make your notion of weak bisimulation compatible with this usage, while also getting rid of the dependency on `optionTheory`, would be to parameterise your definition of weak bisimilarity on an auxiliary predicate `WBISIM (obs: 'label -> bool) ts R` where `obs` is a predicate that specifies whether a label should be treated as an observable or as a tau.

Hi Johannes (@IlmariReissumies)

thanks for your suggestion. (I ever read your PhD desertion «Culling Concurrency Theory ...»)

The datatype definition you gave is indeed the literal definition of transition label in most CCS books. However, in formalizations such a definition is not convenient: when we need to quantifier over just visible actions (e.g. in the definition of weak bisimulation), you have to write things like ``!α. is_visible α ==> ...``. A better way (chosen by M. Nesi [1], whose work in HOL88 was ported and extended by me as the current code in HOL's `examples/CCS`) is to define it as two types:

```
datatype 'a label = Input 'a | Output 'a
datatype 'a action = Label 'a | Tau
```

In this way, things become much easier. Also, in CCS's relabelling operator (ν), the action being relabelled must be visible.   Then I found that, above 2nd datatype definition can be replace by simply using `optionTheory`, with `NONE` overloading `Tau` (and `SOME` overloading `Label`, c.f.  line 55-56 of `examples/CCS/CCSScript.sml`)

```
val _ = overload_on ("tau",   ``NONE :'b Action``);
val _ = overload_on ("label", ``SOME :'b Label -> 'b Action``);
```

Such a change only requires minor changes in all existing proofs: all I need to do is to use the theorems (nchotomy, distinct, one_one, ...) of `optionTheory` in place of theorems generated by `datatype`. In this way, the only two `datatype`-defined types in my CCS formalization are just the CCS type itself, and the type of visible actions shown above.

Speaking of a general formalization of labelled transition system, I think the distinction between INPUT and OUTPUT actions is completely irrelevant with bisimulation, thus the whole `label` type should be just a type variable supplied by the upper theory. But the `tau` (invisible) action must be explicitly provided (and used in definition of weak bisimulation, etc), by augmenting the user-supplied label type with just one more element - this is exactly the purpose of `optionTheory`.  I have little knowledge on CSP, but from what I can see in CCS, Value-passing CCS, Pi-Calculus and P/T (Petri Net), they all have the (unique) `tau` action. Thus my proposal of this general labelled transition system based on `optionTheory` should be able to provide a basis for all these process algebras.

At last, the problem of your `WBISIM (obs: 'label -> bool) ts R` is, by `obs` itself we don't have an unique invisible action.  Thus we have no easy way to write a invisible transition between two processes, say `ts P Tau Q` using just `ts` and `obs`.   Thus I still think the uses of `optionTheory` is optimal, and for the dependency issue I think we can actually move all these new code from `relationTheory` into a new proof script under `src/coalgebras`.

Regards,

Chun

[1] Nesi, M.: A formalization of the process algebra CCS in high order logic. University of Cambridge, Computer Laboratory (1992).

binghe:
Maybe, instead of `(obs :'label -> bool)` we just parameterize `WBISIM` with an explicit `tau` action:

```
WBISIM (tau: 'b) (ts :'a ->'b ->'a -> bool) (R :'a->'a->bool)
```

In this way, the upper theory can still define a single action type in your way, and give one action the special meaning as the "invisible" action. However, this brings some other difficulties in proving lemmas on tau transitions and their RTC versions, where the weak bisimulation is not defined yet) and is not very elegant in my view.

binghe:
> I agree about how it might be nice to have a general "labelled" theory. Incidentally, the `path` theory is designed to support things in this direction. Also, @JamesShaker is due to provide the “obvious” coinduction principle for bisimilarity sometime soon. Anyway, many thanks for this update!

Hi Michael,

I'll see how @JamesShaker proceed by doing that, but there's already a solution in `examples/CCS`: we first use `CoInductive` (or `Hol_coreln`) to define `BISIM_REL` from some rules, then prove `BISIM_REL_def` as a theorem. The needed coinduction principle will be automatically generated by `Hol_coreln`. The `cases` theorem also generated together, what Robin Milner called "the property star", is even more useful. Many bisimulation proofs need that `cases` theorem - which essentially states that bisimulation relation is a fixpoint (but not necessary a greatest fixpoint).

--Chun

IlmariReissumies:
@binghe Yes, I understand your reasons for preferring options when formalising things like the CCS relabelling operator, and I agree that it's annoying for most applications to lose the property that there is a unique unobservable action. Thus on further thought I like your second idea of simply parameterising it on an element better. But I'm not sure how it would make proofs about tau transitions and their RTC versions more difficult --- in your CCS setting you would just instantiate `tau` to `NONE` and everything would be as before, no?

Some parenthetical remarks:

You're right that the distinction between input and output is irrelevant for a general theory of LTS. I gave it mostly as an example of a typical instantiation of the set of labels.

Some process formalisms actually do have more than one unobservable, for example the extensions of CCS with action priorities by Cleaveland and Hennessy needs to distinguish between high- and low-priority internal actions. But in such a setting you're likely to need a more exotic definition of bisimulation anyway, so there's little use catering to that here.

I agree about the other process calculi you mentioned, but I'm not convinced that this theory would form a useful basis for the pi-calculus. To get a sensible notion of bisimulation for the pi-calculus, your definition needs to treat binding names in labels as binding into both the label and the residual process. (I believe Davide sometimes calls such things higher-order transition systems). This is usually done by adding freshness side conditions to the definition. I'm not suggesting to adapt the general theory to accommodate this; that would force us to commit to a theory of name binding, and in general open the floodgates to an ocean of weirdness.

binghe:
Hi Hi Johannes (@IlmariReissumies),

actually... there's no essential difficulties in using `WBISIM (tau: 'b) (ts :'a ->'b ->'a -> bool) (R :'a->'a->bool)`.  What I was thinking is that, many lemmas about `EPS` (invisible transition) and `WEAK_TRANS` transitions now have no logical constant (previously was `NONE`) in it, but this `tau` could be a toplevel universal quantifier. The proof steps are still almost the same.

The conveniences in using two label types for CCS lelabelling operator is, of course, also irrelevant with bisimulation. Thus my uses of `optionTheory` actually cannot support its usage in defining weak bisimulation.

Alright, let me try to actually build it and see if it's small and clean enough to be put into `relationTheory`.

mn200:
@binghe Yes, my expectation was @JamesShaker would just define the underlying constant with `Hol_coreln`, automatically getting the conduction principle.

PR #720, opened 2019-08-07
The basic theory of weak bisimulation and its sample usage in CCS

binghe:
This PR follows #719 and #716, in which I was planing to build a basic theory of (standard) weak bisimulation for general labeled transitions (of type `:'a->'b->'a->bool), up to its equivalence result.  To use this work, the upper theory (e.g. a process algebra) must provide a transition relation and a **unique** invisible (tau) action.

The basic definitions are the following: (`ETS` stands for "*empty* transition"; `WTS` stands for "*weak* transition", better names are welcome)

```
   [ETS_def]  Definition
      ⊢ ∀ts tau. ETS ts tau = (λx y. ts x tau y)⃰
   
   [WTS_def]  Definition
      ⊢ ∀ts tau.
            WTS ts tau =
            (λp l q. ∃p' q'. ETS ts tau p p' ∧ ts p' l q' ∧ ETS ts tau q' q)
   
   [WBISIM_def]  Definition
      ⊢ ∀ts tau R.
            WBISIM ts tau R ⇔
            ∀p q.
                R p q ⇒
                (∀l.
                     l ≠ tau ⇒
                     (∀p'. ts p l p' ⇒ ∃q'. WTS ts tau q l q' ∧ R p' q') ∧
                     ∀q'. ts q l q' ⇒ ∃p'. WTS ts tau p l p' ∧ R p' q') ∧
                (∀p'. ts p tau p' ⇒ ∃q'. ETS ts tau q q' ∧ R p' q') ∧
                ∀q'. ts q tau q' ⇒ ∃p'. ETS ts tau p p' ∧ R p' q'

   [WBISIM_REL_def]  Theorem      
      ⊢ ∀ts tau. WBISIM_REL ts tau = (λp q. ∃R. WBISIM ts tau R ∧ R p q)

```

The coinduction principle is provided by `WBISIM_REL_coind`, generated by `CoInductive` command. And the last theorem:

```
   [WBISIM_REL_IS_EQUIV_REL]  Theorem
      ⊢ ∀ts tau. equivalence (WBISIM_REL ts tau)
```

Then I change the CCS example to define weak bisimulation `WEAK_BISIM` upon `WBISIM`:

```
   [WEAK_BISIM_def]  Definition      
      ⊢ ∀R. WEAK_BISIM R ⇔ WBISIM TRANS τ R
```

Again, several theorems in `example/CCS/WeakEQScript.sml` now become trivial.

I didn't put these work into `relationTheory` because I think they're too many and not really belong to the theory of relations.  Then I chose to put this `bisimulationTheory` under `src/relation/`, mostly because that folder is quite empty and this theory of bisimulation only depends on `relationTheory`. (I suggest also moving the strong bisimulation code into this script, if the PR is accepted, because there're still many ways to further enrich it.)

--Chun


mn200:
This looks very nice—thanks!  

I agree also that the bisimulation stuff now in `relation` can move to this theory.

Can you maybe write a brief paragraph about this material for the “new theories” section of `doc/next-release.md`? 

Both of these improvements would be fine to include in this PR.

binghe:
Now I added a short note in `next-release.md` and moved all `BISIM` and `BISIM_REL` from `relationTheory` to `bisimulationTheory`, with the coinduction principle added:

```
   [BISIM_REL_coind]  Theorem      
      ⊢ ∀ts BISIM_REL'.
            (∀a0 a1.
                 BISIM_REL' a0 a1 ⇒
                 ∀l.
                     (∀p'. ts a0 l p' ⇒ ∃q'. ts a1 l q' ∧ BISIM_REL' p' q') ∧
                     ∀q'. ts a1 l q' ⇒ ∃p'. ts a0 l p' ∧ BISIM_REL' p' q') ⇒
            ∀a0 a1. BISIM_REL' a0 a1 ⇒ BISIM_REL ts a0 a1
```

It's actually a little tricky to make sure the parameter `ts` appear at toplevel, otherwise the generated coinduction principle is wrong; I also fixed a similar problem in the WBISIM_REL version.  I also modified CCS example to directly define `STRONG_EQUIV` upon `BISIM_REL` without using `CoInductive` or `Hol_coreln` at all:

```
   [STRONG_EQUIV_def]  Definition
      ⊢ STRONG_EQUIV = BISIM_REL TRANS
```

Some proofs in the CCS example are further simplified. For `WEAK_EQUIV` I found that doing the same way results into longer proofs, because in its CCS-based definition it directly quantify over visible actions (`:'b Label` type), without the need of testing `l <> tau`. So I decided to not touch it.

mn200:
Fantastic; thank you!  (Is the comment about it being tricky to get the right coinduction principle effectively a bug report for Hol_coreln?)

binghe:
Thanks for merging it.

# Trick 1

For the "tricks" I mentioned, I don't think it's a bug of `Hol_coreln`, but maybe this is something worthy for documentation. Let me explain what I have observed:

My current (correct) co-inductive definition is the following:

```
CoInductive BISIM_REL :
    !p q. (!l.
            (!p'. ts p l p' ==> ?q'. ts q l q' /\ (BISIM_REL ts) p' q') /\
            (!q'. ts q l q' ==> ?p'. ts p l p' /\ (BISIM_REL ts) p' q'))
      ==> (BISIM_REL ts) p q
End
```

Note that, `ts` is a free variable in above coinduction rules. This is on purpose, because somehow `Hol_coreln` recognized `ts` as a schematic variable, such that, it becomes a universal quantifier in all generated theorems, especially the generated coinduction principle:

```
   ⊢ ∀ts BISIM_REL'.
         (∀a0 a1.
              BISIM_REL' a0 a1 ⇒
              ∀l.
                  (∀p'. ts a0 l p' ⇒ ∃q'. ts a1 l q' ∧ BISIM_REL' p' q') ∧
                  ∀q'. ts a1 l q' ⇒ ∃p'. ts a0 l p' ∧ BISIM_REL' p' q') ⇒
         ∀a0 a1. BISIM_REL' a0 a1 ⇒ BISIM_REL ts a0 a1: thm
```

However, if I slightly modify above definition by putting `ts` together with `p` and `q` as a explicitly quantified variable:

```
CoInductive BISIM_REL :
    !ts p q. (!l.
            (!p'. ts p l p' ==> ?q'. ts q l q' /\ (BISIM_REL ts) p' q') /\
            (!q'. ts q l q' ==> ?p'. ts p l p' /\ (BISIM_REL ts) p' q'))
      ==> (BISIM_REL ts) p q
End
```

Now the resulting coinduction principle becomes this:

```
⊢ ∀BISIM_REL'.
         (∀a0 a1 a2.
              BISIM_REL' a0 a1 a2 ⇒
              ∀l.
                  (∀p'. a0 a1 l p' ⇒ ∃q'. a0 a2 l q' ∧ BISIM_REL' a0 p' q') ∧
                  ∀q'. a0 a2 l q' ⇒ ∃p'. a0 a1 l p' ∧ BISIM_REL' a0 p' q') ⇒
         ∀a0 a1 a2. BISIM_REL' a0 a1 a2 ⇒ BISIM_REL a0 a1 a2: thm
```

This is *wrong*, because now the `ts` (shown as `a0`) appears as a universal quantifier as inner clause, the upper theorem cannot instantiate (by `SPEC`) this theorem into a coinduction principle for their particular bisimulation where `ts` is replaced by a logical constant like `TRANS` in CCS.  However, using this wrong coinduction principle I can still literally prove that `BISIM_REL` is the coarsest relation contained in `BISIM`, but now this `BISIM_REL` is not a binary relation any more, instead it's a ternary relation on `ts`, `p` and `q`.

Thus, by using free variable on purposes when defining a coinductive (or maybe also inductive) relation, one can distinguish between two groups of variables, those belong to the targetin relation, and those purely schematic.  I'm not sure if the designer of `Hol_coreln` (and maybe also `Hol_reln` had this idea in mind.

# Trick 2

Another irrelevant "thick" (specified to bisimulation) is, by looking at the following definition of `BISIM` given by @JamesShaker, one may get impressions that `p`, `q`, `l` are quantifiers of the same level:

```
   ⊢ ∀ts R.
         BISIM ts R ⇔
         ∀p q l.
             R p q ⇒
             (∀p'. ts p l p' ⇒ ∃q'. ts q l q' ∧ R p' q') ∧
             ∀q'. ts q l q' ⇒ ∃p'. ts p l p' ∧ R p' q': thm
```

The definition is equivalent with another version where `l` is pushed inner (after `R p q ==>`). However, this is not true when defining `BISIM_REL` by `CoInductive` rules, which should have a schema: `!p q. (!l. X) ==> (BISIM_REL ts) p q`. That is, variable `l` must be an universal quantifier of the inner antecedent clause. If we still put `p`, `q`, `r` together and try to define `BISIM_REL` again, the resulting `_rules` theorem is still the same, but the `_coind` and `_cases` theorems are all *wrong* . The difference is, in these generated theorems, the variable `l` now becomes an *existential* quantifier: (the resulting relation is not a bisimulation equivalence any more)

```
   ⊢ ∀ts BISIM_REL'.
         (∀a0 a1.
              BISIM_REL' a0 a1 ⇒
              ∃l.
                  (∀p'. ts a0 l p' ⇒ ∃q'. ts a1 l q' ∧ BISIM_REL' p' q') ∧
                  ∀q'. ts a1 l q' ⇒ ∃p'. ts a0 l p' ∧ BISIM_REL' p' q') ⇒
         ∀a0 a1. BISIM_REL' a0 a1 ⇒ BISIM_REL ts a0 a1: thm
```

mn200:
Thanks for your comments!

Re: Trick 1.  This was definitely intended in the design of `Hol_reln` (it is needed when defining `RTC` for example).  The existence of schematic variables is described in the DESCRIPTION manual, and the need to use them when defining `RTC` is one of the examples in the same section. 

I will need to think more about Trick 2.

PR #724, opened 2019-08-27
Expand message from Define failure.

talsewell:
Adds one more line to the output message to make it a little
clearer that "goal has been set up with ..." implies that it this goal
is now in the proof manager state, and you should look there.

(I appreciate that this might be a bad time to print the proof
manager state, since that might be long enough to hide what
has happened.)


mn200:
Thanks!

Issue #727, opened 2019-09-02
drule shouldn't move around quantifiers


AndreasLoow:
I just noticed the following, not sure if it's new `drule` behaviour or if I just haven't noticed it before. I have the following assumption (last):

```
   98.  ∀cenv' nl min max.
            sum_foldM cell_run cenv'' nl = INR cenv' ∧
            netlist_ok min max nl ∧ si_lt bsi min ⇒
            ∀var.
                cell_inp_run cenv' (cget_net bsi var) =
                cell_inp_run cenv'' (cget_net bsi var)
```

and when I do `first_x_assum drule` I get the following goal:

```
(∀min max var.
             netlist_ok min max fnl ∧ si_lt bsi min ⇒
             cell_inp_run cenv'³' (cget_net bsi var) =
             cell_inp_run cenv'' (cget_net bsi var)) ⇒ [...]
```

Notice how the quantifier for `var` has been moved to the top. I think it would be better if `drule` did not move around quantifiers like that. (Moving around quantifiers like that will make using `impl_tac` impossible afterwards a series of `drule`s sometimes.)

mn200:
It's not new; the normalisation done by `mp_then`, which is what implements `drule`, will move variables out in this way.  This gives `mp_then` uniform access to all of the theorem's hypotheses. Unfortunately for `drule`, this will unnecessarily move the `var` quantification out in your case. 

You can renormalise the theorem that `drule` puts into the goal with

     drule_then (mp_tac o SIMP_RULE bool_ss [RIGHT_FORALL_IMP_THM])

We might also think about making `impl_tac` smarter: when it sees `(!x. P ==> Q x) ==> R`, it could give you `P` to prove...

AndreasLoow:
Thanks for the info, the more general `impl_tac` seems like a good idea, and that would also be useful in other situations. (The very best thing would be to not have `drule` move around things, but I'm not familiar with the implementation of `drule` and I'm sure there are good reasons for the current behaviour. Now at least I know that it's not a bug.)

PR #730, opened 2019-09-06
Make impl_tac work with universally quantified implications

AndreasLoow:
As suggested in #727, `impl_tac` now handles universally quantified implications:

```
(* Example 1, works as expected: *)
Theorem thm1:
 a ==> (!x. a ==> b x) ==> b y
Proof
 strip_tac \\ impl_tac >- first_assum MATCH_ACCEPT_TAC \\ disch_then MATCH_ACCEPT_TAC
QED

(* Example 2, impl_tac dies (as it should) because x occurs in hypo: *)
Theorem thm2:
 a 5 ==> (!x. a x ==> b x) ==> b 5
Proof
 strip_tac \\ impl_tac (* <-- dies *)
QED

(* Example 3, MATCH_MP does "too much": *)
Theorem thm3:
 a ==> (!x y. a ==> b x) ==> b 5
Proof
 strip_tac \\ impl_tac >- first_assum MATCH_ACCEPT_TAC \\
 (* goal here: "(∀x. b x) ⇒ b 5", i.e. not "(∀x y. b x) ⇒ b 5"! *)
 disch_then MATCH_ACCEPT_TAC
QED
```

(Not well-tested and there's probably a prettier way/more efficient way to do the free vars check in `PROVEHYP_THEN`.)

### DIFF
--- a/help/Docfiles/Thm_cont.PROVEHYP_THEN.doc
+++ b/help/Docfiles/Thm_cont.PROVEHYP_THEN.doc
@@ -15,8 +15,6 @@ requires that {th} be an (universally quantified) implication
 Given an implication {|- !x1..xn. l ==> r x1..xn}, the result is a new
 sub-goal requiring the user to prove {l}, and the application of {th2tac}
 to the theorems with conclusion {l} and {!x1..xn. r x1..xn}.
-(The former has just been proved by the user; the latter is the result
-of applying {MATCH_MP} to {th} and {l}.)
 
 Diagrammatically, one might see this as
 {
--- a/help/Docfiles/Thm_cont.PROVEHYP_THEN.doc
+++ b/help/Docfiles/Thm_cont.PROVEHYP_THEN.doc
@@ -10,12 +10,13 @@ Theorem-tactical.
 
 \DESCRIBE
 An application of the tactic {PROVEHYP_THEN th2tac th} to a goal {g}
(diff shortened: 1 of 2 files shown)


AndreasLoow:
Just wanted to add that I've tested this in some real proofs now, and it seems to work.

AndreasLoow:
Is there something that needs to be clarified before this can be merged?

mn200:
Sorry, this slipped off my radar.  The only thing I'm less keen on is the use of `MATCH_MP`.  I'd prefer, something like

   CONV_RULE (STRIP_QUANT_CONV (LAND_CONV (REWR_CONV (EQT_INTRO nth)) THENC
                                                                 REWR_CONV foo))

where `foo` is the theorem that says that `(T ==> t) <=> t`.

AndreasLoow:
Sorry, not sure I'm following. What's the problem with `MATCH_MP`? (Overkill? But it also makes it easy to describe what the tactic does in the documentation.)

mn200:
Right, `MATCH_MP` is “wrong” because no matching should happen at all. (My first `REWR_CONV` above should rather be `K`.)

AndreasLoow:
Not sure why the Travis tests are failing, seems unrelated to what I have changed as far as I can tell, :s.

mn200:
Great stuff - many thanks.  The failures are to do with my multi-directory Holmake changes.

PR #731, opened 2019-09-15
Latest updates of the CCS example

binghe:
Hi,

this is some updates of the CCS example. Several key lemmas are improved with less antecedents, some definitions in `multivariateTheory` are simplified. TeX outputs are fixed.

--Chun


mn200:
Thanks!

PR #732, opened 2019-09-16
Fixed duplicated definition of "inf" in iterateTheory

binghe:
Hi,

this PR fixed a long-standing issue in `iterateTheory` since it was ported to HOL. In it, the infimum (`inf`) of real sets in `realTheory`:

```
   [inf_def]  Definition
      ⊢ ∀p. inf p = -sup (λr. p (-r))
```

has been redefined here in a different (but correct) way:

```
   [inf]  Definition
      ⊢ ∀s. inf s = @a. (∀x. x ∈ s ⇒ a ≤ x) ∧ ∀b. (∀x. x ∈ s ⇒ b ≤ x) ⇒ b ≤ a
```

As the consequence, if any user theory imported `iterateTheory` after `realTheory`, all theorems using the original `inf` in `realTheory` won't work any more.

`inf` is not a total function, in the sense that the infimum of real sets only exists if the set is not empty and there's a lower bound of it, i.e. the following criteria:

```
val inf_criteria = “s ≠ ∅ ∧ ∃b. ∀x. x ∈ s ⇒ b ≤ x”;
```

The two definitions are equivalent, only under the above criteria, as Micheal Norrish ever explained in HOL-info. Now I first renamed the `inf` of `iterateTheory` to `inf'`, then I proved a local theorem showing their conditional equivalence:

```
 ⊢ ∀s. s ≠ ∅ ∧ (∃b. ∀x. x ∈ s ⇒ b ≤ x) ⇒ (inf s = inf' s)
```

Once this is done, I only needed to fix a small number of proofs in which the original definition of `inf'` is directly used, by rewriting `inf` to `inf'`, then the existing proof steps still work.

The theorem `INF_EQ` holds only under the definition of `inf'`. Now I have to add some more antecedents to make sure the `inf` exist.  Fortunately only one theorem in `real_topologyTheory` needs `INF_EQ`, and I fixed that proof by further showing the involved infimums indeed exist.

Maybe with a more careful work we can even completely remove the global definition of `inf'`, to define it by `Q.ABBREV_TAC` only in those involved proofs, but this is not urgent and optional.

--Chun

### DIFF
--- a/src/probability/iterateScript.sml
+++ b/src/probability/iterateScript.sml
@@ -43,6 +43,18 @@ fun ASSERT_TAC tm = SUBGOAL_THEN tm STRIP_ASSUME_TAC;
 val ASM_ARITH_TAC = REPEAT (POP_ASSUM MP_TAC) THEN ARITH_TAC;
 val ASM_REAL_ARITH_TAC = REPEAT (POP_ASSUM MP_TAC) THEN REAL_ARITH_TAC;
 
+fun wrap a = [a];
+val Rewr = DISCH_THEN (REWRITE_TAC o wrap);
+val Know = Q_TAC KNOW_TAC;
+val Suff = Q_TAC SUFF_TAC;
+val POP_ORW = POP_ASSUM (fn thm => ONCE_REWRITE_TAC [thm]);
+
+local
+  val th = prove (``!a b. a /\ (a ==> b) ==> a /\ b``, PROVE_TAC [])
+in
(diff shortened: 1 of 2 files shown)


mn200:
I agree that it would be nice to get rid of `inf'` entirely; the `INF_EQ` theorem could just replace `inf'` with the RHS of that definition. This is definitely progress nonetheless—thanks!

binghe:
Thanks for the hints. Now I see how to make it perfect.

PR #734, opened 2019-09-17
Fix examples/fun-op-sem

hrutvik:


mn200:
Thanks for your work on this; I'll merge into `develop` and see if I also have local problems with it on `expk`.

PR #736, opened 2019-09-20
Re-defined ext_suminf for positive functions only

binghe:
Hi (Michael),

following our discussions on HOL-info, now I have re-defined `ext_suminf` (extrealTheory) for positive functions only, using `new_specification`:

```
   [ext_suminf_def]  Definition
      ⊢ ∀f.
            (∀n. 0 ≤ f n) ⇒
            suminf f = sup (IMAGE (λn. ∑ f (count n)) 𝕌(:num))
```

This eliminates the weird results like `ext_suminf (\n. -1) = -1`.  Then I fixed all broken probability proofs due to this change. (The html generation commands in probability theories are deleted, additionally.)

During the proof fixing process, on a failed call of `REAL_ARITH_TAC`, I got strange outputs:

```
Exception raised at Lib.trye:
original exn. not a HOL_ERR
Exception-
   HOL_ERR
     {message = "original exn. not a HOL_ERR", origin_function = "trye",
      origin_structure = "Lib"} raised
```
 
I think this indicates that, the SML file (RealArith) where `REAL_ARITH_TAC` is defined, lacks of a definition of `ERR`:

```
val ERR = mk_HOL_ERR "RealArith";
```

 I added the above line, but I didn't test if it actually work. Please confirm.

--Chun


mn200:
Thanks for this.  However, the second commit, binding `ERR` in `RealArith` doesn't do anything as far as I can tell.  (`ERR` is never referenced in `RealArith`, and it can't be exported as there is an explicit `RealArith.sig`.)

mn200:
If you think there's a problem in `RealArith` (which may well be!), please report a separate issue.

mn200:
I've merged your first commit  (thanks again!) manually, and will ignore the second.

binghe:
No problem. It seems that HOL's `REAL_ARITH_TAC` (ported by Joe Hurd from HOL Light to HOL98, long time ago) is not as powerful as its latest version in HOL Light, but in its current scope there's no explicit bug.

PR #738, opened 2019-09-25
HOL-Analysis: univariate differential and integral calculus (with doc)

binghe:
Hi,

this is HOL Light's differential and integral calculus (based on Henstock-Kurzweil Integrals), reduced to univariate case.  This PR adds another 700+ theorems and 20,000+ lines of proof scripts into HOL's code base.

Now we have two more theory files (`derivativeScript.sml` and `integrationScript.sml`) added, both depend on the existing `real_topologyScript.sml`.  I consider the `src/probability` folder not suitable for them any more, so I moved all related files into a directory `src/analysis`. We may call them `HOL-Analysis` library.

I also updated HOL Description with the recent changes in extended reals, Probability Theory and the new calculus stuff. I moved Probability-related contents from the section of "Numbers" to a new section "Higher Mathematics", where I put the new calculus stuff in. (In the future, some details of `real_topologyTheory` should be also described there.)  Notice that, previously the manual wrong said that `transcTheory` contains the Fundamental Theorem of Calculus. But this is not true.  Now the Fundamental Theorem of Calculus is here (in `integrationTheory`):

```
[FUNDAMENTAL_THEOREM_OF_CALCULUS]  Theorem
      ⊢ ∀f f' a b.
            a ≤ b ∧
            (∀x.
                 x ∈ interval [(a,b)] ⇒
                 (f has_vector_derivative f' x)
                   (at x within interval [(a,b)])) ⇒
            (f' has_integral f b − f a) (interval [(a,b)])
```

The classical requirements for a function to be (Riemann-)integrable is that, the function must be continuous and differentiable everywhere in the involved close interval. But Henstock-Kurzweil Integral also has another stronger version, in which a
countable number of exceptional points is permitted:

```
[FUNDAMENTAL_THEOREM_OF_CALCULUS_STRONG]  Theorem
      ⊢ ∀f f' s a b.
            COUNTABLE s ∧ a ≤ b ∧ f continuous_on interval [(a,b)] ∧
            (∀x.
                 x ∈ interval [(a,b)] DIFF s ⇒
                 (f has_vector_derivative f' x)
                   (at x within interval [(a,b)])) ⇒
            (f' has_integral f b − f a) (interval [(a,b)])
```

The porting work was originally done by people from HVG Concordia. I never know in which HOL version their proof script can run, but anyway I further fixed all broken proofs (mostly due to the less-powerful `REAL_ARITH_TAC` in current HOL4, comparing with HOL Light.) As the work is  non-trivial and really time-consuming, I added myself into the copyright notes at the beginning of relevant files.

--Chun


binghe:
NOTE: The current `integralTheory` (in `src/real`, never mentioned in the Manual) should be considered as obsoleted. It's an early version of gauge integral, also from HOL Light, ported by some chinese long time ago. I'll see how to reuse `src/real/Diff.sml` for the new `integrationTheory`.

binghe:
Let me cancel this big PR and prepare some smaller ones instead.

PR #739, opened 2019-09-30
[mosml] fixed HolBdd (and examples) due to "equality type" changes

binghe:
This PR fixes the mosml-only HolBdd library (and examples) due to the "equality type" (and "tight equality") changes in K13 release.


mn200:
Thanks for this!  I'm somewhat surprised that this works at all...

PR #740, opened 2019-10-01
Fixed normalForms due to "tight equality" changes

binghe:
This PR fixes `normalForms` (in `src/metis`) due to "tight equality" changes. Some test cases in comments are moved into `selftest.sml` (to make sure they're not broken again).

### DIFF
--- a/src/metis/normalForms.sml
+++ b/src/metis/normalForms.sml
@@ -83,7 +83,7 @@ fun AVOID_SPEC_TAC (tm, v) =
 
 local
   open tautLib;
-  val th = prove (``(a = b) /\ (c = d) ==> (a /\ c <=> b /\ d)``, TAUT_TAC);
+  val th = prove (``(a <=> b) /\ (c <=> d) ==> (a /\ c <=> b /\ d)``, TAUT_TAC);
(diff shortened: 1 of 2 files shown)


mn200:
Most of the first commit's changes seem to be about making strictly unnecessary changes.  For example, changing `(x = y)` to `(x <=> y)` is always unnecessary. 

I appreciate the movement of test-cases into the selftest; but which of them fails in the old version of the code?

binghe:
`CONTRACT_CONV:    (p \/ r) \/ p \/ ~q <=> p \/ r \/ ~q` fails, due to the changes in the following two helper lemma:

```
 val CONTRACT_DISJ = prove
   (``!a b b'. (~a ==> (b <=> b')) ==> (~a ==> (a \/ b <=> b'))``,
    tautLib.TAUT_TAC);

 val DISJ_CONGRUENCE = prove
   (``!a b b'. (~a ==> (b <=> b')) ==> (a \/ b <=> a \/ b')``,
    tautLib.TAUT_TAC);
```

The type of `b` and `b’` changed under different equalities, as `b = b’` was wrongly considered together, but the same tactic worked.

mn200:
Thanks for the additional information!

PR #741, opened 2019-10-11
Fixed real_topologyTheory for tight equality

binghe:
Hi,

this PR mainly fixed `real_topologyTheory` for (the default) tight equality. The line `ParseExtras.temp_loose_equality()` has been removed while the whole theory still builds.

I also removed its dependency on `util_probTheory` by moving back all definitions and lemmas about `from` to `real_topologyTheory` (But then I have to add `real_topologyTheory` into the dependencies of `probabilityTheory` because it needs `from` lemmas.)  This is in preparation with moving `real_topologyTheory` out of `src/probability` folder in the future, when more theories on real analysis were ported from HOL Light.

Two theorems previously commented in `real_topologyTheory` are now fixed (due to my increased knowledge on the differences of `REAL_ARITH_TAC` between HOL4 and HOL Light):

```
   [IS_INTERVAL_SUMS]  Theorem      
      ⊢ ∀s t.
            is_interval s ∧ is_interval t ⇒
            is_interval {x + y | x ∈ s ∧ y ∈ t}

   [HAUSDIST_BALLS]  Theorem      
      ⊢ (∀a b r s.
             hausdist (ball (a,r),ball (b,s)) =
             if r ≤ 0 ∨ s ≤ 0 then 0 else dist (a,b) + abs (r − s)) ∧
        (∀a b r s.
             hausdist (ball (a,r),cball (b,s)) =
             if r ≤ 0 ∨ s < 0 then 0 else dist (a,b) + abs (r − s)) ∧
        (∀a b r s.
             hausdist (cball (a,r),ball (b,s)) =
             if r < 0 ∨ s ≤ 0 then 0 else dist (a,b) + abs (r − s)) ∧
        ∀a b r s.
            hausdist (cball (a,r),cball (b,s)) =
            if r < 0 ∨ s < 0 then 0 else dist (a,b) + abs (r − s)
```

Regards,

Chun Tian

### DIFF
--- a/src/probability/probabilityScript.sml
+++ b/src/probability/probabilityScript.sml
@@ -14,7 +14,7 @@ open HolKernel Parse boolLib bossLib;
 
 open pairTheory combinTheory optionTheory prim_recTheory arithmeticTheory
      res_quanTheory res_quanTools pred_setTheory realTheory realLib
-     seqTheory transcTheory real_sigmaTheory;
+     seqTheory transcTheory real_sigmaTheory real_topologyTheory;
 
 open hurdUtils util_probTheory extrealTheory sigma_algebraTheory
      measureTheory borelTheory lebesgueTheory martingaleTheory;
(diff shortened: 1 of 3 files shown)


mn200:
Thanks for the ongoing work on this material!

PR #742, opened 2019-10-16
Add DROP_TAKE and TAKE_DROP_SWAP theorems

myreen:
I think `DROP_TAKE` and `TAKE_DROP` would be the most natural names for these theorems. Unfortunately, `TAKE_DROP` was already taken. 

I decided to call the clashing one `TAKE_DROP_SWAP`. 

Alternatively, one could go through the whole development and rename `TAKE_DROP` to `TAKE_APPEND_DROP` and call the current `TAKE_DROP_SWAP` simply `TAKE_DROP` instead. 

Which is preferred?


mn200:
I think it's probably better to leave existing names alone if possible.

PR #746, opened 2019-10-27
Henstock–Kurzweil integral (ported from HOL Light)

binghe:
Hi,

currently HOL4 only has a minimal theory of derivative and integrals (at `src/real/integralTheory`), ported from HOL Light in 2012. This PR adds a 10x bigger formalization of Henstock–Kurzweil integral [1] (or generalized Riemann integral, or **gauge** integral), ported by people at Concordia HVG also from HOL Light, in 2015.  Due to the difference between HOL and HOL Light (e.g. `REAL_ARITH_TAC`, and the definition of `inf`), some further fixes were done by me, several definitions are even slightly changed.  The two theory files were built upon `real_topologyTheory`. They include the famous Fundamental Theorem of Calculus:

```
[FUNDAMENTAL_THEOREM_OF_CALCULUS]  Theorem
      ⊢ ∀f f' a b.
            a ≤ b ∧
            (∀x.
                 x ∈ interval [(a,b)] ⇒
                 (f has_vector_derivative f' x)
                   (at x within interval [(a,b)])) ⇒
            (f' has_integral f b − f a) (interval [(a,b)])
```

The connection between Henstock–Kurzweil integral and the existing Probability materials (i.e. Lebesgue integral) are the following:

1. The construction of the household Lebesgue measure has two major method: one is by guage integral, the other is by Carathéodory's extension theorem. I'm going to support both approaches and prove their equivalences in the future.

2. A function is Lebesgue integrable if and only if the function and its absolute value are Henstock–Kurzweil integrable.  This important connection theorem could be formalized later (by someone or me) in HOL4 using existing theories as basis.

I think this is a piece of beautiful and elegant work, should be kept at HOL official and maintained officially. These theories are not directly found in HOL Light, which now has a multivariate version, more general but less easy to actually use.

--Chun

P. S. In `util_probTheory`, the definition `powr` is duplicated with `rpow` (in `transcTheory`). Now I turned it into a overload.

[1] https://en.wikipedia.org/wiki/Henstock–Kurzweil_integral


mn200:
Thanks for this!  Can you write something (perhaps just what you have above) for the release notes please?

binghe:
> Thanks for this! Can you write something (perhaps just what you have above) for the release notes please?

I have added `derivativeTheory` and `integrationTheory` into release notes.

mn200:
Fantastic!  Thanks again.

PR #747, opened 2019-10-28
Merged "extra" theories between "miller" and "diningcryptos" examples

binghe:
Hi,

between "miller" and "diningcryptos" there are some shared small extra theories with almost the same contents. This PR cleaned up "diningcryptos" example and removed five "extra" theories in `examples/diningcryptos` folder (and use those from the "miller" example instead)

- `extra_bool`: all `xor`-related lemmas are merged into the same theory file of "miller" example.
- `extra_pred_set`, `extra_listTheory`, `extra_num`, `extra_real`: moved and merged into "miller" example with some cleanups.

In addition, many unnecessary `PRE_INCLUDES` in existing Holmakefiles are removed. I also added `CLINE_OPTIONS = -r` to make "miller" example buildable at top-level, learnt from a similar changes in `temporal_deep`.

At the end, about 4,000 lines of duplicated code are removed. Both examples still build, nothing is lost.

--Chun


mn200:
Thanks for this work!

PR #750, opened 2019-11-06
AKS mechanisation

jhlchan:
Import from  my PhD thesis work.


mn200:
Looks good. Two comments:

- please remove `OPTIONS = QUIT_ON_FAILURE` from all the Holmakefiles (this is no longer required)
- please write an addition to the release notes file (`doc/next-release.md`) describing the example

mn200:
In the AKS Holmakefile, you should specify a particular target to represent the culmination of the work.  (Otherwise Holmake won't build anything at all because there are no default targets in the AKS directory.) I'd suggest putting something like the following in:

```
.PHONY: all
all: theories/AKScleanTheory.uo
```

mn200:
I notice that if all you need is `AKScleanTheory` then `AKSmachineTheory` isn't built.  Is that file supposed to be present?

mn200:
Conversely, in the `algebra` example, you don't have a `Holmakefile` at all.  I'd suggest creating one that looks like
```
CLINE_OPTIONS = -r
INCLUDES = field finitefield group lib linear monoid polynomial ring
```
That way, if someone does `Holmake` in `examples/algebra` everything is built.  At the moment, nothing at all happens.

binghe:
Hi, nice work. Where can I find your thesis? I’d like to read it.

mn200:
See: https://openresearch-repository.anu.edu.au/handle/1885/177195

mn200:
Thanks for this!

PR #758, opened 2019-12-05
Holwrap

lxndrcx:
I wrote a 'wrapper' for hol, similar to rlwrap, but with some more features like history, mild syntax highlighting and multiline input.


mn200:
Thanks for this!

PR #760, opened 2019-12-11
Register llist as a datatype in TypeBase

IlmariReissumies:
This is slightly abusive because some of the `TypeBase` entries are the
dual of what you would expect, but on the positive side, it should
allow one to use lazy lists in case expressions, with `CaseEq`, and so
on.

Seems to play nicely with `Cases_on`, but not with `Induct_on`.

An alternative to this approach would be to redesign TypeBase so that
non-datatypes can have case theorems.

### DIFF
--- a/src/coalgebras/llistScript.sml
+++ b/src/coalgebras/llistScript.sml
@@ -2648,7 +2648,7 @@ val LLENGTH_LREPEAT = Q.store_thm(
    Case constant, distinctness etc. for TypeBase
    -------------------------------------------------------------------------- *)
 
-Definition llist_CASE_def:
+Definition llist_CASE_def[nocompute]:
   llist_CASE ll b f =
     case LTL_HD ll of
       NONE => b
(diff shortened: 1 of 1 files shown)


src/coalgebras/llistScript.sml
```diff
@@ -2644,16 +2644,75 @@ val LLENGTH_LREPEAT = Q.store_thm(
   `LLENGTH (LREPEAT l) = if NULL l then SOME 0 else NONE`,
   rw[LREPEAT_def])
 
+(* --------------------------------------------------------------------------
+   Case constant, distinctness etc. for TypeBase
+   -------------------------------------------------------------------------- *)
+
+Definition llist_CASE_def:
```
mn200: If you're going to provide a later `compute` theorem for this constant, the definition should be `nocompute`.

IlmariReissumies: check!

mn200:
Thanks!

PR #762, opened 2019-12-17
Fix "HOL Theory Hierarchy (clickable) disappeared..." #749

### LINKED ISSUE #749: HOL Theory Hierarchy (clickable) disappeared...
Hi (Michael),

I don't know which recent commit caused this, but the HOL Theory Hierarchy (clickable), or Theory Graph, like those in [1], disappeared. In another words, the generated SVG picture (`theories.svg`) from `theories.dot`, although still contains all information*, now it's not clickable or viewable from the HTML page `theories.html`.

Can you reproduce this strange issue?

--Chun

[1] https://hol-theorem-prover.org/kananaskis-13-helpdocs/help/theorygraph/theories.html
* If I open the SVG file by Adobe Illustrator, I can still see the theory graph. If I chose to generate a PDF from the `.dot` file, the PDF looks perfect.

binghe:
This PR fixes #749 by adjusting some parameters in the generated `theories.dot` and `theories.html` files. It's probably also a bug of `dot` (graphviz) but the changes here worked around the potential bug.

See #749 for related discussions and a sample screenshot after this change.

### DIFF
--- a/help/src-sml/DOT
+++ b/help/src-sml/DOT
@@ -100,7 +100,13 @@ fun gen_dotfile file node_info =
      val ostrm = openOut file
  in
    PP.prettyPrint(curry output ostrm, 75)
-                 (pp_dot_file {size="16,16",ranksep="1.0",nodesep="0.30"}
+                 (* HOL Theory Hierarchy (clickable) disappeared (#749)
+
+                    Enlarged SVG size (old: (16,16) inch) to make sure
+                    the success of SVG generation. Further scaling is done
+                    at HTML side. Nodesep 0.2 (old: 0.3) looks better.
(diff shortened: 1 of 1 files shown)


binghe:
Hi ...

People from Graphviz community said [1] that that was a bug of `dot` [2] introduced in graphviz between 2.38 and 2.43 and was fixed about one month ago.

I was using 2.43 but now I downgraded to 2.40 and it works. Thus there's no need to merge anything. Closing the PR.

[1] https://gitlab.com/graphviz/graphviz/issues/1639#note_261928195
[2] https://gitlab.com/graphviz/graphviz/issues/1605

PR #768, opened 2020-01-08
Add PairCases tactic

AndreasLoow:
Previously only Cases, Cases_on and PairCases_on existed, now we have PairCases as well.

(An alternative would be to make Cases recognise pairs and split them appropriately, but that would break old proofs that depend on names generated by Cases.)


konrad-slind:
I am a bit confused. Cases and Cases_on already deal with pairs, What they
don't handle atomically is iterated pairs. (Correct me if wrong!) Not sure
if anyone has added this, but an easy way to implement such a "tupleCases"
tactic would be to simply rewrite with pairTheory.FORALL_PROD.

AndreasLoow:
Cases and Cases_on do, exactly as you say, handle pairs but not iterated pairs. So PairCases is only needed when you have iterated pairs (and for non-iterated pairs if you want the naming scheme used by both PairCases and PairCases_on for the variables that splitting introduces).

To clarify, what I meant by "recognise pairs and split them appropriately" is that it would be better if Cases and Cases_on recognised iterated pairs and split them exactly the same way as PairCases and PairCases_on split now (including the naming scheme). "Better" in the sense that proofs would look nicer, and proof authors would not have to remember this iterated-pairs-special-case that needs a special tactic while most other splitting can be done by Cases/Cases_on.

konrad-slind:
Yes, it's annoying. The basic Cases{_on} tactics are designed to just do
one top-level case analysis. If you want to change them so that, in the
case of iterated product types, the full pair structure gets exposed, then
there may be some backwards incompatibility (as you point out). In that
case, though, then the code that should be changed is in BasicProvers.

konrad-slind:
Probably what would be best is to design and implement a "namedCases{_on}"
where there is an extra parameter specifying the variable names to be used
for each constructor. So

   namedCases ["a b c"]

applied to the goal

    !p:num#num#num. FST(p) < FST(SND p) /\ FST(SND p) < SND(SND p)

would yield

   a < b /\ b < c

and

  namedCases ["", "h t"]

applied to

  !list. P list

would yield

    P [] /\ P h t

AndreasLoow:
I think what one ultimately want is something slightly more general: A tactic that allows for partial specification of new names. This is how (at least some) Coq tactics work. For example, this more general namedCases should allow inputs such as:

```
namedCases ["", "h _"]
```

where the underscore means that we do not care about the name of the list tail (so the tactic should feel free to generate any unique name for it).

konrad-slind:
Have added `namedCases{_on}` entrypoints implementing the above to BasicProvers and bossLib.

AndreasLoow:
Piggybacking on Konrad's `namedCases{_on}` I have a local change to `Cases{_on}` that splits iterated pairs. Initially, I tried giving all "split variables" uniform names, such that `e : ('a # 'b)` would split into `(e0, e1)` and `p : ('a # 'b # 'c)` would split into `(p0, p1, p2)` etc. But this broke too many proofs so now I instead have the following naming scheme: for pairs of length 2, use the old naming scheme, and for pairs of greater length, use the naming scheme p0, p1, p2, ...

The current naming scheme only breaks around 5 files in `src` -- i.e. the change is not backward compatible but the breakage seems small.

Anyone has any opinion on if such breakage can be considered acceptable or not? The pro of the change is that `PairCases{_on}` is no longer needed (i.e. `Cases{_on}` can be used everywhere instead), and the con is that some old proofs might break.

@konrad-slind Btw, when implementing the above I discovered a problem with `namedCases{_on}`. Have not looked into it yet:

```
Theorem foo:
 !(b : num # num # num). b = ARB
Proof
 namedCases ["b0 b1 b2"] (* <-- works *)
QED

Theorem foo':
 !(b : (num # num) # num). b = ARB
Proof
 namedCases ["b0 b1 b2"] (* <-- dies *)
QED
```

konrad-slind:
Error is because I strip along the spine rather than flattening. But the
justification proof might become more complex; will have to think about it.

AndreasLoow:
I just realized that I'm not actually sure what is meant by iterated pairs, or rather that multiple things can be meant by it. For example, for the following types

```
t1 = num # num # num
t2 = (num # num) # num
t3 = num # (num # num) # num
```

I think what I initially had in mind was that t1 would be split into three, t2 into two, and t3 into three -- but maybe splitting t1 into three, t2 into three, and t3 into four is equally sensible.

Maybe this is already clear to everyone else except me, just wanted to add this in case it's not.

mn200:
I haven't tracked this conversation closely enough to now understand what should be done here. I'm disturbed by the merge failure above.  Can that be fixed please if there's still useful material to be included from this PR?

konrad-slind:
I haven't followed either. I know about Cases{_on} and namedCases{_on}. The
first does a minimal top-level case split, inventing names. The second does
the same, using user-given names. One deviation: for products, it
iteratively does the case-split but only works on things of type ty1 # .ty2
# ... # tyn.  With a bit of time I could get namedCases{_on} to handle the
full space of tuples. The issues with ambiguity that Andreas mentions don't
bother me much since they seem rare (I think?) and can be handled by
repeated namedCases{_on}. Will an improved namedCases{_on} suffice for all
tuple case splitting tactics?

Konrad.

AndreasLoow:
I've been using PairCases in my proofs, and I think almost all usages of it could be replaced by an improved Cases except this special case: Consider doing induction on l in "!l x. [...]" where l is a list of 3-tuples and x has some non-tuple type. Then, when simplification is sufficient to solve [...] after Induct and the introduced list head element in the step case has been split, you can do the following "Induct >> TRY PairCases >> rw [...]". If you instead would have used a Cases tactic that splits iterated pairs, that tactic would have done cases on x in the induction base case, which could break things if the tactic succeedes.

Anyhow, I think the current status is this: If namedCases{_on} is fixed such that is only splits "properly" iterated pairs, i.e. (x, y, z) but not ((x, y), z), then I can use the implementation of namedCases{_on} to make Cases{_on} split pairs in a similar manner. The current problem with namedCases{_on} if I recall correctly is that it dies when given a pair of type ((x, y), z) to split -- instead it should just split such pairs into two parts in my opinion.

But given that PairCases has the above use case, maybe this PR can be merged and then I can create another PR with the modified Cases{_on} tactic if/when namedCases{_on} is fixed? (Both Cases{_on} and namedCases{_on} are needed, because they address different problems: Sometimes you just want to split and don't care about the introduced names, then Cases{_on} is appropriate, and sometimes the introduced names are important and then namedCases{_on} is appropriate.)

(Sorry for the long comment.)

mn200:
Instead of two entry-points, I think having an API that allows for the "don't care" name to be easily included is probably better
We might even consider modifying the handling of `Cases_on` so that the name splits can be indicated within the quotation.  E.g.:
```
   Cases_on `mylist (* _ "h t" *)`
```
Anyway, I'll merge this as a stop-gap.

mn200:
Thanks for your work on this!

PR #770, opened 2020-01-13
Improve sptree toSortedAList

talsewell:
Adding run-length-encoding to toSortedAList solves the problem with
sparse trees, so toSortedAList is now respectably fast to EVAL in
all cases. This version is slightly slower than the previous for
packed trees, but it doesn't seem worth trying to fix that.


mn200:
Looks good except for the Unicode violations.

talsewell:
Oops, I need to fix the unicode constraint more deeply into my brain.

mn200:
The Unicode thing bites us all.

mn200:
Thanks for the work!

PR #772, opened 2020-01-13
balanced_map: add a thm about ALOOKUP of toAscList

talsewell:
Adds a previously missing thm which is true if the comparison is antisymmetric.


mn200:
Thanks!

PR #735, opened 2019-09-19
The new HOL-Probability based on [0,+Inf]-measure theory

binghe:
Hi,

in this extremely huge PR, I present the first stage work of HOL's new formal probability theory based on `[0,+Inf]`-measure theory. I have been working on this project since July 2018, immediately after I saw some initial work done by HVG concordia from the artifacts of the DFT paper [1].

HOL's current probability theory is based on `[0,+inf)`-measure theory, i.e. the measure/probability is real-valued, while the integral/expectation is extreal-valued. In practice, the user theories have to involve theorems from both `realTheory` and `extrealTheory` and frequently convert values between them.  Besides, the `[0,+inf)`-measure theory is not sufficient to formalize many theoretical results, since it only supports finite measure spaces.

However, the existing probability theory provides good supports to some official examples: `<holdir>/examples/miller` and `<holdir>/examples/diningcryptos` (this one is currently broken). Also, for potential theories based on Elementary Probability Theory, real-valued probability and expectation values would be enough. Thus, I carefully consider all aspects and propose the following "dual" solution:

- Now HOL provides two Probability formalizations: one is fully based on extreals, at `<holdir>/src/probability/probabilityTheory`, the other is pure real-valued, at `<holdir>/src/probability/real_probabilityTheory`. (this version is based on code from the `diningcryptos` example, original by Aaron Coble).    The two theories cannot be loaded at the same time, as most constants (and theorems) have the same name. But they share some underlying theories, e.g. `util_probTheory` (utilities) and `sigma_algebraTheory`, the latter was part of `measureTheory`, now all definitions and theorems for systems of sets (algebra, sigma algebra, dynkin systems, ring, ...) are moved into this new `sigma_algebraTheory` as shared code. c.f. `<holdir>/src/probability/README.md` for the list of theory files and their purposes.

- Some common tactics used in all proof scripts of Joe Hurd, were now maintained in `<holdir>/src/probability/hurdUtils.sml` (which was in the miller example). The file name follows other similar libraries, e.g. `jrhUtils`, `schneiderUtils`.

- The new `extrealTheory` has been greatly modified with many fixes on definitions. Some of them were done by HVG people, e.g. preventing `PosInf + NegInf = PosInf` (now the result of this sum is unspecified), others were done by me, e.g. division-by-zero is now unspecified.

[...]

PR #774, opened 2020-01-16
Equivalent definitions of Borel sets (real_borelTheory)

binghe:
Hi,

recall in PR #574, HOL-Light's Topology theory was merged into HOL4 as `real_topologyTheory` for the new Probability formalization but even after the huge PR #735 it's still not used. The present PR finally starts to use it.

Currently in `real_borelTheory`, the Borel sets of `univ(:real)` is generated by the smallest σ-algebra containing all right-closed halfspaces:

```
   [borel_def]  Theorem      
      ⊢ borel = sigma 𝕌(:real) (IMAGE (λa. {x | x ≤ a}) 𝕌(:real))
```

Then it can be proven that other shapes of intervals and halfspaces are all in `borel`:

```
   [borel_measurable_sets]  Theorem
      ⊢ (∀a. {x | x < a} ∈ subsets borel) ∧
        (∀a. {x | x ≤ a} ∈ subsets borel) ∧
        (∀a. {x | a < x} ∈ subsets borel) ∧
        (∀a. {x | a ≤ x} ∈ subsets borel) ∧
        (∀a b. {x | a < x ∧ x < b} ∈ subsets borel) ∧
        (∀a b. {x | a < x ∧ x ≤ b} ∈ subsets borel) ∧
        (∀a b. {x | a ≤ x ∧ x < b} ∈ subsets borel) ∧
        (∀a b. {x | a ≤ x ∧ x ≤ b} ∈ subsets borel) ∧
        (∀c. {c} ∈ subsets borel) ∧ ∀c. {x | x ≠ c} ∈ subsets borel
```

But it's well known (see, e.g. [1, p.18-19]) that, the same Borel set can also be generated by all open sets, or other shapes of intervals and halfspaces, even with rational boundaries. However, these equivalent definitions are hard to prove, starting from the above definition.    A complete rework of this part was done by Muhammad Qasim et al. (HVG) in 2015, inspired from Isabelle/HOL. Now I edited and merged their code into HOL's `real_borelTheory`. The *new* definition of Borel set is based on open sets, the most abstract version:

```
   [borel]  Definition
      ⊢ borel = sigma 𝕌(:real) {s | open s}
```

Using advanced theorems provided by `real_topologyTheory`, the following equivalent definitions of Borel sets are then proven as theorems: (the original definition `borel_def` now becomes the synonym of `borel_eq_le`)

```
   [borel_eq_ge_le]  Theorem      
      ⊢ borel = sigma 𝕌(:real) (IMAGE (λ(a,b). {x | a ≤ x ∧ x ≤ b}) 𝕌(:real # real))
   
   [borel_eq_gr]  Theorem      
      ⊢ borel = sigma 𝕌(:real) (IMAGE (λa. {x | a < x}) 𝕌(:real))
   
   [borel_eq_gr_less]  Theorem      
      ⊢ borel = sigma 𝕌(:real) (IMAGE (λ(a,b). {x | a < x ∧ x < b}) 𝕌(:real # real))
   
   [borel_eq_le]  Theorem      
      ⊢ borel = sigma 𝕌(:real) (IMAGE (λa. {x | x ≤ a}) 𝕌(:real))
   
   [borel_eq_less]  Theorem      
      ⊢ borel = sigma 𝕌(:real) (IMAGE (λa. {x | x < a}) 𝕌(:real))
```

Furthermore, I make `real_borelTheory` loadable into the extreal-based probability theories, by moving some existing `real_borelTheory` theorems either into the shared `sigma_algebraTheory` or to `real_measureTheory`. Many duplicated theorems between the two probability formalizations, mostly with names `MEASURABLE_*`, are merged as a single shared version in `sigma_algebraTheory`.

It's important to have both `borel` and `Borel` accessible from the extreal-based probability, because the other `borel$Borel` is generated from the set of extended reals, which cannot be used to define `lborel` and `lebesgue` (the Borel and Lebesgue measurable sets with household Lebesgue measure) in the future. (I'm still working on this part, the next PR will focus on `borelTheory`.)

Then, the constant `eventually` in `real_topologyTheory` and `sequenceTheory` (`examples/miller`) have some conflicts. I resolve this problem by changing their definition names and then overload on `eventually`. This make some sense, because `eventually` is such a common name (also in linear temporal logic) that many theories are trying to use, better to overload instead of directly define it.

Regards,

Chun Tian

[1] Schilling, R.L.: Measures, Integrals and Martingales. Cambridge University Press (2005).

### DIFF
--- a/examples/miller/formalize/sequenceScript.sml
+++ b/examples/miller/formalize/sequenceScript.sml
@@ -32,7 +32,11 @@ val stake_def = Define
 
 val sdrop_def = Define `(sdrop 0 = I) /\ (sdrop (SUC n) = sdrop n o stl)`;
 
-val eventually_def = Define `eventually x y = ?n. sdrop n x = sdrop n y`;
+Definition eventually_def :
+    seq_eventually x y <=> ?n. sdrop n x = sdrop n y
+End
+(* cf. real_topologyTheory.eventually *)
+val _ = overload_on ("eventually", ``seq_eventually``);
 
 (* ------------------------------------------------------------------------- *)
 (* Theorems.                                                                 *)
--- a/examples/miller/formalize/sequenceScript.sml
+++ b/examples/miller/formalize/sequenceScript.sml
@@ -32,11 +32,7 @@ val stake_def = Define
 
 val sdrop_def = Define `(sdrop 0 = I) /\ (sdrop (SUC n) = sdrop n o stl)`;
 
-Definition eventually_def :
-    seq_eventually x y <=> ?n. sdrop n x = sdrop n y
-End
-(* cf. real_topologyTheory.eventually *)
-val _ = overload_on ("eventually", ``seq_eventually``);
+val eventually_def = Define `eventually x y = ?n. sdrop n x = sdrop n y`;
 
 (* ------------------------------------------------------------------------- *)
 (* Theorems.                                                                 *)
--- a/examples/miller/prob/probScript.sml
+++ b/examples/miller/prob/probScript.sml
@@ -9,8 +9,8 @@ open extra_listTheory hurdUtils extra_realTheory extra_boolTheory
 
 open util_probTheory sigma_algebraTheory real_measureTheory real_probabilityTheory;
 open subtypeTheory extra_pred_setTheory extra_pred_setTools;
-open sequenceTheory sequenceTools;
 open prob_algebraTheory prob_canonTools prob_canonTheory;
+open sequenceTheory sequenceTools;
 
 val _ = new_theory "prob";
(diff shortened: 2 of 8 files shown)


binghe:
Sorry, after more experiments I found the overload of "eventually" was unnecessary and I reverted these changes. The key fix is actually the changes in `probScript.sml` by moving `open sequenceTheory` to the last, and maybe also by providing extra type information when `eventually x y` appears in single-quoted terms.

mn200:
Thank for your ongoing work on this!

binghe:
Thanks. The "next PR" that I mentioned is also done, see #779

PR #777, opened 2020-01-29
Move l2n theorems from patricia_casts to numposrep

AndreasLoow:
Just two l2n theorems that should live in numposrep rather than patricia_casts. (Found them because I needed l2n_APPEND.)


mn200:
Thanks!

PR #786, opened 2020-02-18
Convergence (AE/PR/LP) and Weak Law of Large Numbers (uncorrelated r.v.'s)

binghe:
Hi,

this PR partially fulfills my dream since 2016 when I was attending the Advance Probability course of Prof. Massimo Campanino [1] (student of Bruno de Finetti [2]), i.e. an HOL formalization of convergence concepts of (real-valued) random sequences (and the Law(s) of Large Numbers, to be fully finished). The key is to make the convergence concepts presented in HOL terms as elegant as possible. After years of experiments, finally I choose the follow way: a single term `converge` (overloaded as `-->`) plus a new type for "convergence modes".

Given a sequence of real-valued r.v.'s, i.e. `(∀n :num. real_random_variable (X n) p)` and another real-valued random variable `Y`, I have defined the following three concepts of convergence (from `X n` to `Y`):

1. `converge_AE` (convergence “a.e.” or **almost everywhere**)

```
   [converge_AE_def]  Theorem
      ⊢ ∀p X Y.
            (X ⟶ Y) (almost_everywhere p) ⇔
            AE x::p. ((λn. real (X n x)) ⟶ real (Y x)) sequentially
```

2. `converge_PR` (convergence in pr., or **in probability**),

```
   [converge_PR_def]  Theorem
      ⊢ ∀p X Y.
            (X ⟶ Y) (in_probability p) ⇔
            ∀e.
                0 < e ∧ e ≠ +∞ ⇒
                ((λn.
                      real
                        (prob p {x | x ∈ p_space p ∧ e < abs (X n x − Y x)})) ⟶
                 0) sequentially
```

3. LP (convergence In `L^p` or in **Lebesgue integration**).

```
   [converge_LP_def]  Theorem
      ⊢ ∀p X Y r.
            (X ⟶ Y) (in_lebesgue r p) ⇔
            0 < r ∧ r ≠ +∞ ∧
            (∀n. expectation p (λx. abs (X n x) powr r) ≠ +∞) ∧
            expectation p (λx. abs (Y x) powr r) ≠ +∞ ∧
            ((λn. real (expectation p (λx. abs (X n x − Y x) powr r))) ⟶ 0)
              sequentially
```

(Other similar concepts include: **convergence vaguely** (or weakly), and convergence **in distribution**. The later is the key for the other famous limit theorem in Probability, namely the Central Limit Theorem. But this is not my task.)

The above three definitions all look very different with each other. And it's actually very hard to work directly with `converge_AE`, which says, except for a null set (or measure zero), for all other point `x` in the sample space, the limit of `X n (x)` is `Y (x)`. However, several equivalent theorems are proven to convert that "limit" to something else, namely `sup`, `inf`, `limsup` and `liminf`. The most important one is the following:

```
   [converge_AE_alt_limsup]  Theorem
      ⊢ ∀p X Y.
            real_random_variable Y p ⇒
            ((X ⟶ Y) (almost_everywhere p) ⇔
             ∀e.
                 0 < e ∧ e ≠ +∞ ⇒
                 prob p
                   (limsup
                      (λn. {x | x ∈ p_space p ∧ e < abs (X n x − Y x)})) = 0)
```

Recall that, "limsup" means "infinitely often" (cf. Borel-Cantelli Lemma). Now it should be clear why "convergence a.e." implies "convergence in pr.":  the former says that, it's impossible (zero probability) that `|X n x - Y x| > e` "infinitely often" when `n` increases. In another words, after certain `N`, for all `n > N` the probability of `|X n x - Y x| > e` is exactly zero.  Of course this implies "in pr." which only requires this probability be "close to zero".  The converse does not hold, simply because a probability very small and close to zero is still not zero, thus it's not a null set, and cannot fit with the definition of "convergence a.e.".

Indeed, we can prove that, AE implies PR:

```
   [converge_AE_imp_PR]  Theorem
      ⊢ ∀p X Y.
            real_random_variable Y p ⇒
            (X ⟶ Y) (almost_everywhere p) ⇒
            (X ⟶ Y) (in_probability p)

   [converge_AE_imp_PR']  Theorem
      ⊢ ∀p X.
            prob_space p ∧ (∀n. real_random_variable (X n) p) ⇒
            (X ⟶ (λx. 0)) (almost_everywhere p) ⇒
            (X ⟶ (λx. 0)) (in_probability p)
```

Notice that, in all above concepts, a convergence from `X` to `Y` and from `X - Y` to 0 are equivalent, for instance:

```
   [converge_AE_to_zero]  Theorem
      
      ⊢ ∀p X Y.
            real_random_variable Y p ⇒
            ((X ⟶ Y) (almost_everywhere p) ⇔
             ((λn x. X n x − Y x) ⟶ (λx. 0)) (almost_everywhere p))
```

On the other hand, LP convergence implies PR, but I only need and proved a special version (converge to zero, with integer-valued power:

```
   [converge_LP_imp_PR']  Theorem
      ⊢ ∀p X k.
            prob_space p ∧ (∀n. real_random_variable (X n) p) ⇒
            (X ⟶ (λx. 0)) (in_lebesgue (&k) p) ⇒
            (X ⟶ (λx. 0)) (in_probability p)
```

To further verify the correctness of the above definitions, I have proved a simple version of The Weak Law of Large Numbers for uncorrelated random variables with a common bound of variances. (Independence with Identical Distribution, or I.I.D., plus finite variance, implies this case.)

The so-called "Law of Large Numbers" deals with the partial sums of an infinite sequence of random variables `(∀n x. S n x = ∑ (λi. X i x) (count n))`. Suppose the mean (expectation) of `(S n)` is `M n`, and we treat `S (SUC n) x − M (SUC n)) / &SUC n` (their centered average) as another random variable. Then under certain conditions (usually indepent with identical distributions), this centered average converges to 0, in pr. (the weak law) or a.e. (the strong law).

For uncorrelated r.v.'s with a common bound of variance, the proof is simple and based on Markov or Chebeshev Inequalities, and the centered average actually converges to 0 in `L^2`. Thus also in pr.:

```
   [uncorrelated_def]  Definition      
      ⊢ ∀p X Y.
            uncorrelated p X Y ⇔
            finite_second_moments p X ∧ finite_second_moments p Y ∧
            expectation p (λs. X s * Y s) =
            expectation p X * expectation p Y

   [WLLN_uncorrelated_L2]  Theorem
      ⊢ ∀p X S M.
            (∀i j. i ≠ j ⇒ uncorrelated p (X i) (X j)) ∧
            (∃c. c ≠ +∞ ∧ ∀n. variance p (X n) ≤ c) ∧
            (∀n x. S n x = ∑ (λi. X i x) (count n)) ∧
            (∀n. M n = expectation p (S n)) ⇒
            ((λn x. (S (SUC n) x − M (SUC n)) / &SUC n) ⟶ (λx. 0))
              (in_lebesgue 2 p)

   [WLLN_uncorrelated]  Theorem
      ⊢ ∀p X S M.
            (∀i j. i ≠ j ⇒ uncorrelated p (X i) (X j)) ∧
            (∃c. c ≠ +∞ ∧ ∀n. variance p (X n) ≤ c) ∧
            (∀n x. S n x = ∑ (λi. X i x) (count n)) ∧
            (∀n. M n = expectation p (S n)) ⇒
            ((λn x. (S (SUC n) x − M (SUC n)) / &SUC n) ⟶ (λx. 0))
              (in_probability p)
```

All informal proofs are from the excellent probability book of Kai Lai Chung [3], who illustrates the Laws of Large Numbers better than all other books that I ever read.

Under the same condition, the "strong" law also holds, but the proof is slightly harder. I'll finish this work later, say, in a month or two.

P. S. for other changes, to make `converge_LP` possible, I have modified the definition of `extreal_ln`, ensuring `ln 0 = NegInf`, because otherwise `0 powr a` is not defined at all.  I also exported many extreal lemmas into automatic rewrites. The current chosen set of lemmas did not break any existing proof.

Regards,

Chun Tian

[1] https://www.unibo.it/sitoweb/massimo.campanino/en
[2] https://en.wikipedia.org/wiki/Bruno_de_Finetti
[3] Chung, K.L.: A Course in Probability Theory. Academic Press (2001).


binghe:
The theorem `WLLN_uncorrelated_L2` is broken after https://github.com/HOL-Theorem-Prover/HOL/commit/8751db679fc50b5d741f5c63a1a5cfa7d1808080 .

The following commit can be Cherry-Picked to fix this issue: https://github.com/binghe/HOL/commit/3aae96bd22ef0cd9ba5b98b90eff9e3bd68c9d09 (after merging #787).

binghe:
The previously mentioned patch has been merged into the submitted branch, which is synchronized with recent `develop` branch.

src/probability/extrealScript.sml
```diff
@@ -1133,28 +1178,38 @@ val extreal_sub_add = store_thm
...
-val sub_0 = store_thm
-  ("sub_0", ``!x y. (x - y = 0) ==> (x = y)``,
+Theorem sub_0 :
+    !x y :extreal. (x - y = 0) ==> (x = y)
+Proof
     rpt Cases
  >> RW_TAC std_ss [extreal_sub_def, num_not_infty, extreal_of_num_def, extreal_11,
-                   REAL_SUB_0]);
+                   REAL_SUB_0]
+QED
```
mn200: Can't this be an iff (and a [simp])?

binghe: No, `sub_0` cannot be an iff theorem without adding more antecedents: if `x = y = PosInf` or `x = y = NegInf`, then `x - y` is not specified, thus `x - y = 0` doesn't hold ...

mn200: OK - that's fair.  Thanks.

binghe:
> Please remove `Reverse` from `hurdUtils` and replace all uses with either `reverse` or `REVERSE` (both from `Tactical`). I don't want a third synonym...

OK, I'm removing all uses of `Reverse` in `hurdUtils` and proofs in `src/probability` (and also in `example/CCS`).  As `REVERSE` is also a name of a theorem in `rich_listTheory`, my only choice is `reverse`.

mn200:
Thanks.  Arguably, that choice of theorem name is a mistake in `rich_list`.  It is just an alias for `listTheory.REVERSE_DEF`, so I'm tempted to remove it.

binghe:
> Thanks. Arguably, that choice of theorem name is a mistake in `rich_list`. It is just an alias for `listTheory.REVERSE_DEF`, so I'm tempted to remove it.

Great. I'll make a separate PR to remove this problematic theorem. It's used in some proofs in at least `rich_listTheory` itself, and the removals may break some examples, thus requires more tests.

mn200:
Thanks for all the work you've been putting into this ongoing project!

PR #787, opened 2020-02-20
Fixed extrealTheory.inv_inj after 8751db6

binghe:
The original proof of `extrealTheory.inv_inj` strangely broke into three subgoals, of which the first two are identical. Now after 8751db679fc50b5d741f5c63a1a5cfa7d1808080 the proof gets simplified to just two subgoals. Now all probability code builds fine.


mn200:
Already fixed.

PR #791, opened 2020-03-06
Add some simple theorems about real numbers

HeikoBecker:
These theorems are taken from the FloVer repository (https://gitlab.mpi-sws.org/AVA/FloVer).
I renamed some of them to fit the style used in `realScript.sml`

### DIFF
--- a/src/real/realScript.sml
+++ b/src/real/realScript.sml
@@ -4293,8 +4293,7 @@ QED
 
 Theorem REAL_DIV_ZERO:
   !a b.
-    b <> 0 ==>
-    ((a / b = 0) <=> (a = 0))
+    (a / b = 0) <=> ((a = 0) \/ (b = 0))
 Proof
   rpt strip_tac \\ EQ_TAC \\ fs[REAL_DIV_LZERO, real_div]
 QED
--- a/src/real/realScript.sml
+++ b/src/real/realScript.sml
@@ -4237,34 +4237,12 @@ Proof
   simp[REAL_POW_POS]
 QED
 
-Theorem REAL_LE_LT_EQ:
-  ! a b.
-    (a <= b) = ((a < b) \/ (a = b))
-Proof
-  rpt strip_tac >> EQ_TAC >> strip_tac
-  >> qspecl_then [`a`, `b`] assume_tac REAL_LT_TOTAL >> fs[real_lte, REAL_LE_REFL]
-  >- (metis_tac[])
-  >- (metis_tac[REAL_LT_GT])
-  >- (`a < a` by (irule REAL_LT_TRANS >> qexists_tac `b` >> fs[])
-      >> metis_tac [REAL_LT_REFL])
-  >> metis_tac [REAL_LT_REFL]
-QED
-
-Theorem REAL_LE_EQ:
-  !a b.
-    a <= b /\ b <= a ==> (a = b)
-Proof
-  fs[REAL_LE_LT_EQ] >> rpt strip_tac >> fs[]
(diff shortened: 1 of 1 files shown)


binghe:
Hi, I have some comments:

1. `REAL_LE_LT_EQ` is duplicated with the existing `REAL_LE_LT`:
```
   [REAL_LE_LT]  Theorem   
      ⊢ ∀x y. x ≤ y ⇔ x < y ∨ (x = y)
```

2. `REAL_LE_EQ` is less general than the existing `REAL_LE_ANTISYM`:
```
   [REAL_LE_ANTISYM]  Theorem      
      ⊢ ∀x y. x ≤ y ∧ y ≤ x ⇔ (x = y)
```

A tactic `MATCH_MP_TAC REAL_LE_EQ` can be rewritten as `REWRITE_TAC [GSYM REAL_LE_ANTISYM]`.

3. `REAL_ABS_LE0` is a good addition. In `extrealTheory` there's a similar `abs_le_0`:
```
   [abs_le_0]  Theorem      
      ⊢ ∀x. abs x ≤ 0 ⇔ x = 0
```

4. `REAL_INV_LE_ANTIMONO` is a good addition, but I think the two other versions (`*_IMPL` and `*_IMPR`) are not that necessary in practice.

4. `REAL_MUL_LE_COMPAT_NEG_L` may be renamed to `REAL_LT_LMUL_NEG_IMP` following the conventions of other existing theorems:
```
   [REAL_LE_LMUL]  Theorem      
      ⊢ ∀x y z. 0 < x ⇒ (x * y ≤ x * z ⇔ y ≤ z)
   
   [REAL_LE_LMUL_IMP]  Theorem      
      ⊢ ∀x y z. 0 ≤ x ∧ y ≤ z ⇒ x * y ≤ x * z
   
   [REAL_LE_LMUL_NEG]  Theorem      
      ⊢ ∀x y z. x < 0 ⇒ (x * y ≤ x * z ⇔ z ≤ y)
```

--Chun

HeikoBecker:
Thanks @binghe for taking the time to review and your detailed comments.  I have addressed them as follows in the latest commit:
1. `REAL_LE_LT_EQ` is removed
2. `REAL_LE_EQ` is removed
5. renamed the theorem

On 4.: In the development from which I am porting these theorems, `REAL_INV_LE_ANTIMONO` was basically proven to obtain the `_IMPR/L` versions which I then make use of in proofs.
The main idea is that conditional rewriting was causing some pain there, whereas MP matching is straight-forward. Could one keep them for the ease of automation then?

binghe:
Thanks @HeikoBecker for making the additional changes. Now these new theorems all look reasonable to me, just I have no right to decide anything here:)

mn200:
Thanks for your work on this!

src/real/realScript.sml
```diff
@@ -4291,4 +4291,12 @@ Proof
...
 QED
 
+Theorem REAL_DIV_ZERO:
+  !a b.
+    b <> 0 ==>
+    ((a / b = 0) <=> (a = 0))
+Proof
+  rpt strip_tac \\ EQ_TAC \\ fs[REAL_DIV_LZERO, real_div]
+QED
+
```
mn200: I think you could make this an iff: `(a/b = 0) <=> (a = 0) \/ (b = 0)`

Though this then depends on the way `inv 0 = 0`, I think the added convenience is worth it.

HeikoBecker: Done in the latest commit. I did not even need to change the proof script.

PR #590, opened 2018-09-18
Fixed PSL examples 

binghe:
Hi,

The PSL (Property Specification Language) formal theory (`examples/PSL`) is a huge formalization work done by Thomas Tuerk, Klaus Schneider and Mike Gordon [1], currently it doesn't build in HOL from at least K10 to K12 releases.

In this PR, I have successfully fixed most scripts, under the following top-level directories (and their dependencies):

- `examples/PSL/1.01/official-semantics`
- `examples/PSL/1.01/executable-semantics`
- `examples/PSL/1.1/official-semantics`

I have tested and fixed the builds in Poly/ML with both stdknl and expk, using `rename1` tactic that I've learnt from Thomas Tuerk occasionally, yesterday:) 

In PSL's `regexpTheory`, special term syntax ``p | q`` has broken the GSPEC of `pred_setTheory`  and maybe others. I changed it to "p || q", just like what I did in CCS theory.

A single script `ExecuteSemanticsScript.sml` under `examples/PSL/1.1/executable-semantics` doesn't work. The last big changes were done by Joe Hurd. But in his commits, at least one newly added theorem `S_ANY` has something missing:

```
val S_ANY = store_thm
  ("S_ANY",
   ``!p. US_SEM p S_ANY = BOTTOM_FREE_LIST p``,
   RW_TAC std_ss [S_ANY_def, BOTTOM_FREE_LIST_def]
   ++ Induct_on `p`
   ++ ONCE_REWRITE_TAC [US_SEM]
   ++ RW_TAC std_ss [S_EMPTY, EVERY_DEF, LENGTH_NIL]
   ++ RW_TAC std_ss [S_TRUE_def]
   ++ Cases_on `p = []` >> RW_TAC std_ss [EVERY_DEF]
   ++ RW_TAC std_ss []
   ++ REVERSE EQ_TAC
   >> (RW_TAC std_ss []
       ++ Q.EXISTS_TAC `METIS_TAC [APPEND]
  ...
```

(After `Q.EXISTS_TAC`, a term is missing.)  I don't really understand these theories currently, thus I have no way to fix it. (If  Prof. Joe Hurd could see this message, please help)

Actually my initial goal was to fix `examples/temporal_deep` (it depends on `examples/PSL/1.1/official-semantics`), but it is much harder to fix this example. A partial work is at [2]. I hope their original author (Thomas Tuerk) could help me in fixing them.   (My ultimate plan is to formalize some of my results in LTL Runtime Verification based on these scripts.)

Regards,

[...]

PR #792, opened 2020-03-07
`temporal_deep` fixes, selftests and reformats

binghe:
Hi,

This PR fixes some remaining issues (due to tight equality changes) in the `temporal_deep` example, greatly re-formated, also with more selftests added. (These days I was reading it, and all changes are actually side effects during the reading process.)

The `temporal_deep` example [1] by Thomas Türk (master project, under the supervision of Prof. Klaus Schneider) provides currently the only formalization of the translation from Linear Temporal Logic (LTL) to **symbolic** Omega automata (aka "automata formula", essentially generalized Büchi). Other formalizations found in HOL (`examples/logic/ltl`) and Isabelle AFP [2] are based on **explicit-state** automata, i.e. graph-like structures.

All theorems were fixed in #590 and #596, but some ML functions in `temporal_deep/src/translations/translationsLib` are still broken due to the "tight equality" changes but the issues were not found as I never tried those API functions. Now, for the first time, I saw how the ML function `ltl2omega` translates an LTL formula `p U q` (p until q),  or `LTL_SUNTIL (LTL_PROP (P_PROP p),LTL_PROP (P_PROP q))`, into a symbolic Omega automata:

```
∀sv.
         IS_ELEMENT_ITERATOR sv 1 {q; p} ⇒
         ∀i.
             LTL_SEM i (LTL_SUNTIL (LTL_PROP (P_PROP p),LTL_PROP (P_PROP q))) ⇔
             A_SEM i
               (A_UNIV
                  (symbolic_semi_automaton (λs. ∃n. n < 1 ∧ s = sv n)
                     (P_BIGAND [])
                     (XP_BIGAND
                        [XP_EQUIV
                           (XP_PROP (sv 0),
                            XP_OR
                              (XP_PROP q,
                               XP_AND (XP_PROP p,XP_NEXT_PROP (sv 0))))]),
                   A_IMPL (A_BIGAND [],ACCEPT_COND_PROP (P_PROP (sv 0)))))
```

(In theory, the above translation result can be further synthesized by an ML function into an explicit-state Büchi automaton.)

I made all tests in `temporal_deep/src/examples/ltl2omega.sml` into selftests, triggered by the `selftest.sml` in the same folder.

The rest of changes are just code style/indent fixes, partially turned into HOL's new Theorem-defining syntax, with referenced papers added, etc.  All good changes for future learners (including myself), in my opinion.

Please squash all commits into a single commit if the PR will be merged.

Regards,

Chun Tian

[1] Tuerk, T.: A hierarchy for Accellera's property specification language, Thesis, University of Kaiserslautern (2005).
[2] https://www.isa-afp.org/entries/LTL_to_GBA.html


mn200:
Many thanks for this!

binghe:
By the way, how do you make sure `examples/temporal_deep/src/examples/selftest.exe` being actually used in the regression tests?   Is there a "test sequence" file to put it on?   I found that the other `selftest.exe` in examples/temporal_deep/src/model_check` depends on BDD and thus cannot run in PolyML, so obviously the current regression tests do not include it.

mn200:
If 
- a directory is mentioned in the build sequence at a level <= the build tool runs at;
- that directory includes a selftest.exe file after `Holmake` finishes; and
- the test level >= 1

then the selftest.exe file will be run.

To get your example to run in the regression machinery, add its directory to the `large-theories` file with an appropriate test-level annotation.

binghe:
> If
> 
>     * a directory is mentioned in the build sequence at a level <= the build tool runs at;
> 
>     * that directory includes a selftest.exe file after `Holmake` finishes; and
> 
>     * the test level >= 1
> 
> 
> then the selftest.exe file will be run.
> 
> To get your example to run in the regression machinery, add its directory to the `large-theories` file with an appropriate test-level annotation.

Thanks. Let me confirm one more thing. So the machinery doesn't look into sub-directories? For instance, currently we have this line in `large-theories`:

```
!examples/temporal_deep
```

But since the `selftest.exe`s that I mentioned are all in some sub-directories, thus actually none of them will be running? And to resolve this issue, I should add something (after the above line) like:

```
[mosml]!!examples/temporal_deep/src/model_check
!!examples/temporal/src/examples
```

And the "level" you mentioned is the number of `!` (excalmatory mark)?

mn200:
Yes to all of your questions.  Holmake will recurse, following `INCLUDES` links, but the running of selftest.exe is handled by `build`.

binghe:
> Yes to all of your questions. Holmake will recurse, following `INCLUDES` links, but the running of selftest.exe is handled by `build`.

Thanks again. Then I was wrong in removing `examples/miller/miller` from the test sequence (replaced by just `examples/miller`) and caused the `selftest.exe` there disabled.

I'll fix all these issues, making sure all desired level 1 and 2 tests run in my own computer, then submit another short PR.

PR #793, opened 2020-03-10
Add some floating-point roundtripping thms

HeikoBecker:


mn200:
It looks as if this PR depends on the other one; the error is about a theorem not being present.

HeikoBecker:
Arghs, sorry. Maybe then I should hold back on the PR until my other pull request is finished.

Issue #721, opened 2019-08-08
Reorient "bad" equalities in simplifier


mn200:
Implement a conversion that flips equalities which have `l = r` where `l` has no free variables and `r` does. I think this conversion could then be on by default in the stateful simpset.

arolle:
I implemented `REORIENT_EQ_CONV` (in a08046e60e23a69fd8f7b455f8d84b5143f643a0). In which script file should the stateful simpset be extended?

mn200:
Please embed the conversion in a simpset fragment (something like `REORIENT_EQ_ss`) in `boolSimps` and then separately add that fragment into the stateful set up in `BasicProvers`.

Make sure the conversion and the fragment have names so that they can be readily removed.

arolle:
`boolSimps.NORMEQ_CONV` and `boolSimps.NORMEQ_ss` seem to already implement this functionality.

mn200:
Closed by #794 and follow-up work.

PR #794, opened 2020-03-10
Reorient equalities through the simplifier

arolle:
Addresses github issue #721


binghe:
The addition of `NORMEQ_ss` into default simp sets seems to potentially break many existing proof scripts.

mn200:
It looks to me as if the damage is not too great.  With the changes in d8cccb5 it now seems as if it is just some examples that are failing.  If @arolle can get those fixed as well, I'd be happy to merge this in.

mn200:
But, one other thing, please make the PR to merge into the `develop` branch; not `master`.

mn200:
Thanks for your work on this!

PR #798, opened 2020-03-20
Strong Law of Large Numbers for uncorrelated r.v.'s (SLLN_uncorrelated)

binghe:
Hi,

this PR continues #786 and formally proved a version of the Strong Law of Large Numbers (SLLN) (for uncorrelated random variables with a common bound of variance).  Perhaps this is the first formalization of SLLN in any theorem prover, even the present one is just a simple version: (hard version = no assumptions on variance at all, just with finite expectation)

```
   [SLLN_uncorrelated]  Theorem
      ⊢ ∀p X S M.
            (∀i j. i ≠ j ⇒ uncorrelated p (X i) (X j)) ∧
            (∃c. c ≠ +∞ ∧ ∀n. variance p (X n) ≤ c) ∧
            (∀n x. S n x = ∑ (λi. X i x) (count n)) ∧
            (∀n. M n = expectation p (S n)) ⇒
            ((λn x. (S (SUC n) x − M (SUC n)) / &SUC n) ⟶ (λx. 0))
              (almost_everywhere p)
```

The statement is the same with the version Weak Law of Large Number proved in #786 except for `almost_everywhere` in place of `in_probability` in the conclusion. Note that (pairwise) independence implies uncorralation, thus from the present theorem we can derive SLLN/WLLN for I.I.D. (independent with identical distribution) r.v.s with finite variance.

Again, as usual, the formalization of such a fundamental theorem in probability theory has stimulated the additions of some other useful lemmas in related theory files, such as `logrootTheory`, `extrealTheory` etc. I believe these additions are all reasonable.  Besides, I added the following overloads in `realTheory` for `x pow 2` and `x pow 3`:

```
(* from arithmeticTheory.EXP *)
val _ = overload_on (UnicodeChars.sup_2, “\x. x pow 2”);
val _ = overload_on (UnicodeChars.sup_3, “\x. x pow 3”);
```

These overloads and pretty printings are actually defined in `arithmeticTheory`, and the extreal versions of `pow` also used them already. Adding them for reals seems natural.

Future work: for a sequence of (totally) independent real-valued (i.e. `extreal` type but never takes `+/-Inf` as values) random variables without finite variances, the related Laws of Large Numbers are the really complicated and interested ones.  Their proofs are several times larger and stand at a chain of other useful results, still need a long time to complete.

Regards,

Chun Tian

### DIFF
--- a/src/num/extra_theories/logrootScript.sml
+++ b/src/num/extra_theories/logrootScript.sml
@@ -113,6 +113,14 @@ val ROOT_UNIQUE = Q.store_thm("ROOT_UNIQUE",
    THEN METIS_TAC [DECIDE ``a < b ==> SUC a <= b``, exp_lemma3, LESS_EQ_TRANS,
                    DECIDE ``a <= b ==> ~(b < a:num)``, ROOT]);
 
+Theorem ROOT_EXP :
+    !n r. 0 < r ==> ROOT r (n ** r) = n
+Proof
+    rpt STRIP_TAC
+ >> MATCH_MP_TAC ROOT_UNIQUE
+ >> RW_TAC arith_ss []
+QED
+
 val log_exists = Q.prove(
    `!a n. 1 < a /\ 0 < n ==> ?log. a ** log <= n /\ n < a ** SUC log`,
    REPEAT STRIP_TAC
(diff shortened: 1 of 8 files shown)


mn200:
Thanks for this!  Incidentally, please do write up a summary of all this recent work for the release notes.

binghe:
> Thanks for this! Incidentally, please do write up a summary of all this recent work for the release notes.

I thought about this, but the entire new probability (based on [0,+inf]-measure theory) is new since the last release, and thus it seems that any individual new theorems are too "small" to be mentioned in `next-release.md`.

However, if you agree, I can add some key theorem names into the release notes, together with my next PR. Besides, I would be more than happy to completely revise the probability-related part of "The HOL Description", starting from the new `extreal` arithmetics with fixed semantics of `+`, `-`, `/`, `inv`, etc.

mn200:
Both of these ideas sound great.  Many thanks.

PR #803, opened 2020-04-01
Fixed building of various HOL OpenTheory packages

binghe:
Hi,

HOL's OpenTheory building (`--otknl`) is actually broken, the building process didn't fail only because those key `art` files for uploadable OpenTheory packages were not built within the building process.  I've learnt a lot about HOL's OpenTheory support from the code changes of #702, and now I'm able to fix some obvious issues and get new versions of uploadable HOL OpenTheory packages and install them into my local OT installation:

```
hol-base-1.3
hol-floating-point-1.2
hol-integer-1.1
hol-monad-1.1
hol-quotient-1.1
hol-real-1.1
hol-res-quan-1.1
hol-ring-1.2
hol-sort-1.0
hol-string-1.2
hol-words-1.2
```

The key fixes are in `src/boss/prove_base_assumsScript.sml` and `src/real/prove_real_assumsScript.sml`, in which some theorems were broken due to the "tight equality" changes. The latest version of OpenTheory's `base` package is changed, this is also reflected in `prove_base_assumsScript.sml`.

The other key fix is at `src/opentheory/postbool/Opentheory.sml`. According to the build sequence, this file is built before `listTheory`, yet it depends on `listTheory` (to use its syntax). I cannot move the entire directory into a later position in the build sequence, because other files in the directory are needed much earlier. So I created a new directory `src/opentheory/reader` and moved that file there, also with a more reasonable structure name `OpenTheoryReader`, then added this new direcotyr into the build sequence after `listTheory`.

Then I fixed many `Holmakefile`s for the changed theory dependencies, changed theory names and directories. If an HOL OpenTheory package (e.g. `hol-base`) does not depend on other HOL packages, I let them be built by default. This increased the total needed time for finishing the entire build, by about half hour in a normal computer.

Finally, I added two new OpenTheory package definitions (`hol-analysis` and `hol-probability`) in `src/probability`. The long-term goal is to have my probability formalization work loadable into other theorem provers (so that they don't need to redo it again). But unfortunately I can't confirm if the packing work is really working, because the building of `real_topology.ot.art` requires more than 16GB memory that I currently have in my laptop. But all other `.art` and `.ot.art` files can be built correctly.

One last issue: the recently added `itself` type in `boolTheory` is a problem for OpenTheory packaging, because HOL's `boolTheory` is NOT part of `hol-base`.  An easy solution (my suggestion) is to move the `itself` type stuff into `oneTheory` or something like it, and we prevent touching `boolTheory` with essential new things from now on.

Regards,

Chun Tian


mn200:
This looks reasonable to me.  Can you make an issue for discussing the remaining issues please?  In particular, I don't like the idea of freezing our `bool` theory just for the sake of OpenTheory compatibility.  The particular suggestion you make of moving `itself` to `one` is a nice idea though.

binghe:
> This looks reasonable to me. Can you make an issue for discussing the remaining issues please? In particular, I don't like the idea of freezing our `bool` theory just for the sake of OpenTheory compatibility. The particular suggestion you make of moving `itself` to `one` is a nice idea though.

Thanks. I have filed #805 for three remaining issues that I found.

PR #804, opened 2020-04-02
Move subtopology theorems from real_topology to (general) topology

binghe:
Hi,

this PR slightly updated `real_topologyTheory` by moving general theorems about "subtopology" (independent of real numbers) back to the more general `topologyTheory`, which is now further enriched.

It also redefined `ball` and `dist` using definitions from `metricTheory`. This change doesn't help too much at this moment, but could be very useful the day when `real_topologyTheory` tries to support multi-dimensional Euclidean spaces.

Also, I think the two old scripts (`analysis.ml` and `transc.ml`) in `src/real` can be removed now: for whatever leftover in them, we should look for the `real_topologyTheory` and more theories to be ported from HOL Light.

--Chun


binghe:
In addition, `real_topologyTheory` has defined more general versions of `suminf`, `summable` and `sums` on restricted (infinite) integer sets. I made some connections between them and the corresponding definitions in `seqTheory` (and `realTheory`), so that user can easily migrate between related theorems:

```
   [suminf_univ]  Theorem
      
      ⊢ ∀f. suminf 𝕌(:num) f = suminf f
   
   [summable_univ]  Theorem
      
      ⊢ ∀f. summable 𝕌(:num) f ⇔ seq$summable f
   
   [sums_univ]  Theorem
      
      ⊢ ∀f l. (f sums l) 𝕌(:num) ⇔ seq$sums f l
```

binghe:
On a second thought, I shouldn't just remove `analysis.ml` and `transc.ml` without checking what exact leftovers are there.  I think I can do a better job here: do a full review of the two legacy scripts and port all theorems to HOL4 if anything is not ported yet (then remove old files). Close this PR for now, other changes will be part of next wave of probability additions.

PR #806, opened 2020-04-07
Move [compute] annotation for subspt to right place

HeikoBecker:
This pull request moves the `[compute]` annotation from `subspt_def` to `subspt_eq` which is more useful for `EVAL`

### DIFF
--- a/src/finite_maps/sptreeScript.sml
+++ b/src/finite_maps/sptreeScript.sml
@@ -1434,7 +1434,7 @@ val subspt_eq = Define `
      (spt_center t = SOME x) /\
      subspt t1 (spt_left t) /\ subspt t2 (spt_right t))`
 
-val _ = save_thm("subspt_eq[compute]",subspt_eq);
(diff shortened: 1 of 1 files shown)


mn200:
Please try omitting the `compute` on the `_eq` theorem; it should be in the compset already.

HeikoBecker:
I am testing it locally and will report back as soon as possible

mn200:
Many thanks!

HeikoBecker:
Sorry for taking so long on this. I have tested it locally, and removing the `compute` annotation from `subspt_def` is sufficient.

HeikoBecker:
One more thing: Is there a reason why `lookup_def` is not added to the compset? Can I add a `[compute]` annotation there too?

mn200:
There's a `lookup_compute` that does a better job.

HeikoBecker:
Thanks. I did not see that one.

mn200:
Thanks for this!

PR #809, opened 2020-04-13
Hyperlinks and sections (based on initial letters) in «REFERENCE»

binghe:
Hi,

«The HOL REFERENCE» (PDF) was having no bookmarks nor hyperlinks but one single big chapter. I wanted to add single-letter sections into it so that I can easily navigate into the document. Also, by loading `hyperref` package all page numbers in the index is now clickable.  In addition, the "Index" itself is now a chapter shown in PDF bookmark.

(This is a rare PR where I write real SML program code instead of formal proofs.)

See attached screenshot for the typesetting results.

--Chun

2020-04-13 11 36 48 pm

### DIFF
--- a/Manual/Reference/reference.tex
+++ b/Manual/Reference/reference.tex
@@ -6,7 +6,7 @@
 %\usepackage{amsmath}
 \usepackage{alltt}
 \usepackage{../LaTeX/layout}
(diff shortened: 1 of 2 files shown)


mn200:
Very nice!  Thanks for this.

PR #812, opened 2020-04-18
Small list theory fixes

AndreasLoow:
The TAKE_* fixes should be uncontroversial I think. Not sure if the DROP_NIL->DROP_EQ_NIL commit should be merged. I think the name is better, and the statement no longer includes >=, but it might break old proofs.

### DIFF
--- a/doc/next-release.md
+++ b/doc/next-release.md
@@ -136,6 +136,15 @@ New examples:
 Incompatibilities:
 ------------------
 
+*  `listTheory` and `rich_listTheory`: Some theorems have been generalized.
+
+	For example, `EVERY_{TAKE, DROP, BUTLASTN, LASTN}` had unnecessary preconditions.
+	As a result of some theorems being generalized, some `_IMP` versions of the same
+	theorems have been dropped, as they are now special cases of the non-`_IMP` versions.
+
+	Also, `DROP_NIL` has been renamed to `DROP_EQ_NIL`, to avoid having both `DROP_nil`
+	and `DROP_NIL` around. Furthermore, `>=` in the theorem statement has been replaced with `<=`.
+
 * * * * *
 
 <div class="footer">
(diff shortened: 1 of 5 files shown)


AndreasLoow:
Seems that problems similar to what the TAKE_* commit fixes are everywhere in rich_listTheory:

We have e.g. `BUTLASTN_TAKE` as `[] ⊢ ∀n l. n ≤ LENGTH l ⇒ BUTLASTN n l = TAKE (LENGTH l − n) l`, but one can prove:

```
Theorem BUTLASTN_TAKE':
 ∀n l. BUTLASTN n l = TAKE (LENGTH l − n) l
Proof
 rpt gen_tac \\ Cases_on ‘n ≤ LENGTH l’ >- metis_tac [BUTLASTN_TAKE] \\
 simp [BUTLASTN_def] \\
 ‘LENGTH l - n = 0’ by decide_tac \\ pop_assum (rewrite_tac o single) \\
 simp [DROP_LENGTH_TOO_LONG]
QED
```

Also, we have `BUTLASTN_compute` as:

```
   ⊢ ∀n l.
         BUTLASTN n l =
         (let
            m = LENGTH l
          in
            if n ≤ m then TAKE (m − n) l
            else FAIL BUTLASTN $var$(longer than list) n l): thm

```
                                                           
So we get the following strangeness:

```
(REWRITE_CONV [BUTLASTN_def] THENC EVAL) “BUTLASTN 5 [1]” evals to []
EVAL “BUTLASTN 5 [1]” evals to FAIL BUTLASTN $var$(longer than list) 5 [1]
```

mn200:
Once upon a time, the theory was that these functions should not be total and they were defined accordingly.  But they've since been redefined to be total (returning `[]` when the numbers are wrong), but many theorems and behaviours haven't been adjusted to catch up with this.

mn200:
Please write an backwards-incompatibility paragraph for the release notes documenting that a bunch of theorems have had their statements change.

mn200:
Thanks for all the work on this!

PR #815, opened 2020-05-02
Add compute cases for 0 for BIT_UPDATE

AndreasLoow:
Previously we had

```
> EVAL ``(0 :+ T) (0w:word8)``;
val it =  [] ⊢ (0 :+ T) 0w = (0 :+ T) 0w: thm
```

Now we instead have

```
> EVAL ``(0 :+ T) (0w:word8)``;
val it =  [] ⊢ (0 :+ T) 0w = 1w: thm
```


mn200:
Thanks!

PR #818, opened 2020-05-22
lemmata on llist

arolle:
relating mainly LFILTER, every and LAPPEND

The proofs of the following propositions are originally by Johannes Åman
Pohjola:
exists_thm_strong
LFILTER_fromList
LFILTER_EQ_CONS
every_LFILTER

### DIFF
--- a/src/coalgebras/llistScript.sml
+++ b/src/coalgebras/llistScript.sml
@@ -743,6 +743,12 @@ val LAPPEND = new_specification
 val _ = export_rewrites ["LAPPEND"]
 val _ = computeLib.add_persistent_funs ["LAPPEND"]
 
+Theorem LCONS_LAPPEND:
+  !x ll. x:::ll = LAPPEND [|x|] ll
+Proof
+  fs[LAPPEND]
+QED
+
 (* properties of map and append *)
 
 val LMAP_APPEND = store_thm(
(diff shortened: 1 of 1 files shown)


arolle:
Some of the proofs are ATM in the CakeML repo. The names are inspired from listTheory.
Let me know if the style is acceptable and the lemmas are relevant. 
@IlmariReissumies: FYI

IlmariReissumies:
Thanks, it's good to get the `llist` theory more fleshed out! It looks like a few of these might overlap with what @myreen and I are currently doing on the `llist` branch here.

  https://github.com/HOL-Theorem-Prover/HOL/tree/llist

For example, both branches have an `LFILTER_fromList`.

I also wonder if `every_fromList_EVERY` would be more useful as an equality instead of an implication? Moreover, do we really need `LCONS_LAPPEND`? It looks like it follows immediately from two applications of the `LAPPEND` definition.

mn200:
I agree that `LCONS_LAPPEND` looks unnecessary.  

Could @myreen, @IlmariReissumies and @arolle figure this all out on the other branch and eventually make a PR from there please?

arolle:
I was unaware of the developments on the `llist` branch and will push thereto.

Although trivial, the theorem `LCONS_LAPPEND` was motivated as being the `llist` variant of `rich_listTheory.CONS_APPEND`.

mn200:
*“Do as I say, not as I do.”*. 

There is historical cruft in older theories that doesn't deserve to be replicated...

PR #821, opened 2020-05-27
Cleanup of append thms for SORTED

AndreasLoow:
I think this is an improvement over the old situation. The main improvement is that SORTED_APPEND is now an (conditional) equality instead of an implication.


mn200:
Looks good to me; thanks!

Issue #661, opened 2019-02-18
Lets for pure values for monad syntax


AndreasLoow:
Would it be possible to add a way to bind pure values in HOL's monad syntax in the same way one can do in Haskell's monad syntax? For example, something such that the following

```
compile_exp n (BUOp Not e) =
 let (n, tmpvar) = compile_new_name n in do (* bad order... *)
  (n, inp, cs) <- compile_exp n e;
  return (n, VarInp tmpvar, SNOC (Cell1 CNot tmpvar inp) cs)
od
```

could be written as

```
compile_exp n (BUOp Not e) = do
 let (n, tmpvar) = compile_new_name n; (* bad order... *)
 (n, inp, cs) <- compile_exp n e;
 return (n, VarInp tmpvar, SNOC (Cell1 CNot tmpvar inp) cs)
od
```

mn200:
I'd like something in this vein myself, but unfortunately the parser is not clever enough to allow for reuse of the `let` keyword.  Is there some variant you can think of that would still be preferable to having the `let` on the outside?

AndreasLoow:
`mlet` maybe? That is,

```
compile_exp n (BUOp Not e) = do
 mlet (n, tmpvar) = compile_new_name n; (* bad order... *)
 (n, inp, cs) <- compile_exp n e;
 return (n, VarInp tmpvar, SNOC (Cell1 CNot tmpvar inp) cs)
od
```

mn200:
Looks reasonable to me.

mn200:
Should such `let`s compile to a standard HOL `LET` underneath, or should it compile to a bind of a unit?  For example, 

    “let x = e1 in monadic_e2”

*vs*

    “x <- unit e1; monadic_e2”

AndreasLoow:
Hm, what I had in mind was the LET-based translation, but I'm not really sure about the pros and cons between the different alternatives.

mn200:
Closed by da74731

PR #825, opened 2020-06-05
Add monadic let syntax

hrutvik:
Enable parsing and printing of equivalent `mlet x <- y` and `x <<- y` syntax, which expand to regular let-bindings. Progress towards #661.

Users have the option of printing in either style using `monadsyntax.print_explicit_monad_lets : bool -> unit` (default `false`, i.e. `x <<- y` form).
Note that expansion to regular let-binding means that `do mlet x <- y; ... od` will print as `let x = y in do ... od`. Regular let-bindings in do-blocks will be printed as monadic lets where scoping permits, e.g. `do a <- b; let c = d in do g <- e; f od od` prints as `do a <- b; c <<- d; g <- e; f od`.

### DIFF
--- a/src/monad/monadsyntax.sig
+++ b/src/monad/monadsyntax.sig
@@ -19,6 +19,8 @@ sig
   val disable_monadsyntax : unit -> unit
   val temp_disable_monadsyntax : unit -> unit
 
+  val print_explicit_monadic_lets : bool -> unit
+
   type monadinfo =
        { bind : term,
          ignorebind : term option,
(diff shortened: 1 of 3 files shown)


mn200:
Many thanks for your work on this!

AndreasLoow:
> Note that expansion to regular let-binding means that `do mlet x <- y; ... od` will print as `let x = y in do ... od`.

I just ran into this and got confused by it (I thought it was a bug). Any particular reason not to print it as `do mlet x <- y; ... od`? I guess you get the opposite problem then, i.e. that things given as `let x = y in do ... od` will be printed as `do mlet x <- y; ... od` -- but this seems less harmful since the mlet syntax is more readable.

mn200:
It's a bit tricky to have this work given the current architecture, but I do think it should be possible.

PR #829, opened 2020-06-12
Make TFL return a tailored induction theorem for primitive recursive functions too

myreen:
This `ind_for` function is a cleaned up version of a function that
has existed for sometime in the CakeML translator. I'm moving the
function into `DefnBase` because it can be of more general use.

Example use:

    Datatype:
      prog = Assign num num | Seq (prog list)
    End

    Definition eval_def:
      eval s (Assign n v) = s⦇n ↦ v⦈ ∧
      eval s (Seq progs) = eval_list s progs ∧
      eval_list s [] = s ∧
      eval_list s (p::ps) = eval_list (eval s p) ps
    End

The definition of eval does not produce an induction theorem.
However, in proofs about eval it is often desirable to have
an induction theorem that is tailor to the types and exact
arguments used by eval.

With the new addition to `DefnBase`, one can get hold of such
an induction theorem as follows:

    Theorem eval_ind = DefnBase.ind_for eval_def


myreen:
@mn200 I wonder if a call to `DefnBase.ind_for` could be baked into the `Definition` syntax.

mn200:
> @mn200 I wonder if a call to `DefnBase.ind_for` could be baked into the `Definition` syntax.

Will it return the induction theorem that is generated by TFL for its recursions, and what will it return for definitions that aren't even recursive at all?

mn200:
If it uniformly returned something sensible, the Definition syntax could call it every time...

myreen:
> Will it return the induction theorem that is generated by TFL for its recursions, and what will it return for definitions that aren't even recursive at all?

This current implementation only derives an induction theorem for definitions of recursive functions for which TFL does not produce an induction theorem (the case of primitive recursive). 

The tool in this PR produces a slightly more general induction theorem than the one TFL would produce in case one were to "tricky" TFL into the general case that always produces an induction theorem. 

I haven't tried to apply this tool to non-recursive functions.

**Here's a suggestion:** how about moving this tooling into TFL? In particular: how about making the primitive recursive function case of TFL use this routine to produce an induction theorem where it currently would not produce an induction theorem? With this modification all recursive functions defined by TFL would come with a nice induction theorem.

mn200:
Sounds a good plan to me!

konrad-slind:
So if a function is prim-rec on one of its arguments, just stick the
datatype induction theorem in? This is easy. In fact, the code probably
makes a test for prim-rec-ness and deliberately doesn't build an ind-thm.
Could just be a matter of removing the test.

 --Konrad.

konrad-slind:
What is the use-case for this feature? Is it that you always want an
induction theorem for a definition, if possible?

Konrad.

konrad-slind:
Looks like changing just

https://github.com/HOL-Theorem-Prover/HOL/blob/0f363774a407fb6c84f2228918813c94dc374395/src/tfl/src/Defn.sml#L500

might do the trick. That will store the ind_thm in the theory. So later
references to "foo_ind" should work for prim.rec defns.

In original concept, the DefnBase would hold defns, and then brittle names
wouldn't have to be used.

Konrad.

myreen:
In reply to @konrad-slind:

> So if a function is prim-rec on one of its arguments, just stick the datatype induction theorem in?

Yes, but in many cases it is convenient to have an induction theorem that is tailored to the exact arguments (and their types) as used by a function definition. This is useful in automation such as the CakeML translator and in interactive proofs. 

This `ind_for` function has been lived in the CakeML translator, but I find that I want to use it in files that should not depend on CakeML. That's why I want to find a place to put this that is more generally accessible.

> This is easy.

The automation in this pull request is simple. It just instantiates the datatype induction theorem to the arguments and types used in the function definition.

Would a satisfactory change be to (1) not change `DefnBase` but instead (2) change the lines:
```
  | save_defn (PRIMREC{bind,eqs, ...})       = been_stored (bind,eqs)
  | save_defn (NONREC {eqs, ind, stem, ...}) = store(stem,eqs,ind)
```
to: 
```
  | save_defn (PRIMREC{bind,eqs, ...})       = store(stem, eqs, ind_for eqs)
  | save_defn (NONREC {eqs, ind, stem, ...}) = store(stem, eqs, ind)
```
where `ind_for` is the new code of this pull request now moved to live in `HOL/src/tfl/src/Defn.sml`?

konrad-slind:
That is possible. Or else, closer to where the actual primrec defn is made:

https://github.com/HOL-Theorem-Prover/HOL/blob/0f363774a407fb6c84f2228918813c94dc374395/src/tfl/src/Defn.sml#L1174

Your choice!

Konrad.

myreen:
Thanks @konrad-slind! 

I have renamed this pull request to reflect that it is now about adding this functionality to TFL rather than `DefnBase`.

I will close this pull request while I move `ind_def` inside TFL and test my changes.

PR #832, opened 2020-06-18
Remove makeidx from holtexbasic

AndreasLoow:
For some reason I cannot build acmart-based articles when makeidx is loaded.

(Not sure how much sense it makes that the same file also loads fancyvrb and other things, but those things do not interfere in this specific case so I did not touch them.)


mn200:
Thanks - I've seen the same thing too.

Issue #831, opened 2020-06-16
Create an encapsulated `vimhol` entrypoint


hrutvik:
This issue follows on from Arve Gengelbach's post on the CakeML Slack.

A `vimhol` entrypoint to HOL in `tools/vim` would allow Vim-users to get up and running with HOL immediately after building. At the very least, this should provide a split pane Vim/HOL with a fresh shared FIFO, with convenient exit and so on as in Arve's version. This will also enable easy interactive use of multiple concurrent HOL sessions. The relevant Vim and HOL configs should be taken care of without user intervention.

A good starting point for this would be [Arve Gengelbach's script](http://user.it.uu.se/~arvge836/holtmux.html). He has said it would be worth removing the `-n` flag to ensure a new session is always created, and noted that `--no-run-if-empty` is a GNU `xargs` feature.

No doubt there are other features that could be conveniently encapsulated - suggestions are welcome!

arolle:
An alternative option to consider is to wrap hol in vim. Since [version 8.1 (released May 2018)](https://www.vim.org/vim-8.1-released.php) Vim supports terminal buffers. Although it seems that tmux hooks have been around for longer, than vim 8.1. Wrapping hol in vim might avoid quirks with colouring (i.e. ANSI escape codes) and tmux, as @oskarabrahamsson earlier experienced and solved by adding to the hol config file:
```
Parse.current_backend := PPBackEnd.vt100_terminal
```

mn200:
I await the relevant Pull Request with eager anticipation!

arolle:
The script in the pull-request is tested to work even with vim 7.4 (from 2013) and tmux 1.6, which does not support tmux hooks nor `-c` flag.  `vim` sources `$HOLDIR/tools/vim/filetype.vim`.  And in order for `hol` to use `$HOLDIR/tools/vim/hol-config.sml` the `HOME` directory is set to `$HOLDIR/tools/vim`, which I consider a hack.

mn200:
Closing this now that PR #833 has been merged.

PR #833, opened 2020-06-21
adds vimhol entry point to HOL, with vim and tmux

arolle:
addresses github issue  #831


mn200:
Please consider writing some documentation, either for release notes or perhaps a longer-form companion to the existing "Interaction" manual (which is mostly about the emacs mode).

Issue #842, opened 2020-07-14
ex/l3/riscv: RISC-V ISA specification of RVC `lui` instruction is broken


zaklogician:
This issue concerns the the L3 Specification of the RISC-V ISA located under [examples/l3-machine-code/riscv](examples/l3-machine-code/riscv) and by extension affects the RISC-V version of the binary-to-graph decompiler in [examples/machine-code/graph](examples/machine-code/graph).

**Summary:** The current L3 specification of the RISC-V ISA contains an incorrect decoding for the RVC instruction `LUI`, resulting in incorrect binary-to-graph decompilation of any binary that uses the compressed format `LUI` instruction.

**Example:**
The binary-to-graph decompiler translates the RISC-V assembly instruction
```
ffffffff84010426:	6705                	lui	a4,0x1
```
into the graph-lang node
```
0x84010426 Basic 0x84010428 1 r14 Word 64 Num 16777216 Word 64
```
instead of the expected
```
0x84010426 Basic 0x84010428 1 r14 Word 64 Num 4096 Word 64
```

Notice how the actual constant `16777216` (`0x1000000`) is left-shifted compared to the expected constant `4096` (`0x1000`).

**Cause:**

According to the RISC-V compressed instruction set specification, `C.LUI` should load the non-zero 6-bit immediate field into bits 17 to 12 of the destination register, while clearing the bottom 12 bits and sign-extending bit 17 into all higher bits of the destination register.

In the current L3 specification of `C.LUI`, clearing the bottom 12 bits is performed twice: once in the definition of `ArithI > LUI` at [riscv.spec:2156](examples/l3-machine-code/riscv/src/riscv.spec), and again in `DecodeRVC` at [riscv.spec:4956](examples/l3-machine-code/riscv/src/riscv.spec).

This is a bug in the specification. The binary-to-graph decompiler uses this specification, which results in incorrect translation of compressed format `LUI` instructions.

mn200:
In the absence of anyone else stepping forward to solve this or dispute it, we'd be happy to take a PR to fix this.

zaklogician:
@mn200 seL4 verification already have a patch for this, just doing some final tests before submitting a PR.

mn200:
Excellent; looking forward to it.  And no, I haven't been giving it any time: I don't feel qualified to do so!

zaklogician:
With the merging of pull request #846, this issue is resolved.

PR #846, opened 2020-07-27
Fix RISC-V ISA specification of RVC `lui` instruction (#842)

zaklogician:
Fixes Issue #842 by reimplementing the formal L3 specification of C.LUI to match the informal specification found in The RISC-V Compressed Instruction Set Manual, Version 1.9. The L3 specification is recompiled into sml files.

Signed-off-by: Zoltan Kocsis <zoltan.kocsis@data61.csiro.au>


mn200:
Could you write more explanation of the fix in the commit message please?

mn200:
Actually, let's assume the links to GitHub will last as long as people are interested.  The reference via the #-tagging is probably good enough.

PR #850, opened 2020-08-05
Equivalent definitions of extreal-based Borel sets

binghe:
Hi,

this PR continues #774 and further provides theorems for alternative (equivalent) definitions of extreal-based Borel sets (`Borel`), originally defined by the smallest σ-algebra generated from half space `{x | x < Normal a}`:

```
   [Borel_def]  Theorem      
      ⊢ Borel = sigma 𝕌(:extreal) (IMAGE (λa. {x | x < Normal a}) 𝕌(:real))
```

Sometimes we may need other equivalent definitions generated from different shapes, but it's very hard to prove other equivalent definitions from the above one.  Now, following [1, p.61], I have re-defined `Borel` in a different way, starting from the real-based version `borel`:

```
   [Borel]  Definition      
      ⊢ Borel =
        (𝕌(:extreal),
         {B' | ∃B S. B' = IMAGE Normal B ∪ S ∧ B ∈ subsets borel ∧ S ∈ {∅; {−∞}; {+∞}; {−∞; +∞}}})
```

Within the above definition, the original definition of `Borel` now becomes a theorem (the above "Borel_def"). Besides I have proven the following alternative equivalent definition based on half-closed space `{x | x ≤ Normal a}`:

```
   [Borel_eq_le]  Theorem      
      ⊢ Borel = sigma 𝕌(:extreal) (IMAGE (λa. {x | x ≤ Normal a}) 𝕌(:real))
```

Other possible shapes essentially follow with similar proofs (but it's hard to extract common lemmas). I think they can be added later when needed. NOTE: Actually the only remaining cases are `{x | Normal a < x}` and `{x | Normal a <= x}` (plus the rational versions of all the four cases.) Unlike the case of `borel`, bounded systems like `{x | Normal a < x /\ x <= Normal b}` are only contained in `Borel` but cannot be used as the very generator.

--Chun

[1] Schilling, R.L.: Measures, Integrals and Martingales (Second Edition). Cambridge University Press (2017).


binghe:
Since there are only two (obvious) cases left, I have added them with duplicated proof steps eliminated as much as possible:

```
   [Borel_eq_ge]  Theorem
      ⊢ Borel = sigma 𝕌(:extreal) (IMAGE (λa. {x | Normal a ≤ x}) 𝕌(:real))
   
   [Borel_eq_gr]  Theorem     
      ⊢ Borel = sigma 𝕌(:extreal) (IMAGE (λa. {x | Normal a < x}) 𝕌(:real))
 ```

**Comment by binghe** [comment]
I further added extreal-based Borel measure space (`ext_lborel`) by the following direct definition:
```
   [ext_lborel_def]  Definition
      ⊢ ext_lborel = (space Borel,subsets Borel,lambda ∘ real_set)
```

Then it can be (easily) proven that `ext_lborel` is indeed a (sigma-finite) measure space:
```
   [MEASURE_SPACE_LBOREL]  Theorem      
      ⊢ measure_space ext_lborel

   [SIGMA_FINITE_LBOREL]  Theorem      
      ⊢ sigma_finite ext_lborel
```
The above proofs are feasible only because of the new abstract definition of `Borel` in this PR.

Then, in `probabilityTheory`, the probability density function (PDF) for extreal-based random variables can now be defined:
```
   [pdf_def]  Definition      
      ⊢ ∀p X. pdf p X = distribution p X / ext_lborel
```

Finally, I have proved the following basic properties of `pdf`, following the existing proofs for real-based r.v.'s:

```
   [pdf_le_pos]  Theorem     
      ⊢ ∀p X x.
            prob_space p ∧ random_variable X p Borel ∧
            distribution p X ≪ ext_lborel ⇒
            0 ≤ pdf p X x

   [expectation_pdf_1]  Theorem      
      ⊢ ∀p X.
            prob_space p ∧ random_variable X p Borel ∧
            distribution p X ≪ ext_lborel ⇒
            expectation ext_lborel (pdf p X) = 1
```

Note that, a key difference between the probability theory in HOL4 and Isabelle is that, in HOL4, most deep theorems about r.v.'s are based on extreal-valued r.v.'s, even if their actual value range only in reals (see `real_random_variable_def`).  Thus in the future, the main theorems about probability density functions to be added, should be based on `pdf` instead of the old `PDF` (from HVG's `normal_rvTheory`, not ported yet).

--Chun

mn200:
Thanks for all the continuing work on this!

PR #854, opened 2020-08-17
Add a new example: a small verified bootstrapped compiler

myreen:
This commit contains a development of a "minimal" verified
bootstrapped compiler that targets x86-64 assembly code.

I have added this to the build-sequence as one of the final
examples. The tactic proofs build under both the standard and the
experimental kernels. However, the bootstrap evaluation (in
compiler_evalScript.sml) seems to get stuck on the experimental
kernel. (Is it possible to exclude an example from the build
sequence if HOL is built using the experimental kernel?)

### DIFF
--- a/examples/bootstrap/Holmakefile
+++ b/examples/bootstrap/Holmakefile
@@ -1,0 +1,1 @@
+INCLUDES = $(HOLDIR)/examples/fun-op-sem/lprefix_lub
--- a/examples/bootstrap/README.md
+++ b/examples/bootstrap/README.md
@@ -1,0 +1,36 @@
+# A Small Verified Bootstrapped Compiler
+
+The source language is defined in files:
+ - [source_syntaxScript.sml](source_syntaxScript.sml)
+ - [source_valuesScript.sml](source_valuesScript.sml)
+ - [source_semanticsScript.sml](source_semanticsScript.sml)
+ - [source_propertiesScript.sml](source_propertiesScript.sml)
+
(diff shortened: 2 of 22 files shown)


mn200:
Yes, as _per_ instructions in `tools/build-sequence`:

```
# If a directory name is preceded by a string inside ()-brackets, then
# that directory/target will only be considered if the string is the
# name of the current kernel.  Thus the line (otknl)foo will cause foo to
# be built only if the current kernel is otknl. (The other supported
# kernels at the time of writing are stdknl and expk.)
```

The "failure" to evaluate may just be inefficiency in the `expk` EVAL: I've seen it elsewhere.  (You could do the experiment by waiting for however long, but I don't think I want the regression machinery to take absolutely forever, so I'd be happy to have your bootstrap-eval directory marked as `stdknl` only.)

binghe:
Hi ... this is a compiler of what language? CakeML?

myreen:
@mn200 thanks for the pointer. I've now updated the build sequence entry with the relevant entry.

    [poly](stdknl)!!!examples/bootstrap

With the standard kernel `compiler_evalScript.sml` takes **70 seconds** to build. With the experimental kernel I gave up after **75 minutes**.

@binghe it's a compiler for a small Lisp language that is just large enough to conveniently host a bootstrapped compiler.

mn200:
Thanks - I'll see how it fares in the automatic regression test overnight!

PR #856, opened 2020-08-18
Make pretty-printing of goalTrees a little nicer

HeikoBecker:
This pull request replaces the occurences of `THEN` and `THENL` in `src/proofman/goalTree.sml` with `>>` and `>-` respectively.

### DIFF
--- a/src/proofman/goalTree.sml
+++ b/src/proofman/goalTree.sml
@@ -100,17 +100,16 @@ fun pp_gtree gt =
        | pp (vGoal g) = pp_goal g
        | pp (vThen(vt1,vt2)) =
              block CONSISTENT 1 (
-               pp vt1 >> add_string " THEN" >> add_break (1,0) >>
+               pp vt1 >> add_string " \\\\" >> add_break (1,0) >>
                pp vt2
              )
        | pp (vThenl(vt,tlist)) =
              block CONSISTENT 1 (
-               pp vt >> add_string " THENL" >> add_break (1,0) >>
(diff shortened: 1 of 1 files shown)


binghe:
Hi, just curious, how do you use this feature (goal tree)?

mn200:
Me too!  (I'm also curious, I mean.) I've never used this code.

konrad-slind:
The original idea behind goal trees was for the tactic-under-construction
to be printed out by the REPL after making tactic steps. That way, for
simple tactic proofs, one wouldn't really need to have two buffers at all:
everything could happen in the interactive buffer, and the text of the
tactic would be maintained by the goaltree implementation. I didn't use it
much, since a little bit of magic was needed in order to trap the input
string before it got sent to PolyML. Some minor amount of emacs trickery
(or maybe magic similar to what happens with the quote filter) would be
needed to make it nice to use.

Konrad.

konrad-slind:
Heiko, how are you using goaltrees?

HeikoBecker:
I am using them to "explain" HOL4 tactics. I have a simple DSL for untrained HOL4 users which then is parsed into HOL4 tactics. By capturing the REPL input, I can then parse (DSL-) tactics send to the REPL and decorate the goal tree with their HOL4 counterpart.

konrad-slind:
Thanks! Sounds like a nice application for them.      --Konrad.

mn200:
Thanks for your work on this!

PR #857, opened 2020-08-22
Name inductive cases

talsewell:
Adds a markerTheory constant for doing "which case is this" annotations. This is something
of a pattern in CakeML for proofs with many cases.

The most useful bit is a gadget that modifies induction theorems so that the this-case
annotation will be present in each goal.

### DIFF
--- a/src/boss/bossLib.sig
+++ b/src/boss/bossLib.sig
@@ -155,6 +155,9 @@ sig
   val REABBREV_TAC     : tactic
   val WITHOUT_ABBREVS  : tactic -> tactic
 
+  (* name cases of an induction theorem *)
+  val name_ind_cases : term list -> thm -> thm
+
   (* more simplification variants *)
   val fsrw_tac : simpLib.ssfrag list -> thm list -> tactic
   val simp : thm list -> tactic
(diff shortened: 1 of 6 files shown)


talsewell:
Hmm, I agree with you, but I don't have any bright ideas what a test case might look like. I think it should work for rule induction, what's a good example that's available in base HOL? RTC maybe?

mn200:
RTC is certainly fine to use, but equally, feel free to invent something dumb but illustrative for the purposes of a test.  It's not like it will live on past the execution of the test.

talsewell:
OK, I fixed a bug, but I'm out of time, basically, handing back all my computer gear soon.

talsewell:
Well, I had some distraction time available, and I figured out some tests.
I think I fixed the remaining bugs as well, and rebased a little for
cleanliness. Anything else that needs doing?

Memo to self: Adding things to core places like bossLib is painful, the
build process is a really slow clunky way to test things, and trying to
develop within the REPL leads to silly scoping errors.

mn200:
I'm pretty well happy to take it as is, but any documentation on use strategies would be gratefully received too.

talsewell:
Roger. This is a simple gadget though, it just tweaks your induction (etc) rule so that
you'll know which case is which. Applying it to list_induction as suggested in the
comment will explain what's going on, though it's only really useful if you have a lot
more cases and you'd rather select them by pattern than by position.

talsewell:
Perhaps I should add here, this will come in handy when doing an induction
as a collection of syntactically independent proofs as @myreen  and I have
prototyped, but I can't point at that in the repository yet.

mn200:
If possible, it might be nice to link this to the stuff I just committed that labels rules in inductive relations.

mn200:
Thanks for this!

PR #859, opened 2020-08-29
improvements for using hol with vim

arolle:


mn200:
Thanks for this!  If you're so inclined, it would be nice to have some documentation of the `vim` mode in the interaction manual (see `Manual/Interaction`).  There's already a section there on how to do things with emacs (and sections after that include bits of emacs advice as well), so the `vim` material could fit in parallel.

PR #860, opened 2020-08-29
adds 5 theorems (about lists, relations, and sets)

arolle:


src/pred_set/src/pred_setScript.sml
```diff
@@ -965,6 +965,13 @@ val DIFF_UNION = store_thm(
 ``!x y z. x DIFF (y UNION z) = x DIFF y DIFF z``,
 SRW_TAC[][EXTENSION] THEN METIS_TAC[])
 
+Theorem UNION_DIFF_EQ:
```
mn200, line 968: Please make this an automatic rewrite (add `[simp]` after theorem name), and add an extra conjunct with the  union the other way round.

mn200 (changes requested):

mn200:
The CI failures above appear to be genuine.  Can you fix the affected scripts please?  Use `temp_delsimps` if you like, but actually adjusting the affected proofs would be better if it looks practical.

arolle:
As I notice, this will fail at unification, again. Let me fix that.

binghe:
I think the **fact** that so many `Excl"UNION_DIFF_EQ"` are needed to fix broken proofs, indicates that the previous idea of setting `UNION_DIFF_EQ` as automatic rewrite is a little problematic.

In `s UNION (t DIFF s)`, maybe `s` and `t DIFF s` are disjoint sets, which is required to finish the proof, while `s` and `t` themselves are not.  Thus it's even possible that in the previous proof step `s UNION t` was rewritten to `s UNION (t DIFF s)` on purpose, but then it's reverted by `fs []`...

mn200:
You can imagine scenarios akin to this for just about any rewrite that turns a complicated term into a simpler term.  My instinct says that in cases like this where the RHS is strictly smaller than the LHS, the rewrite should almost always be automatic.  If the play of the proof has led to a goal including the sub-term `s UNION (t DIFF s)`, the user is much more likely to say “Huh; shouldn't that be equal to something simpler?” than “Oh yes; exactly what I wanted!”. 

Of course, sometimes you do want the complicated form (as you say), and then we have the tooling (`Excl` or other tricks) to allow that.  And of course, establishing that your union can be seen in this way is as easy as writing `‘s UNION t = s UNION (t DIFF s)’ by simp[]`.  It is just the follow-up simp (presumably with a theorem that has a `DISJOINT` precondition) that will need to use `Excl` to stop an infinite loop. 

Given that the commit has had to fix only 4 theorems (by adding `Excl` calls), I certainly don’t see this as an invasive change.

mn200:
You can imagine scenarios akin to this for just about any rewrite that turns a complicated term into a simpler term.  My instinct says that in cases like this where the RHS is strictly smaller than the LHS, the rewrite should almost always be automatic.  If the play of the proof has led to a goal including the sub-term `s UNION (t DIFF s)`, the user is much more likely to say “Huh; shouldn't that be equal to something simpler?” than “Oh yes; exactly what I wanted!”. 

Of course, sometimes you do want the complicated form (as you say), and then we have the tooling (`Excl` or other tricks) to allow that.  And of course, establishing that your union can be seen in this way is as easy as writing `‘s UNION t = s UNION (t DIFF s)’ by simp[]`.  It is just the follow-up simp (presumably with a theorem that has a `DISJOINT` precondition) that will need to use `Excl` to stop an infinite loop. 

Given that the commit has had to fix just 5 theorems (by adding `Excl` calls), I certainly don’t see this as an invasive change.

binghe:
@mn200, thanks for your further explanation, now I'm convinced. Sometimes I do find that `fs []` or `rw []` does too much, but then it's natural to use `FULL_SIMP_TAC std_ss` or `ASM_SIMP_TAC std_ss` instead, if I know the precise lemmas for rewriting, or `Excl` if I know what to exclude.

mn200:
Thanks for your work on this!

PR #861, opened 2020-09-03
New example: a sound and complete Hoare logic for diverging programs

myreen:
This is joint work with @IlmariReissumies


mn200:
Thanks for your work on this!

PR #863, opened 2020-09-13
Add support for using implications as rewrites in computeLib. Add glossary to Developer’s Manual.

marioxcc:


mn200:
Thank you for coming to grips with this code!  Do you limit the number of antecedents the system will be willing to try to prove?  (The simplifier has a stack limit to stop too much time being spent on what may end up being completely unproductive.)

Manual/Developers/developers.md
```diff
@@ -183,7 +183,80 @@ Unless otherwise noted, they are built by the configuration process.
...
+:   A theorem.
+
+`_then`
+:   Optional suffix for theorem-tacticals. See the section on tactics in the HOL description manual.
+
+`thm`
+:   A theorem.
+
+`x_`
+:   Optional prefix to indicate a variant that takes a term or quotation. E.g.: The tactical `X_GEN_TAC` compared to `GEN_TAC`.
```
mn200, line 259: These functions (X_GEN_TAC, X_CHOOSE..) take a term/quotation to allow the user to specify something that the system will otherwise guess/invent.

Note also that `x` is used within names of thm-tacticals to indicate that assumptions are removed as they are used (rather than left in place).

marioxcc:
> Thank you for coming to grips with this code! Do you limit the number of antecedents the system will be willing to try to prove? (The simplifier has a stack limit to stop too much time being spent on what may end up being completely unproductive.)

mn200:
Please add regression tests (including old and new behaviours), either in a `selftest.sml` file or in a `theory_tests` sub-directory.

marioxcc:
I added 2 basic tests are comparable to the previously-existing test, which was preserved. What changes are required, if any?

mn200:
Thanks for your work on this!

PR #866, opened 2020-09-24
Revert "Add support for using implications as rewrites in computeLib. Add glossary to Developer’s Manual."

mn200:
Reverts HOL-Theorem-Prover/HOL#863

Regression testing reveals that the changes in #863 put `EVAL` into an infinite loop on line 169 of `examples/ARM/v7/arm_stepScript.sml`.  I will separately cherry-pick the unrelated glossary change back into `develop`.


mn200:
If @marioxcc can think of a nice way to fix this issue, I'm happy to see that proposed instead.

marioxcc:
Fixed in https://github.com/HOL-Theorem-Prover/HOL/pull/867.

mn200:
Not required.

PR #867, opened 2020-09-25
computeLib: Avoid generating the equational rewrite “T ⇔ T” that loops.

marioxcc:
Fixes looping introduced in 0c164018217af52873cf4f250dfd51c361e74da7 and mentioned in https://github.com/HOL-Theorem-Prover/HOL/pull/866: When an user-provided theorem was “⊢ T” or a chain of implications where the consequent was “T”, computeLib would add the equational rewrite “T ⇔ T” which is useless and causes looping in reductions.

This was triggered by `arm_coretypesTheory.NOT_IN_EMPTY_SPECIFICATION` which is the theorem `⊢ T`. I assume this theorem is broken in the distant past and not detected because the proof script uses `save_thm` instead of the new `Theorem` notation or `store_thm`.

### DIFF
--- a/src/compute/src/clauses.sml
+++ b/src/compute/src/clauses.sml
@@ -329,6 +329,15 @@ fun enter_thm compset thm0 = let
       | NONE => ()
     end
 in
+  (*
+     First try to add an useful rewrite. After stripping antencedents:
+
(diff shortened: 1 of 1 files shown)


mn200:
Thanks for this swift response!

PR #869, opened 2020-10-03
Windmill

jhlchan:
Provide scripts for HOL4 proofs of Fermat's Two Squares Theorem, as well as Fermat's Little Theorem.


mn200:
Problems:

* Many changes in `groupAction` seem to just rework a bunch of bound variables to be `o`. This seems
   1. unnecessary;
   2. annoying because of the need to use `$`-escaping

* Some changes in lib (in at least `helperSet` and `helperList`) undo changes I made in commits after you first pushed this material.  This suggests a bad merge.  Please be more careful.  For example, `DROP_NON_NIL` is reinstated, and a number of theorems have their use of new `Theorem`-`Proof`-`QED` syntax undone.

jhlchan:
Understood. I shall close this one, rework and request another one.

PR #873, opened 2020-10-30
Fixes of algebra and fermat examples under experimental kernel

binghe:
Hi,

this PR fixes the broken proof (under `--expk` build option) of two theory files from the `algebra` and `fermat` examples, in the obvious way (using `rename1` to forcely change the variable names back). 

Besides, the two `Holmakefile` of the `fermat` example (wrongly) assumes the toplevel HOL directory is just "HOL", which is not true in my machine. Using the variable `$(HOLDIR)` is the correct way.

Chun


mn200:
Thanks for this!  Incidentally, `rename` is slightly more robust than `rename1` (see the documentation for their aliases, `Q.RENAME1_TAC` and `Q.RENAME_TAC`).

binghe:
I will learn and use `rename` instead of `rename1` from now on.

PR #874, opened 2020-11-24
make sure wordsLib.WORD_GROUND_ss covers "word_ge/gt/hi/hs"

acjf3:
For example, previously ``3w >=+ 2w: word32`` does not simplify.


mn200:
Thanks!

mn200:
I thought I would write a test-case for this code, and first made sure that `WORD_GROUND_ss` was exported from `wordsLib` (66a2b80f7).  I tried a simplifier invocation on the same cases as tested in the existing test for `WORD_GROUND_CONV`.  This proved harder than expected: `WORD_ss` now rewrites `255w : word8` to `-1w` (urk), and if I just have `WORD_GROUND_ss` on its own, it doesn't cope with `~14w = 1w : word4`.

PR #878, opened 2020-12-10
Add tactical for multiple goal selection, and vim support for existing single goal selection

hrutvik:
**Summary of changes:**
- Add `SELECT_LT` list-tactical. This is similar to the existing `FIRST_LT`, but it will apply the tactic to *all* succeeding goals (rather than the first) and reorder them.
- Add `SELECT_GOALS_LT` (and corresponding `>>~` combinator). This is similar to the existing `SELECT_GOAL_LT`, but can select multiple goals and cannot fail.
- Add `SELECT_GOALS_LT_THEN`. This allows the user to supply an additional tactic, the success of which is an extra criterion for goal selection. In the final goal state, this tactic will have already been applied to the selected goals.
- Add basic vim support for existing single goal selection (`SELECT_GOAL_LT`) via the `hP` key binding.

I have not yet added Emacs/vim support for the new feature, pending discussion about how best to integrate this.

### DIFF
--- a/help/Docfiles/Tactical.SELECT_LT.doc
+++ b/help/Docfiles/Tactical.SELECT_LT.doc
@@ -1,0 +1,34 @@
+\DOC
+
+\TYPE {SELECT_LT : tactic -> list_tactic}
+
+\SYNOPSIS
+Applies a tactic to the all the goals in the goal-list for which the
+tactic succeeds.
+
+\KEYWORDS
+
+\DESCRIBE
+Given a list of goals {gl}, an application of {SELECT_LT tac} to {gl}
+will try to apply {tac} to each goal in {gl} in turn. If no goal lets
+{tac} succeed, the goal state remains unchanged. Otherwise, the goals
+for which {tac} succeeds will generate (possibly empty) list(s) of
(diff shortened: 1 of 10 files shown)


mn200:
What's the difference between 
    
    SELECT_LT (tac1 >> tac2) 

and

    SELECT_LT_THEN tac1 tac2

?

hrutvik:
Apologies for the delay! I'm not quite sure I understand - I don't believe there is a `SELECT_LT_THEN`, but I assume you're referring to `SELECT_GOALS_LT_THEN`? In that case, it builds on `SELECT_GOALS_LT` and has type `tmquote list -> tactic -> list_tactic` - it uses both renaming via the `tmquote list` (as in `SELECT_GOAL_LT` and `SELECT_GOALS_LT`) and the success of the supplied `tactic` as goal selection criteria. It's just a shorthand which I thought might be convenient, and defined as:
```
SELECT_GOALS_LT_THEN pats tac = SELECT_LT (RENAME_TAC pats >> tac)
```

mn200:
You're right; I misunderstood.  All looks good to me.  Thanks!

PR #880, opened 2021-01-10
Alternative definitions of 2-dimensional Borel sets (+ other fixes)

binghe:
Hi,

this PR continues the previous work in #774 and #822. From product measures, two-dimensional Borel sets (real-valued) can be easily stated as `borel × borel`, where `CROSS` is overloaded on `prod_sigma` with the following definition (previously in `martingaleTheory`, now moved to `sigma_algebraTheory`):

```
   [prod_sigma_def]  Definition      
      ⊢ ∀a b.
          a × b =
          sigma (space a × space b) (prod_sets (subsets a) (subsets b))
```

But this definition is not convenient when trying to show a two-variable function (of the type `:real # real -> real`) is Borel-measurable. To resolve this issue, I first followed `src/real/metricTheory` and defined the standard 2D Euclidean metric space (in `real_borelTheory`. It cannot be moved to `metricTheory` because of the use of `sqrt` defined in `transcTheory`)

```
   [mr2]  Definition      
      ⊢ mr2 = metric (λ((x1,x2),y1,y2). sqrt ((x1 − y1)² + (x2 − y2)²))

   [MR2_DEF]  Theorem      
      ⊢ ∀x1 x2 y1 y2.
          dist mr2 ((x1,x2),y1,y2) = sqrt ((x1 − y1)² + (x2 − y2)²)
```

Then, following a very hard path with many intermediate lemmas, I have proved that 2D Borel sets (`borel × borel`) can be generated by all 2D open sets (w.r.t. `mr2`) and also all open rectangles:

```
   [borel_2d]  Theorem      
      ⊢ borel × borel = sigma 𝕌(:real # real) {s | open_in (mtop mr2) s}
   
   [borel_2d_alt_box]  Theorem      
      ⊢ borel × borel = sigma 𝕌(:real # real) {box a b × box c d | T}
```

These two results can be **later** used in `probabilityTheory` to show that a pair of independent random variables X and Y satisfies E[XY] = E[X] E[Y], where E denotes the mathematical expectation (and thus forms a path towards the Law of Large Numbers for I.I.D. random variables).

I moved some theorems (all related to sigma algebras and prod sigma algebras) from `martingaleTheory` to upward `sigma_algebraTheory`, mostly because the latter file is relatively small.

Beside the above main results, in case HOL is preparing a new release, I also did some minor fixes and other small improvements in HOL core theories, but mostly in `src/probability`.   The only important change outside of the probability folder is the following handy theorem in `cardinalTheory` that I needed in proving the above `borel_2d` theorem:

```
   [CARD_LE_COUNTABLE]  Theorem      
      ⊢ ∀s t. COUNTABLE t ∧ s ≼ t ⇒ COUNTABLE s
```

Previously the types of both `s` and `t` are `:'a -> bool`. But I have two sets of different types. It turns out that the same proof still work when I changed `t:'a->bool` to `t:'b->bool`. Now this theorem is more useful and nothing is broken.

Regards,

Chun Tian

### DIFF
--- a/src/pred_set/src/more_theories/cardinalScript.sml
+++ b/src/pred_set/src/more_theories/cardinalScript.sml
@@ -2991,13 +2991,17 @@ val COUNTABLE_CASES = store_thm ("COUNTABLE_CASES",
  >> ONCE_REWRITE_TAC[COUNTABLE_ALT_cardleq, FINITE_CARD_LT]
  >> METIS_TAC [CARD_LE_LT]);
 
-val CARD_LE_COUNTABLE = store_thm ("CARD_LE_COUNTABLE",
- ``!s:'a->bool t:'a->bool. COUNTABLE t /\ s <=_c t ==> COUNTABLE s``,
-  REWRITE_TAC[COUNTABLE, ge_c] THEN REPEAT STRIP_TAC THEN
-  KNOW_TAC ``?(t :'a -> bool).
-      (s :'a -> bool) <=_c t /\ t <=_c univ((:num) :num itself)`` THENL
-  [EXISTS_TAC ``t:'a->bool`` THEN ASM_REWRITE_TAC[],
-   METIS_TAC [CARD_LE_TRANS]]);
+(* changed ‘t:'a->bool’ to ‘t:'b->bool’ *)
+Theorem CARD_LE_COUNTABLE :
+    !s:'a->bool t:'b->bool. COUNTABLE t /\ s <=_c t ==> COUNTABLE s
+Proof
+    REWRITE_TAC [COUNTABLE, ge_c]
+ >> rpt STRIP_TAC
+ >> KNOW_TAC ``?(t :'b -> bool).
+      (s :'a -> bool) <=_c t /\ t <=_c univ((:num) :num itself)``
+ >- (EXISTS_TAC ``t:'b->bool`` >> ASM_REWRITE_TAC[])
+ >> METIS_TAC [CARD_LE_TRANS]
+QED
 
 val CARD_EQ_COUNTABLE = store_thm ("CARD_EQ_COUNTABLE",
  ``!s:'a->bool t:'a->bool. COUNTABLE t /\ s =_c t ==> COUNTABLE s``,
--- a/src/probability/borelScript.sml
+++ b/src/probability/borelScript.sml
@@ -2714,6 +2714,17 @@ Proof
  >> ASM_SET_TAC [space_def]
 QED
 
+Theorem IN_MEASURABLE_BOREL_EQ' :
+    !a f g. (!x. x IN space a ==> (f x = g x)) /\
+            g IN measurable a Borel ==> f IN measurable a Borel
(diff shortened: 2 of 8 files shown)


mn200:
Thanks for your continuing work on this material!

mn200:
Something now breaks in `examples/miller/*/probScript.sml`, perhaps a result of something in your most recent commit.  Do you have any easy fix to make, if so?  I can look into it in the next little while myself if I need to.

binghe:
> Something now breaks in `examples/miller/*/probScript.sml`, perhaps a result of something in your most recent commit. Do you have any easy fix to make, if so? I can look into it in the next little while myself if I need to.

Hi, I think something I moved into `sigma_algebraTheory` (thus visible to `real_probabilityTheory`) must have duplicated names with those defined by `miller` examples, and the proofs fail because different definitions (or theorems) now come into the place. Let me have a deeper look and see what's the best way to fix it (with duplicated definitions possibly unified.)

binghe:
> > Something now breaks in `examples/miller/*/probScript.sml`, perhaps a result of something in your most recent commit. Do you have any easy fix to make, if so? I can look into it in the next little while myself if I need to.
> 
> Hi, I think something I moved into `sigma_algebraTheory` (thus visible to `real_probabilityTheory`) must have duplicated names with those defined by `miller` examples, and the proofs fail because different definitions (or theorems) now come into the place. Let me have a deeper look and see what's the best way to fix it (with duplicated definitions possibly unified.)

OK, it's actually caused by my modification of the theorem `MEASURABLE_PROD_SIGMA` in `sigma_algebraTheory`:

```
[MEASURABLE_PROD_SIGMA]  Theorem      
      ⊢ ∀a a1 a2 f.
          sigma_algebra a ∧ FST ∘ f ∈ measurable a a1 ∧
          SND ∘ f ∈ measurable a a2 ⇒
          f ∈
          measurable a
            (sigma (space a1 × space a2)
               (prod_sets (subsets a1) (subsets a2)))
```

Previously this theorem wasn't used by any other theorem in `src/probability` and I didn't realized that it's actually from the very beginning by Hurd himself and used by the miller example. The build failure is caused by my modification/simplification of this single theorem. Now it's fixed in the latest commit of my branch [1]. Feel free to cherry-pick it into `develop`.

[1] https://github.com/binghe/HOL/commit/c75684837b9839c7a5987263ec232d86f17d68b7

The miller example relies on the concept of "product sigma-algebras" but didn't explicitly define this concept. When I move `prod_sigma_def` back to `sigma_algebraTheory`, it was natural to simplify the above theorem using the new definition. But now I created a new variant of the above theorem instead, for future use in `probabilityTheory`:

```
[MEASURABLE_PROD_SIGMA']  Theorem      
      ⊢ ∀a a1 a2 f.
          sigma_algebra a ∧ FST ∘ f ∈ measurable a a1 ∧
          SND ∘ f ∈ measurable a a2 ⇒
          f ∈ measurable a (a1 × a2)
```

PR #883, opened 2021-01-19
Add roundoff error theorems for cases where floating-point evaluation becomes denormal

HeikoBecker:
The theorems in `lift_ieeeScript` currently only prove roundoff error bounds for the case where either the result is 0 or a normal number. This adds roundoff error theorems for the third, denormal case.


mn200:
Thanks for these!

PR #884, opened 2021-01-26
Add Lassie's HOL4 tutorial and the Lassie lib to the release notes

HeikoBecker:


Manual/Lassie-Tutorial/intro.tex
```diff
@@ -0,0 +1,99 @@
+
+\paragraph{Note for CPP 2021 reviewers}
```
arolle, line 2: You could omit this paragraph.

HeikoBecker: Thanks! That somehow slipped through.

mn200:
Thanks!

mn200:
Building the manual fails because a `setup_hol4.sh` file is not present.

HeikoBecker:
Sorry, I forgot to add the file.

Thank you for fixing it already!

PR #886, opened 2021-02-01
Add higher-order case splits

talsewell:
The higher-order elim rule combines the case_elim and case_rand
theorems already available from Prim_rec. The implication form
uses negation to flip the quantifiers, from a disjunction of
existentially-quantified possibilities to a conjunction of
universally-qualified implications.

### DIFF
--- a/src/1/Prim_rec.sig
+++ b/src/1/Prim_rec.sig
@@ -45,9 +45,16 @@ sig
    val case_cong_thm               : thm -> thm -> thm
    val prove_constructors_distinct : thm -> thm option list
    val prove_constructors_one_one  : thm -> thm option list
+
+   (*-------------------------------------------------------------------------
+      Prove various equalities on case distinction terms.
+    -------------------------------------------------------------------------*)
+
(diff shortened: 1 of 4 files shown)


mn200:
Rather than `GEN` and `ISPEC` in the first one, I'd just use `INST_TYPE` to make the range type of the "predicate" variable be `:bool`.    And then, use `REWR_CONV` in the right position (`RAND_CONV`) rather than `REWRITE_RULE` to use the `elim` theorem.  I can do this myself if you don't get around to it in the near future.

talsewell:
Hmm, you're right, that would be simpler.

I fiddled with trying to simplify the conversion in the second one also. It's possible to apply not-exists rule without higher-order mechanisms by first eta-contracting it, but this has to be alternated with beta expansion, so I'm not sure there are any wins there.

mn200:
Given conversation in Slack @ cakeml, I'll leave this a bit longer to give you a chance to make TypeBase changes as well.

mn200:
To be consistent with other names in `TypeBase`, the new entry points could perhaps pick up names ending in `_of`.

mn200:
Awesome; thanks!

PR #888, opened 2021-02-05
cleanup disj_norm workarounds in probability scripts

binghe:
Hi,

this PR removed the following lines in probability related scripts and fixed the broken proofs, so that all theorems are proven under default definitions of `std_ss` and `real_ss`:

```
val std_ss = std_ss -* ["lift_disj_eq", "lift_imp_disj"]
val real_ss = real_ss -* ["lift_disj_eq", "lift_imp_disj"]
val _ = temp_delsimps ["lift_disj_eq", "lift_imp_disj"]
```

Overall speaking, the new proofs became slightly smaller with the new default simp sets.

--Chun

### DIFF
--- a/src/probability/borelScript.sml
+++ b/src/probability/borelScript.sml
@@ -36,11 +36,7 @@ open hurdUtils util_probTheory extrealTheory sigma_algebraTheory
 
 val _ = new_theory "borel";
 
-val std_ss = std_ss -* ["lift_disj_eq", "lift_imp_disj"]
-val real_ss = real_ss -* ["lift_disj_eq", "lift_imp_disj"]
-val _ = temp_delsimps ["lift_disj_eq", "lift_imp_disj"]
-
-val ASM_ARITH_TAC = REPEAT (POP_ASSUM MP_TAC) >> ARITH_TAC; (* numLib *)
+val ASM_ARITH_TAC = rpt (POP_ASSUM MP_TAC) >> ARITH_TAC; (* numLib *)
 val DISC_RW_KILL = DISCH_TAC >> ONCE_ASM_REWRITE_TAC [] >> POP_ASSUM K_TAC;
 fun METIS ths tm = prove(tm, METIS_TAC ths);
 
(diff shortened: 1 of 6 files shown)


mn200:
Thanks for this; it's good to hear that the `disjnorm` change did indeed make shorter proofs possible!

binghe:
> Thanks for this; it's good to hear that the `disjnorm` change did indeed make shorter proofs possible!

However, I have to say that it is sometimes a little annoy that, whenever there's a goal like `x <> PosInf \/ x <> NegInf` (and the next step is to do either `DISJ1_TAC` or `DISJ2_TAC` after some preparations), the new `std_ss` automatically rewrites the goal to `x = PosInf ==> x <> NegInf` and then strip `x = PosInf` into assumptions. Previously this striping was achieved by `STRONG_DISJ_TAC` if the proof really goes in this way.   But if the proof actually needs to strip `x = NegInf` and prove `x <> PosInf` instead, more proof steps are needed here after the new `std_ss`.   In general, the "symmetry" in a goal like `A <> B \/ C <> D` is broken by the new `std_ss`. This is sometimes good, other times bad.

binghe:
If ... the rewriting rules could be a little smarter, e.g., it only (automatically) rewrites `~A \/ B` but keep `~A \/ ~B` unchanged, where `B` is a term without top-level negation, then it will be perfect to me (i.e. some proofs in probability).

mn200:
Your example is not great (turning `x <> Posinf \/ x <> Neginf` into `x = Posinf ==> x <> Neginf` will subsequently simplify to `x = Posinf ==> Posinf <> Neginf`, which, I assume is trivially true). But you are right that there are surely awkward cases.  If there were an easy criterion the simplifier could assess...

One scenario I can imagine is having a goal of the form `e1 <> e2 \/ P`, where you know that you want to prove that `e1 <> e2`.  The simplifier will turn this into `e1 = e2 ==> P'` (where `P'` is the simpler form of `P` caused by knowing that `e1 = e2`).  There is no easy tactic to then focus one’s attention on just the `e1 = e2` term. 

Note that unless you are using `rw` or `srw_tac` or `RW_TAC`, the `e1=e2` term will *not* move into the assumptions.  

I suppose the best answer for the moment is to locally exclude the responsible rewrite: `simp[Excl "lift_disj_eq"]`.

binghe:
Hi Machael,

thanks for the explanations. I'll keep in mind `simp [Excl "lift_disj_eq"]` and use it whenever the proof goal is over simplified.

Chun

PR #889, opened 2021-02-08
Add `examples/algorithms`

hrutvik:
For now containing only a closure operation and a pseudo-topological sort, both implemented
over `num_set num_map`.


mn200:
You need to add this to the build sequence, with preceding !s as a rough measure of how long the files take to build.

mn200:
Thanks!

PR #891, opened 2021-02-14
Added dependability proof scripts under examples in HOL repo

ahmedwaqar:
This PR includes the proof scripts developed during the formalization of dependability modeling and analysis techniques, such as Reliability Block Diagrams (RBD) and Fault Trees (FT).  Both of these dependability techniques have been recommended by several safety standards, like IEC 61508, ISO 26262 and EN50128, for developing safe hardware and software systems.  The detail of the formalization can be found in the [link](http://save.seecs.nust.edu.pk/Downloads/wahmad_thesis.pdf).

A README file is added for brief detail about each proof script.


mn200:
Can you make a number of changes please? 

- There is a `Makefile` in one directory that has a path referring to your own directory.  Of course, the reference to your directory should go, but in fact, the `Makefile` should go entirely.  Rather, use a `Holmakefile` that uses `INCLUDES` lines to cause recursive makes to happen. 
-  The old Joe Leslie-Hurd style of infixes needs to be replaced by current ones (*e.g.*, `++` should become `>>` and `>>` should become `>-`).  See 505ee6d5da1 for an example of what needs to be done.
- If possible, please use the modern `Theorem .. Proof .. QED` style rather than the `store_thm` primitive directly. 
- Please encapsulate the `.ml` file properly.  If it's supposed to be an executable that does something on top of the theories, please build it as such.

mn200:
And the material needs to be part of the build sequence so that it gets regression-tested.

ahmedwaqar:
Thanks, Michael. Please, see my replies below:

>* There is a Makefile in one directory that has a path referring to your own directory. Of course, the reference to your directory should go, but in fact, the Makefile should go entirely. Rather, use a Holmakefile that uses INCLUDES lines to cause recursive makes to happen.

The ``Holmakefile`` is already triggering the build in each subdirectory and doesn't depend on the ``Makefile``.  I put the ``Makefile`` only for convenience for selective build and also to clean the directory. I see some directories have one similar to it, but I can remove it if that's not a convention.

>* The old Joe Leslie-Hurd style of infixes needs to be replaced by current ones (e.g., ++ should become >> and >> should become >-). See 505ee6d for an example of what needs to be done.

Yes. I'm aware of this change. I'll replace them in the next commit.

>* If possible, please use the modern Theorem .. Proof .. QED style rather than the store_thm primitive directly.

Sure, that's more aesthetically pleasing on the eyes.

>* Please encapsulate the .ml file properly. If it's supposed to be an executable that does something on top of the theories, please build it as such.

I'll think about it whether to keep it or let it go.

>* And the material needs to be part of the build sequence so that it gets regression-tested.

Thanks for pointing that out. I can extend the ``$(HOLDIR)/tools/sequences/final-examples`` to be triggered by the ``tools/build-sequence``. However, I see none of the automatic regression testing ``Build job`` trigger the ``build-sequence`` in ``$(HOLDIR)/examples``. Do I need to modify somewhere else too? I'm afraid other build sequences might not have required dependencies. I need ``examples/miller`` and ``src/probability`` to go successful.

mn200:
The `final-examples` file is probably a reasonable place to list your dependability example. The `build` will build examples if you use its `-t<n>` option. If you use `INCLUDES` directives to list `miller` as a dependency, then that will get built before the build of your own material proceeds.  Alternatively, just make sure you are later in the file than `miller` and whatever else you need. 

Also: please write something for the `doc/next-release.md` file describing the work.

ahmedwaqar:
The last commit contains the following fixes:
* Replaced ``++`` and ``>>`` with ``>>`` and ``>-`` in all proof scripts
* Used modern defns and thms formatting
* Deleted ``Makefile`` and ``auto_smart_grid.ml``
* Added a note about *Dependability Analysis* in ``next_release.md``
* Added include statement in the ``final-examples`` so that dependability proof scripts pass through regression

ahmedwaqar:
@mn200 I've made the suggested changes. Please, feel free to let me know if I miss something.

examples/dependability/RBD/Holmakefile
```diff
@@ -0,0 +1,3 @@
+HOLDIR = ../../
+
+INCLUDES = $(HOLDIR)/miller/formalize
```
mn200, line 3: This should surely be `$(HOLDIR)/examples/formalize`

examples/dependability/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Installation
+
+To use the proof script, follow the steps below:
+
+1. Install latest version of HOL by downloading it from  https://hol-theorem-prover.org/ or
+(Follow the steps mentioned in http://save.seecs.nust.edu.pk/Downloads/Installation%20of%20HOL%20&%20HOL-LIGHT%20in%20Linux.pdf)
+2. Fork and clone the repo ``https://github.com/ahmedwaqar/Formal-Dependability.git``
+3. Goto directory folder ``cd Formal-Dependability``
+4. Modify ``Makefile`` by specifying the ``HOLDIR`` variable to the path of your ``Holmake`` binary
```
mn200, line 38: Are these instructions still correct given that the material is now just part of HOL?

mn200 (changes requested):

examples/dependability/RBD/Holmakefile
binghe: If it's still not clear to @ahmedwaqar, this means that the first line `HOLDIR = ../../` can be just deleted, because the variable `HOLDIR` will be setup by `Holmake` internally to the actual root directory of HOL4.

ahmedwaqar: Thanks. I'm aware of the ``HOLDIR`` environment variable and the HOL directory structure, but got clumsy to fix it before making a commit. I corrected the  path in ``RBD/Holmakefile`` to ``INCLUDES = $(HOLDIR)/examples/miller/formalize``.

examples/dependability/README.md
ahmedwaqar: Thanks. I've removed all the installation instructions and also simplified *README*.

ahmedwaqar:
@mn200 I believe that I addressed the suggested changes. Please, let me know if I miss something. Thanks

mn200:
Thanks for your work on this!

PR #894, opened 2021-02-22
Fixed DIFF_CONV (derivative of real functions) for transcendental functions

binghe:
Hi,

previously the support of transcendental functions by `DIFF_CONV` (`src/real/Diff.sml`) for calculating derivatives of real functions, is not properly exported after `transcTheory` is loaded.  This PR has fixed this issue. Now, starting from a fresh HOL session, if only `Diff` is loaded:
```
> load "Diff";
val it = (): unit
> open Diff;
val DIFF_CONV = fn: conv
val basic_diffs = ref []: thm list ref
```

Then the user can call `DIFF_CONV` to calculate the derivative of basic arithmetic functions including `pow`:
```
> DIFF_CONV ``\x. x pow 2``;
val it = ⊢ ∀x. ((λx. x²) diffl (2 * x pow (2 − 1) * 1)) x: thm
> DIFF_CONV ``\x. inv x``;
<<HOL message: more than one resolution of overloading was possible>>
val it = ⊢ ∀x. x ≠ 0 ⇒ ((λx. x⁻¹) diffl -(1 / x²)) x: thm
> DIFF_CONV ``\x. (x + 1) pow 2``;
val it = ⊢ ∀x. ((λx. (x + 1)²) diffl (2 * (x + 1) pow (2 − 1) * (1 + 0))) x: thm
```

If `transcTheory` is further loaded (no need to open it):
```
> load "transcTheory";
val it = (): unit
```

Then the user can additionally call `DIFF_CONV` to calculate the derivative of some transcendental functions (`sin`, `cos`, `exp` and `ln`):
```
> DIFF_CONV ``\x. sin x``;
val it = ⊢ ∀x. ((λx. sin x) diffl (cos x * 1)) x: thm
> DIFF_CONV ``\x. exp (x pow 2)``;
val it = ⊢ ∀x. ((λx. exp x²) diffl (exp x² * (2 * x pow (2 − 1) * 1))) x: thm
> DIFF_CONV ``\x. ln (1 + x)``;
val it = ⊢ ∀x. 0 < 1 + x ⇒ ((λx. ln (1 + x)) diffl ((1 + x)⁻¹ * (0 + 1))) x: thm
```

Actually in many proofs of `transcTheory` the function `DIFF_CONV` does support transcendental functions (whenever the corresponding functions are defined), by adding new theorems into the ref variable `basic_diffs` defined in `Diff.sml`, but this facility wasn't exported to other theories. This can be resolved by adding the following code at the end of `transcTheory.sml` (by calling `Theory.quote_adjoin_to_theory` in `transcScript.sml`):
```
local in
  open Diff Drule;

  val _ = basic_diffs := !basic_diffs@[DIFF_EXP, DIFF_SIN, DIFF_COS];
  val _ = basic_diffs := !basic_diffs@(CONJUNCTS DIFF_COMPOSITE);
  val _ = basic_diffs := !basic_diffs@[DIFF_LN];
  val _ = basic_diffs := !basic_diffs@[SPEC_ALL DIFF_LN_COMPOSITE];
end
```

I took this chance to fix the entire `transcTheory` removing the call of `ParseExtras.temp_loose_equality()` at the beginning. But more importantly, I fixed the previously commented two theorems `MCLAURIN_LN_POS` and `MCLAURIN_LN_NEG` with changes to their existing proof scripts minimized:
```
   [MCLAURIN_LN_NEG]  Theorem      
      ⊢ ∀x n.
          0 < x ∧ x < 1 ∧ 0 < n ⇒
          ∃t. 0 < t ∧ t < x ∧
              -ln (1 − x) =
              sum (0,n) (λm. x pow m / &m) + x pow n / (&n * (1 − t) pow n)
   
   [MCLAURIN_LN_POS]  Theorem      
      ⊢ ∀x n.
          0 < x ∧ 0 < n ⇒
          ∃t. 0 < t ∧ t < x ∧
              ln (1 + x) =
              sum (0,n) (λm. -1 pow SUC m * x pow m / &m) +
              -1 pow SUC n * x pow n / (&n * (1 + t) pow n)
```

I also found the reason why previous attempts to fix these proofs failed: at certain point the proof needs `DIFF_CONV` to calculate the derivative of `ln (1 + x)` and `ln (1 - x)` but a key theorem `DIFF_LN` was somehow missing in `basic_diffs`. Once I added it, the existing proofs just worked, with only minor rewriting issues.

I have also added some selftests for `DIFF_CONV`.   I will investigate how to support more transcendental functions (`rpow`, e.g.) in `DIFF_CONV` and (in long term) how to port `Diff.sml` to the more advanced `derivativeTheory`.

Regards,

Chun Tian


binghe:
Hi,

In the additional commit, I have added support of functions `tan`, `asn` (arcsin), `acs` (arccos), and `atn` (arctan) to `DIFF_CONV` with unit tests.   Besides, I have moved the definitions of convex/concave functions and Jenson's inequalities from "miller" example to core real theories (`transcTheory` and `real_sigmaTheory`, plus a little to `util_probTheory`), as these concepts/theroems are important in basic calculus.

Chun

binghe:
d2ce818 added `EXP_LE_X_FULL` (in `extrealTheory`, together with its two extreal versions) as an application of the just recovered powful `MCLAURIN_LN_NEG`:

```
   [EXP_LE_X_FULL]  Theorem      
      ⊢ ∀x. 1 + x ≤ exp x
```
which is a stronger version of the existing theorem `EXP_LE_X` in `transcTheory`:
```
   [EXP_LE_X]  Theorem      
      ⊢ ∀x. 0 ≤ x ⇒ 1 + x ≤ exp x
```

(I was unable to prove `EXP_LE_X_FULL` until we have `MCLAURIN_LN_NEG` recovered.)

mn200:
I don’t expect you to have realised this (I’ve hardly announced it), but I’m now deprecating `adjoin_to_theory`.  If you want to update a global theorem list resource, please use the entry-points in `ThmSetData`, in particular `export_with_ancestry`.  This will set up an attribute, so that users can do things like:
```
Theorem DIFF_EXP[difftool]: 
  ...
Proof
  ...
QED
```
You have to implement the input record to `export_with_ancestry` by providing a functional implementation wrt the `'value` abstract type.  In your case, that type could well just be `thm list`.  If this all seems too much to get your head around, let me know, and I’ll do it after merging this.

You can see two instances of this in the handling of monotonicity and rule-induction theorems in the `IndDefLib` implementation.

binghe:
Hi Michael,

thanks for letting me know this. But I'm afraid that I don't understand how to use `export_with_ancestry` instead just by looking at existing sample code in `IndDefLib`. So please help coding it (after merging this PR).

Please note that the following line:
```
val _ = basic_diffs := !basic_diffs@[SPEC_ALL DIFF_LN_COMPOSITE];
```
means that `DIFF_CONV` may not work correctly if you just export the original theorem `DIFF_LN_COMPOSITE` without doing that `SPEC_ALL`, thus the ongoing `[difftool]` property shouldn't be put as part of `DIFF_LN_COMPOSITE` but another local lemma defined as `SPEC_ALL DIFF_LN_COMPOSITE`.

--Chun

binghe:
Also, please feel free to quash all commits when merging this PR, because I made many times of mistakes leaving whitespaces or TABs in the committed code.

mn200:
Right: for things like the need to do `SPEC_ALL`, I would indeed do something like 
```
Theorem foo'[difftool] = SPEC_ALL foo
```

As for TABs, I strongly recommend setting the global emacs variable `indent-tabs-mode` to `nil`.  That will stop the automatic indentation from ever inserting a TAB.

Thanks for all the work getting us this far!

mn200:
See b24817032a5 for how to use `ThmSetData`.

binghe:
> See [b248170](https://github.com/HOL-Theorem-Prover/HOL/commit/b24817032a5f085b6301ede5c5966046045ec78f) for how to use `ThmSetData`.

Thanks, now I started to "feel" how to use `ThmSetData` (would need some practices, one day).

PR #897, opened 2021-03-05
Refine proofs in Fermat's Two Squares.

jhlchan:
Improve proofs of theorems in Fermat's Two Squares.


mn200:
Thanks for the updates!

jhlchan:
Very good.

PR #903, opened 2021-03-17
Fix sporadically occuring bug in lassie related to sempre

HeikoBecker:
SEMPRE sporadically adds line-breaks to the information it prints on stdout.
Due to the way how SEMPRE and Lassie communicate, this meant that there was a slight chance for terms like `x <= y <=> ...` to be understood as the SEMPRE prompt instead of being a part of the full response. This lead to inconsistencies between SEMPRE's and Lassie's internal states.
This PR fixes this randomly occuring bug by making the SEMPRE prompt syntactically different from HOL4 symbols.


mn200:
Many thanks for your work on this!

PR #905, opened 2021-04-06
Add tactic cheatsheet

hrutvik:
Ported from the cheatsheet at https://hrutvik.co.uk/hol4-tactic-cheatsheet, which is itself adapted from http://sange.fi/~magnus/cheatsheet.txt due to @myreen.

The main changes are:
 - Github-flavoured markdown doesn't appear to permit tables without headers, so each table has an empty header row.
 - Line-breaks may occur mid-tactic.
 - No table of contents.


mn200:
Can you 
- move it into a separate directory under Manual, using the naming convention such that sub-dir `Foo` can generate a `foo.pdf`
- link to that directory in the master `Manual/Holmakefile`
- write a Holmakefile in the new sub-directory to generate a PDF file (assume and use `pandoc`)

?

Additionally: I can see the value in using Github flavoured markdown, but I think `pandoc`’s supported features are more generous, and you could generate an HTML locally if you wanted to support people looking at the cheatsheet in their browsers.

myreen:
@mn200 do you want (1) a pandoc-generated PDF or (2) a nice github rendering of markdown? 

I would like to prioritise (2). Yes, `pandoc` can produce a html and yes I guess I can build my own html file for viewing, but wouldn't it be much nicer to have a markdown file that one can just have a bookmark for in the browser and that will display the _latest_ github version directly (to anyone, i.e. I could share that link with students without them having to have pandoc installed or build things).

Having chatted with @hrutvik, it seems that pandoc doesn't support code snippets as well as github markdown. In other words, it might be difficult to have both (1) and (2). One could possibly have both, but one might look sub-optimal.

mn200:
I can just as easily make a bookmark to a local HTML file that is generated as HOL builds.  That will then contain information that is correct for my installation.

mn200:
`pandoc` supports writing ```
Here is some embedded code with italicised arguments: <code>tactic _arg_</code>
```
We could skip the generation of the PDF to focus on the HTML, and for this, the above seems good enough.  Or are there are other features still a problem?

hrutvik:
To summarise/document some offline discussion and some decisions:

## Motivation
@myreen's [cheatsheet](http://sange.fi/~magnus/cheatsheet.txt) has been very useful to those learning HOL, but has become outdated. I produced an updated version on [my website](https://hrutvik.co.uk/hol4-tactic-cheatsheet), with input from other HOL users. @myreen suggested PRing this into HOL itself so that it would be accessible and could be maintained.

## Usage
At first, the thinking was a GitHub-flavoured Markdown version could display nicely on GitHub itself - hence the restrictions in the initial pull request. @mn200 above suggested output to a PDF file using `pandoc`, allowing the cheatsheet to be built locally as part of the documentation.

The use of italics to represent "metavariables" in usage of tactics etc. was an issue. Markdown does not seem to permit italics in code blocks without use of the HTML `<code>` tag - but `pandoc` was not translating this to PDF output.

A more experienced `pandoc` user may be able to get this working, but for now the output is HTML only. In any case, this will hopefully give the best of both worlds - the HTML file can be built locally and viewed using a browser, and perhaps a relatively up-to-date version can be hosted on the HOL website. The accompanying CSS is ported almost directly from my website.

**Result:** a HTML page generated from Markdown using `pandoc` (with a little custom CSS), built as part of the documentation.

## Tables vs. definition lists
- The first iteration of this cheatsheet used tables to display tactics etc. alongside their descriptions.
  - This made CSS super easy on my website when prototyping
  - But resulted in source which is difficult to maintain due to restrictions of Markdown tables
- When moving to `pandoc`, there were [a few options](https://pandoc.org/MANUAL.html#tables)
  - `multiline_tables` and `pipe_tables` appear to require some element of making column widths in the source match the desired output widths
  - `simple_tables` has similar poor readability as the GitHub-flavoured Markdown pipe tables
  - `grid_tables` could work, but would still be difficult to edit
- Instead, I moved to definition lists - these are syntactically cleaner and semantically more appropriate
  -  The only cost I can see is that the CSS becomes a little hacky

mn200:
Awesome.  Thanks for this!

PR #908, opened 2021-04-08
merge productTheory into iterateTheory (with fixes)

binghe:
Hi,

the main purpose of this PR is to cleanup `src/probability` folder by merging the small `productTheory` (593 lines) into `iterateTheory` (3796 lines), both ported from HOL Light (to support `real_topologyTheory`).

Currently `productScript.sml` is the smallest file in `src/probability`, and it's very unlikely to add more contents into it in the future, thus I consider merging it into `iterateScript.sml` should be a good idea. Actually, the title of `iterateTheory` (Generic iterated operations and special cases of sums over N and R) has already indicated the merge (special cases of products (like sums) over N and R were defined in `productTheory`).

Furthermore, in `extrealTheory` the finite product of extreals `EXTREAL_PROD_IMAGE` (`PI`) was defined by `ITSET` (following the same ideas of `EXTREAL_SUM_IMAGE`). Now I have changed to use `iterate` instead:
```
Definition ext_product_def :
    ext_product = iterate (( * ):extreal->extreal->extreal)
End
Overload EXTREAL_PROD_IMAGE = “\f s. ext_product s f”
```

By re-proving the main theorem `EXTREAL_PROD_IMAGE_THM` using theorems from iterateTheory, all other proofs about `EXTREAL_PROD_IMAGE` remain the same. But if in the future we need more properties, they can be more easily derived from `iterateTheory`.

Once `extrealTheory` starts to depend on `iterateTheory`, I immediately found that there's a theorem `REAL_LE_SUP` (in `iterateTheory`) conflicting with a theorem with the same in `realTheory`. Thus I renamed it to `REAL_LE_SUP'` and fixed all proofs using it (similarly with previous treatments to `REAL_SUP_LE`). In this way, when `iterateTheory` were open by other end users, both `REAL_LE_SUP` and `REAL_LE_SUP'` will be available.

The removal of **`productTheory`** has been noted in `next-release.md`.  There are also some other small additions in `extrealTheory`.

Regards,

Chun Tian


konrad-slind:
Hi Chun Tian,

 All this is very nice, thank you for doing this. I had a look
over iterateScript.sml. Lots of the definitions and theorems,
e.g. segments and sums, would be helpful in theories that use
only natural numbers. At some point it might be worth migrating
some of the theory down to src/pred_set. It's a question of
visibility, really. I tend not to think about probability
theories when I go looking for definitions and theorems about
sets of nums. But there is obviously lots of general stuff in the
probability theories.

Maybe the solution is just to have better search facilities.

Konrad.

binghe:
Hi Konrad,

I believe `DB.match` can effectively find useful real/integer/set theorems in `iterateTheory` but you're right that end users will  never considering "iterateTheory" into the theory list of `DB.match` when searching theorems.

I see `iterateTheory` mainly as a HOL-Light compatibility layer, which contains many alternative definitions or theorems of those already in `pred_setTheory` and `realTheory`. Having it, porting new theories from HOL Light would be much easier.

It's also true that currently the following theories are not directly related to probability theory (and lebesgue integration) but belongs to general real analysis:

```
     real_topologyScript.sml      * Elementary Topology in Euclidean Space (from HOL Light)
     derivativeScript.sml         * Univariate Derivative Theory in R^1 (from HOL Light)
     integrationScript.sml        * Henstock-Kurzweil (gauge) Integral (from HOL Light)
     iterateScript.sml            * Generic iterated operations and special cases of sums over N and R
```

I would suggest moving them into a separate `$(HOLDIR)src/analysis` directory as the first major step of cleaning up `src/probability`.    Of course, it's also possible to just move `iterateTheory` into `src/num/theories` with set- and real-stuff moving to other places.

Let me know what do you prefer.

Chun Tian

konrad-slind:
Thanks for responding so quickly. Let's see what Michael wants to do.
Moving stuff earlier in the theory hierarchy is, long-term, good hygiene.
But may easily be deferred.

Konrad.

binghe:
I forgot to add that, in current `iterateTheory` there're some definitions which belongs to algebra:
```
[monoidal]  Definition      
      ⊢ ∀op.
          monoidal op ⇔
          (∀x y. op x y = op y x) ∧
          (∀x y z. op x (op y z) = op (op x y) z) ∧
          ∀x. op (neutral op) x = x
   
   [neutral]  Definition      
      ⊢ ∀op. neutral op = @x. ∀y. op x y = y ∧ op y x = y
```

I believe, if we can move something from `examples/algebra` to HOL's core library (into `$(HOLDIR)/src/algebra`, e.g.) , at least those up to Group Theory, although it may sound "using canon to shoot mosquito", but the above `monoidal` and `neutral` wouldn't be needed any more.   For this reason, I suggest delaying moving `iterateTheory` to live with arithmetic theories, not until HOL4 got a core algebra library.

Chun

mn200:
There's definitely stuff that can be done in the medium or longer term here, but I'm happy to do it in small steps.   Thanks for the work so far!

PR #909, opened 2021-04-19
Move find_consts and find_consts_thy to DB

HeikoBecker:
Minor PR moving the functions `find_consts` and `find_consts_thy` to the DB file instead of bossLib.
@AndreasLoow and I discussed this on the CakeML slack. Ideally this functionality should be in the DB structure to make it easier to find.


mn200:
Please fix the Travis-detected regressions.

HeikoBecker:
They should be fixed now. Funnily enough, it looks like I found a bug in the `find_consts` functions. The `val` assignment used was catching the call to `ancestry` too early, and therefore not done when the function is called, but rather at the time when the `DB` structure is build. This should be fixed by changing it into a `fun` binding.

mn200:
Nice find!  Thanks for your work on this.

PR #913, opened 2021-05-14
Merge integrationTheory.indicator with real_borelTheory.indicator_fn

binghe:
Hi,

An indicator_function `𝟙 s` simply returns either 0 or 1 _indicating_ if a given value `x` belongs to a given set `s`, but it has enormous applications in real analysis (both Lebesgue and HK integrations).  Currently there are three versions of indicator functions in HOL core theories:

```
(* integrationTheory, line 4526- *)
val indicator = new_definition ("indicator",
  ``indicator s :real->real = \x. if x IN s then 1 else 0``);

(* real_borelTheory, line 57- *)
val indicator_fn_def = Define
   `indicator_fn s = \x. if x IN s then (1:real) else (0:real)`;

(* borelTheory, line 52- *)
Definition indicator_fn_def :
    ext_indicator_fn s = \x. if x IN s then (1 :extreal) else (0 :extreal)
End
val _ = overload_on ("indicator_fn", “ext_indicator_fn”);
```

The purpose of this PR is to merge the above definitions as much as possible. The overloading of `ext_indicator_fn` is especially problematic: from just the types of `s` and `x` the type system cannot decide the type of `indicator s x`, which could either be `:real` or `:extreal`. Many existing proofs were not broken only because the last definition comes at last (in the chain of theory dependencies) and thus `:extreal` is preferred. But once I try to move the definition to an early theory (`extrealTheory`), many proofs are broken by this type confusion.

Also note that `integrationTheory.indicator` was originally ported (by HVG concordia people) from HOL Light. Below is the original definition:
```
(* Multivariate/vectors.ml, line 4132- *)
let indicator = new_definition
   `indicator s :real^M->real^1 = \x. if x IN s then vec 1 else vec 0`;;
```
It looks more general (i.e. supporting multi-dimensional real sets) but actually not. Our `borelTheory.indicator_fn` is even more general, as it simply supports any type of sets (of `:'a -> bool`). But this generalization shouldn't change the proofs of properties of indicator functions.  In fact, no proofs in `integrationTheory` is broken if I simply use the general definition from `real_borelTheory` instead!

After this PR, we have just two (closely-related) versions of indicator functions. The `:real` version is now in `iterateTheory` (which is in preparation for moving into `src/real` in future work):

```
(* This is originally from HOL Light (Multivariate/vectors.ml). Generalized. *)
Definition indicator :
    indicator (s :'a -> bool) :'a -> real = \x. if x IN s then 1 else 0
End
```

The `:extreal` version is now in `extrealTheory` (together with `fn_plus` and `fn_minus` moved in from `borelTheory`) with the following simple definition (its old definition now becomes an equivalent theorem):
```
Definition indicator_fn :
    indicator_fn s = Normal o indicator s
End
```

Now the legacy real-based probability theories (`real_measure`, `real_lebesgue` and `real_probability`) needs to use the following local overloading to make sure all `indicator_fn` in the existing theorem statements and proofs are still there:
```
Overload indicator_fn[local] = “indicator”
Theorem indicator_fn_def[local] = indicator
```
But new user code of real-based probability should directly use `indicator` instead.

As a (good) side effect, which was actually my original intension, the size of the file `borelScript.sml` is reduced to less than 10,000 lines, with pure-extreal stuff moved out to their more suitable places.    Besides, the definition and lemms of `from n` is also moved into `iterateTheory` (to live with `numseg`) from `real_topologyTheory`.

Regards,

Chun Tian


mn200:
Thanks for continuing to work on this impressive project!

mn200:
I believe this change may have broken the dining-cryptos example.

binghe:
> I believe this change may have broken the dining-cryptos example.

Confirmed. Sorry for not testing enough. I'm isolating this issue, which looks weird (after opening `iterateTheory` for "indicator_fn_def").

binghe:
> I believe this change may have broken the dining-cryptos example.

I've fixed the issue. There are two issues actually. The easy one is the missing of `indicator_fn` and "indicator_fn_def", which can be fixed by putting the following two lines in "informationScript.sml" as the compatibility fix:
```
Overload indicator_fn[local] = “indicator”
Theorem indicator_fn_def[local] = indicator
```

The hard one is actually caused by the overloading of "CROSS" of `prod_sigma` in `sigma_algebraTheory` long time ago.  In "leakageLib.sml" there are several occurrences of "$CROSS" which is supposed to mean `pred_set$CROSS` but wrongly interpreted as `sigma_algebra$CROSS`.  The fix is to explicitly use `pred_set$CROSS` instead.   I failed to see what exact change trigger this issue but I think it should fail immediately since the overloading of "CROSS" happens.

Anyway, I have put the fix as part of PR #918

PR #916, opened 2021-05-19
Replaced 'throw(..)' (removed in C++17) with 'noexcept(false)' in minisat.

HAKarlsson:
I replaced 'throw(Exception_EOF)' with 'noexcept(false)' in minisat. Dynamic exception specification (i.e., throw(..)) was removed from C++17, the replacement is 'noexcept(false)' according to [Removing Deprecated Exception Specifications from C++17](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2015/p0003r0.html).

This commit ensures that minisat can be compiled by g++ version 11 and later.

### DIFF
--- a/src/HolSat/sat_solvers/minisat/File.C
+++ b/src/HolSat/sat_solvers/minisat/File.C
@@ -118,8 +118,11 @@ void putUInt(File& out, uint64 val)
 }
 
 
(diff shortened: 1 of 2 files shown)


mn200:
Thanks for this!

PR #918, opened 2021-05-21
Redefine REAL_SUM_IMAGE based on iterateTheory (now moved to src/real)

binghe:
Hi,

In `iterateTheory` ("src/probability", ported from HOL Light) there's the definition `sum` of the values of real functions over a (finite) set:
```
   [sum_def]  Definition      
      ⊢ sum = iterate $+
```
while in HOL's own `real_sigmaTheory` ("src/real") there's a similar concept `REAL_SUM_IMAGE` (or `SIGMA`) widely used by user code and other HOL core theories:
```
   [REAL_SUM_IMAGE_DEF]  Definition      
      ⊢ ∀f s. SIGMA f s = ITSET (λe acc. f e + acc) s 0
```
They only differ at the order of arguments: `sum s f = SIGMA f s`, but `iterateTheory` is more advanced (to be explained below) with more useful lemmas. The main purpose of this PR is to re-define `REAL_SUM_IMAGE` by `sum` so that:

1.  many existing `SIGMA` theorems now have trivial proofs based on their `sum` version;
2. new `SIGMA` theorems can be easily migrated from their `sum` versions.

```
   [REAL_SUM_IMAGE_DEF]  Definition (NEW)      
      ⊢ ∀f s. SIGMA f s = sum s f
```

As the result, the size of `real_sigmaTheory` is reduced with more `SIGMA` theorems provided. And we take this chance to move `iterateTheory` from "src/probability" to "src/real" (so that `real_sigmaTheory` can depend on it).   (Note that it seems possible to further define `SIGMA` as an overloading of `\f s. sum s f`, but when I tries this, some proofs are broken, then I gave up.)

It turns out that `iterateTheory` is more advanced: now some `SIGMA` theorems can be proved without finiteness assumptions on the involved sets! For example, the following "primed" versions are now added:
```
   [REAL_SUM_IMAGE_0']  Theorem      
      ⊢ ∀s. SIGMA (λx. 0) s = 0

   [REAL_SUM_IMAGE_EQ']  Theorem      
      ⊢ ∀s f f'. (∀x. x ∈ s ⇒ f x = f' x) ⇒ SIGMA f s = SIGMA f' s

   [REAL_SUM_IMAGE_IMAGE']  Theorem      
      ⊢ ∀P f'.
          INJ f' P (IMAGE f' P) ⇒
          ∀f. SIGMA f (IMAGE f' P) = SIGMA (f ∘ f') P

   [REAL_SUM_IMAGE_POS']  Theorem      
      ⊢ ∀f s. (∀x. x ∈ s ⇒ 0 ≤ f x) ⇒ 0 ≤ SIGMA f s
```
These theorems are impossible to prove if `SIGMA` were defined by `ITSET` based on induction of finite sets. But they are all mathematically true, even the involved set is infinite or even uncountable!  (I didn't dig into `iterateTheory` to see how this can be achieved).  On the other hand, it's also true that not every `SIGMA` theorem can drop their finiteness assumptions. For example, the following theorem on addition
```
   [REAL_SUM_IMAGE_ADD]  Theorem      
      ⊢ ∀s. FINITE s ⇒
            ∀f f'. SIGMA (λx. f x + f' x) s = SIGMA f s + SIGMA f' s
```
doesn't hold for the sum of infinite series (even countable), if I remember correctly from the textbook.   Any way, by dropping some finiteness antecedents from the existing theorems may lead to shorter proofs while the proof is still mathematically precise (the present case is totally different with division-by-zero of reals).

I did a small "editor-of-choice" selection of `sum` lemmas and translated them into `SIGMA` lemmas (using a simple ML function `translated` I defined in "real_sigmaScript.sml"). The most impressive one is the following lemma about permutation (self bijection):
```
   [REAL_SUM_IMAGE_PERMUTES]  Theorem      
      ⊢ ∀f p s. p PERMUTES s ⇒ SIGMA f s = SIGMA (f ∘ p) s
```
It's amazing that this theorem universally holds for whatever sets, even uncountably large (as long as the `SIGMA` result is specified).

Recall that the finite product of extreals (`extreal$EXTREAL_PROD_IMAGE`) was changed to based on also `iterateTheory`. I took this chance to re-prove some of its lemmas using advanced lemmas from `iterateTheory` and slightly shorten the size of "extrealScript.sml".  This is basically why in this PR more theorems are added but more lines are deleted than added.

Currently there are still many useful lemmas about real numbers that could be moved to `realTheory` for easier finding by end users. But many of these lemmas need `real_ss`, which is not available in `realTheory`.  I think for now it's better to leave them as is, since `realTheory` is already very big.

Regards,

Chun Tian

### DIFF
--- a/src/real/real_sigmaScript.sml
+++ b/src/real/real_sigmaScript.sml
@@ -28,16 +28,8 @@ val art = ASM_REWRITE_TAC;
     Where f's range is the real numbers and P is finite.
    ---------------------------------------------------------------------- *)
 
-(* The old definition (only used by REAL_SUM_IMAGE_THM):
-
 Definition REAL_SUM_IMAGE_DEF :
     REAL_SUM_IMAGE f s = ITSET (\e acc. f e + acc) s (0:real)
-End
-
-   The new definition (based iterateTheory):
- *)
-Definition REAL_SUM_IMAGE_DEF :
-    REAL_SUM_IMAGE f s = iterate$Sum s f
 End
 
 Overload SIGMA = “REAL_SUM_IMAGE”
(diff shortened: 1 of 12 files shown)


binghe:
Please ... squash all commits if they are going to be merged.

src/real/real_sigmaScript.sml
```diff
@@ -26,126 +28,90 @@ val POP_ORW = POP_ASSUM (fn thm => ONCE_REWRITE_TAC [thm]);
...
+(* The old definition (only used by REAL_SUM_IMAGE_THM):
+
+Definition REAL_SUM_IMAGE_DEF :
+    REAL_SUM_IMAGE f s = ITSET (\e acc. f e + acc) s (0:real)
+End
+
+   The new definition (based iterateTheory):
+ *)
+Definition REAL_SUM_IMAGE_DEF :
+    REAL_SUM_IMAGE f s = iterate$Sum s f
```
ahmedwaqar, line 40: Both definitions are essentially the same. `REAL_SUM_IMAGE` is defined using `ITSET` to sum over uncountable as well as uncountable sets. On the other hand, `iterate.sum` is defined using `iterate (+)` which is potentially boils down to the same definition in `SUM_CLAUSES`as `REAL_SUM_IMAGE_THM`.


 I think instead of redefining `REAL_SUM_IMAGE` on `sum`. You can simply prove an equivalency relationship between `iterate.sum` and `REAL_SUM_IMAGE` and switch to `sum` at your ease. The use of `REAL_SUM_IMAGE` is prevalent in most of the subsequent theories, which might break with your change.

ahmedwaqar (changes requested):

binghe:
Hi @ahmedwaqar , thanks for your review. Now I have taken your advice and recovered the previous definition of `REAL_SUM_IMAGE_DEF` and use an equivalent theorem to link with `sum`:
```
   [REAL_SUM_IMAGE_sum]  Theorem      
      ⊢ ∀f s. FINITE s ⇒ SIGMA f s = sum s f
```
Note that, the above theorem does not hold without `FINITE s`.  Actually there's no magic behind `sum`: without finiteness assumptions `sum s f = 0` by definition.  Thus many of my previous words (related to "amazing") are not true.

Anyway, as the final result, all those "primed" versions are now gone. But still we have all those simplified proofs in `real_sigmaTheory` due to the reduction to the related proofs from `iterateTheory`.

binghe:
Commit ae58b66 should have fixed the left issues (broken "diningcryptos" example) in PR #913

mn200:
Thanks for this!

PR #920, opened 2021-05-26
Add theory-qualification to the TRUTH theorem in QuoteFilter

oskarabrahamsson:
Prior to this addition, `Holmake` would fail with a non-informative error message ("Static Errors") on theories containing Definition-syntax but not having `TRUTH` in scope. The same script would run without error in an interactive session. Below is an example when Holmake will fail, but an interactive session succeed.

Here is the script:
```sml
(* fooScript.sml *)
open HolKernel;

val _ = new_theory "foo";

Definition foo_def:
  foo = ()
End

val _ = export_theory ();
```
And here is the failure:
```
oskar@macintosh: ~/tmp Holmake
fooTheory                                                                 real:    0s  user:    0sFAIL<1>
 <<HOL message: Created theory "foo">>
 Saved definition __ "foo_def"
 error in quse /Users/oskar/tmp/fooScript.sml : Fail "Static Errors"
 error in load /Users/oskar/tmp/fooScript : Fail "Static Errors"
 Uncaught exception: Fail "Static Errors"
```
And here is the same script in an interactive session:
```
oskar@macintosh: ~/tmp hol

---------------------------------------------------------------------
       HOL-4 [Kananaskis 14 (stdknl, built Wed Mar 24 14:14:03 2021)]

       For introductory HOL help, type: help "hol";
       To exit type <Control>-D
---------------------------------------------------------------------
[Use-ing configuration file /Users/oskar/.hol-config.sml]
> Vim input /tmp/vimhol0Script.sml
HOLLoadSendQuiet HolKernel completed
Vim input /tmp/vimhol0Script.sml
<<HOL message: Created theory "foo">>
Definition has been stored under "foo_def"
val foo_def =  [] ⊢ foo = (): thm
Exporting theory "foo" ... done.
Theory "foo" took 0.09913s to build
```

### DIFF
--- a/tools/Holmake/QuoteFilter
+++ b/tools/Holmake/QuoteFilter
@@ -198,7 +198,7 @@ fun magic_bind (arg as QFS{induction_thm,...}) =
     Systeml.bindstr ("val " ^ !induction_thm ^
                      " = DB.fetch \"-\" \"" ^
                      !induction_thm ^
-                     "\" handle HOL_ERR _ => TRUTH;") ^ ";"
+                     "\" handle HOL_ERR _ => boolTheory.TRUTH;") ^ ";"
 
 type lexresult = int*string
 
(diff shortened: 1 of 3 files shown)


mn200:
Nice catch!  But can you fix the regression failures?  (They arise because they check for exact string matches in the filter's output; see the `expected` files in `tools/Holmake/tests/quote-filter`.)

mn200:
Thanks for your work on this!

PR #923, opened 2021-06-05
Merged Q_set (rational subset of reals) proofs of real and extreal

binghe:
Hi,

ration numbers (or rational subset of reals/extreals) are invaluable for proving the Borel measurability of some sets. Currently there are two versions of `Q_set`, one is in `real_borelTheory`:
```
   [real_rat_set_def]  Definition      
      ⊢ q_set =
```
the other was in `borelTheory` but then moved to `extrealTheory`:
```
   [Q_set_def]  Definition      
      ⊢ ℚ =
```

The proofs of Q_set lemmas for the extreal version were directly ported from the real version, but this is unnecessary. The purpose of this PR is remove these duplicated proofs, by re-defining the extreal version of `Q_set` by the real version:
```
   [Q_set]  Definition      
      ⊢ ℚ = IMAGE Normal q_set
```
and then many proofs for the extreal-valued `Q_set` are simplified.  Since now `extrealTheory` depends on some thing that is in `real_borelTheory`, I have to move "real_rat_set_def" and related lemmas to upper theory, now in `util_probTheory`.

P. S. I'm also thinking how to use `ratTheory` to achieve the same goal. But anyway the present work is the first step of reducing duplicated proofs.

Regards,

Chun Tian


mn200:
Thanks for this!

PR #924, opened 2021-06-09
Thibault wip

barakeel:
Latest version of TacticToe:
 -  can now propose a tactic script even when failing with the function suggest ()
 -  updated search tree (now uses a real search tree instead of a dictionary)
 -  support for tree neural networks advice
 - and many small changes


barakeel:
Re-opening it to trigger a rebuild.

PR #925, opened 2021-06-10
Thibault wip

barakeel:
Latest version of TacticToe:

- can now propose a tactic script even when failing with the function suggest ()
- support for tree neural networks advice   
- updated search tree for both TacticToe and psMCTS (now uses a real search tree instead of a dictionary)
- ongoing reinforcement learning experiments in examples/AI_tasks.
- last commit fix missing directory


mn200:
I’m not keen on linking into `sigobj` from `examples` directories.  sigobj is supposed to be for `src` material, which is guaranteed to be built with HOL is built. If you want people to be able to refer to material in `examples`, get them to use explicit `INCLUDES` lines referring to those directories.  E.g.,
```
INCLUDES = $(HOLDIR)/examples/AI_tasks/...
```
The action of including that line in a `Holmakefile` will cause those files to be built as an automatic dependency.

Alternatively, if it’s material that really should be built all the time and really should appear in `sigobj`, move it into `src`.

barakeel:
Thanks, I'll try to see how I can perform external parallelization (spawn multiple instances of buildheap on generated files) without linking to sigobj.

barakeel:
I think I addressed the problem by generating automatically Holmakefile files in some cases or telling the user to do it in other cases either in the documentation or in a README.

mn200:
Thanks!

PR #927, opened 2021-06-17
Add support for passing variable defs to Holmake as arguments

### LINKED ISSUE #100: Allow VAR=value command-line parameters for Holmake
Among other things, this functionality would allow build to call Holmake with a SELFTESTLEVEL variable defined.  
## 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/25096603-allow-var-value-command-line-parameters-for-holmake?utm_campaign=plugin&utm_content=tracker%2F495335&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F495335&utm_medium=issues&utm_source=github).
</bountysource-plugin>

oskarabrahamsson:
Adds support for `VAR=value` style command-line arguments to Holmake. If it works, then this PR should close #100.

Here is a summary of the change: When Holmake is done parsing command-line options, the remaining arguments are separated into those that match `<ANYTHING>=<ANYTHING>` and those that do not. The latter are treated as before, while the former are inserted into the base environment _after_ the holpathdb is read, and _before_ the Holmakefile is read, each time Holmake reads a Holmakefile (this is the intent, at least).

### DIFF
--- a/tools/Holmake/Holmake.sml
+++ b/tools/Holmake/Holmake.sml
@@ -66,16 +66,22 @@ end
 (*** parse command line *)
 fun apply_updates fs v = List.foldl (fn (f,v) => #update f (warn,v)) v fs
 
-fun getcline args = let
-  open GetOpt
-in
-  getOpt {argOrder = Permute,
-          options = HM_Cline.option_descriptions,
-          errFn = die}
-         args
-end
+fun getcline args =
+  let
+    open GetOpt
+    val (opts, rest) = getOpt {argOrder = Permute,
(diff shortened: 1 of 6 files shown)


mn200:
Sounds awesome!  Can you write test-cases for the feature under `Holmake/tests` and then out the test into the build sequence please?

mn200:
Awesome!  Thanks again.

PR #929, opened 2021-07-05
EmitTeX: handle primed variables that contain unicode numerals

acjf3:


src/TeX/EmitTeX.sml
```diff
@@ -269,6 +269,16 @@ local
       case overrides s of
         NONE => string_map (s,sz)
       | SOME r => r
+  fun translate_prime s =
```
mn200, line 272: As  `x'¹⁰'` and similar is possible, it seems like you should check for `sup_1` and `sup_0` too.

acjf3: Agreed

mn200:
Awesome; thanks!

PR #930, opened 2021-07-05
Small tweaks to the cheatsheet

hrutvik:
- Add `measureInduct_on` on suggestion from @konrad-slind 
- Prefer `namedCases_on` over `tmCases_on`
- Add `Excl` on correction from @myreen


mn200:
Thanks for this!

PR #931, opened 2021-07-12
Add various small theorems

xrchz:


src/bag/bagScript.sml
```diff
@@ -1816,6 +1831,30 @@ val BAG_IMAGE_EQ_EMPTY = store_thm(
   simp[]);
 val _ = export_rewrites ["BAG_IMAGE_EQ_EMPTY"]
 
+Theorem ITSET_BAG_INSERT_BAG_UNION_BAG_IMAGE_BAG_OF_SET:
```
mn200, line 1834: This theorem name is kind of gross.

mn200 (approved):

src/bag/bagScript.sml
xrchz: Agreed, but I think it's more of a lemma for the theorem below. (And both of these theorems aren't important enough to have nice names.)

PR #932, opened 2021-07-21
Define the field of real numbers

xrchz:
These are defined in separate theories (as opposed to the existing
Instances theories) to avoid parsing confusions between other numeric
types and reals.


mn200:
Cool!

PR #933, opened 2021-07-22
Merged some duplicated real lemmas; new NUM_FLOOR lemmas

binghe:
Hi,

currently there are multiple versions or copies of REAL_POW (`pow`) related lemmas found in `iterateTheory`, `integralTheory` and `real_topologyTheory`. I believe they all came from HOL Light, and the duplications must be from the porting work of these theories when they were not merged into HOL4 together. The most important one is REAL_POW_ARCH which is not easy to prove if an end user couldn't find it (e.g. only from `realTheory`):

```
   [REAL_ARCH_POW]  Theorem      
      ⊢ ∀x y. 1 < x ⇒ ∃n. y < x pow n
```

In this PR, I have cleaned up many such duplications in the above mentioned theories, and then I moved a single copy of them into `realTheory`.  Whenever some proofs require things like `REAL_ARITH` or `real_ss` (not accessible yet in `realTheory`), I reworked the proofs removing the dependencies.

(Previously `realTheory.REAL_ARCH_POW`, moved from `util_probTheory`, was proven only for `x pow 2`, now the old theorem was renamed to "REAL_ARCH_POW2", following the one in `real_topologyTheory`), with corresponding user code fix in `extrealTheory`.)

In addition, I also added some new NUM_FLOOR (`flr`)-related lemmas into `realTheory`, to try to complete the existing lemmas. They are already proven useful (in my other ongoing project(s)) for eliminating inequalities about `flr(x)` or `&flr(x)` which the existing lemmas do not cover. The theorem names were chosen following naming schemas of similar lemmas:

```
   [NUM_FLOOR_LT]  Theorem      
      ⊢ ∀x. x − 1 < &flr x
   
   [NUM_FLOOR_MONO]  Theorem      
      ⊢ ∀x y. 0 ≤ x ∧ 0 ≤ y ∧ x ≤ y ⇒ flr x ≤ flr y
   
   [NUM_FLOOR_MUL_LOWERBOUND]  Theorem      
      ⊢ ∀n x. 1 < n ∧ 1 ≤ x ⇒ x < &(n * flr x)
   
   [NUM_FLOOR_lower_bound]  Theorem      
      ⊢ ∀x. 1 ≤ x ⇒ x / 2 < &flr x
```

Besides, the following lemma (about related NUM_CEILING or `clg`) is now promoted from `util_probTheory` to `realTheory` with a new name like the above:

```
   [NUM_CEILING_UPPER_BOUND]  Theorem      
      ⊢ ∀x. 0 ≤ x ⇒ &clg x < x + 1
```

Hope this helps,

Chun Tian


mn200:
Thank you!  It's nice to have a monotonically cleaner system emerge like this.

Issue #851, opened 2020-08-06
Polyscripter reads ~/.hol-config.sml when building manual PDFs


binghe:
Hi,

when I try to build HOL manuals, the `polyscripter` command reads my local `~/.hol-config.sml` file which may contain custom settings like `Globals.linewidth` which may change the final typesetting results.  As a workaround, I have to comment all harmful settings in my `~/.hol-config.sml` to make sure that I see the same manual PDFs as those built by others.

Is it possible to let `polyscripter` ignore `~/.hol-config.sml`?

Regards,

Chun Tian

mn200:
Good point.   Note that I think the options from your config file may be read at compile time, not runtime.

binghe:
Thanks.

PR #935, opened 2021-07-23
hol config

arolle:
Adds support for a `HOL_CONFIG` environment variable to allow for a custom `hol-config` configuration at a non-standard location or potentially ignoring any present hol config. If the variable is set, any other hol-config file will be ignored. If the value of `HOL_CONFIG` is a readable file, it will be used.
These changes are motivated by #851 and avoid resetting `$HOME` within the vimhol script.


mn200:
Awesome, thank you for this!

PR #936, opened 2021-07-26
Add some more bag theorems

xrchz:


src/bag/bagScript.sml
```diff
@@ -1585,6 +1597,12 @@ Proof
   rw[BAG_FILTER_DEF, FUN_EQ_THM, BAG_OF_SET] \\ rw[] \\ fs[]
 QED
 
+Theorem BAG_FILTER_SPLIT:
```
mn200, line 1600: I feel like this should be oriented the other way round so that it can be used as a rewrite directly.

xrchz: Agreed - done.

mn200:
Closing due to having been subsumed by #937

PR #937, opened 2021-07-26
Multivariate polynomials

xrchz:


examples/algebra/multipoly/multipolyScript.sml
```diff
@@ -0,0 +1,2232 @@
...
+  \\ pop_assum SUBST1_TAC
+  \\ `(λP. GBAG g (BAG_FILTER P b)) = λP. GBAG g (BAG_IMAGE I (BAG_FILTER P b))`
+  by simp[FUN_EQ_THM]
+  \\ pop_assum SUBST1_TAC
+  \\ irule GBAG_IMAGE_PARTITION
+  \\ simp[]
+  \\ metis_tac[]
+QED
+
+Theorem ring_mult_lsum:
```
mn200, line 417: Indentation here is gross.

mn200:
Does this subsume #936 ?

xrchz:
> Does this subsume #936 ?

It builds on top of it. So yes, in a sense.

examples/algebra/multipoly/multipolyScript.sml
xrchz: I don't see the issue - what's the proposed change?

xrchz: oh maybe GitHub was showing me the wrong quote - do you mean the statement of ring_mult_lsum? I will improve that.

mn200:
At some point, you will also need to put this into the build sequence.

xrchz:
`examples/algebra` is already in `tools/sequences/final-examples` -- is that not sufficient? I don't know if those sequences imply recursion into subdirectories (if not, then the current entry doesn't accomplish much).

mn200:
The sequences don't imply recursion, but the `Holmakefile` in `examples/algebra` achieves the recursion by mentioning the relevant directories in its `INCLUDES =` line.

PR #939, opened 2021-08-10
Add `unionWith` function to `balanced_mapTheory`

hrutvik:
As in Haskell's `Data.Map`, `unionWith` takes the union of two balanced maps, but uses its function parameter to combine values for keys present in both maps.

Theorems are stated in terms of a new constant in `finite_mapTheory`, `FMERGE_WITH_KEY` - this is like the existing `FMERGE`, but its function parameter processes the key as well as the values in the case where a key is present in both maps.

Some other theorems are also added to `finite_mapTheory` - these are mostly ported from https://github.com/cakeml/pure.

### DIFF
--- a/src/finite_maps/finite_mapScript.sml
+++ b/src/finite_maps/finite_mapScript.sml
@@ -669,6 +669,11 @@ val DRESTRICT_FUPDATE = Q.store_thm
             EXTENSION] THEN PROVE_TAC []);
 val _ = export_rewrites ["DRESTRICT_FUPDATE"]
 
+Theorem DRESTRICT_EQ_FEMPTY:
+  !m s. DISJOINT s (FDOM m) <=> DRESTRICT m s = FEMPTY
+Proof
+  ho_match_mp_tac fmap_SIMPLE_INDUCT >> rw[] >> eq_tac >> rw[]
+QED
 
 val STRONG_DRESTRICT_FUPDATE = Q.store_thm
 ("STRONG_DRESTRICT_FUPDATE",
@@ -1802,6 +1807,12 @@ val FUPDATE_LIST_SNOC = store_thm("FUPDATE_LIST_SNOC",
   ``!xs x fm. fm |++ SNOC x xs = (fm |++ xs) |+ x``,
   Induct THEN SRW_TAC[][FUPDATE_LIST_THM])
 
+Theorem DOMSUB_FUPDATE_LIST:
+  !l m x. (m |++ l) \\ x = (m \\ x) |++ (FILTER ($<> x o FST) l)
+Proof
+  Induct >> rw[FUPDATE_LIST_THM, fmap_eq_flookup] >>
+  PairCases_on `h` >> gvs[] >> simp[DOMSUB_FUPDATE_THM]
+QED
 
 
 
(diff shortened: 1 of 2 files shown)


src/finite_maps/finite_mapScript.sml
```diff
@@ -2943,6 +3129,91 @@ Proof
...
+
+Theorem FDIFF_FUPDATE:
+  FDIFF (fm |+ (k,v)) s =
+  if k IN s then FDIFF fm s else (FDIFF fm s) |+ (k,v)
+Proof
+  rw[fmap_eq_flookup, FLOOKUP_FDIFF, FLOOKUP_UPDATE] >>
+  EVERY_CASE_TAC >> gvs[]
+QED
+
+Theorem FDIFF_FEMPTY:
```
mn200, line 3187: This should be `simp` (if something else doesn't already achieve that)

src/finite_maps/finite_mapScript.sml
```diff
@@ -2943,6 +3129,91 @@ Proof
...
+  FDIFF FEMPTY s = FEMPTY
+Proof
+  rw[fmap_eq_flookup, FLOOKUP_FDIFF]
+QED
+
+Theorem FDIFF_EMPTY:
```
mn200, line 3193: Also `simp`

src/finite_maps/finite_mapScript.sml
```diff
@@ -2339,6 +2518,13 @@ val fmap_rel_FEMPTY2 = store_thm(
   METIS_TAC [FDOM_EQ_EMPTY]);
 val _ = export_rewrites ["fmap_rel_FEMPTY2"]
 
+Theorem fmap_rel_FEMPTY3:
```
mn200, line 2521: This should be `simp`

src/finite_maps/finite_mapScript.sml
```diff
@@ -1031,6 +1031,163 @@ val FLOOKUP_DRESTRICT = store_thm("FLOOKUP_DRESTRICT",
...
+
+Theorem FMERGE_WITH_KEY_DRESTRICT:
+  DRESTRICT (FMERGE_WITH_KEY f m1 m2) vs =
+  FMERGE_WITH_KEY f (DRESTRICT m1 vs) (DRESTRICT m2 vs)
+Proof
+  rw[fmap_eq_flookup, FLOOKUP_DRESTRICT, FLOOKUP_FMERGE_WITH_KEY] >>
+  every_case_tac >> gvs[]
+QED
+
+Theorem FMERGE_WITH_KEY_EQ_EMPTY:
```
mn200, line 1182: This should be `simp`.

mn200:
Looks good; would just add more `simp`s (as above)

binghe:
Hi @hrutvik,

Nice work! May I suggest you to also add a little documentation for the existing `FMERGE` and your `FMERGE_WITH_KEY` into HOL Description (at the end of Section 5.5.5 "Finite maps (finite_map)", for example)?   I never knew this functionality of finite maps, as otherwise I may have already used it in the CCS example.

In case you refuse then I can try to do the documentation instead, but obviously you are the expert here.

Regards,

Chun Tian

mn200:
I will merge this now, and await documentation as a separate PR.  Thanks for your work on this, Hrutvik!

Issue #941, opened 2021-08-14
Cannot start "hol" session in "$(HOLDIR)/src/boss"


binghe:
Hi,

after HOL building finished, if I go to "$(HOLDIR)/src/boss" and start "hol", I got the following error saying **Structure (bossLib) has not been declared**, although bossLib is just in the same directory:
```
binghe@binghe-mac6:~/ML/HOL/src/boss$ hol

---------------------------------------------------------------------
       HOL-4 [Kananaskis 14 (stdknl, built Thu Jul 29 17:16:26 2021)]

       For introductory HOL help, type: help "hol";
       To exit type <Control>-D
---------------------------------------------------------------------
[Use-ing configuration file /Users/binghe/.hol-config.sml]
/Users/binghe/ML/HOL/tools-poly/prelude2.ML:4: error: Structure (bossLib) has not been declared Found near open bossLib
Uncaught exception: Fail "Static Errors"
```

The same issue doesn't happen in any other "$(HOLDIR)/src/" subfolder that I randomly tried.  Can you reproduce it?

Regards,

Chun Tian

barakeel:
Yes got the same error.
---------------------------------------------------------------------
       HOL-4 [Kananaskis 14 (stdknl, built Mon Jul 26 16:44:29 2021)]

       For introductory HOL help, type: help "hol";
       To exit type <Control>-D
---------------------------------------------------------------------
/home/thibault/HOL/tools-poly/prelude2.ML:4: error: Structure (bossLib) has not been declared Found near open bossLib
Uncaught exception: Fail "Static Errors"

mn200:
This is a known problem with the way the `Holmakefile` in that directory interacts with starting up a session.  You will get similar problems in directories that refer to the internal `numheap` (roughly, between building arithmetic and boss).

binghe:
> This is a known problem with the way the `Holmakefile` in that directory interacts with starting up a session. You will get similar problems in directories that refer to the internal `numheap` (roughly, between building arithmetic and boss).

But sometimes I need to debug "src/boss/prove_base_assumsScript.sml", which needs to read two OpenTheory articles in current directory. If I don't start "hol" session in "src/boss" there seems no way to replay the scripts without manually changing the path of OT articles.   To workaround of this present issue (in another ongoing PR) I'm moving OT-files outside of "src/boss" to somewhere else.

mn200:
[...]

PR #915, opened 2021-05-16
Add some theorems about PROD_IMAGE

xrchz:

PR #942, opened 2021-08-18
Fixed OpenTheory building of hol-base (new version: 1.3)

binghe:
Hi,

1. currently the `--otknl` build is broken at "src/boss", when building "$(HOLDIR)/src/boss/hol4-base-unint.art", saying **different constants named "HOL4.divides.prime** (see [1] for more details). This is not the only issue, however it's directly caused by PR #915, where `pred_setTheory` started to depend on `dividesTheory` while in "$(HOLDIR)/src/boss/hol4-base-unint.thy" this dependency wasn't updated correspondingly.

2. Once the above issue is fixed, the building process will again stop at the same place saying **different constants named "HOL4.marker.using**, and this is caused by some missing imports of `markerTheory` in "$(HOLDIR)/src/boss/hol4-base-unint.thy" due to other historical updates of core theories. The way I fixed it is to try to add `marker` imports into every components and let the `opentheory` tool to tell me which ones are redundant (and then revert the changes).

3. Once the above issue is fixed, the building process will stop when building "$(HOLDIR)/src/boss/prove_base_assumsScript.sml": starting from theorem `th85` the existing proof scripts do not work any more.  This is actually because, due to changes of "$(HOLDIR)/src/boss/hol4-base-unint.thy", the order of HOL4 assumption theorems (to be proved in this file using OpenTheory basic theorems) has changed, also the number (there's one more assumption).   To fix this issue, I had to first check out a working old --otknl build and then add all statements of each assumption (from `th1` to `th105`) into "prove_base_assumsScript.sml" as comments, and then once any proof of `th*` theorem is broken now we can find its proof at another place in this file.  Finally, I found the following new assumption without a proof, so I made one:
```
val exists_simp = hd(amatch “(?x. t) <=> t”);

(* |- (?!x. F) <=> F *)
val th105 = store_thm("th105", el 105 goals |> concl,
  PURE_REWRITE_TAC [BETA_RULE (SPEC “\x:'a. F” ex_unique_thm)] \\
  PURE_REWRITE_TAC [SPEC “F” exists_simp, F_and] \\
  REFL_TAC);
```

4. However, to debug "prove_base_assumsScript.sml", which reads two OT articles ("base-theorems.art" and "hol4-assums.art") in the same directory, I couldn't start "hol" session in "$(HOLDIR)/src/boss" because of a known "won't fix" issue (see [2] for details).  After some thoughts, I decide to move all OT-related files out, from "$(HOLDIR)/src/boss" to a new directory at the same level "$(HOLDIR)/src/base" (named after the OT targets "hol4-base.thy" or `hol-base-1.x`). This movement is not necessary but makes things easier. (**This movement is now cancelled**.)

5. Once the above issue is fixed, again, the building process will stop in "src/probability" saying something like **different constants named mono_increasing**.  I try to understand why `mono_increasing` is special and it turns out that its definition was in `seqTheory` but is then redefined in `real_lebesgueTheory` with exactly the same statements. This is not any meaningful overloading but simply a duplication (or mistake). After cleaning up the duplication the OT building issue gets fixed.  However, there are other dependency mistakes in "$(HOLDIR)/hol4-analysis-unint.thy" et al, due to the moving of `iterateTheory` to "src/real". They are fixed too.

Eventually, after all above fixes, now the `--otknl` building works again, and I let it auto-install `hol-base` (new version: 1.3, as the current version is 1.2 [3]) into local opentheory repository (`opentheory list` command will show `hol-base-1.3`).  There are still some issues:

- In `markerTheory` some constants are not supposed to be exported into OpenTheory level, but currently the related code doing this work seems do not work (I can still see all these constants from the HTML documentation of `hol-base-1.3`.)

- other OT packages beyond `hol-base` are still need to investigate and fix, for them to auto-install after the OT articles building. Some recent core theory additions (like `cardinalTheory`) are not touched yet.

Regards,

Chun Tian

[1] https://github.com/gilith/opentheory/issues/9
[2] https://github.com/HOL-Theorem-Prover/HOL/issues/941
[3] http://opentheory.gilith.com/?pkg=hol-base-1.2


mn200:
Thanks for all this great work!  My only kvetch is that I am not very keen on the use of `src/base`; can you instead use (a subdirectory of) the existing `src/opentheory` directory please?

binghe:
> Thanks for all this great work! My only kvetch is that I am not very keen on the use of `src/base`; can you instead use (a subdirectory of) the existing `src/opentheory` directory please?

You're welcome. My long-term goal is to make my probability work available (through OpenTheory) to other theorem provers.

Regarding "src/base", I have moved all OT-files back to "src/boss" as now I know how to use `hol.bare` to replay "prove_base_assumsScript.sml".  (Note that a directory holding those files must be at the same level with "src/boss", because otherwise we would need to modify all relative file names in some `.thy` files, e.g. when it refers to files like "../marker/marker.ot.art" or "../opentheory/hol4.int".)

binghe:
I forgot to mention that `OpenTheoryReader` is now available in all kernels, since this "reading" functionality doesn't involve anything from the logging kernel.  (However, I haven't started to investigate how a public OpenTheory package can be loaded into HOL4 with all its definitions and theorems treated the same level as native HOL theorems.)

mn200:
Thanks again!

PR #944, opened 2021-08-23
Add some finite map constants to Description manual

hrutvik:
Include definitions and a few lines on `FDIFF`, `FMERGE_WITH_KEY`, and `FMERGE`.
Thanks to @binghe for the suggestion in the comments of HOL-Theorem-Prover/HOL#939.


mn200:
Nice; thanks for this!

PR #946, opened 2021-08-24
More algebra and polynomials

xrchz:


src/pred_set/src/pred_setScript.sml
```diff
@@ -4951,6 +4951,14 @@ val SUM_IMAGE_PERMUTES = store_thm(
   `SIGMA (f o g) s = SIGMA f (IMAGE g s)` by SRW_TAC[][SUM_IMAGE_INJ_o] THEN
   SRW_TAC[][]);
 
+Theorem SUM_IMAGE_ADD:
```
mn200, line 4954: Seems unused elsewhere so it's not clear that this is the right orientation for the equality.

src/sort/sortingScript.sml
```diff
@@ -1445,6 +1445,24 @@ Proof
   \\ PROVE_TAC[]
 QED
 
+Theorem LIST_REL_PERM:
```
mn200, line 1448: - Depends on generated name `M`
- seems to be expressed to suit `ho_match_mp_tac`

Suggest:
```
Theorem LIST_REL_PERM:
  !l1 l2 l3. LIST_REL P l1 l2 /\ PERM l2 l3 ==>
             ?l4. PERM l1 l4 /\ LIST_REL P l4 l3
Proof
  Induct_on ‘LIST_REL’
  \\ rw[PERM_CONS_EQ_APPEND, PULL_EXISTS]
  \\ ntac 2 (irule_at Any listTheory.LIST_REL_APPEND_suff)
  \\ simp[]
  \\ first_x_assum $ drule_then strip_assume_tac
  \\ gs[LIST_REL_SPLIT2, SF SFY_ss]
QED
```

examples/algebra/ring/ringDividesScript.sml
```diff
@@ -29,7 +29,10 @@ open ringIdealTheory;
 open ringUnitTheory;
 open ringTheory;
 open groupTheory;
-open monoidTheory;
+open monoidTheory gbagTheory bagTheory;
+val MEMBER_NOT_EMPTY = pred_setTheory.MEMBER_NOT_EMPTY;
```
mn200, line 33: There's an `open pred_setTheory` earlier in this file; do you need this rebind of `MEMBER_NOT_EMPTY` because the name is masked by some other `open`?  If so, perhaps we should avoid the name clash in the relevant theories.

mn200 (commented):
Minor inline comments

examples/algebra/ring/ringDividesScript.sml
xrchz: It's masked by `bagTheory.MEMBER_NOT_EMPTY`. Yes perhaps we should avoid the name clash but could be done separately from this PR (and not by me :))?

mn200: Totally.  I can change the one in `bagTheory`.

src/sort/sortingScript.sml
xrchz: Very nice! Done.

src/pred_set/src/pred_setScript.sml
xrchz: I don't mind either way - `EXTREAL_SUM_IMAGE_ADD` and `REAL_SUM_IMAGE_ADD` have it the other way around so I will change it here to match.

mn200:
Thanks for this!

PR #947, opened 2021-08-30
Minor completion of Lebesgue integration (+integral_neginf, etc.)

binghe:
Hi,

this PR adds a few more theorems to `lebesgueTheory` with the purpose of extending the parameter of `probabilityTheory.expectation_const` from normal reals to all extended reals:
```
   [expectation_const]  Theorem (old)      
      ⊢ ∀p c. prob_space p ⇒ expectation p (λx. Normal c) = Normal c

   [expectation_const]  Theorem (new)
      ⊢ ∀p c. prob_space p ⇒ expectation p (λx. c) = c
```

Besides, I did some cleanup to `topologyTheory` (a local lemma is removed), and `real_topologyTheory`, in which the "lift_disj_eq", "lift_imp_disj"-related compatibility fixes are narrowed down to only 5 affected theorems, rendering `std_ss` unchanged.

In addition, after 9fe1ff9f2635920e0b708b9a2c0949a6ea559c27, the changed definition `topologyTheory.limpt` now can be used to re-define `real_topology.limit_point_of` in the following way:
```
   [limit_point_of_def]  Definition      
      ⊢ ∀x s. x limit_point_of s ⇔ limpt euclidean x s
```
while the previous definition now becomes a theorem. (It remains to convert more definitions in `real_topologyTheory` using those from general topology.)

--Chun

### DIFF
--- a/src/probability/extrealScript.sml
+++ b/src/probability/extrealScript.sml
@@ -5433,6 +5433,23 @@ Proof
     REWRITE_TAC [IN_APP, lt_inf_epsilon]
 QED
 
+Theorem inf_num :
+    inf (\x. ?n :num. x = -&n) = NegInf
+Proof
+    rw [GSYM le_infty, inf_le]
+ >> CCONTR_TAC
+ >> fs [GSYM extreal_lt_def, GSYM lt_infty]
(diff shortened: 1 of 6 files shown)


mn200:
Thanks for your work on this!

PR #948, opened 2021-09-29
Fix Theorem-Proof goal selection in Vim mode

oskarabrahamsson:
This is intended to fix an issue when goal terms containing the text `Proof` were not properly selected by `<LocalLeader>a`. In the following example, only the first two lines would be selected:
```
Theorem foo:
  foo /\
  barProof$foo ==>
    baz
Proof
...
```


xrchz (approved):

mn200:
Thanks Oskar!

PR #950, opened 2021-10-06
Size combinator eqs

talsewell:
These changes:
  - update pair and option size functions to match DataSize.define_size
  - prove & store a normalisation theorem `_size_eq` for every type with a size definition `_size_def`
  - slightly simplify a proof I was fixing when I tried a more aggressive version which actually changed `_size_def`


talsewell:
I don't understand that build error, it looks like something has gone wrong in communicating with minisat, and I can't see how this change can be relevant.

mn200:
I agree that you're fine wrt the regression error.  I may move that check to "may fail" category as I've seen it before.  (It'd be nice to fix it too of course.)

Thanks for your work on this!

PR #955, opened 2021-10-22
Fix sum size (again)

talsewell:
This time with an additional patch to the way TotalDefn.guessR uses type_size,
giving it access to both the "old" and "new" behaviour. This fixes the MLsyntax
problem. I haven't been able to run the full tests here, maybe the regression
server can check it.


mn200:

PR #964, opened 2021-11-05
Add deprecate_real and prefer_real functions to realLib

HeikoBecker:
This adds functions similar to those in `intLib.sml` and `numLib.sml` to make the parser give precedence to reals if needed.

### DIFF
--- a/src/real/realLib.sml
+++ b/src/real/realLib.sml
@@ -15,6 +15,37 @@ struct
 
   local open transcTheory in end;
 
-  open RealArith realSimps Diff isqrtLib;
+  open HolKernel RealArith realSimps Diff isqrtLib;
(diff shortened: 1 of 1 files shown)


mn200:
Thanks for your work on this!

binghe:
I think I need to copy this code also into `ratLib` and adapt it for rational operators, because currently simply opening `ratTheory` in `util_probScript.sml` will break some proofs about real numbers.

binghe:
Hi, I believe the additional functions are not accessible unless we also put their declarations into `realLib.sig`:

```
   (* syntax *)
   val prefer_real     : unit -> unit
   val deprecate_real  : unit -> unit
```

Chun

PR #965, opened 2021-11-07
Fix incremental builds by moving jrhUtils from src/integer to src/res_quan/src

binghe:
Hi,

currently some libraries in "src/real" (e.g. `RealArith`) depends on "src/integer/jrhUtils" but this dependency is not reflected in `INCLUDES` of "src/real/Holmakefile".  As a result, if I touch some core theories (e.g. `fcpScript.sml`) and rebuild HOL4, real scripts will complain that "`$(HOLDIR)/sigobj/jrhUtils.ui` cannot be loaded."

Given that `realTheory`, etc. do not use `integerTheory` at all, simply moving `jrhUtils` to earlier folders like "src/res_quan/src" (together with the newly moved `hurdUtils`) could resolve all related issues.

Furthermore, currently the probability heap doesn't contain any thing useful, this is due to a small mistake in related `Holmakefile`, now fixed.    The `EXTRA_CLEAN` of `examples/probability/Holmakefile` also doesn't work. Fixed too.

Regards,

Chun Tian


mn200:
There is certainly a problem here, and moving `jrhUtils` is independently a good idea, but I'm not sure about your diagnosis of the real problem. The `real` Holmakefile does include `INCLUDES = ../integer` so I'm not sure what you mean by the claim that the dependency is not reflected there.  Moreover, that include is necessary because of the `real_of_int` constant (in `intrealTheory`).

mn200:
If you touch `fcpScript.sml`, the build fails in `src/probability` when the system attempts to rebuild `util_prob`. The failure arises because of the attempt to find `sigobj/jrhUtils.uo` as you have observed. 

This happens because the dependency analysis of util_prob knows that it depends on both `RealArith` and `fcp` (the latter changing causes the rebuild). The corresponding `.uo` file for `RealArith` mentions `sigobj/jrhUtils` (as well as `src/integer/jrhUtils` (!!)).  `RealArith` doesn’t need to be recompiled or rebuilt because it does not depend on anything that has changed, but when the system tries to load it as a preliminary to executing `util_probScript.sml`, it can’t find `jrhUtils` in `sigobj` because the `link-to-sigobj` target hasn’t executed in `src/integer` yet (non-deterministically).

Ultimately, the problem is the automatic dependency analysis that led to `RealArith.uo` containing `sigobj/jrhUtils`. *That*’s what needs to change.  Your fix isn’t guaranteed to solve the problem because `RealArith` may well end up with a dependency on `src/res_quan/src/jrhUtils`, and there’s no guarantee that `src/res_quan/src`’s `link-to-sigobj` target will have fired when `util_probScript.sml` comes to run.

mn200:
Thanks for bringing this to my attention!

binghe:
Thanks for the detailed analysis and the actual solution. (Somehow I failed to see `../integer` as the first element of `INCLUDES` in `src/real/Holmakefile` and I was even trying to add `../integer` again…)

PR #967, opened 2021-11-08
Add new realLib functions into realLib.sig; fixed overloading of "~"

binghe:
Hi,

this small PR fixes some issues left by #964

1. The two functions `prefer_real` and `deprecate_real` should have their signatures in `realLib.sig`, otherwise there should be no way to actually call them from outside.

2. The list of arithmetic operator symbols are not enough. There should be also "~" which behaves the same as "-" (negation). `numLib` doesn't need this because `~(0 : num)` is a illegal term but `~(0 :real)` is legal (= `-(0 :real)`).    Similarily, the related code in `intLib` has also the same issue. Now it's fixed.    (I found this issue by trying to open `ratTheory` in probability scripts and deprecate `rat` and `int`.)

3. I copied the code in `realLib` and added two similar functions into `ratLib`.  They worked (but the related code changes are not in this PR.)    (All other code changes in `ratLib.sml` are just some reformating, making the code reading easier.)

Regards,

Chun Tian

### DIFF
--- a/src/integer/intLib.sml
+++ b/src/integer/intLib.sml
@@ -5,6 +5,7 @@ struct
 
   val operators = [("+", intSyntax.plus_tm),
                    ("-", intSyntax.minus_tm),
+                   ("~", intSyntax.negate_tm),
                    ("numeric_negate", intSyntax.negate_tm),
                    ("*", intSyntax.mult_tm),
                    ("/", intSyntax.div_tm),
--- a/src/rational/ratLib.sig
+++ b/src/rational/ratLib.sig
@@ -44,7 +44,7 @@ sig
         val rat_rewrites : thm list
         val rat_num_rewrites : thm list
 
-	(* syntax *)
+        (* syntax *)
         val prefer_rat     : unit -> unit
         val deprecate_rat  : unit -> unit
 end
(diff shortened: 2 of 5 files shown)


mn200:
Thanks for this.  Ideally, it would be nice to have regression tests for all these.

binghe:
> Thanks for this. Ideally, it would be nice to have regression tests for all these.

sorry I have no idea on how to write these regression tests...  For example, if `deprecate_real()` is called, then in theory terms like ``0 < x`` shouldn't be ``(0 :real) < (x :real)`` but something else, however there seems no way to know that "else" a priori.

mn200:
Your selftest.sml file will set up a particular context, so your test will know what should happen then.  The "standard" real context only has natural numbers and reals, so the test could check that the parsed term is indeed of the natural type.  I'll attempt to demonstrate with a forthcoming commit.

mn200:
The tests might also check that bouncing backwards and forwards does reasonable things (it resets the grammars between the two tests at the moment).  But it's certainly valuable: the test I wrote revealed a bug.

binghe:
I have added similar tests for `{prefer,deprecate}_{rat|int}` as part of #968.

PR #968, opened 2021-11-09
Enhanced lemmas for `real$sup` (supremum), etc.

binghe:
Hi,

there are 3 independent parts of this PR:

1. In `realTheory`, the theorem `REAL_IMP_LE_SUP` has a unnecessary antecedent `?r. p r` which is subsumed by another antecedent `?r. p r /\ x <= r`, thus can be eliminated:
```
   [REAL_IMP_LE_SUP]  Theorem (new version)      
      ⊢ ∀p x. (∃z. ∀r. p r ⇒ r ≤ z) ∧ (∃r. p r ∧ x ≤ r) ⇒ x ≤ sup p
```
This theorem is used several in probability scripts. After the changes the related proofs are slightly shorter.   In addition, more lemmas about `sup` are also added (one in `realTheory`, others in `iterateTheory`).

2. In `borelTheory`, the theorem `sup_sequence` is actually the same thing with `extrealTheory.sup_seq`, except for different definitions of real sequence convergence (`f --> l`). (The one from `real_topologyTheory` is more general.)
```
   [sup_sequence (old name)]  Theorem (borelTheory)     
      ⊢ ∀f l.
          mono_increasing f ⇒
          ((f --> l) sequentially ⇔
           sup (IMAGE (λn. Normal (f n)) 𝕌(:num)) = Normal l)

   [sup_seq]  Theorem (extrealTheory)     
      ⊢ ∀f l.
          mono_increasing f ⇒
          (f ⟶ l ⇔ sup (IMAGE (λn. Normal (f n)) 𝕌(:num)) = Normal l)
```
By proving the following equivalence theorem between the two notions (in `real_topologyTheory`), the long duplicated proof of `sup_sequence` can be eliminated (then the name can be changed to `sup_seq'` indicating it's now a simple corollary of `sup_seq`):
```
   [LIM_SEQUENTIALLY_SEQ]  Theorem (real_topologyTheory)      
      ⊢ ∀s l. (s ⟶ l) sequentially ⇔ seq$--> s l
```

3. Following discussion in previous PR #967, now I have added some regression tests for `prefer` and `deprecate` of rations and integers, following @mn200 's sample code for reals in `src/real/selftests.sml`.   I changed/fixed two term definitions `ratSyntax.sml` (for rationals 0 and 1), to make the related tests pass.

Regards,

Chun Tian

### DIFF
--- a/src/rational/ratSyntax.sml
+++ b/src/rational/ratSyntax.sml
@@ -1,9 +1,17 @@
 structure ratSyntax :> ratSyntax =
 struct
 
-open HolKernel boolLib Parse ratTheory;
+open HolKernel boolLib ratTheory;
 
 val ERR = mk_HOL_ERR "ratSyntax";
+
+(* Fix the grammar used by this file (when it's really needed)
+structure Parse = struct
+  open Parse
+  val (Type,Term) = parse_from_grammars ratTheory.rat_grammars
+end
+ *)
+open Parse;
 
 (*--------------------------------------------------------------------------*
  * constants
@@ -11,13 +19,6 @@ val ERR = mk_HOL_ERR "ratSyntax";
 
 (*val int_ty = intSyntax.int_ty;*)
 val rat_ty = mk_thy_type{Tyop = "rat", Thy="rat", Args = []};
-
-(* old definitions:
-val rat_0_tm = prim_mk_const {Name="rat_0",Thy="rat"};
-val rat_1_tm = prim_mk_const {Name="rat_1",Thy="rat"};
- *)
-val rat_0_tm = “0q”;
-val rat_1_tm = “1q”;
 
 val rat_nmr_tm = prim_mk_const {Name="rat_nmr",Thy="rat"};
 val rat_dnm_tm = prim_mk_const {Name="rat_dnm",Thy="rat"};
@@ -36,6 +37,13 @@ val rat_gre_tm = prim_mk_const {Name="rat_gre",Thy="rat"};
 val rat_leq_tm = prim_mk_const {Name="rat_leq",Thy="rat"};
 val rat_geq_tm = prim_mk_const {Name="rat_geq",Thy="rat"};
 val rat_of_num_tm = prim_mk_const {Name = "rat_of_num", Thy = "rat"}
+
+(* old definitions:
+val rat_0_tm = prim_mk_const {Name="rat_0",Thy="rat"};
+val rat_1_tm = prim_mk_const {Name="rat_1",Thy="rat"};
+ *)
+val rat_0_tm = mk_comb(rat_of_num_tm, numSyntax.term_of_int 0);
+val rat_1_tm = mk_comb(rat_of_num_tm, numSyntax.term_of_int 1);
 
 (*--------------------------------------------------------------------------*
  * constructors
--- a/src/rational/ratSyntax.sml
+++ b/src/rational/ratSyntax.sml
@@ -1,14 +1,7 @@
 structure ratSyntax :> ratSyntax =
 struct
 
-open HolKernel boolLib Parse;
-
-(* interactive mode
-app load ["ratTheory"];
-*)
-
-open ratTheory;
-
+open HolKernel boolLib Parse ratTheory;
 
 val ERR = mk_HOL_ERR "ratSyntax";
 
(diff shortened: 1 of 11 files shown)


src/rational/ratSyntax.sml
```diff
@@ -17,10 +10,14 @@ val ERR = mk_HOL_ERR "ratSyntax";
...
 
 (*val int_ty = intSyntax.int_ty;*)
-val rat = mk_thy_type{Tyop = "rat", Thy="rat", Args = []};
+val rat_ty = mk_thy_type{Tyop = "rat", Thy="rat", Args = []};
 
+(* old definitions:
 val rat_0_tm = prim_mk_const {Name="rat_0",Thy="rat"};
 val rat_1_tm = prim_mk_const {Name="rat_1",Thy="rat"};
+ *)
+val rat_0_tm = “0q”;
```
mn200, line 19: You shouldn't call the parser in library files without having ensured the parser’s behaviour at the head of the file.  For the sort of thing you need to do, see [this example](https://github.com/HOL-Theorem-Prover/HOL/blob/1d5cce12600fa6c2b79f88f35406216e78ef0d2b/src/taut/tautLib.sml#L21-L26). Otherwise, you run the risk that this code will be executed in a context where the global grammars make a mess of your attempt to parse any given string.

mn200:
I wonder what your opinion is on the use of `IN` *vs* just predicate application in this material. There seem to be both styles present now.  My instinct is to prefer `IN` in most places where a normal person would view the “predicate” as a set.  Certainly, I think `sup P` implies that `P` is a set, so that we should have `x IN P` rather than just `P x`.  

Of course, backwards compatibility means that I wouldn't delete the old theorems, but perhaps there's a naming convention we could adopt to indicate in which style a given theorem is.  (If the `IN` version was proved first, the “pred” version is only a `SIMP_RULE std_ss [IN_DEF]` away.)

binghe:
Thanks for the explanations. Now I have redefined `rat_0_tm` and `rat_1_tm` using purely functional interface:
```
val rat_0_tm = mk_comb(rat_of_num_tm, numSyntax.term_of_int 0);
val rat_1_tm = mk_comb(rat_of_num_tm, numSyntax.term_of_int 1);
```
meanwhile you suggested code has been also put into the same file but inside comments, for potential future uses.

I think the choice of defining `sup` and `inf` by predicate application instead of `IN` was due to historical reasons - perhaps at the time when the original `realTheory` was created (by John Harrison) the predicate-based set theory `pred_setTheory` was not there yet, and this is perhaps when `realTheory` does not open `pred_setTheory` at all.

I totally agree that the `IN` versions of these theorems are more natural in understanding and also in practical applications.  Furthermore, that fact that the `IN` versions can be easily converted to "predicate application" versions by simple rewrite using `IN_DEF` (or `IN_APP`), indicates only the `IN` version should be in the library if we were to keep only one of them.

The original authors of `extrealTheory` seems have "blindly" learnt this bad historical convention and also defined and proved all `sup` and `inf` theorems in predicate application versions. As a result, they constantly need to convert their goals back to `IN` versions after the application of their `sup` or `inf` theorems.  (To show these "funny" proof code, one can simply run `grep ONCE_REWRITE_TAC.*GSYM SPECIFICATION *Script.sml` in `$(HOLDIR)/src/probability`.)

When I started to extend `extrealTheory` with new theorems, the *naming convention* I chose is to add a sibling "primed" version of each existing sup/inf theorems as the `IN` versions. For example, check the statements of the following theorems in `extrealTheory`:
```
    val inf_eq : thm
    val inf_eq' : thm
    val inf_le : thm
    val inf_le' : thm
    val inf_le_imp : thm
    val inf_le_imp' : thm
    val inf_lt : thm
    val inf_lt' : thm
    val le_sup : thm
    val le_sup' : thm
    val le_sup_imp : thm
    val le_sup_imp' : thm
    val sup_eq : thm
    val sup_eq' : thm
    val sup_le : thm
    val sup_le' : thm
    val sup_le_sup_imp : thm
    val sup_le_sup_imp' : thm
    val sup_lt : thm
    val sup_lt' : thm
    val sup_lt_epsilon : thm
    val sup_lt_epsilon' : thm
    val lt_inf_epsilon : thm
    val lt_inf_epsilon' : thm
```
The `IN` versions are much easier to use, because when dealing with the `sup` of a set, usually the set is given in form of `IMAGE f UNIV`, and latter theorems like `IN_IMAGE` must be used to reduce `x IN IMAGE f UNIV` into further forms.

Back to `realTheory`, below are my current choices:

1. I don't put any `IN` theorems into `realTheory`, also I don't open `pred_setTheory` in `realTheory`.  (The fact that `IN_DEF` is now in `boolTheory` is a little weired, I try to now use it if `pred_setTheory` is not opened.)

2. And if a new sibling theorem about `sup`/`inf` is to be added in `realTheory`, like the case of `REAL_IMP_LT_SUP`, I take the predicate application form to match other related theorems.

3. For the ease of applications, when I add some "primed" (i.e. they only differ by having one `'` at the end of existing theorem names) versions of `sup`/`inf` (for real sets) theorems, I only add them into secondary real theories, i.e. `iterateTheory`, which contains alternative definitions of `sup` and `inf` and some related theorems.

4. New `sup`/`inf` theorems (e.g. `SUP_MONO`) who do not share any name parts with existing "historical" theorems, will only be provided in `IN` versions, in secondary real theorie(s).   They can be easily converted to the other form by end users (e.g. by `REWRITE_RULE [IN_DEF]`).

Chun

mn200:
Thanks for the detailed discussion of your methods/philosophy! I don't think there'd be a problem with putting "primed"/IN versions of theorems into `realTheory`; there's nothing particularly sacrosanct that must be preserved there.  Having theories grow is fine.

Thanks also for all the other work in this PR.

Issue #963, opened 2021-11-02
Weird «HOL Description» sub-section names


binghe:
Hi,

In «The HOL System Description», the section "DERIVATION OF THE STANDARD RULES", there are two weird (identical) sub-section names: (around page 61-62)

- 2.3.21 Use of a definition (the content is `RIGHT_BETA`)
- 2.3.22 Use of a definition (the content is `RIGHT_LIST_BETA`)

I hardly can connect these drules with "use of a definition". What are the correct subsection names supposed to be?  I think this is an old issue, because they were like now even in this manual for Kananaskis-7 that I found on Internet.

Also, I occasionally found this issue when searching "definition" in the manual, aiming to find out what's the corresponding "modern" syntax for `zDefine` (something like `Definition[...]` "without computation"?). Can we document this somewhere in the manual?

Thanks.

Chun Tian

konrad-slind:
Re: making definitions. The new syntactic forms for declaring
definitions, theorems, overloading, etc are really nice, but
their documentation is scattered over the Description. Pointers
are given in the Index, under

  special syntactic forms for scripts

However, I think what you are looking for is the
nocompute "pseudo-attribute", and I eventually found it
documented in the computeLib part of the manual. The example
given is

  Definition f_def[nocompute]:
    f x = x + 10
  End

Note there is no space allowed between the definition name and
the pseudo-attribute list, and the "Definition" and "End"
keywords need to be in the first column.

Konrad.

mn200:
I'll look at the manual soon.  As for the syntax; Konrad is right. 

(It does seem unnecessarily fussy to require the square brackets to be up against the identifier; I'll change that to allow intervening whitespace.)

binghe:
Thanks for your answer and manual changes!

PR #970, opened 2021-11-12
HOL Manual: a section for the overall syntax of Definition

binghe:
Hi,

This is a small updates to «HOL Description».   The background (see #963 for more details) is that I was unable to find the corresponding "modern" syntax of `zDefine` by searching in that manual.    When the new syntax is introduced, old ML functions are all removed from the manual, but these functions are still in many core theory or other old user code, causing potential difficulties.

Furthermore, in «HOL Description», the introduction of the `Definition` facility and its syntax is current under the section "Recursive Functions", but more often a `Definition` is just used to define abbreviations.   So I think maybe it's better to move some contents to the beginning of chapter 6 (Advanced Definition Principles) as a new initial section, putting also some related contents from K13 release notes into it, to give a overall description of `Definition` and its possible attributes.

In the updated manual, I try to briefly mention all the `Define`-siblings (`xDefine`, `zDefine` and `tDefine`) so that by searching the manual code readers may find a link from the old API to the new syntax sugar. Cross-references are added as much as possible.

Also, the PDF bookmarks in «HOL Reference» should have the letter Z (for `zDefine`).

Regards,

Chun Tian

### DIFF
--- a/Manual/Description/definitions.stex
+++ b/Manual/Description/definitions.stex
@@ -1,5 +1,63 @@
 \chapter{Advanced Definition Principles}\label{HOLdefinitions}
 
+\section{General-Purpose Definition}
+\label{sec:definition}
+
+\ml{Definition} takes a high-level specification of a \HOL{} function, and attempts to define the
+function in the logic. If this attempt is successful, the specification is derived from the
+definition. The derived specification is returned to the user, and also stored in the current
+theory. \ml{Definition} may be used to define \emph{abbreviations}, \emph{recursive functions},
+and \emph{mutually recursive functions}.
+
+\index{Definition@\ml{Definition}}
+The basic form of a \ml{Definition} is
+\begin{alltt}
+  Definition defname[\textit{attrs}]:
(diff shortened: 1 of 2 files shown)


mn200:
Nice.  Lifting the details up and out to head the section makes perfect sense. 

In the service of absolutely accuracy, you might say that the new syntax only maps to `tDefine` or `xDefine`; it never goes to `Define`.  Indeed, it does all of this *via* an entirely new entrypoint, `TotalDefn.qDefine`. (It's not clear to me that documenting the latter really adds much value.)

binghe:
> Nice. Lifting the details up and out to head the section makes perfect sense.
> 
> In the service of absolutely accuracy, you might say that the new syntax only maps to `tDefine` or `xDefine`; it never goes to `Define`. Indeed, it does all of this _via_ an entirely new entrypoint, `TotalDefn.qDefine`. (It's not clear to me that documenting the latter really adds much value.)

Please make further revision by yourself. I don’t know such deep details…

PR #971, opened 2021-11-15
Some convexity-derived inequalities including AGM and Young's inequality (ported from HOL Light)

binghe:
Hi,

This (clean) PR first ports some convexity-derived inequalities including AGM ("arithmetic and geometric means") and Young's inequality from HOL Light ("Multivariate/realanalysis.ml", line 7757-7869), and then promotes `YOUNG_INEQUALITY` to `extrealTheory` (for my later uses):
```
   [AGM]  Theorem      (real_sigmaTheory)
      ⊢ ∀s x n.
          s HAS_SIZE n ∧ n ≠ 0 ∧ (∀i. i ∈ s ⇒ 0 ≤ x i) ⇒
          product s x ≤ (sum s x / &n) pow n
   
   [AGM_2]  Theorem
      
      ⊢ ∀x y u v.
          0 < x ∧ 0 < y ∧ 0 ≤ u ∧ 0 ≤ v ∧ u + v = 1 ⇒
          x rpow u * y rpow v ≤ u * x + v * y
   
   [AGM_GEN]  Theorem      
      ⊢ ∀a x s.
          FINITE s ∧ sum s a = 1 ∧ (∀i. i ∈ s ⇒ 0 ≤ a i ∧ 0 < x i) ⇒
          product s (λi. x i rpow a i) ≤ sum s (λi. a i * x i)
   
   [AGM_ROOT]  Theorem      
      ⊢ ∀s x n.
          s HAS_SIZE n ∧ n ≠ 0 ∧ (∀i. i ∈ s ⇒ 0 ≤ x i) ⇒
          root n (product s x) ≤ sum s x / &n
   
   [AGM_RPOW]  Theorem      
      ⊢ ∀s x n.
          s HAS_SIZE n ∧ n ≠ 0 ∧ (∀i. i ∈ s ⇒ 0 < x i) ⇒
          product s (λi. x i rpow (1 / &n)) ≤ sum s (λi. x i / &n)
   
   [AGM_SQRT]  Theorem      
      ⊢ ∀x y. 0 ≤ x ∧ 0 ≤ y ⇒ sqrt (x * y) ≤ (x + y) / 2

   [YOUNG_INEQUALITY]  Theorem      
      ⊢ ∀a b p q.
          0 < a ∧ 0 < b ∧ 0 < p ∧ 0 < q ∧ p⁻¹ + q⁻¹ = 1 ⇒
          a * b ≤ a rpow p / p + b rpow q / q

[...]

PR #972, opened 2021-11-17
Hölder's Inequality and Cauchy-Schwarz Inequality for Lebesgue Integration

binghe:
Hi,

this PR adds some important inequalities for Lebesgue integration, namely Hölder's Inequality and Cauchy-Schwarz Inequality (at the end of `martingaleTheory`):
```
   [hoelder_inequality]  Theorem      
      ⊢ ∀m u v p q.
          measure_space m ∧ 0 < p ∧ 0 < q ∧ p⁻¹ + q⁻¹ = 1 ∧ p ≠ +∞ ∧
          q ≠ +∞ ∧ u ∈ function_space p m ∧ v ∈ function_space q m ⇒
          ∫ m (λx. abs (u x * v x)) ≤ norm p m u * norm q m v

   [cauchy_schwarz_inequality]  Theorem      
      ⊢ ∀m u v.
          measure_space m ∧ u ∈ L2_space m ∧ v ∈ L2_space m ⇒
          ∫ m (λx. abs (u x * v x)) ≤ norm 2 m u * norm 2 m v
   
   [cauchy_schwarz_inequality']  Theorem      
      ⊢ ∀m u v.
          measure_space m ∧ u ∈ L2_space m ∧ v ∈ L2_space m ⇒
          ∫⁺ m (λx. abs (u x * v x)) ≤
          sqrt (∫⁺ m (λx. (abs (u x))²) * ∫⁺ m (λx. (abs (v x))²))
```
Note that Cauchy-Schwarz Inequality is just a special case of Hölder's Inequality with `p = q = 2`, while the current statements (and proof) of Hölder's Inequality haven't covered the extreme cases (`p = PosInf` or `q = PosInf`, left as TODO in the future).  Cauchy-Schwarz Inequality can be used in probability theory, for example, to show that, if a sequence of random variables all have finite second moments (i.e. in L2 space), their summation has also finite second moments. The core proof requires Young's Inequality added in #971.

As the necessary preliminaries, the previously unused definition (`lebesgue$norm`) are refined and moved to `martingaleTheory`. The definition of `function_space` (and its two overloads `L1_space` and `L2_space`) is now defined:
```
   [function_space_def]  Definition      
      ⊢ ∀p m.
          function_space p m =
          {u | u ∈ Borel_measurable (m_space m,measurable_sets m) ∧
           if p = +∞ then
             ∃c. 0 < c ∧ measure m {x | x ∈ m_space m ∧ c ≤ abs (u x)} = 0
           else ∫⁺ m (λx. abs (u x) powr p) ≠ +∞}

   [norm_def]  Definition      
      ⊢ ∀p m u.
          norm p m u =
          if p = +∞ then
            inf {c | 0 < c ∧ measure m {x | x ∈ m_space m ∧ c ≤ abs (u x)} = 0}
          else ∫⁺ m (λx. abs (u x) powr p) powr p⁻¹
```

However, the most important contribution of this PR, in my option, is the following theorem stating that "any mono-increasing (extreal-valued) function is Borel measurable":
```
   [IN_MEASURABLE_BOREL_BOREL_MONO_INCREASING]  Theorem      
      ⊢ ∀f. (∀x y. x ≤ y ⇒ f x ≤ f y) ⇒ f ∈ Borel_measurable Borel
```
This important theorem is used in proving the Borel-measurability of extreal-valued power functions `powr` in the proof of Hölder's Inequality.  (In the future, when normal random variables are involved, this theorem can easily show that `exp` is also Borel measurable, simply because it's mono-increasing.):
```
   [IN_MEASURABLE_BOREL_BOREL_ABS_POWR]  Theorem      
      ⊢ ∀p. 0 < p ∧ p ≠ +∞ ⇒ (λx. abs x powr p) ∈ Borel_measurable Borel
```

Such a "sharp" result requiring so little antecedents (`∀x y. x ≤ y ⇒ f x ≤ f y`) is invaluable.  The proof of `IN_MEASURABLE_BOREL_BOREL_MONO_INCREASING` is not very long but requires a large number of case analysis with extensive uses of `sup` and `inf` lemmas. (Textbook only gives some picture for illustrations, never full proofs, thus I had to work out the full proof by myself.)  This is also one of the two powerful tools for proving Borel measurability. P. S. The other one is that "any continuous function" is Borel measurable, already done by HVG people for real-valued functions:
```
   [in_borel_measurable_continuous_on]  Theorem (real_borelTheory)
      ⊢ ∀f. f continuous_on 𝕌(:real) ⇒ f ∈ borel_measurable borel
```

In the beginning of `lebesgueTheory` and `real_lebesgueTheory`, some comments (on the purposes of definitions) are corrected or added, for math learners. Recent new additions in `extrealTheory` are slightly re-ordered.

In the second commit, many duplicated real-analysis theorems (ported from HOL-Light) were removed. I found them by analyzing HOL's generated theorem list (THEORY BINDINGS).  One such theorem (`DIFF_BIGINTER`), only internally used by `real_topologyTheory` has conflicts with another existing theorem in `pred_setTheory`. I renamed it to `DIFF_BIGINTER2`.

Regards,

Chun Tian


src/probability/extrealScript.sml
```diff
@@ -2890,14 +2892,28 @@ Proof
...
+
 (***************************)
 (*      Exp and powr       *)
 (***************************)
 
 Theorem exp_pos :
     !x :extreal. 0 <= exp x
 Proof
-    GEN_TAC >> Cases_on `x`
+    Q.X_GEN_TAC ‘x’ >> Cases_on `x`
```
mn200, line 2916: If the name already occurs in the original goal (as here), using `Q.X_GEN_TAC` is unnecessarily paranoid.  (Also note that the easier to type `qx_gen_tac` also exists.)

mn200 (approved):
Very nice!

src/probability/extrealScript.sml
binghe: Yes, you are right (this is necessary). It's totally my (bad) personal style that 1) use upper-name tactics to distinguish between "inner" (logics) and "outer" languages (ML code), 2) use `X_GEN_TAC`  instead of `GEN_TAC` (and `Induct_on` instead of `Induct`, etc.) to make the proof a bit more readable.  But this is totally meaningless in very simple proofs like the present case.

PR #974, opened 2021-11-19
Add CEILING_{DIV,MOD} to arithmeticTheory

myreen:


binghe:
> Good point! Will fix.

But I see `qspec_then` is already in `bossLib`:
```
val qspec_then : term quotation -> thm_tactic -> thm -> tactic
```
Is this what you can use directly?

Chun

binghe:
> > Good point! Will fix.
> 
> But I see `qspec_then` is already in `bossLib`:
> 
> ```
> val qspec_then : term quotation -> thm_tactic -> thm -> tactic
> ```
> 
> Is this what you can use directly?
> 
> Chun

Ah I see the problem: `bossLib` is not available at the stage when `arithmeticTheory` is building...  sorry for disturbing.

mn200:
Thanks!

PR #975, opened 2021-11-21
Update WF_REL_TAC to use size_eq, MEM_SPLIT etc

talsewell:
Have WF_REL_TAC be able to clean up termination conjuncts as
well as the whole goal, and also make use of size_eqs, and
also have an extra (configurable) set of rewrites used only
if they remove a conjunct, which by default includes
MEM_SPLIT and list_size_append.

This allows a number of custom termination proofs to be reduced to
the relevant `WF_REL_TAC`. There's only a little breakage, for some
reason mostly in lisp-related examples. I've checked most CakeML
definitions are not affected.

### DIFF
--- a/src/boss/bossLib.sml
+++ b/src/boss/bossLib.sml
@@ -222,6 +222,16 @@ fun name_ind_cases nm_tms thm = let
 end
 
 (* ----------------------------------------------------------------------
+    useful for proving termination in rose-tree settings
+   ---------------------------------------------------------------------- *)
+
+val size_comb_tac =
+  full_simp_tac boolSimps.bool_ss [listTheory.MEM_SPLIT]
+  THEN CONV_TAC TotalDefn.size_eq_conv
+  THEN simp_tac boolSimps.bool_ss
+    [listTheory.list_size_append, listTheory.list_size_def]
+
+(* ----------------------------------------------------------------------
     convenient simplification aliases
    ---------------------------------------------------------------------- *)
 
--- a/src/boss/bossLib.sml
+++ b/src/boss/bossLib.sml
@@ -222,7 +222,7 @@ fun name_ind_cases nm_tms thm = let
 end
 
 (* ----------------------------------------------------------------------
-    useful for proving termination in rose-tree settings
+    useful for proving termination in fold and rose-tree settings
    ---------------------------------------------------------------------- *)
 
 val size_comb_tac =
@@ -230,6 +230,10 @@ val size_comb_tac =
   THEN CONV_TAC TotalDefn.size_eq_conv
   THEN simp_tac boolSimps.bool_ss
     [listTheory.list_size_append, listTheory.list_size_def]
+
+val _ = let
+  val sref = TotalDefn.termination_solve_simps
+in sref := ([listTheory.MEM_SPLIT, listTheory.list_size_append] @ ! sref) end
 
 (* ----------------------------------------------------------------------
     convenient simplification aliases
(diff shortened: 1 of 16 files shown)


src/boss/bossLib.sml
```diff
@@ -221,6 +221,20 @@ fun name_ind_cases nm_tms thm = let
...
+    useful for proving termination in fold and rose-tree settings
+   ---------------------------------------------------------------------- *)
+
+val size_comb_tac =
+  full_simp_tac boolSimps.bool_ss [listTheory.MEM_SPLIT]
+  THEN CONV_TAC TotalDefn.size_eq_conv
+  THEN simp_tac boolSimps.bool_ss
+    [listTheory.list_size_append, listTheory.list_size_def]
+
+val _ = let
```
mn200, line 234: With `ThmSetData` in use, this addition should happen in `listScript.sml`

src/list/src/listScript.sml
```diff
@@ -1325,6 +1325,10 @@ Induct
      THEN REWRITE_TAC [MEM] THEN REPEAT STRIP_TAC
      THEN FIRST_ASSUM MATCH_MP_TAC THEN ASM_REWRITE_TAC[]]);
 
+val list_size_append = Q.store_thm("list_size_append",
```
mn200, line 1328: Please use “modern” `Theorem` syntax for this.

src/num/termination/TotalDefn.sml
```diff
@@ -161,6 +161,12 @@ val termination_simps =
           pairTheory.RPROD_DEF,
           SUB_LESS_I,DIV_LESS_I,MOD_LESS_I];
 
+(*---------------------------------------------------------------------------*)
+(* Extra rewrites, used only if they would solve a termination conjunct.     *)
+(*---------------------------------------------------------------------------*)
+
+val termination_solve_simps = ref ([] : thm list);
```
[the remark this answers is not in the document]
mn200: I appreciate that the old code uses a reference; maybe fix that too :)

mn200 (changes requested):

talsewell:
Thanks for looking Michael. Some if this isn't all that helpful. There's a reason I echo what's done in the surrounding code when I'm this far away from familiar contexts: it's just not clear to me which features are unavailable (we're before bossLib here), which are unwanted (e.g. Unicode), etc.

I'll update listScript right away. I now notice that if I'd scanned further in that file I'd have found plenty of other uses of the Theorem syntax. I guess that's available everywhere Q is (without [] attributes)?

I'll get around to reading about `ThmSetData` .. eventually.

mn200:
Don't worry about the `ThmSetData` stuff; I'll do it myself.  `Theorem` syntax and attributes is available after `Q` (it maps to `Q.store_thm`). `Definition` is available after `num/termination`.

talsewell:
Thanks for looking into that Michael, and sorry about the grumpy comment.

I'd assumed that the use of `thm ref` was because this theorem set was only ever changed temporarily - I see now that I was wrong about that, and it's a huge improvement that some of the theory-hacking is gone.

It does look like there's a bit of boilerplate to set up the theorem set though. I was thinking of mentioning that in Isabelle there's a thing called named_theorems, which is just a wrapper around the more generic theory data mechanism. It simplifies things in the most common case, in which the dynamic data is just a flat list of theorems. I wonder if something like that might help here?

mn200:
I made heavy weather of it because I wanted to additionally support removing theorems by name, and then I decided that the  underlying data structure should be a map from names to theorems.  But I agree that this sort of thing is likely to be a common case; it'd be worth providing such instances "for free" in `ThmSetData`.  Also an a-list form with names I suppose so that situations where the order is important I suppose.  Thanks for the suggestion.

PR #976, opened 2021-11-22
Move two common HOL-Light lemmas (LT and LT_LE) to arithmeticTheory

binghe:
Hi,

This small PR moves two common HOL-Light lemmas (`LT` and `LT_LE`) to arithmeticTheory, so that many copies of them can be eliminated from various core theories. This change may also ease future porting work of theories from HOL Light.

Regards,

Chun Tian


mn200:
Nice optimisation, thanks!

PR #978, opened 2021-12-04
Elementary proofs of some `sqrt` properties

binghe:
Hi,

in b85c65b194653e9c8ccc874df88cee073ff5f21c, I saw the definition of `sqrt` (square root of real numbers) has been moved to `realTheory`. I think this is a good action, as `sqrt` is not transcendental function, and many subsequent theories requiring it do not need other truly transcendental functions, which must be built upon differential and integral calculus, limits, metric and topology. (In HOL Light, real transcendental functions are even derived from their complex-valued versions.)

In this PR, I re-worked some `sqrt`-related theorems in `realTheory` (more can be done in the future if needed) by elementary methods without using any transcendental function. The basic idea is that `inf  {x | 0 < x /\ c <= x pow 2} = sqrt c` for `0 < c`, and then the existence of square roots can be proved by properties of `pow` and `inf`. Of course, the total amount of proofs are a bit longer than their previous versions in `transcTheory`.
```
   [SQRT_0]  Theorem      
      ⊢ sqrt 0 = 0
   
   [SQRT_MUL]  Theorem      
      ⊢ ∀x y. 0 ≤ x ∧ 0 ≤ y ⇒ (sqrt (x * y) = sqrt x * sqrt y)
   
   [SQRT_POS_LE]  Theorem      
      ⊢ ∀x. 0 ≤ x ⇒ 0 ≤ sqrt x
   
   [SQRT_POS_LT]  Theorem      
      ⊢ ∀x. 0 < x ⇒ 0 < sqrt x
   
   [SQRT_POS_NE]  Theorem      
      ⊢ ∀x. 0 < x ⇒ sqrt x ≠ 0
   
   [SQRT_POW2]  Theorem      
      ⊢ ∀x. ((sqrt x)² = x) ⇔ 0 ≤ x
   
   [SQRT_POW_2]  Theorem      
      ⊢ ∀x. 0 ≤ x ⇒ ((sqrt x)² = x)
```

As a side effect, once we have enough `sqrt` properties in `realTheory`, I managed to move the canonical `mr2` (Euclidean metric in 2-D real space) from `real_borelTheory` to `metricTheory` (to live with `mr1`):
```
   [mr2]  Definition
      ⊢ mr2 = metric (λ((x1,x2),y1,y2). sqrt ((x1 − y1)² + (x2 − y2)²))

   [MR2_DEF]  Theorem      
      ⊢ ∀x1 x2 y1 y2.
          dist mr2 ((x1,x2),y1,y2) = sqrt ((x1 − y1)² + (x2 − y2)²)
```

Regards,

Chun Tian


binghe:
In the 2nd commit I have re-proved more `SQRT` theorems and moved them to `realTheory`. However, I found the following theorem is "impossible" to prove:
```
[SQRT_DIV]  Theorem      
      ⊢ ∀x y. 0 ≤ x ∧ 0 ≤ y ⇒ sqrt (x / y) = sqrt x / sqrt y
```
Note that, then `0 < x /\ y = 0`, the above theorem becomes `sqrt (x / 0) = sqrt x / 0`.  No real theorem can do future rewrites...

binghe:
> Note that, _when_ `0 < x /\ y = 0`, the above theorem becomes `sqrt (x / 0) = sqrt x / 0`. No real theorem can do future rewrites...

Actually this is possible by `|- x / y = x * inv y` and `|- inv 0 = 0`...

binghe:
Hi, in the latest commit 8300f5d I have ported `SQRT_DIV` to `realTheory` and finally removed dependency on `transcSyntax` (and `transcTheory`) from `isqrtLib`. I also added some basic tests in `src/real/selftest.sml`.

I moved the definition of `root` back to `transcTheory`, by using the following direct definition of `sqrt`:
```
   [sqrt_def]  Definition      
      ⊢ ∀x. sqrt x = @u. (0 < x ⇒ 0 < u) ∧ (u² = x)
```
The original definition of `sqrt` now becomes an equivalent theorem in `transcTheory`:
```
   [sqrt]  Theorem      
      ⊢ ∀x. sqrt x = root 2 x
```

src/real/transcScript.sml
```diff
@@ -727,6 +727,16 @@ QED
 (* Some properties of roots (easier via logarithms)                          *)
 (*---------------------------------------------------------------------------*)
 
+Definition root:
```
mn200, line 730: Couldn't `root` stay in real?  It's no more transcendental than `sqrt`?

binghe: That's because so far we don't have other `ROOT` theorems (with elementary proofs) moved into `realTheory`, and a definition itself is hardly useful without its supporting theorems.   In general, I think it's not a very good style to isolate a definition from all its properties.

If, in the future, some or all `ROOT` theorems could be re-proved (not as easy as the case of `sqrt`) without using transcendental functions, then of course they should be moved out of `transcTheory`.   But I'm fine to move `root` back if you insist on...

mn200: OK; this is fair.

binghe:
Thanks. By the way, do you prefer `iterateTheory` depending on `transcTheory`, or the other way (`transcTheory` depending on `iterateTheory`)?  I think with little efforts it's also possible to make `iterateTheory` getting rid of topology stuff,  by moving `LN_PRODUCT` and `ROOT_PRODUCT` to `transcTheory`.

mn200:
At the moment, `iterate` theory seems to be a random grab-bag.  I'm planning to move the `numseg` constant back to `pred_set` for example, and it seem as if there's other material there that should also be earlier.  So, yes, I think if iterate continues to exist at all, it should definitely predate `transc`.

binghe:
> At the moment, `iterate` theory seems to be a random grab-bag. I'm planning to move the `numseg` constant back to `pred_set` for example, and it seem as if there's other material there that should also be earlier. So, yes, I think if iterate continues to exist at all, it should definitely predate `transc`.

Great. My previous plan is done by the just merged PR #980, except that the two theorems `LN_PRODUCT` and `ROOT_PRODUCT` were moved to `real_sigmaTheory` where they are used in proving `AGM` etc, because of some issues on the two `sum` definitions: one is in `realTheory`, the other is in `iterateTheory`: simply opening `iterateTheory` in `transcTheory` will break the proofs of `FTC1`. Changing the overload of `sum` in `iterateTheory` to inferior is one solution, however.

At least half of `iterateTheory` doesn't use real numbers at all, e.g. `NSUM` and `NPRODUCT`. We have too many duplicated definitions and theorems on the summation and products of numbers....

PR #979, opened 2021-12-07
Match cheatsheet changes in hol-website

dwRchyngqxs:
Same changes as https://github.com/HOL-Theorem-Prover/hol-webpages/pull/10


mn200:
Thanks!

PR #980, opened 2021-12-08
Dependency cleanup (iterateTheory and transcTheory)

binghe:
Hi,

this small PR doesn't add any new theorem nor modify any existing proof. It only removes dependency on `transcTheory` from `iterateTheory`, making it a direct secondary theory after `realTheory`, and also from `src/floating-point` (the 2nd commit), as the further step of a previous commit 2fa1a06eb7e3b3bf67c20baa42eb70e4347a39c4.

In the attached sreenshot, we can see that now ieee-floating theories are completely separated from topology- and transcendental functions, and the position of `iterateTheory` is much "higher", as an immediate child of `realTheory`. This makes more senses.

P. S. currently we have two versions of differential- and intergral- calculus, the old one is hidding in `src/real` (`seq`, `lim`, `transc` and `integral`), the new one is in `src/probability` (`derivative` and `integration`). Merging them without breaking any compatibility is possible but requires a long time to work on (without any academic value, of course.)

Regards,

Chun Tian

Schermata 2021-12-08 alle 12 54 37

### DIFF
--- a/src/float/floatScript.sml
+++ b/src/float/floatScript.sml
@@ -5,7 +5,8 @@
 open HolKernel boolLib bossLib
 open pairTheory pred_setTheory prim_recTheory numTheory arithmeticTheory
      realTheory ieeeTheory
-open numLib realLib Ho_Rewrite
+
+open numLib realSimps RealArith Ho_Rewrite
 
 val () = new_theory "float"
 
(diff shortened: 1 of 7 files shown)


binghe:
With the 2rd commit now `src/float` also does not depend on `transcTheory`, as it only uses `sqrt`. Now the relevant part of the core dependency graph looks like this: (even better)

Schermata 2021-12-08 alle 15 02 30

mn200:
Thanks for cleaning this up.  I hadn't realised that the (outdated) `float` stuff also depended on `transc`.

PR #981, opened 2021-12-08
Simulation proofs relating L3 and ASL-derived models of Armv8 ISA

hrutvik:
- A script to clone a pre-generated model of the Armv8 ISA, hosted on the [Arm Research GitLab](https://gitlab.com/arm-research/security)
  - The model was extracted from Arm's ASL (architecture specification language) to [Sail](https://www.cl.cam.ac.uk/~pes20/sail/) using [`asl_to_sail`](https://github.com/rems-project/asl_to_sail)
  - This was then extracted to [Lem](https://github.com/rems-project/lem) via the Sail ecosystem, and then to HOL4 using Lem
  - Some manual modifications have also been made:
    - Simplification of the state/exception monad (in particular model non-determinism using choice operator rather than sets of outcomes)
    - Removal of address translation and ensuring physical addresses are 64-bit to align with L3 models
    - Removal off a certain diagnostic-style feature ("trickbox")
- Proofs relating the existing L3 model to the ASL-derived one for all instructions generated by the [CakeML](https://cakeml.org/) Armv8 backend:
  - *Given*
    - L3 and ASL states which are suitably equivalent (i.e. they agree on all parts of the L3 state, since it is a subset of the ASL state)
    - certain system register configurations for the ASL model
    - the L3 model does not produce an exception
  - *Then*
    - the ASL model does not produce an exception
    - the resulting L3 and ASL states are suitably equivalent
    - none of the specified system register configurations have changed

With thanks to @acjf3 and @myreen for their help!


hrutvik:
Note that the models take ~4 hrs to build, and (to the best of my knowledge) require a machine with 64 GB RAM

mn200:
In an ideal world, our existing `-t3` builds are done on a machine capable of handling this, so you could add the appropriate directory to the `EX3DIRS` definition in `src/parallel_builds/core/Holmakefile`.   However, the recursive invocation of `Holmake` necessary after the `get-mra-snapshot` call may not interact super-well with this. 

Moreover, the need to base our builds on calls to remove infrastructure makes me antsy; if the intention is that the snapshot shouldn't change, can't we just have it in the distribution directly?

acjf3:
The model builds on my machine with just 16GB or RAM. However, it does take a long time to build (at least 3 hours) and the source is fairly large (36M). Obviously, the equivalence proofs themselves add to these times. 

@mn200 How do you feel about skipping this development from the self tests entirely? Would that be a blocker to including it in the examples directory?

mn200:
I don’t have a huge problem with not having it in the regression testing sequence, but I then naturally worry that it will bit-rot and it’ll end up unmaintained and irretrievable.   And even if the sources in the `examples` directory are fine, what happens when the remote URL changes/disappears? 

If it depends so heavily on a remote resource that is not itself practical to distribute inside HOL, would it make more sense to have @hrutvik’s work distributed there as part of that instead?

hrutvik:
On the bit-rot aspect of things - If it helps, the intention is to put it in the CakeML regression test I believe (perhaps @myreen can confirm). I don't have anything useful to contribute on the remote/non-remote side of things, happy to defer

acjf3:
The HOL Arm ASL model is currently hosted in a GitLab [repository](https://gitlab.com/arm-research/security/mra-hol-snapshots) which only Arm employees can commit to. This was part of the appeal, as it provides provenance/integrity for the specification.

For various reasons, I'm not keen on replicating the contents under the `examples` directory, to be distributed with HOL. One reason is that it would become lost within the ever growing clutter of the `examples` directory. It feels like we should consider using separate repositories under `github.com/HOL-Theorem-Prover` for hosting large examples? (Obviously, this would introduce synchronisation issues, but developments such as CakeML already have solutions for that.) One advantage is that it would help folks when it comes to referencing, locating and downloading these development for inspection. Another advantage is that it could help with regression testing, e.g., providing more control over build frequency. In particular, I would like to see these equivalence proofs (and associated CakeML proofs) built with fairly low frequency. How often are `-t3` builds done at present?

@mn200 , @myreen , @hrutvik Thoughts?

mn200:
I do rather like the idea of a separate repo under `HOL-Theorem-Prover` as this can then be regression tested completely independently of HOL and CakeML.

acjf3:
Cool. Hopefully @hrutvik can help with putting things together for that (no rush though).

mn200:
Thanks for getting this all done!

PR #982, opened 2021-12-10
Re-support old Emacs versions by eliminating IF-LET*

binghe:
Hi,

currently in `tools/hol-mode.src` there's a single use of `IF-LET*` Lisp macro (introduced in Emacs 26.1, I think) stopping the entire HOL emacs lisp mode from running on older Emacs versions (24 and 25).  This issue has been driving me crazy, because the **best** Emacs environment on macOS, Aquamacs [1], its latest versions (3.5 and 3.6 beta 2) are still based on Emacs 25.x. It may still take years to see Aquamacs 4 coming out (based on Emacs 28) as the author has planned.

I hope you can do me a favor by accepting this PR so that I don't need to keep the local changes in every of my HOL working directories.  The patch can also benefit Emacs+HOL users on Ubuntu 18.04, Solaris 10 and 11, and other operating systems in which no official prebuilt Emacs 27 is easily available.

Regards,

Chun Tian

[1] http://aquamacs.org

### DIFF
--- a/tools/hol-mode.src
+++ b/tools/hol-mode.src
@@ -844,19 +844,24 @@ region instead. With prefix ARG prompt for a file-name to load."
 
 (defun hol--find-alternate-executable (dir0 original-name0)
   (let ((dir (expand-file-name dir0))
-        (original-name (expand-file-name original-name0)))
+        (original-name (expand-file-name original-name0))
+        executable)
     ;; check for lastmaker file
-    (if-let*
-        (((file-readable-p ".HOLMK/lastmaker"))
-         ((> 500 (file-attribute-size (file-attributes ".HOLMK/lastmaker"))))
-         (lines (with-temp-buffer
-                  (insert-file-contents-literally ".HOLMK/lastmaker")
-                  (split-string (buffer-string) "\n" t)))
-         ((not (null lines)))
-         (dir1 (file-name-directory (car lines)))
-         (name (concat dir1 "hol"))
-         ((not (equal name original-name))))
-        (list (concat dir1 "hol") "(honouring last build in this directory)")
+    (if (and (file-readable-p ".HOLMK/lastmaker")
+             (> 500 (file-attribute-size (file-attributes ".HOLMK/lastmaker")))
+             (let ((lines (with-temp-buffer
+                            (insert-file-contents-literally ".HOLMK/lastmaker")
+                            (split-string (buffer-string) "\n" t))))
+               (and (not (null lines))
+                    (let* ((dir1 (file-name-directory (car lines)))
+                           (name (concat dir1 "hol")))
+                      (and (not (equal name original-name))
+                           (setq executable
+                                 (list (concat dir1 "hol")
+                                       "(honouring last build in this directory)")))))))
+        ;; then
+        executable
+      ;; else
       (while (and (not (hol--looks-like-root-holdir dir))
                   (not (equal "/" dir)))
         (setq dir (file-name-directory (directory-file-name dir))))


binghe:
I believe the function `hol--find-alternate-executable` has a bug (even before my changes): the file name of the alternative executable constructed from the current HOL working directory is always `hol`, even when the user has set `hol-executable` to `hol.bare` from whatever HOL working directories.  As a result, I think currently there's actually no way to start up `hol.bare` from within a HOL working directory.

mn200:
I think this behaviour (wrt `hol.bare`) is by design.  You can specify use of `hol.bare` by using `hol-set-executable`.

binghe:
> I think this behaviour (wrt `hol.bare`) is by design. You can specify use of `hol.bare` by using `hol-set-executable`.

You're right. My previous judgements were wrong as I was wrongly setting a different `hol.bare` from another working directory. I tried again (to starting a HOL session in the directory of `pred_setTheory` and it worked.) Sorry.

mn200:
I don't have any strong objection to this, though it's clear that the new code is much uglier than the `if-let*` code (which is why I used that facility in the first place).  Is Aquamacs really that much better than the emacs you'd get from [emacsformacosx.com](http://emacsformacosx.com)?

binghe:
> I don't have any strong objection to this, though it's clear that the new code is much uglier than the `if-let*` code (which is why I used that facility in the first place). Is Aquamacs really that much better than the emacs you'd get from [emacsformacosx.com](http://emacsformacosx.com)?

The answer to your question is yes [1] but I admit this is very subjective and unconvincing. I use Aquamacs everyday, every minute when my Mac computer is on. (And whenever it has a new release, I always click the donation link.)   People choose to use another software than another, mostly because of some personal opinions, and it was these personal opinions making me think that 1) OCaml is a bad platform (thus HOL Light is only good at its theory library), and 2) Isabelle is also a bad platform (simply because it doesn't provide a on-disk file format to hold already proved theorem (at a finer level, the heap dumping provided by PolyML doesn't resolve most difficulities.)

For the first part, I think `IF-LET*` is a very bad macro. As a multiple bindings version of `IF-LET`, it never existed in Common Lisp, but something weird from Clojure community [2]. The problem is `IF-LET*` is that, when its _else_ branch is executed, only a partial set of variable bindings are established (the rest of them are still `NIL`). `IF-LET` is still good, no such concerns as it only supports one variable binding. In another related macro `WHEN-LET*` (and `WHEN-LET*`), such issues also do not exist.

In your original code using `IF-LET*`, all those established bindings are not used in the _else_ branch, and this is supposed to be a sign that `IF-LET*` should be used in this case. In my modified code, yes it's longer, but all irrelevant bindings are only inside the _cond_ part of the top `IF`, naturally avoiding unintended access from _then_ and _else_ branches.

Anyway, thanks for your "don't have any strong objection", because otherwise I have to spend much more time with Aquamacs author to push the efforts for its next major release.

Regards,

Chun Tian

[1] http://aquamacs.org/features.html
[2] https://clojuredocs.org/clojure.core/if-let

binghe:
I have cut a piece of macro definition code from the latest Emacs (`lisp/emacs-lisp/subr-x.el`) into a new elisp file `if-let.el` (see attached file) and then load it in my `~/.emacs`:
```
;; hol-mode.el needs IF-LET*
(when (<= emacs-major-version 25)
  (byte-recompile-file "~/.emacs.d/if-let.lisp" :load t))
```
Now I can normally use `hol-mode` in older Emacs. According to the notes found in byte-compilation, even Emacs 19 can be supported. `EmacsForMacOSX` has been deleted from my computer.  No need to waste more time on this matter.

[if-let.txt](https://github.com/HOL-Theorem-Prover/HOL/files/7703299/if-let.txt)

PR #983, opened 2021-12-10
Convert HOL-Light's ITSET into an equivalent theorem

binghe:
Hi,

This PR resolved a small but long-standing issue in `iterateTheory` where the important operator `iterate` was defined by HOL Light's version of `ITSET`, ever renamed to `ITSET'` by me during the cleanup of this theory:
```
   [ITSET']  Definition      
      ⊢ ∀f s b.
          ITSET' f s b =
          (@g. g ∅ = b ∧
               ∀x s.
                 FINITE s ⇒
                 g (x INSERT s) = if x ∈ s then g s else f x (g s)) s
```
Now I have successfully removed this definition and turned the above definition into the following equivalent theorem w.r.t HOL4's version of `ITSET`:
```
   [ITSET_alt]  Theorem      
      ⊢ ∀f s b.
          (∀x y z. f x (f y z) = f y (f x z)) ∧ FINITE s ⇒
          ITSET f s b =
          (@g. g ∅ = b ∧
               ∀x s.
                 FINITE s ⇒
                 g (x INSERT s) = if x ∈ s then g s else f x (g s)) s
```
And then the new definition of `iterate` just uses HOL4's version of `ITSET`.  Usually the definition of `iteate` itself is not directly used, but there's one exception in `derivativeTheory`, in which the above `ITSET_alt` theorem is now used to fix the proof (with minimal efforts).

P.S. I tried to move `ITSET_alt` to `pred_setTheory` but failed, because this theorem requires a lot of other theorems in `iterateTheory`.

Regards,

Chun Tian

### DIFF
--- a/src/probability/derivativeScript.sml
+++ b/src/probability/derivativeScript.sml
@@ -1778,15 +1778,14 @@ val REAL_MUL_NZ = store_thm ("REAL_MUL_NZ",
 
 (* `sum (0, SUC n)` is defined by realTheory.sum
    `sum {0 ..   n}` is defined by iterateTheory.sum_def
-
-   TODO: re-prove this lemma once `iterate` is re-defined by `pred_set$ITSET`
  *)
-val lemma_sum_eq = prove (
-  ``!n x:real. sum (0, SUC n) (\n. (\n. inv(&(FACT n))) n * (x pow n)) =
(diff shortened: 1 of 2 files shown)


mn200:
Nice!  Thanks for this.  The general tidying up of late is very satisfying.

PR #984, opened 2021-12-10
Automatic proofs that datatypes are countable

talsewell:
Some setup theory and a library that automatically proves that
datatypes defined in the TypeBase are countable.

I think I put the theory in the right place, it's unclear though whether
the relevant code belongs under 'more_theories'.

### DIFF
--- a/src/pred_set/src/more_theories/countable_typesLib.sig
+++ b/src/pred_set/src/more_theories/countable_typesLib.sig
@@ -1,0 +1,7 @@
+signature countable_typesLib =
+sig
+  include Abbrev
+
+  val mk_countable : hol_type -> thm
+
+end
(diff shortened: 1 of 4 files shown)


mn200:
Thanks!

mn200:
If you're going to appeal to `bossLib` or `listTheory`, then you definitely need to be in `pred_set/src/more_theories`, and so too do your tests.   As _per_ de7a160.

xrchz:
I thought there was a library for this in examples, but I can't find it any
more. Was it moved/deleted?

mn200:
Commit f55fd784a053e12 mentions removal of something called `countable` (“unused and untested” apparently), so I guess that may have been this.

Issue #962, opened 2021-11-02
Possible issues on Linux/ARM64


binghe:
Hi,

I don't have any ARM64 hardware, but I have an ARM64 version of Ubuntu 20.04 LTS driven by QEMU (you can easily get one with my scripts [1], also learnt from others) on Mac (it's a pity that currently I have no way to run macOS on Apple M1), mainly for building ARM64 releases of my other programs.

Since Poly/ML 5.9 is coming with native ARM64 support, I tried to build HOL in my ARM64 Linux box, and it strangely stops in `src/relation` with strange errors from `TAUT_PROVE` (?!):
```
Building directory src/relation [02 Nov, 09:27:47]
Holmake: Linking /home/binghe/ML/HOL/src/relation/relationScript.uo to produce theory-builder executable
error in quse /home/binghe/ML/HOL/src/refute/Canon.sml : HOL_ERR {message = "", origin_function = "TAUT_PROVE", origin_structure = "tautLib"}
error in load $(HOLDIR)/src/refute/Canon : HOL_ERR {message = "", origin_function = "TAUT_PROVE", origin_structure = "tautLib"}
error in load $(HOLDIR)/sigobj/refuteLib : HOL_ERR {message = "", origin_function = "TAUT_PROVE", origin_structure = "tautLib"}
error in load $(HOLDIR)/src/IndDef/InductiveDefinition : HOL_ERR {message = "", origin_function = "TAUT_PROVE", origin_structure = "tautLib"}
error in load $(HOLDIR)/sigobj/IndDefLib : HOL_ERR {message = "", origin_function = "TAUT_PROVE", origin_structure = "tautLib"}
error in load $(HOLDIR)/sigobj/BasicProvers : HOL_ERR {message = "", origin_function = "TAUT_PROVE", origin_structure = "tautLib"}
error in load /home/binghe/ML/HOL/src/relation/relationScript : HOL_ERR {message = "", origin_function = "TAUT_PROVE", origin_structure = "tautLib"}
Uncaught exception: HOL_ERR {message = "", origin_function = "TAUT_PROVE", origin_structure = "tautLib"}
Holmake: Failed script build for /home/binghe/ML/HOL/src/relation/relationScript - exited with code 1
*** FATAL: Build failed in directory /home/binghe/ML/HOL/src/relation (exited with code 1)
```

[...]

PR #985, opened 2021-12-11
Fixed incorrect results of MiniSat 1.x on Linux/ARM where "char" is unsigned

binghe:
See #962 for the original issue report and related discussions. The credits are totally of Oskar Abrahamsson (@oskarabrahamsson) who found the root causes and the solution.

Below are something that I also never knew before: it seems that in C90 ("seems" as I have no access to its ISO standard documentation), there are 3 _distinct_ types: `char`, `signed char` and `unsigned char`, and whether `char` is signed or unsigned is implementation-specific (by "implementation" it means `compiler+OS+arch`). For example, I found the following notes in «Oracle® Developer Studio 12.6: C User's Guide» (Appendix F. Implementation-Defined ISO/IEC C90 Behavior) [1]:
```
F.1.4.8 (6.2.1.1) Whether a plain char has the same range of values as signed char or unsigned char:
A char is treated as a signed char.
```

So far all compilers I can access on Linux/ARM (GCC and LLVM) treat `char` as `unsigned char`, which is different with what the MiniSat authors were expecting. Changing all string-irrelevant `char` uses to `signed char` should also work, but changing them to modern `int8_t` is more reasonable.

Lessions (I, not sure others) learnt: 

1. `char`, `signed char` and `unsigned char` ...
2. Never trust SAT solvers blindly (e.g. Isabelle includes MiniSat 2.x which has no proof logging features).
3. Don't trust SAT people who optimize their programs in _this_ way (to win SAT competitions) can lead to polynomial-time-complexity algorithms (thus prove `P = NP`) one day in the future...

and

4. Try UndefinedBehaviorSanitizer (UBSan) [2] first on weird-behavior C/C++ programs next time.

[1] https://docs.oracle.com/cd/E77782_01/html/E77788/bjazt.html#scrolltoc
[2] https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html

### DIFF
--- a/src/HolSat/sat_solvers/minisat/Solver.C
+++ b/src/HolSat/sat_solvers/minisat/Solver.C
@@ -253,7 +253,7 @@ public:
 
 void Solver::analyze(Clause* confl, vec<Lit>& out_learnt, int& out_btlevel)
 {
-    vec<char>&     seen  = analyze_seen;
+    vec<int8_t>&   seen  = analyze_seen;
     int            pathC = 0;
     Lit            p     = lit_Undef;
     bool           sconfl;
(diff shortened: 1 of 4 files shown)


mn200:
Thanks!

PR #987, opened 2021-12-16
Alternative thms to large quantHeuristics rules

talsewell:
Provide alternatives to the (very large) LIST_LENGTH_n and
(slightly large) HD_TL_EQ_n thms in quantHeuristics. The
alternative theorems depend on arithmetic but work for any
size by repeated rewriting.

This makes it possible that the large theorems might be
removed in the future.


mn200:
Thanks for this!

PR #993, opened 2022-01-19
A canonical metric for (one-dimensional) extreal space

binghe:
Hi,

this is the first step to have a proper definition of "continuous function" for extreal-valued functions (for future work such as weak convergence of extreal-based probability distributions): a canonical metrics for one-dimensional extreal spaces.

It's important to realize that, for no matter what abstract spaces, the "distance" (or metric) between any two points in this space must be a real number.  For one-dimensional real spaces, a trivial definition `\(x,y). x-y` simply works by satisfying the triangle inequality. For extreals, however, it's not that straightforward.

The "distance" between extreal values here are defined in the following way:

1. the distance of two infinities are either 1 or 0, depending on if they are the same.
2. the distance between an infinity and a normal extreal is also 1.
3. the distance between two normal extreals is the "reduced" metric of their real values which is always smaller than 1, by taking `d/(1+d)`, where `d = x-y`:

```
Definition extreal_dist_def :
    extreal_dist (Normal x) (Normal y) = dist (reduced_metric mr1) (x,y) /\
    extreal_dist  PosInf     PosInf    = 0 /\
    extreal_dist  NegInf     NegInf    = 0 /\
    extreal_dist  _          _         = 1
End

   [extreal_mr1_def]  Definition      
      ⊢ extreal_mr1 = metric (UNCURRY extreal_dist)

   [reduced_metric_def]  Definition      
      ⊢ ∀m. reduced_metric m =
            metric (λ(x,y). dist m (x,y) / (1 + dist m (x,y)))
```

By proving the triangle inequalities of `extreal$extreal_mr1` and `metric$reduced_metric`, now we are ready to proceed with a possible definition of "continous extreal-valued function".

Regards,

Chun Tian

### DIFF
--- a/src/probability/extrealScript.sml
+++ b/src/probability/extrealScript.sml
@@ -13,7 +13,7 @@ open metisLib combinTheory pred_setTheory res_quanTools pairTheory jrhUtils
      prim_recTheory arithmeticTheory tautLib pred_setLib hurdUtils;
 
 open realTheory realLib real_sigmaTheory RealArith seqTheory limTheory
-     transcTheory iterateTheory;
+     transcTheory iterateTheory metricTheory;
 
 open util_probTheory cardinalTheory;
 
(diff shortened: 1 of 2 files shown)


mn200:
Very nice; thanks!

PR #994, opened 2022-01-24
Remove q which caused problems.

lir108:
The additional q causes issue when loading HOL into emacs.


mn200:
Thanks!  Clearly the regression tests aren't super careful for this stuff

PR #977, opened 2021-11-28
More general topological and metric lemmas ported from HOL Light

binghe:
Hi,

this PR can be seen as a small improvement of HOL4's general topology and metric library by porting more related lemmas from HOL Light (mainly its `sets.ml` and `metric.ml`).  (I do this for future improvements in measure theory, however.) Previously there were few theorems in HOL4's `topologyTheory` and `metricTheory`, now they are charged with more results.

The end of the chain of newly ported theorems is the following non-trivial result (in `metricTheory`) showing that, in any metrizable (topological) space, any closed set is in _gdelta_, i.e., can be expressed as a countable intersection of open sets.  
```
   [CLOSED_IMP_GDELTA_IN]  Theorem      
      ⊢ ∀top s. metrizable_space top ∧ closed_in top s ⇒ gdelta_in top s
```
This generalized the following existing theorem in `real_topologyTheory` (and making its proof greatly simplified):
```
   [CLOSED_AS_GDELTA]  Theorem      
      ⊢ ∀s. closed s ⇒ gdelta s
```

The above theorem `CLOSED_IMP_GDELTA_IN` requires the following definitions and related lemmas, all ported from HOL Light:
```
   [gdelta_in]  Definition      
      ⊢ ∀top.
          gdelta_in top =
          countable INTERSECTION_OF open_in top relative_to topspace top

   [metrizable_space]  Definition      
      ⊢ ∀top. metrizable_space top ⇔ ∃m. top = mtop m

   [INTERSECTION_OF]  Definition      
      ⊢ ∀P Q.
          P INTERSECTION_OF Q =
          (λs. ∃u. P u ∧ (∀c. c ∈ u ⇒ Q c) ∧ BIGINTER u = s)

   [relative_to]  Definition      
      ⊢ ∀P s t. (P relative_to s) t ⇔ ∃u. P u ∧ s ∩ u = t
```
There are also `fsigma_in` and `UNION_OF`, as dual versions of `gdelta_in` and `INTERSECTION_OF`.   HOL Light has very nice formalization of set generators such as `{FINITE|COUNTABLE|ARBITRARY UNION_OF ...`, now they are also available in HOL4. Below is list of their theorems in `topologyTheory`:
```
    val ARBITRARY_INTERSECTION_OF_COMPLEMENT : thm
    val ARBITRARY_INTERSECTION_OF_EMPTY : thm
    val ARBITRARY_INTERSECTION_OF_IDEMPOT : thm
[...]

PR #997, opened 2022-02-01
Basic facts of complete measure spaces

binghe:
Hi,

currently there is a wrong definition (not used anywhere) in `measureTheory`, related to complete measure spaces:
```
   [complete_of_def]  Definition (old)
      ⊢ ∀m. complete_of m =
            (m_space m,
             {s ∪ n | s ∈ measurable_sets m ∧ ∃t. n ⊆ t ∧ null_set m t},
             measure m)
```

The most counter-intuitive fact in measure theory is that the subset of a "null" set (i.e. having zero measure) may not be measurable at all (and thus has no measure in the considered measure space). Given such a measure space, there's an obvious "fix" by adding all subsets of null sets into consideration, resulting into the so-called "complete measure spaces" constructed from the incomplete ones.   The above definition implements this idea, but the measure part `measure m` is wrong: the measure of previously incomplete measure space is not defined on the measurable sets of the resulting complete  measure space.  Thus, in general, the "completion" process can only produce a new system of sets, without a new measure built automatically:
```
   [completion_def]  Definition (new)      
      ⊢ ∀m. completion m =
            (m_space m,
             {s ∪ n | s ∈ measurable_sets m ∧ ∃t. n ⊆ t ∧ null_set m t})
```

Given any measure space `m`, what we can say in general for `completion m` are the following (new theorems):

1. The system of sets `completion m` is indeed a sigma-algebra:
```
   [SIGMA_ALGEBRA_COMPLETION]  Theorem      
      ⊢ ∀m. measure_space m ⇒ sigma_algebra (completion m)
```

2. The system of sets `completion m` contains all measurable sets of `m`:
```
   [COMPLETION_SUBSET_SUBSETS]  Theorem      
      ⊢ ∀m. measure_space m ⇒ measurable_sets m ⊆ subsets (completion m)
```

3.  Doing `completion` on a already-complete measure space is useless, i.e. the resulting system of sets are the same:
```
   [COMPLETION_STABLE]  Theorem      
      ⊢ ∀m. complete_measure_space m ⇒
            space (completion m) = m_space m ∧
            subsets (completion m) = measurable_sets m
```

This is what has been done in the present PR.   But note that, in `borelTheory` there's the Lebesgue measure space based on Gauge integrals:
```
   [lebesgue_def]  Definition      
      ⊢ lebesgue =
        (𝕌(:real),{A | ∀n. indicator A integrable_on line n},
         (λA. sup {Normal (integral (line n) (indicator A)) | n ∈ 𝕌(:num)}))
```

One may ask if `lebesgue` is the same thing as the completion of real-valued Borel measure space `lborel` (constructed by Carathéodory's Extension Theorem in a difficult way)
```
   [lborel_def]  Definition      
      ⊢ (∀s. s ∈ subsets right_open_intervals ⇒ lambda s = lambda0 s) ∧
        (m_space lborel,measurable_sets lborel) = borel ∧
        measure_space lborel
```

The answer is *yes*, but so far this result is not proved yet.  This proof is not in any probability book (nor measure theory books, as far as I read). Instead, the only proof I found is in «A Modern Theory of Integration» [1] (the introduction book of Henstock-Kurtsweil integration), Theorem 18.19 (p.314), saying that any measurable set `E IN lebesgue` can be decomposed into `E = K UNION Z`, where `Z` is a null set of `lebesgue`, `K` is an *F-sigma* set in `borel` (and therefore must be contained in `completion lborel`, the other side is trivial).  This would be my next move (_the previous PR #977 porting lemmas of G-delta and F-sigma sets from HOL-Light was actually in preparation of this work_), but I think the existing wrong definition must be fixed first.

Regards,

Chun Tian

[1] Bartle, R.G.: A Modern Theory of Integration. American Mathematical Soc. (2001).

### DIFF
--- a/src/probability/measureScript.sml
+++ b/src/probability/measureScript.sml
@@ -4715,15 +4715,12 @@ QED
 val null_set_def = Define
    `null_set m s <=> s IN measurable_sets m /\ (measure m s = 0)`;
 
-(* MATHEMATICAL SCRIPT CAPITAL N, not very meaningful
-val _ = Unicode.unicode_version {u = UTF8.chr 0x1D4A9, tmnm = "null_set"};
- *)
-
-(* a measure space m which is not yet complete can be completed *)
-val complete_of_def = Define
-   `complete_of m =
-      (m_space m, {s UNION n | s IN measurable_sets m /\ ?t. n SUBSET t /\ null_set m t},
-       measure m)`;
+(* NOTE: the type of ‘completion m’ is “:'a algebra” *)
+Definition completion_def :
+    completion (m :'a m_space) =
+      (m_space m, {s UNION n | s IN measurable_sets m /\
+                               ?t. n SUBSET t /\ null_set m t})
+End
 
 (* the measure space m is called complete iff any subset of a null set is again
    in `subsets m` (thus also a null set) see [1] (p.29], [5] (p.382) *)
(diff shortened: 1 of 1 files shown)


mn200:
Thanks for your work on this!

PR #998, opened 2022-02-16
vimhol terminal

arolle:
For supported vim (or neovim) editors these changes make it possible to open a HOL repl session within the editor next to an opened file buffer (at the focussed buffer's working directory). In addition the vimhol shortcuts are disabled for files other than HOL4 script files.


mn200:
I'm curious about 
> Non-standard heaps do not load the code to setup the vim-hol repl connection.

In what way do "standard" heaps load the code, and why don't non-standard heaps?

arolle:
Actually, this originates from working with different versions of HOL (prior to and after 3f9b66ce0ec2c5e109a27ab0ba1f34454f242ae5 ).

I've removed that part from this PR.

mn200:
Thanks!

PR #1001, opened 2022-02-22
Add `SELECT_GOALS_LT_THEN1` with `>>~-` combinator

hrutvik:
As discussed offline. Thanks to @myreen and @samsa1 for the suggestion

### DIFF
--- a/help/Docfiles/Tactical.SELECT_LT_THEN.doc
+++ b/help/Docfiles/Tactical.SELECT_LT_THEN.doc
@@ -3,9 +3,9 @@
 \TYPE {SELECT_LT_THEN : tactic -> tactic -> list_tactic}
 
 \SYNOPSIS
-Applies a first tactic to the all the goals in the goal-list for which the
+Applies the first tactic to all goals in the goal-list for which the
 tactic succeeds.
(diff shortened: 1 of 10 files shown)


hrutvik:
Currently `SELECT_LT_THEN` will not fail if its first tactic does not select any subgoals, or if it solves all the subgoals it selects. This is unlike `THEN1`, which will complain if its first tactic solves all subgoals. This could be changed, though it may break usages of `SELECT_LT` and the derived `>>~` combinator.

help/Docfiles/Tactical.SELECT_LT_THEN.doc
```diff
@@ -0,0 +1,47 @@
+\DOC
+
+\TYPE {SELECT_LT_THEN : tactic -> tactic -> list_tactic}
+
+\SYNOPSIS
+Applies a first tactic to the all the goals in the goal-list for which the
```
mn200, line 6: Grammar.  Perhaps “Applies first tactic to all the goals in the goal-list ...”

mn200 (commented):
Looks awesome!

PR #1002, opened 2022-02-23
Use SHELL as replacements of "/bin/sh" if necessary

binghe:
Hi,

This is Solaris 10 x86. When building recent HOL it throws the following error when executing shell commands:
```
for i in *.uo *.ui *.sig ; do ln -fs `pwd`/$i /export/home/binghe/ML/HOL/sigobj ; done && for i in *.sig ; do echo `pwd`/$(basename $i .sig) >> /export/home/binghe/ML/HOL/sigobj/SRCFILES ; done
sh: syntax error at line 1: `(' unexpected
*** FATAL: Build failed in directory /export/home/binghe/ML/HOL/src/parallel_builds/core (exited with code 1)
```

This is because Solaris 10 uses the original Bourne shell `/bin/sh` which doesn't support the shell command syntax (I guest the part `$(basename $i .sig)` is the problem).    All Linux systems, on the other hand, uses the GNU Bourne Again Shell (bash) which works well with HOL.   Note that `/bin/bash` is not available in all Unix systems, e.g. FreeBSD (where user can install `/usr/local/bin/bash`.)

Just like the previous issues with the original BSD `make` vs GNU `gmake`, the present PR resolves the building issue by allowing user to customize the shell command program with the `SHELL` variable.   By default, the value of `SHELL` inherits the value of the environment variable of the same name.

HOL also generates several script files like `$(HOLDIR)/bin/hol` and `$(HOLDIR)/bin/hol.bare` which has `#!/bin/sh` as the first line (the running shell). There's no need to change them, but in `$(HOLDIR)/bin/hol` the use of `$()` is not compatible with the original Bourne shell. I changed it to use backquotes instead, which is supported.

P. S. currently Poly/ML also has some issues. See https://github.com/polyml/polyml/issues/171 for one of them. (The other is the uses of `/bin/sh` in `polyc`, whose syntax also goes beyond the original Bourne shell.)   But any way, now I can successfully build and use HOL on Solaris 10.

Chun

### DIFF
--- a/tools-poly/Holmake/unix-systeml.sml
+++ b/tools-poly/Holmake/unix-systeml.sml
@@ -97,6 +97,7 @@ val OS = ""
 
 val DEPDIR = ""
 val GNUMAKE = ""
+val SHELL = ""
 val DYNLIB = ""
 val version = ""
 val ML_SYSNAME = ""
--- a/tools-poly/Holmake/winNT-systeml.sml
+++ b/tools-poly/Holmake/winNT-systeml.sml
@@ -53,6 +53,7 @@ val MOSMLDIR =
 val OS =
 val DEPDIR =
 val GNUMAKE =
+val SHELL =
 val DYNLIB =
 val version =
 val release =
--- a/tools-poly/configure.sml
+++ b/tools-poly/configure.sml
@@ -253,6 +253,7 @@ in
    "val OS ="       --> ("val OS = "^quote OS^"\n"),
    "val DEPDIR ="   --> ("val DEPDIR = "^quote DEPDIR^"\n"),
    "val GNUMAKE ="  --> ("val GNUMAKE = "^quote GNUMAKE^"\n"),
+   "val SHELL ="    --> ("val SHELL = "^quote SHELL^"\n"),
    "val DYNLIB ="   --> ("val DYNLIB = "^Bool.toString dynlib_available^"\n"),
    "val version ="  --> ("val version = "^Int.toString version_number^"\n"),
    "val ML_SYSNAME =" --> "val ML_SYSNAME = \"poly\"\n",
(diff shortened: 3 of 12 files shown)


mn200:
Given that some of the core Holmakefile technology relies on shell syntax that is specific to the `/bin/sh` family, I worry about people thinking that they can set `SHELL` to arbitrary things (_e.g._ `tcsh` or something weird like that).  If they do that, various Holmakefiles throughout the system will simply break.

It seems like it would be safer to allow people to locally set `SHELL` in a `Holmakefile` (as _per_ what [GNU make does](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html#Choosing-the-Shell)), and to make sure that all of our existing files do indeed work with vanilla `/bin/sh`.

binghe:
> Given that some of the core Holmakefile technology relies on shell syntax that is specific to the `/bin/sh` family, I worry about people thinking that they can set `SHELL` to arbitrary things (_e.g._ `tcsh` or something weird like that). If they do that, various Holmakefiles throughout the system will simply break.
> 
> It seems like it would be safer to allow people to locally set `SHELL` in a `Holmakefile` (as _per_ what [GNU make does](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html#Choosing-the-Shell)), and to make sure that all of our existing files do indeed work with vanilla `/bin/sh`.

Yes, this sounds a better solution. I close this PR and look forward to see it.

mn200:
If you want the credit, do feel free to make a PR to fix the broken use of `$(..)`.  I'll get around to fixing that otherwise at some point as it's definitely wrong.

binghe:
> If you want the credit, do feel free to make a PR to fix the broken use of `$(..)`. I'll get around to fixing that otherwise at some point as it's definitely wrong.

no… you can cherry-pick whatever changes in my branch and commit by yourself. I just want a working HOL on Solaris 10.

PR #1004, opened 2022-03-11
Remove company mode integration, update snippets

HeikoBecker:
Back when adding the company mode integration, I was not aware that it shadows other backends. This is clearly undesirable.
This PR removes this integration.
Also, I am updating the HOL4 snippets in `tools/yasnippets`. When using the emacs mode, they are accessible by pressing `M-h Tab` under a supported keyword.


mn200:

PR #1005, opened 2022-03-11
Cheatsheet updates

hrutvik:
- Add `pairarg_tac` and `disj{1,2}_tac` to the cheatsheet (thanks to @myreen and @samsa1 respectively for the suggestions)
- Superficial change: convert a couple of instances of `tactic1` to `tactic` when there is no `tactic2`


mn200:
Thanks!

PR #1006, opened 2022-03-12
Lemmas/trivial theorems to ultimately be used for extrealTheory and borelTheory results.

jdyeager:
Specifically some manipuations of ITSET (for working with EXTREAL_SUM_IMAGE) and showing some functions are bijections (for a later result about suminf).

### DIFF
--- a/src/coretypes/pairScript.sml
+++ b/src/coretypes/pairScript.sml
@@ -160,7 +160,7 @@ val SWAP_def = new_definition ("SWAP_def", ``SWAP a = (SND a, FST a)``)
 
 (* Theorem the SWAP inverts itself *)
 Theorem SWAP_SWAP[simp]:
-    ∀x. SWAP (SWAP x) = x
+    !x. SWAP (SWAP x) = x
 Proof
     simp[SWAP_def]
 QED
--- a/src/coretypes/pairScript.sml
+++ b/src/coretypes/pairScript.sml
@@ -158,8 +158,8 @@ val PAIR_FST_SND_EQ = store_thm(
 
 val SWAP_def = new_definition ("SWAP_def", ``SWAP a = (SND a, FST a)``)
 
-(* Theorem the SWAP invert itself *)
-Theorem SWAP_SWAP:
+(* Theorem the SWAP inverts itself *)
+Theorem SWAP_SWAP[simp]:
     ∀x. SWAP (SWAP x) = x
 Proof
     simp[SWAP_def]
--- a/src/coretypes/pairScript.sml
+++ b/src/coretypes/pairScript.sml
@@ -158,6 +158,13 @@ val PAIR_FST_SND_EQ = store_thm(
 
 val SWAP_def = new_definition ("SWAP_def", ``SWAP a = (SND a, FST a)``)
 
+(* Theorem the SWAP invert itself *)
+Theorem SWAP_SWAP:
+    ∀x. SWAP (SWAP x) = x
+Proof
+    simp[SWAP_def]
+QED
+
 (*---------------------------------------------------------------------------*)
 (* CURRY and UNCURRY. UNCURRY is needed for terms of the form `\(x,y).t`     *)
 (*---------------------------------------------------------------------------*)
--- a/src/pred_set/src/pred_setScript.sml
+++ b/src/pred_set/src/pred_setScript.sml
@@ -4657,6 +4657,13 @@ val COMMUTING_ITSET_RECURSES = store_thm(
     ASM_SIMP_TAC bool_ss [COMMUTING_ITSET_INSERT, delete_non_element]
   ]);
 
+(* Corollary *)
+Theorem ITSET_SING:
+    ∀f x a. ITSET f {x} a = f x a
+Proof
+    rw[] >> fs[ITSET_THM]
+QED
+
 (* ----------------------------------------------------------------------
     SUM_IMAGE
 
@@ -6451,6 +6458,14 @@ val pair_to_num_inv = Q.store_thm ("pair_to_num_inv",
    (!x y. num_to_pair (pair_to_num (x, y)) = (x, y))`,
   SRW_TAC [][pair_to_num_def, num_to_pair_def]);
 
+(* More generally applicable version of the above *)
+Theorem pair_to_num_inv':
+    (∀x. pair_to_num (num_to_pair x) = x) ∧
+    (∀x. num_to_pair (pair_to_num x) = x)
+Proof
+    simp[FORALL_PROD,pair_to_num_inv]
+QED
+
 val num_cross_countable = Q.prove (
   `countable (UNIV:num set CROSS UNIV:num set)`,
   RWTAC [countable_surj, SURJ_DEF, CROSS_DEF] THEN
@@ -6987,6 +7002,26 @@ Proof
 QED
 
 (* end PREIMAGE lemmas *)
+
+(* Miscellaneous bijections *)
+
+Theorem BIJ_NUM_TO_PAIR:
+    BIJ num_to_pair UNIV (UNIV CROSS UNIV)
+Proof
+    simp[BIJ_IFF_INV] >> Q.EXISTS_TAC ‘pair_to_num’ >> simp[pair_to_num_inv']
+QED
+
+Theorem BIJ_PAIR_TO_NUM:
+    BIJ pair_to_num (UNIV CROSS UNIV) UNIV
+Proof
+    simp[BIJ_IFF_INV] >> Q.EXISTS_TAC ‘num_to_pair’ >> simp[pair_to_num_inv']
+QED
+
+Theorem BIJ_SWAP:
+    BIJ SWAP (UNIV CROSS UNIV) (UNIV CROSS UNIV)
+Proof
+    simp[BIJ_IFF_INV] >> Q.EXISTS_TAC ‘SWAP’ >> simp[SWAP_SWAP]
+QED
 
 (* "<<=" is overloaded in listTheory, cardinalTheory and maybe others,
    we put its Unicode and TeX definitions here to make sure by loading any of the
--- a/src/pred_set/src/pred_setScript.sml
+++ b/src/pred_set/src/pred_setScript.sml
@@ -4659,7 +4659,7 @@ val COMMUTING_ITSET_RECURSES = store_thm(
 
 (* Corollary *)
 Theorem ITSET_SING[simp]:
-    ∀f x a. ITSET f {x} a = f x a
+    !f x a. ITSET f {x} a = f x a
 Proof
     rw[] >> fs[ITSET_THM]
 QED
@@ -6460,8 +6460,8 @@ val pair_to_num_inv = Q.store_thm ("pair_to_num_inv",
 
 (* More generally applicable version of the above *)
 Theorem pair_to_num_inv'[simp]:
-    (∀x. pair_to_num (num_to_pair x) = x) ∧
-    (∀x. num_to_pair (pair_to_num x) = x)
+    (!x. pair_to_num (num_to_pair x) = x) /\
+    (!x. num_to_pair (pair_to_num x) = x)
 Proof
     simp[FORALL_PROD,pair_to_num_inv]
 QED
--- a/src/pred_set/src/pred_setScript.sml
+++ b/src/pred_set/src/pred_setScript.sml
@@ -4561,7 +4561,7 @@ Theorem ITSET_THM =
 W (GENL o rev o free_vars o concl)
   (DISCH_ALL(ASM_REWRITE_RULE [ASSUME ``FINITE s``] (SPEC_ALL ITSET_def)));
 
-Theorem ITSET_EMPTY =
+Theorem ITSET_EMPTY[simp] =
         REWRITE_RULE []
                      (MATCH_MP (SPEC ``{}`` ITSET_THM) FINITE_EMPTY);
 
--- a/src/pred_set/src/pred_setScript.sml
+++ b/src/pred_set/src/pred_setScript.sml
@@ -4658,7 +4658,7 @@ val COMMUTING_ITSET_RECURSES = store_thm(
   ]);
 
 (* Corollary *)
-Theorem ITSET_SING:
+Theorem ITSET_SING[simp]:
     ∀f x a. ITSET f {x} a = f x a
 Proof
     rw[] >> fs[ITSET_THM]
@@ -6459,7 +6459,7 @@ val pair_to_num_inv = Q.store_thm ("pair_to_num_inv",
   SRW_TAC [][pair_to_num_def, num_to_pair_def]);
 
 (* More generally applicable version of the above *)
-Theorem pair_to_num_inv':
+Theorem pair_to_num_inv'[simp]:
     (∀x. pair_to_num (num_to_pair x) = x) ∧
     (∀x. num_to_pair (pair_to_num x) = x)
 Proof
@@ -7008,19 +7008,19 @@ QED
 Theorem BIJ_NUM_TO_PAIR:
     BIJ num_to_pair UNIV (UNIV CROSS UNIV)
 Proof
-    simp[BIJ_IFF_INV] >> Q.EXISTS_TAC ‘pair_to_num’ >> simp[pair_to_num_inv']
+    simp[BIJ_IFF_INV] >> Q.EXISTS_TAC ‘pair_to_num’ >> simp[]
 QED
 
 Theorem BIJ_PAIR_TO_NUM:
     BIJ pair_to_num (UNIV CROSS UNIV) UNIV
 Proof
-    simp[BIJ_IFF_INV] >> Q.EXISTS_TAC ‘num_to_pair’ >> simp[pair_to_num_inv']
+    simp[BIJ_IFF_INV] >> Q.EXISTS_TAC ‘num_to_pair’ >> simp[]
 QED
 
 Theorem BIJ_SWAP:
     BIJ SWAP (UNIV CROSS UNIV) (UNIV CROSS UNIV)
 Proof
-    simp[BIJ_IFF_INV] >> Q.EXISTS_TAC ‘SWAP’ >> simp[SWAP_SWAP]
+    simp[BIJ_IFF_INV] >> Q.EXISTS_TAC ‘SWAP’ >> simp[]
 QED
 
 (* "<<=" is overloaded in listTheory, cardinalTheory and maybe others,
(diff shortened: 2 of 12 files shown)


src/probability/borelScript.sml
```diff
@@ -9293,6 +9293,207 @@ Proof
...
+(***********************)
+
+(*  Some of these do require addition simplifier manipulations.
+      I try to have the simplifier modified for as little as possible in order to make
+      it easier to relocate things to more appropriate places above.
+      - Jared Yeager                                                                   *)
+
+(*** IN_MEASURABLE_BOREL Theorems ***)
+
+Theorem IN_MEASURABLE_BOREL_CONG:
```
mn200, line 9307: Please make this an iff oriented from f to g. That way it satisfies the constraints to be a Definition congruence, used in termination analysis.

jdyeager: I'm happy to (especially since that form is redundant with _EQ'), but I'll need you to elaborate on what exactly that entails. I'm still very fuzzy on congruences. Would that just be:
∀a f g. (∀x. x ∈ space a ⇒ f x = g x) ⇒ (f ∈ Borel_measurable a ⇔ g ∈ Borel_measurable a)

mn200:
There's a description in the DESCRIPTION manual, but roughly yes.  The other thing would be to add a `a = b` precondition and to replace the `a` with the `b` on the RHS.

src/probability/borelScript.sml
binghe: Similar changes can be done also for `IN_MEASURABLE_CONG` in `sigma_algebraTheory`.

src/probability/measureScript.sml
```diff
@@ -5394,6 +5394,103 @@ val limsup_suminf_indicator_space = store_thm
...
+(*  These do not require addition simplifier manipulations on my part. It would
+      probably be more appropriate to add these in the proper places above.
+      - Jared Yeager                                                                   *)
+
+val _ = reveal "C";
+
+(*** measure_space Theorems ***)
+
+Theorem measure_space_cong:
+    ∀sp sts mu nu. measure_space (sp,sts,mu) ∧ (∀s. s ∈ sts ⇒ nu s = mu s) ⇒ measure_space (sp,sts,nu)
```
binghe, line 5410: How about `|- (∀s. s ∈ sts ⇒ mu s = nu s) ⇒ (measure_space (sp,sts,mu) <=> measure_space (sp,sts,nu))` ? In this way it may be directly used in a rewrite rule.

jdyeager: I'll actually probably go so far as the make the sp0 and sts0 on the LHS and sp1 and sts1 on the RHS (maybe '1' and '2' to match convention, maybe just adding prime on the RHS to match congruence rule convenion, still thinking about that).
Actually, I'll rename these rules with "_eq" or some such (because I undoubtedly use them irule), and then make proper congruence rules for measurable, measure_space, and Borel_measurable.

mn200: If to be used as a congruence in the simplifier, you will need to also have equalities for all the other sub terms (`sp` and `sts`).  If you omit those, the simplifier won't even look at those sub-terms.  (This can be a powerful and useful trick, but is unlikely to be what you want here.)

jdyeager: Indeed, hence my comment about renaming the _cong stuff to _eq (when I still want those results for use with irule) and then making actual congruence results with the _cong suffix. That should be in the most recent (4th) commit.

mn200:
Thanks for your work on this!

src/coretypes/pairScript.sml
```diff
@@ -158,6 +158,13 @@ val PAIR_FST_SND_EQ = store_thm(
 
 val SWAP_def = new_definition ("SWAP_def", ``SWAP a = (SND a, FST a)``)
 
+(* Theorem the SWAP invert itself *)
+Theorem SWAP_SWAP:
```
mn200, line 162: This should be `[simp]`

src/pred_set/src/pred_setScript.sml
```diff
@@ -4657,6 +4657,13 @@ val COMMUTING_ITSET_RECURSES = store_thm(
     ASM_SIMP_TAC bool_ss [COMMUTING_ITSET_INSERT, delete_non_element]
   ]);
 
+(* Corollary *)
+Theorem ITSET_SING:
```
mn200, line 4661: This should be `[simp]`

src/pred_set/src/pred_setScript.sml
```diff
@@ -6451,6 +6458,14 @@ val pair_to_num_inv = Q.store_thm ("pair_to_num_inv",
    (!x y. num_to_pair (pair_to_num (x, y)) = (x, y))`,
   SRW_TAC [][pair_to_num_def, num_to_pair_def]);
 
+(* More generally applicable version of the above *)
+Theorem pair_to_num_inv':
```
mn200, line 6462: This should be `[simp]`

src/probability/extrealScript.sml
```diff
@@ -9463,6 +9463,268 @@ Proof
...
+    simp[extreal_of_num_def]
+QED
+
+Theorem normal_minus1:
+    Normal (-1) = -1
+Proof
+    ‘Normal (-1) = -(Normal 1)’ suffices_by simp[normal_1] >> simp[extreal_ainv_def]
+QED
+
+Theorem extreal_le_simp:
```
mn200, line 9501: This should be a permanent `[simp]`.

src/probability/extrealScript.sml
```diff
@@ -9463,6 +9463,268 @@ Proof
...
+    (∀x y. Normal x ≤ Normal y ⇔ x ≤ y) ∧ (∀x. −∞ ≤ x ⇔ T) ∧ (∀x. x ≤ +∞ ⇔ T) ∧
+    (∀x. Normal x ≤ −∞ ⇔ F) ∧ (∀x. +∞ ≤ Normal x ⇔ F) ∧ (+∞ ≤ −∞ ⇔ F)
+Proof
+    rw[extreal_le_def] >> Cases_on ‘x’ >> simp[extreal_le_def]
+QED
+
+val _ = mk_local_simp "extreal_le_simp";
+
+Theorem extreal_lt_simp:
```
mn200, line 9510: Again, `[simp]` it.

src/probability/extrealScript.sml
```diff
@@ -9463,6 +9463,268 @@ Proof
...
+    (∀x y. Normal x < Normal y ⇔ x < y) ∧ (∀x. x < −∞ ⇔ F) ∧ (∀x. +∞ < x ⇔ F) ∧
+    (∀x. Normal x < +∞ ⇔ T) ∧ (∀x. −∞ < Normal x ⇔ T) ∧ (−∞ < +∞ ⇔ T)
+Proof
+    simp[extreal_lt_eq] >> rw[extreal_lt_def]
+QED
+
+val _ = mk_local_simp "extreal_lt_simp";
+
+Theorem extreal_0_simp:
```
mn200, line 9519: And this too.

src/probability/extrealScript.sml
```diff
@@ -9463,6 +9463,268 @@ Proof
...
+
+Theorem EXTREAL_PROD_IMAGE_MONO:
+    ∀f g s. FINITE s ∧ (∀x. x ∈ s ⇒ 0 ≤ f x ∧ f x ≤ g x) ⇒ ∏ f s ≤ ∏ g s
+Proof
+    NTAC 2 strip_tac >> simp[GSYM AND_IMP_INTRO] >> Induct_on ‘s’ >>
+    rw[EXTREAL_PROD_IMAGE_THM,DELETE_NON_ELEMENT_RWT] >> irule le_mul2 >>
+    simp[EXTREAL_PROD_IMAGE_POS]
+QED
+
+Theorem EXTREAL_PROD_IMAGE_COUNT_ZERO:
```
mn200, line 9651: `[simp]`

src/probability/extrealScript.sml
```diff
@@ -9463,6 +9463,268 @@ Proof
...
+    ∀f. ∏ f (count 0) = 1x
+Proof
+    simp[COUNT_ZERO,EXTREAL_PROD_IMAGE_EMPTY]
+QED
+
+Theorem EXTREAL_PROD_IMAGE_COUNT_ONE:
```
mn200, line 9657: `[simp]` (unless there is already a simp for `PI f {x}`; and if there isn't, why isn't there?)

src/probability/extrealScript.sml
```diff
@@ -9463,6 +9463,268 @@ Proof
...
+    rw[] >>
+    map_every (fn tms => qspecl_then tms assume_tac ext_suminf_2d_full)
+        [[‘λm n. f m n’,‘(λm. suminf (λn. f m n))’,‘num_to_pair’],
+        [‘λn m. f m n’,‘(λn. suminf (λm. f m n))’,‘SWAP ∘ num_to_pair’]] >>
+    rfs[BIJ_NUM_TO_PAIR,INST_TYPE [alpha |-> “:num”,beta |-> “:num”] BIJ_SWAP,BIJ_COMPOSE,SF SFY_ss] >>
+    NTAC 2 $ pop_assum $ SUBST1_TAC o SYM >> irule ext_suminf_eq >>
+    rw[o_DEF] >> Cases_on `num_to_pair n` >> simp[SWAP_def]
+QED
+
+Theorem exp_mono_le:
```
mn200, line 9698: `[simp]`

src/real/real_sigmaScript.sml
```diff
@@ -633,6 +633,39 @@ Proof
...
+(*   (generally for use in descendent theories)                              *)
+(* ------------------------------------------------------------------------- *)
+
+Definition REAL_PROD_IMAGE_DEF:
+    REAL_PROD_IMAGE f s = ITSET (λe acc. f e * acc) s (1:real)
+End
+
+val _ = overload_on ("∏", Term ‘REAL_PROD_IMAGE’);
+
+Theorem REAL_PROD_IMAGE_EMPTY:
```
mn200, line 647: `[simp]`.

binghe:
@mn200 can you remind me (and @jdyeager ) how to detect unicode violations? I believe there are many here.

mn200:
The easiest way is: when you build HOL (`bin/build`) pass the `-t` flag and the selftest that checks for these will run as part of the process.

src/coretypes/pairScript.sml
jdyeager: Done.
I'm always hesitant about making really simple lemmas like this simps. Basically, I tend to use them in one theorem or lemma later on, and so I figure the one proof I use them in doesn't justify adding the simplification rule to consideration for every other proof.
That said, if you say it should be a `[simp]`, I am happy to make it as such (since I trust your judgement a LOT more).

src/pred_set/src/pred_setScript.sml
jdyeager: Done.
Raises the question though: Should `ITSET_EMPTY` be a `[simp]`?

src/pred_set/src/pred_setScript.sml
jdyeager: Done.

src/probability/extrealScript.sml
jdyeager: I agree, about this and the following two. That said, I suspect they will break a number of existing proofs, so I'll tackle them last.
I'm guessing I can use something like diminish to just remove them from consideration for areas where they would cause problems, and augment them back in afterwards.

src/probability/extrealScript.sml
jdyeager: This raises questions about:
`ITSET_EMPTY` (again)
`COUNT_ZERO`
`REAL_PROD_IMAGE_EMPTY`
`EXTREAL_PROD_IMAGE_EMPTY`
`REAL_SUM_IMAGE_EMPTY` (which doesn't exist, I could do that quickly)
`EXTREAL_SUM_IMAGE_EMPTY`
`EXTREAL_SUM_IMAGE_COUNT_ZERO`

src/probability/extrealScript.sml
jdyeager: This raises questions about:
`COUNT_ONE`
`REAL_PROD_IMAGE_SING` (doesn't exist, I could make it)
`EXTREAL_PROD_IMAGE_SING` (the thing you ask about existing)
`REAL_SUM_IMAGE_SING`
`EXTREAL_SUM_IMAGE_SING`
`EXTREAL_SUM_IMAGE_COUNT_ONE`

As for why `EXTREAL_PROD_IMAGE_SING` isn't a `[simp]`, that calculus predates me. I suspect whoever initially proved it either didn't know about `[simp]`s, or was worried about making too many things `[simp]`s (although perhaps I am projecting).

That said, I'm happy to make whatever `[simp]`s. Incidentally, how do you do that with the old `store_thm()` syntax?

jdyeager:
Could someone give me a brief lesson about this unicode violation business? I ran with the `-t` flag, there are some violations. As best I can some combination of ⇒, ≤, maybe λ or ∑, and ⇔ are bad news. Should I just be replacing them with more ascii variants?

mn200:
Yes.  The policy is that Unicode other than the magic quotes and the lambda character shouldn't appear under `src`.

src/probability/extrealScript.sml
binghe: In old `store_thm()` syntax you can do the same: just put `[simp]` as part of the theorem name string (only with that string but SML variable name). The older way is to call `export_rewrites`, in `pred_setScript.sml` there should be such examples.

I personally tend to not declare too much `[simp]` for less common logical constants. Initially I consider `EXTREAL_PROD_IMAGE` is only needed in very few cases, and I should bother too much the automatic simplifier.  Furthermore, sometimes I meet the cases that a goal is simplified too much due to some later-added `[simp]` and as a solution I may remove some unwanted `[simp]`.

I suggest for now you can just follow what @mn200 suggests and get the PR merged ASAP. Later in proving other theorems it is always possible to adjust set `[simp]` properties (before next official release).

mn200:
There's an emacs mode command in HOL/Misc that replaces lots of Unicode with ASCII equivalents.

src/pred_set/src/pred_setScript.sml
mn200: Yes!

src/probability/extrealScript.sml
mn200: I'd hope that automatic rules for `<` and `<=` would just make everyone's life simpler. My annotations may not have been clear enough, but I certainly wasn't suggesting that the rewrites for numerals inside and outside `Normal` should be `simp`.

jdyeager: I absolutely agree. I just suspect that there are some proofs that produced subgoals that having these simps will now automatically solve. Or some assumption gets simplified and so a dependent drule or irule doesn't work right. I suspect something will break. So I wanted to get easier stuff addressed first.

That said, perhaps I will tackle these while I wait for judgement to be rendered on the various `(EXT)REAL_{SUM,PROD}_IMAGE_{EMPTY,SING,COUNT_{ZERO,ONE}}` theorems.

(Also, thankfully, I didn't misunderstand about the `Normal 0 = 0` type theorems. Otherwise, I would have commented about how many times I've used those backwards.)

src/probability/extrealScript.sml
jdyeager: I absolutely intend to follow Michael's suggestions, but I just also wanted to solicit his judgement on other similar theorems while I was at it.

But I also want to get the PR to an acceptable state ASAP. Especially since this is only the first batch of results. I wanted to do the Lebesgue stuff in a separate batch, because I suspected there would be some issues. I'll start working on isolating the stuff tomorrow.

jdyeager:
Funny story: I wrote a lousy perl script to do some of this (as an excuse to learn some perl) before I saw your HOL/Misc comment. But unicode is now sanitised.

mn200:
Heh.  I'm not sure I'd recommend learning Perl to anyone.  (You should have done it in SML :) )  But hey, add something else to your quiver of acquired CS skills.

src/probability/extrealScript.sml
jdyeager: Done

src/probability/extrealScript.sml
jdyeager: Can confirm, all of these break things. So, @mn200, what are the magical incantations to disable and re-enable simps? I'm not seeing anything obvious in simpLib.

mn200: Use `Excl "thmname"` as an argument in the list of theorems passed to `simp` and friends.   Or, to induce a more long-lasting change, use `val _ = temp_delsimps ["thmname1", "thmname2"]`.

jdyeager: The `temp_delsimps` is more what I was looking for. Anything to bring a simp back later on after that? (Basically allowing me to disable them for a number of proofs within the script.)

mn200: Yes (`augment_srw_ss`), but it's super gross to leave a script in a state where non-obvious chunks of it will behave differently because of theorems jumping in and out of the global simpset value.  If you're going to use `temp_delsimps` put it at the top of the file and don't undo it later on.

binghe: I think, if one has to prove something by excluding some theorems instead of applying some, this would be counter-intuitive to what humans usually do.

src/probability/extrealScript.sml
jdyeager: Done

src/probability/extrealScript.sml
jdyeager: Done

src/real/real_sigmaScript.sml
jdyeager: Done

jdyeager:
Some setup remains to be done for making other things simps.
Making EXTREAL_SUM_IMAGE_EMPTY a simp breaks things, I will try to address that.
Also added an extreal_1_simp to go with the extreal_0_simp.

mn200:
As discussed, I'll rebase and merge this now, and we can collectively fix any remaining problems.

mn200:
I just built HOL with this merged and there were no errors.  So #remaining-problems = 0, it appears.

mn200:
I take it back: SLLN_uncorrelated now fails to prove in `examples/probability`.  (My first build was only at `-t1`.)

binghe:
> I take it back: SLLN_uncorrelated now fails to prove in `examples/probability`. (My first build was only at `-t1`.)

Small problem. Tomorrow I will send a PR to fix this, together with more `[simp]` conversions.

Issue #999, opened 2022-02-18
vim-mode can no longer send contents of regular .sml files to HOL


oskarabrahamsson:
I used to be able to send the contents of any buffer of any filetype into the HOL4 session, not just proof scripts. This is useful when e.g. debugging library code.

I'm guessing this feature disappeared somewhere around 544d89da.

arolle:
If that is a desired feature, then the alternatives are to revert said commit or to change `*?Script.sml` to `*?.sml` in the [`filetype.vim`](https://github.com/HOL-Theorem-Prover/HOL/blob/b46e29d269344e1d2981f1c850969b4f6ff7a0ac/tools-poly/configure.sml#L519).

mn200:
It'd be good if the `vim` community could figure out what they want here.

mn200:
Closed by 8413d64

PR #1007, opened 2022-03-14
vimhol: enable vimhol for any `*.sml` file

arolle:
The changes of 544d89d only enable the vim keymappings for buffers of
`*Script.sml` files. It should also be possible to interact with the HOL
repl from other files, like `*Lib.sml`.

Fixes Github issue #999


mn200:
Thanks!

PR #1008, opened 2022-03-15
Add a theory defining interaction trees (itree)

myreen:
The added file defines a type for potentially infinite interaction trees. We take inspiration from the itree type of Xia et al.'s POPL'20 paper titled [Interaction Trees: Representing Recursive and Impure Programs in Coq](https://popl20.sigplan.org/details/POPL-2020-Research-Papers/14/Interaction-Trees-Representing-Recursive-and-Impure-Programs-in-Coq).

Interaction trees are interesting because they can both represent a program's observable I/O behaviour and also model of the I/O behaviour of the external world.

Our version of the type for interaction trees, `itree`, has the following constructors.  Here `Ret` ends an interaction tree with a return value; `Div` is internal silent divergence (an infinite run of `Tau`s); `Vis` is a visible event that returns a value that the rest of the interaction tree can depend on.

    ('a,'e,'r) itree  =
       Ret 'r                          -- termination with result 'r
    |  Div                             -- end in silent divergence
    |  Vis 'e ('a -> ('a,'e,'r) itree) -- visible event 'e with answer 'a,
                                          then continue based on answer

The POPL paper includes a silent Tau action:

    |  Tau (('a,'e,'r) itree)          --  a silent action, then continue

We omit `Tau` since it makes reasoning about intereaction trees messy. It causes a mess because one then has to deal with equality up to deletion of finite runs of `Tau` actions, ugh. We model an infinite run of `Tau`s using `Div`.

### DIFF
--- a/src/coalgebras/itreeScript.sml
+++ b/src/coalgebras/itreeScript.sml
@@ -1,0 +1,582 @@
+(*
+  This file defines a type for potentially infinite interaction
+  trees. We take inspiration from the itree type of Xia et al.'s
+  POPL'20 paper titled "Interaction Trees".
+
+  Interaction trees are interesting because they can both represent a
+  program's observable I/O behaviour and also model of the I/O
+  behaviour of the external world.
+
+  Our version of the type for interaction trees, itree, has the
+  following constructors.  Here Ret ends an interaction tree with a
+  return value; Div is internal silent divergence (an infinite run of
+  Taus); Vis is a visible event that returns a value that the rest of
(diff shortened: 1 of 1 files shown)


mn200:
Thanks!  It'd be good to have the type described in the DESCRIPTION manual section on all our built-in theories as well.

mn200:
The right place would be `Manual/Description/theories.stex`

mn200:
The reference from 21546e9 is a typo; that should have been a reference to #1088

PR #1013, opened 2022-03-28
Fix script compatibility with vanilla /bin/sh

binghe:
currently HOL generates shell scripts code which is not compatible with vanilla `/bin/sh`, i.e. the original Bourne Shell. (see also #1002 for some old discussions.)  It seems that the `$(...)` syntax is not supported. Instead, we can use backquotes instead. The changes here shouldn't break anything but will make HOL buildable in old Unix systems like Solaris 10.


mn200:
Thanks for coming back to this!

PR #1014, opened 2022-03-30
Fix broken proofs after changing several extreal lemmas [simp]

binghe:
Hi,

the main purpose of this PR is to fix those leftovers in #1006, where the following extreal lemmas should be put into automatic simplifier `[simp]`:
```
        EXTREAL_SUM_IMAGE_EMPTY
        extreal_le_simp(s)
        extreal_lt_simp(s)
        extreal_0_simp(s)
        extreal_1_simp(s)
```
but several largest proofs will be broken (once `[simp]` were added), either because some proof branches now disappeared, or tactics like `rw [GSYM extreal_le_eq]` now causes loops.  But I found all these broken proofs easy to fix, just I have to replay these long proofs to find the broken places.

The following example (retrieved from the proof of `borelTheory.Borel_def`) shows why automatically simplify `Normal x <= Normal y` to `x <= y` sometimes is not good. Consider the following proof state (note that here the goal is *above* the dash line):
```
        a ≤ real y
   ------------------------------------
    8.  y ≠ +∞
    9.  y ≠ −∞
   10.  Normal a ≤ y
```
The proof is the following steps
```
1.   REWRITE_TAC [GSYM extreal_le_eq] \\
2.   Suff ‘Normal (real y) = y’ >- ˙ \\
3.   MATCH_MP_TAC normal_real >> ASM_REWRITE_TAC []
```
First step, the goal is rewritten to `Normal a ≤ Normal (real y)`, then if one can prove that `Normal (real y) = y` (by `normal_real`, easy), the goal will match the assumption 10 and the proof completes.  At the beginning of the 2nd step, right after `hurdUtils.Suff` (= `Q_TAC SUFF_TAC`, we have:
```
        Normal (real y) = y ⇒ Normal a ≤ Normal (real y)
   ------------------------------------
    8.  y ≠ +∞
    9.  y ≠ −∞
   10.  Normal a ≤ y
```
Previously the proof uses `rw []` (instead of `RW_TAC std_ss []` now) to simply complete the proof, but now unfortunately `rw []` will first rewrite `Normal a ≤ Normal (real y)` back to the initial goal `a ≤ real y`, causing the broken proof.   I would say this is a *mistake* of previous (broken) proofs, because using `rw []` here, although literally very short, is actually a waste.

In the `extrealSimps` structure (newly added in #1006 but contents are disabled), I have re-enabled (and renamed)  `extreal_SS` which contains `["extreal_le_simps", "extreal_lt_simps", "extreal_0_simps", "extreal_1_simps"]`. I think we can add some selftests in the future. Currently it's not used anywhere but I think this structure can be a good place for other non-trivial extreal simplification or decision procedure in the future.   Note that the simplification of extreal expressions is harder than the case of reals, because of 1) unspecific arithmetics of infinities, 2) unspecific devision-by-zero. In particular, AC (`add_comm` and `add_assoc`) do not hold without further antecedents.  (I found `examples/pgcl/src/posrealTools.sml` a good start to work on.)

Besides, I have renamed a few theorems added in #1006, or renamed their quantifiers, or removed some unnecessary antecedents. In particular, the direct reason that `large_numberTheory` were broken currently, is that there are two `IN_MEASURABLE_BOREL_CONG` in `borelTheory`, i.e. the newly added one overrides the old one. Now I have changed the new one to `IN_MEASURABLE_BOREL_CONG'` (also with simplified statements).

## Other changes from ongoing work

In addition, some new properties about `finite_second_moments` are added in `probabilityTheory`, with supporting lemmas in dependent theories.

Also, previously the theorem `integrable_mul_indicator` has two many unnecessary antecedents. By using new proof methods, now `measure m s < +∞` and `(∀x. x ∈ m_space m ⇒ f x ≠ −∞ ∧ f x ≠ +∞)` are all unnecessary. 
```
    [integrable_mul_indicator]  Theorem      
      ⊢ ∀m s f.
          measure_space m ∧ s ∈ measurable_sets m ∧ measure m s < +∞ ∧
          (∀x. x ∈ m_space m ⇒ f x ≠ −∞ ∧ f x ≠ +∞) ∧ integrable m f ⇒
          integrable m (λx. f x * 𝟙 s x)
```
Since this theorem is only used in a few large proofs in examples. I directly removed these antecedents and then simplified the related proofs in `large_numberTheory`.

### DIFF
--- a/examples/probability/large_numberScript.sml
+++ b/examples/probability/large_numberScript.sml
@@ -18,6 +18,8 @@ open util_probTheory sigma_algebraTheory extrealTheory measureTheory
 
 val _ = new_theory "large_number";
 
+(* val _ = intLib.deprecate_int(); *)
+
 (* "In the formal construction of a course in the theory of probability, limit
     theorems appear as a kind of superstructure over elementary chapters, in
     which all problems have finite, purely arithmetical character. In reality,
@@ -2998,11 +3000,10 @@ val LLN_IID_shared_tactics =
      MATCH_MP_TAC bounded_imp_finite_second_moments >> art [] \\
      CONJ_TAC >- FULL_SIMP_TAC std_ss [real_random_variable_def] \\
      Q.EXISTS_TAC ‘&SUC n’ \\
-     NTAC 3 (POP_ASSUM K_TAC) \\
+     NTAC 3 (POP_ASSUM K_TAC) (* useless assumptions *) \\
     ‘Y = (\n. f n o X n)’
        by (rw [Abbr ‘Y’, Abbr ‘f’, truncated_def, o_DEF]) >> POP_ORW \\
-     rw [Abbr ‘f’]
-     >- (rw [abs_0, extreal_of_num_def, extreal_le_eq, extreal_abs_def]) \\
+     rw [Abbr ‘f’] \\
      MATCH_MP_TAC lt_imp_le \\
      FULL_SIMP_TAC std_ss [GSYM extreal_lt_def, extreal_of_num_def,
                            extreal_le_eq, extreal_abs_def])
@@ -3154,7 +3155,6 @@ Proof
          FULL_SIMP_TAC std_ss [prob_space_def, p_space_def, events_def,
                                prob_def, real_random_variable_def] \\
          HO_MATCH_MP_TAC integrable_mul_indicator >> art [] \\
-         CONJ_TAC >- METIS_TAC [abs_not_infty] \\
          MATCH_MP_TAC (REWRITE_RULE [o_DEF] integrable_abs) >> art []) >> DISCH_TAC \\
      Know ‘!i. i < k ==>
                expectation p (\x. Y i x pow 2) <= a n *
@@ -3175,7 +3175,6 @@ Proof
              HO_MATCH_MP_TAC integrable_mul_indicator \\
              FULL_SIMP_TAC std_ss [prob_space_def, p_space_def, events_def,
                                    prob_def, real_random_variable_def] \\
-             CONJ_TAC >- METIS_TAC [abs_not_infty] \\
              MATCH_MP_TAC (REWRITE_RULE [o_DEF] integrable_abs) >> art []) >> Rewr' \\
          MATCH_MP_TAC (REWRITE_RULE [GSYM expectation_def] integral_mono) >> simp [] \\
          STRONG_CONJ_TAC >- FULL_SIMP_TAC std_ss [prob_space_def] \\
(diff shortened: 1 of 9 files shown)


jdyeager:
My apologies about the `IN_MEASURABLE_BOREL_CONG` issue.



For the automatic simp issue, this does show an intrinsic issue with making some things simps is that different people go about things in different ways. So what is natural to one person may be odd to another. And the simps may favour one of those intuitions.

For example, for that specific instance you mentioned, I would have written a lemma `real_mono` or `real_le`:
```
Theorem real_mono:
    ∀x y. x ≠ +∞ ∧ x ≠ −∞ ∧ y ≠ +∞ ∧ y ≠ −∞ ∧ x ≤ y ⇒ real x ≤ real y
Proof
    Cases_on ‘x’ >> Cases_on ‘y’ >> simp[]
QED
```

And then solved
```
        a ≤ real y
   ------------------------------------
    8.  y ≠ +∞
    9.  y ≠ −∞
   10.  Normal a ≤ y
```
with the tactic `dxrule_at (Pos $ el 5) real_mono >> simp[]`. (Actually, on second glance, I'd just do `Cases_on ‘y’ >> fs[]`.)

Basically, to me it is more natural to wrap the assumption with `real` and have `real (Normal a)` become `a`, tending toward the reals. Whereas for you it seems it is more natural to wrap the goal with `Normal` and have `Normal (real y)` become `y`, tending toward the extended reals. Those simps favour tending toward the reals.



I assume from the comment about `integrable_mul_indicator`,
```
integrable_mul_indicator':
   ⊢ ∀m s f.
       measure_space m ∧ s ∈ measurable_sets m ∧ integrable m f ⇒
       integrable m (λx. f x * 𝟙 s x)
```
would now be very redundant, so I'll make sure to not add that.



(P.S. very cool to see someone else do goal-on-top printing.)

binghe:
@jdyeager thanks for your comments. Anyway I think it's generally positive to have these extreal simps.

As for `integrable_mul_indicator'`, I actually didn't notice it in your list. But yes, it should be added now, given `integrable_mul_indicator` is improved.  But note that this case is not the same with `integral_add'` (the one in your list)  whose proof is essentially the existing `integral_add` plus strictly more steps handling AE.  Even for education purposes I think we should keep both of them.

Note also that, in this PR, I added an `integral_add3` which currently has some unnecessary antecedents (like `(∀x. x ∈ m_space m ⇒ f x ≠ +∞ ∧ f x ≠ −∞)`):
```
[integral_add3]  Theorem      
      ⊢ ∀m f g h.
          measure_space m ∧ integrable m f ∧ integrable m g ∧
          integrable m h ∧ (∀x. x ∈ m_space m ⇒ f x ≠ +∞ ∧ f x ≠ −∞) ∧
          (∀x. x ∈ m_space m ⇒ g x ≠ +∞ ∧ g x ≠ −∞) ∧
          (∀x. x ∈ m_space m ⇒ h x ≠ +∞ ∧ h x ≠ −∞) ⇒
          ∫ m (λx. f x + g x + h x) = ∫ m f + ∫ m g + ∫ m h
```
With your `integral_add'` which I look forward for a new PR, I can also improve this `integral_add3`.  (I didn't prove `integral_add'` myself mostly because in all the high-level proofs I met so far, there isn't a case where the proof cannot go on with only the existing `integral_add`.  But since you have done a better one, I don't need to repeat your work.)

Chun

binghe:
On your comments about "favour tending toward the reals or extreals", my answer is that I don't have any favour.

My principle is that "whenever I can rewrite goals I don't rewrite assumptions", because in large proofs there's only one goal but may be 30-40 assumptions, and a single step `fs []` may cost visible time (1-2 seconds) to complete but then all these hard changes are immediately abandoned since the current (small) proof branch has completed.   Of course, one can always only rewrite the needed assumption(s), but then this requires tactics like `Q.PAT_X_ASSUM` to precisely match an assumption, resulting into longer proof texts.

binghe:
Also, I believe the issue of duplicated theorems can be resolved by `open Sanity` at the beginning of proof scripts. This structure has provided changed version of `save_thm` and `store_thm` which throws exceptions if the same theorem name has been already used.   I didn't use it in HOL's core proof files mostly because I think I'm careful enough.  I guess `Sanity` is mainly designed for end users.

mn200:
Thanks for this!  I begin to think that we might have the system use `Sanity`'s behaviour by default on redefinition using the same name, and to allow such redefinitions by either having `Globals.interactive` set to true (as one usually wants this behaviour interactively when work may be replayed repeatedly), or by setting that Sanity flag.

PR #1017, opened 2022-04-01
Added some stronger versions of existing integral and integrability results.

jdyeager:
This PR aims to provide versions of `integral_` and `integrable_` results for additions, finite summation, and subtraction with as few preconditions as possible. Specifically removing the preconditions that the functions involved be finite everywhere. Intuitively, these preconditions are unnecessary because these functions are already integrable, and thus already finite almost everywhere.

In order to help with this, and number of lemmas about almost everywhere are also proven in `borelTheory`. And of course, some random mathematical lemmas in `extrealTheory`.

Also, I renamed the extreal simpset fragment to `EXTREAL_ss`, so the naming convention is consistent with other simpset fragments.

Finally, this PR should not be excepted as-is, and I make it to solicit feedback. Before this PR is accepted, there is a single theorem in `measureTheory` that I think should perhaps be a [simp] (and was a [simp] in my personal work):
```
MEASURE_SPACE_SIGMA_ALGEBRA:
   ⊢ (∀m. measure_space m ⇒ sigma_algebra (measurable_space m)) ∧
     (∀sa mu. measure_space (space sa,subsets sa,mu) ⇒ sigma_algebra sa) ∧
     ∀sp sts mu. measure_space (sp,sts,mu) ⇒ sigma_algebra (sp,sts)
```
The first conjunct has frequently proved useful to me, but the later two are less common and would require `SF SFY_ss` for the simplifier to leverage them. So I wanted to get some judgement on if this should become a [simp] as-is, become something else that is a [simp], or not become a simp at all.

### DIFF
--- a/examples/probability/large_numberScript.sml
+++ b/examples/probability/large_numberScript.sml
@@ -2184,13 +2184,10 @@ Proof
      [ (* goal 1 (of 3) *)
        MATCH_MP_TAC IN_MEASURABLE_BOREL_CMUL >> BETA_TAC \\
        qexistsl_tac [‘\x. SIGMA (\i. W i x) (count (SUC k))’, ‘inv r’] \\
-       FULL_SIMP_TAC std_ss [prob_space_def, p_space_def, events_def] \\
-       CONJ_TAC >- FULL_SIMP_TAC std_ss [measure_space_def] \\
+       FULL_SIMP_TAC std_ss [prob_space_def, p_space_def, events_def] >> simp [] \\
        MATCH_MP_TAC (INST_TYPE [“:'b” |-> “:num”] IN_MEASURABLE_BOREL_SUM) >> rw [] \\
-       qexistsl_tac [‘W’, ‘count (SUC k)’] >> rw [FINITE_COUNT, IN_COUNT] >|
-       [ FULL_SIMP_TAC std_ss [measure_space_def],
-         FULL_SIMP_TAC std_ss [real_random_variable, p_space_def, events_def],
-         FULL_SIMP_TAC std_ss [real_random_variable, p_space_def, events_def] ],
+       qexistsl_tac [‘W’, ‘count (SUC k)’] >> rw [FINITE_COUNT, IN_COUNT] \\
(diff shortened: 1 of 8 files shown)


binghe:
@jdyeager thanks for your work. I can imagine that if this `MEASURE_SPACE_SIGMA_ALGEBRA` is `[simp]`, many of my existing proofs will be broken, simply because the proof branch for `sigma_algebra (measurable_space m)` will cease to exist. But this is only a short-term pain that can be accepted.

Have you confirmed that `examples/probability` and perhaps also `examples/miller` they are still buildable? If so, I think the PR can be indeed merged AS IS, unless you want to fix those broken proofs by yourself. Otherwise I will find some time to fix them by sending another PR where `MEASURE_SPACE_SIGMA_ALGEBRA` is changed to `[simp]`.

jdyeager:
Honestly, it was because of number of times I had to prove exactly that branch that I eventually snapped and made this simp.

I was able to do a build with the `-t2` flag, and it looks like everything in `examples` builds at the moment (making that theorem a `[simp]` will undoubtedly break some things there as well).

Naturally, I'm happy to have the merge done. But I do feel guilty off-loading the work of fixing things broken by a `[simp]` I proposed. And I feel that I can't just keep having other people do that.

jdyeager:
I should say that at this point `src` seems to build fine if I make `MEASURE_SPACE_SIGMA_ALGEBRA` a `[simp]`. But unsurprisingly, `examples` has issues. I'm working on those.

jdyeager:
It's building with the `-t2` flag with `MEASURE_SPACE_SIGMA_ALGEBRA` as a `[simp]`.

mn200:
Given the uselessness of the other conjuncts (modulo presence of `SFY_ss`), can you isolate the useful conjunct and make just that the `simp` please?

binghe:
I don't see any (obvious) remaining issue in the code changes of this PR, after the only useful part of `MEASURE_SPACE_SIGMA_ALGEBRA` has been kept while others are simply deleted (which didn't break any proof).

mn200:
Thank you both!

PR #1018, opened 2022-04-05
Bosslib shorthands

hrutvik:
As discussed offline, add the following shorthands:
- `qrefine` for `Q.REFINE_EXISTS_TAC`
- `qexists` and `qexistsl` for `qexists_tac` and `qexistsl_tac` respectively
- `SCONV ths tm` for `SIMP_CONV (srw_ss()) ths tm`

Also update the cheatsheet to match.

### DIFF
--- a/Manual/Cheatsheet/cheatsheet.md
+++ b/Manual/Cheatsheet/cheatsheet.md
@@ -331,12 +331,12 @@ In many cases, we may want to state exactly how the goal should be taken apart (
 : For a goal of the form `(A ==> B) ==> C`, splits into the two subgoals `A` and `B ==> C`.
   `impl_keep_tac` is a variant which keeps `A` as an assumption in the `B ==> C` subgoal.
 
-<code>qexists_tac &grave;<i>term</i>&grave;</code>
+<code>qexists &grave;<i>term</i>&grave;</code>
 : Instantiates a top-level `∃` quantifer with the supplied term.
 
-<code>Q.REFINE_EXISTS_TAC &grave;<i>term</i>&grave;</code>
+<code>qrefine &grave;<i>term</i>&grave;</code>
 : Refines a top-level `∃` quantifer using the supplied term - any free variables in the term become`∃`-quantified.
-  For example, for a goal `∃ n : num. if n = 0 then P n else Q n`, applying ``Q.REFINE_EXISTS_TAC `SUC k` >> simp[]`` produces the goal `∃ k : num. Q (SUC k)` (where `SUC` is the successor function).
+  For example, for a goal `∃ n : num. if n = 0 then P n else Q n`, applying ``qrefine `SUC k` >> simp[]`` produces the goal `∃ k : num. Q (SUC k)` (where `SUC` is the successor function).
 
 <code>goal_assum $ drule_at Any</code>
 : For a goal of the form `∃ vars . P1 /\ ... /\ Pn` (where the `vars` may be free in the `Pi`), attempts to match the `Pi` against the assumptions.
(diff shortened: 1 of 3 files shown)


binghe:
`qexists_tac` and `qexistsl_tac` are not short enough? I think tactics (which is identified by their type) should always have `_tac` suffix, to distinguish from tools of forward proofs.

binghe:
If you really want something short, I suggest we follow some "standards", i.e. HOL Mizar mode [1]

assume: `DISCH TAC`
let: `X_GEN_TAC`
take: `Q.EXISTS_TAC`
consider: `Q.X_CHOOSE_TAC`
given: `DISCH_THEN o Q.X_CHOOSE TAC`
suffices_to_show: `MATCH_MP_TAC`
set: `Q.ABBREV_TAC`

[1] Harrison, John. "A Mizar mode for HOL." International Conference on Theorem Proving in Higher Order Logics. Springer, Berlin, Heidelberg, 1996

mn200:
We already have `simp`, `rw` and others that have lost their `_tac` suffix so I'm not averse to the idea of others losing theirs. A Mizar mode style vocabulary would be good, but those particular choices aren't all so great in our setting.  We can't use `let` because it's a reserved word in SML; `suffices_to_show` is too similar to `suffices_by`, and the latter is better in that you quote what it is that suffices; `take` overlaps with the standard function in `List`; and `assume` overlaps with the primitive rule of inference `ASSUME`.  (The reason this all works in the Mizar mode is because that language is interpreted by machinery outside of SML; our tactic language is entirely _within_ SML.)

binghe:
> We already have `simp`, `rw` and others that have lost their `_tac` suffix so I'm not averse to the idea of others losing theirs. A Mizar mode style vocabulary would be good, but those particular choices aren't all so great in our setting. We can't use `let` because it's a reserved word in SML; `suffices_to_show` is too similar to `suffices_by`, and the latter is better in that you quote what it is that suffices; `take` overlaps with the standard function in `List`; and `assume` overlaps with the primitive rule of inference `ASSUME`. (The reason this all works in the Mizar mode is because that language is interpreted by machinery outside of SML; our tactic language is entirely _within_ SML.)

`simp` and `rw`, etc. are frequently used simplifiers, which has complicated underlying code. Thus I think they do not account for (simple) tactics like others.   I copied that map of (from english words to tactics) directly from Harrison's paper but I didn't mean to use them directly. (How can I not know `let` is reserved in SML?)  In general, I think Harrison has provided good hints on which English words to use when you want to have more human-readable tactical proofs.  Most of the overlaps will not hold if the following first-capital names were used instead: `Assume`, `Let`, `Take`, `Consider`, `Given`, ...  but once again, I don't mean to really put them into `bossLib`.

Moving from `qexists_tac` to `qexists` means nothing in my opinion, because `qexists` is not even a word and it doesn't help at all for new users to quickly memorize, while only creates another synonym when old users trying to understand others' proofs.

mn200:
Thanks for these.  I agree that it's not ideal to have proliferating synonyms for the same basic tactics, but dropping `_tac` seems worth it given the number of key-strokes and amount of space consumed on the screen.  

Random, somewhat unserious, thought: perhaps `qexists_tac` should just be the (valid SML) identifier `?`.  Similarly, `qx_gen_tac` could be `!`.  (This conflicts with SML's dereferencing operator; but I'm pretty confident scripts have far more `qx_gen_tac` occurrences than they do references.)  And the versions that take lists would add `*` characters.  

The result becomes line noise tactics:  `?* [’a’, ’b + 2’] >> simp[foo]`

PR #1016, opened 2022-03-31
Fix OpenTheory builds due to DefnBase changes

binghe:
Hi,

this PR fixes OpenTheory (OT) builds when proving base assumptions (`src/boss/prove_base_assumsTheory`) due to core theory changes, after DefnBase changes (fa6ea5c7). (Last OT fixes was #958.)

Let's recall how the OpenTheory package `hol-base` is built (in `src/boss`), mainly for education purposes:

1. an OT article `hol4-base-unint.art` is generated by combining a list of HOL core theories (defined by `ARTS` variable in `src/boss/Holmakefile`).
2. an OT article `hol4-assums.art` is generated by extracting only assumptions from `hol4-base-unint.art`. This process is deterministic but we have no control on the contents and order of these assumptions. Any changes in HOL core theory may cause minor changes in these assumptions. Currently there are 109 of them.
3. an OT article `base-theorems.art` is generated from OpenTheory's own base package (currently it's `base-1.221`).
4. an OT arcile `bool_defs.art` is generated from `bool_defsScript.sml`, which provides the "definition" part of the final `hol4-base.art` (step 6).
5. an OT article `prove_base_assums.art` is generated by playing proofs in `prove_base_assumsScript.sml`, which provides the "proofs" part of the final `hol4-base.art`.
6. finally, `hol4-base.art` is generated by combining `bool_defs.art` and `prove_base_assums.art`.

The following changes happened in base assumptions extracted in the above step 2: (this is for sure caused by DefnBase changes, see the first comment)

- The previous assumption No.29 (`|- Data_Option_none <> Data_Option_some x`) disappeared, causing assumptions No.30-51 shifted to No.29-50.

- A new assumption appeared at No.51 but it looks like a combination of assumptions No.102 `Data_Sum_right x = Data_Sum_right y <=> x = y` and 103 `Data_Sum_left x = Data_Sum_left y <=> x = y` with quantifiers: (NOTE: `Data_Sum_isLeft` = `ISL`, `Data_Sum_left` = `INL`, etc.)
```
|- (!x x'. Data_Sum_left x = Data_Sum_left x' <=> x = x') /\
      !y y'. Data_Sum_right y = Data_Sum_right y' <=> y = y'
```
- Assumptions No.62-65 has changed their orders and also slightly changed statements. For instance, the old assumption
```
|- (!x. Data_Sum_isLeft (Data_Sum_left x)) /\
      !y. ~Data_Sum_isLeft (Data_Sum_right y)
```
has now become
```
[...]

PR #938, opened 2021-07-28
The Strong Law of Large Numbers for IID random variables

binghe:
Hi,

recall in PR #910, the Weak Law of Large Numbers (WLLN) for IID random variables were proven in `examples/probability/large_numberScript.sml`, as `WLLN_IID`. In this PR I *finally* completed the Strong Law of Large Numbers (SLLN) for IID random variables (but only for the convergence part, there is also a divergence part of this theorem):
```
    [SLLN_IID]
      ⊢ ∀p X.
          prob_space p ∧ (∀n. real_random_variable (X n) p) ∧
          pairwise_indep_vars p X (λn. Borel) 𝕌(:num) ∧
          identical_distribution p X Borel 𝕌(:num) ∧ integrable p (X 0) ⇒
          LLN p X almost_everywhere
```

Applying `LLN_alt_converge_AE_IID` one can get a more useful form without the constant `LLN`:
```
> SIMP_RULE std_ss [LLN_alt_converge_AE_IID] SLLN_IID;
val it =
   ⊢ ∀p X.
       prob_space p ∧ (∀n. real_random_variable (X n) p) ∧
       pairwise_indep_vars p X (λn. Borel) 𝕌(:num) ∧
       identical_distribution p X Borel 𝕌(:num) ∧ integrable p (X 0) ⇒
       ((λn x. ∑ (λi. X i x) (count (SUC n)) / &SUC n) ⟶
        (λx. expectation p (X 0))) (almost_everywhere p): thm
```

The proof essentially follows E. Etemadi's 1981 elementary proof [1], based on truncated sequences and elementary variance analysis.  (There are at least four different proofs of `SLLN_IID`, the one found in Isabelle AFP is an application of the Ergodic theory.) There are some big gaps near the end of the textbook proof (not only by the original paper author but also by all subsequent copiers in two textbooks [2,3] with this proof), and I had to think as a mathematician to fix it.

Note that convergence almost everywhere implies convergence in probability, thus in theory the above theorem directly implies `WLLN_IID`, whose long proofs seem unnecessary now. But I think the proof of `WLLN_IID` is still useful for education purposes. More importantly, the proofs of `SLLN_IID` and `WLLN_IID` share the same device and even a large piece of tactics, defined by a ML variable).

As usual, there are many related new dependent lemmas proven together, e.g. more lemmas about almost everywhere convergence, e.g.: (new lemmas about `NUM_FLOOR` or `flr`, needed here, were already described in PR #933)
```
[...]

PR #910, opened 2021-04-21
The Weak Law of Large Numbers for IID random variables

binghe:
Hi,

recall in previous PR #786 the Weak Law of Large Numbers (LLN) for uncorrelated random variables (r.v.'s) was added, with AE/PR/LP convergence concepts and properties. The present PR continuous the work and finally formalized the Weak Law of Larger Numbers for pairwise independent random variables with identical distributions (`large_numberTheory.WLLN_IID`), also called "I.I.D" but this abbreviation usually denotes totally (or mutually) independence (which implies pairwise independence). This proof is somehow difficult, because the involved r.v.'s are not required to have finite second moments (variances). (The core formal proof follows Theorem 5.2.2 of [2, p.114], originally due to Khintchine.)

Due to the increasing size (and compilation time) of `probabilityScript.sml`, I have added a new script `large_numberScript.sml` (initial size: **4022** lines) for all versions of LLN theorems so far (and perhaps other limit theorems in the future.) Some code are moved here from `probabilityTheory`, nostly with updated proofs and statements.

Now all LLN theorems share the same conclusion based on `LLN`: (where the uses of `SUC n` instead of `n` is unnecessary but can prevent division-by-zero when being used to prove further results.)
```
   [LLN_def]  Definition      
      ⊢ ∀p X convergence_mode.
          LLN p X convergence_mode ⇔
          (let
             Z n x = ∑ (λi. X i x) (count n)
           in
             ((λn x. (Z (SUC n) x − expectation p (Z (SUC n))) / &SUC n) -->
              (λx. 0)) (convergence_mode p))
```

The advantage of the above definition are twofold:

  - The reduced term size and reduced number of quantifiers in LLN theorems which now have unified conclusions;
  - There are different alternative definitions (as theorems) of `LLN` for different conditions of LLN theorems.

For instance, in the IID version of LLN theorems, since all involved r.v.'s have the same expectations, the following simplier equivalent definitions of `LLN` (without using `let`) are provided for end users:
```
   [LLN_alt_converge_AE_IID]  Theorem      
      ⊢ ∀p X.
          prob_space p ∧ (∀n. real_random_variable (X n) p) ∧
          identical_distribution p X Borel 𝕌(:num) ∧ integrable p (X 0) ⇒
          (LLN p X almost_everywhere ⇔
           ((λn x. ∑ (λi. X i x) (count (SUC n)) / &SUC n) -->
            (λx. expectation p (X 0))) (almost_everywhere p))

[...]

PR #1022, opened 2022-04-08
Cumulative fixes of HOL4 OpenTheory builds (--otknl)

binghe:
Hi,

HOL's OpenTheory support, at least the "export" part, represents a *hope* that the hard work (i.e. formal proofs) by HOL4 users can be exported into an intermediate format (OT articles) which can be either verified by third-party tools or translated into theorems of other theorem provers (not limited to HOL famility, in theory). The OT builds (`--otknl`) have been broken for quite some months, although previous fixes (#702, #803, #942, #958) were never complete.

The present PR throughly fixed all the remaining (well, not true: there are still some remaining issues mentioned in the comments) and accumulated issues (due to core theory changes) of HOL4's OpenTheory exporting support (NOTE: exporting only,  there are still issues with the OT importing support.)

As a result, in "examples/probability" I created a sample OT theory file to export everything from OpenTheory's own `base-1.221` (Boolean logic and other basic things) and `real-1.67` (construction of real numbers, originally from HOLLight, I think), almost all HOL4's core math theories, until my proofs of Law of Large Numbers of I.I.D. random variables in Probability Theory (#910, #938, #956). The resulting OT article has 856MB, can be downloaded [here](https://www.dropbox.com/s/uv0fxtdw5oyprvx/hol4-large-numbers.7z?dl=0).

To build HOL4 with `--otknl`, it will take about 6 hours on a very good (laptop) computers with 64GB memory, and at most `-j 2` should be used when calling `bin/build`.   (A simple math: each building process takes about 25GB peak memory, and `-j 3` will take 75GB and blow up your computer's memories and swap spaces.)  For a computer with 32GB memory, the total time should be doubled to 12 hours (with `-j 1` build option).

Another 6-7 hours are needed for generating the the sample OT article in "examples/probability" mentioned above. And this also requires manually installation of all the related OT packages by executing `Holmake -j1 install` in their directory, e.g. for `hol4-base.thy`, `hol4-real.thy`, etc.  The core building process has been modified by NOT installing any OT packages during the build, which saves a lot of time and prevent subtle dependency issues.

OpenTheory should be rebuilt by Mlton by adding `MLTON_OPTS += -default-type intinf` into its `Makefile`. But this is only for building the huge (sample) OT articles in `examples/probability`, not required for core theory builds.   Also note that OpenTheory built by PolyML usually has double memory consumptions, thus if you insist on using it, only `-j 1` is feasible even on computers with 64GB memory!

The code changes are many, across 63 files, and this is a very precise work. It's precise because the leaking of assumptions (which then become an unexpected axiom in the exported OT articles) can easily happen if there's any mistake made in the code. In fact, there were many such issues, and identifying them is not always easy. Below I will describe the code changes in details, classifying them in groups, each in form of a comment of this PR.

P. S. Ramana Kumar @xrchz has provided some helps in understanding some SML code. I believe some remaining issues here can only be fixed by him as the original author of many related code.

Chun Tian

### DIFF
--- a/examples/probability/.gitignore
+++ b/examples/probability/.gitignore
@@ -1,3 +1,4 @@
 *Theory.html
 *Theory.lst
 /heap
+*.html
--- a/examples/probability/Holmakefile
+++ b/examples/probability/Holmakefile
@@ -4,7 +4,9 @@ INCLUDES = $(HOLDIR)/src/probability $(HOLDIR)/src/n-bit $(HOLDIR)/src/real \
 EXTRA_CLEANS = heap \
 	$(patsubst %Theory.uo,%Theory.html,$(DEFAULT_TARGETS)) \
 	$(patsubst %Theory.uo,%Theory.lst,$(DEFAULT_TARGETS)) \
-	hol4-large-numbers.art hol4-large-numbers.html hol4-large-numbers.log
+	hol4-large-numbers.art  hol4-large-numbers-only.art \
+	hol4-large-numbers.html hol4-large-numbers-only.html \
+	hol4-large-numbers.log  hol4-large-numbers-only.log
 
 ifdef POLY
 HOLHEAP = heap
@@ -54,8 +56,25 @@ hol4-large-numbers.art: hol4-large-numbers.thy large_number.ot.art \
 			$(HOLDIR)/src/opentheory/hol4.int
 	opentheory info --article -o $@ $<
 
-# This optional step takes about ~52GB memory but (only) two hours to complete.
+# With the rebuilt OpenTheory/Mlton (-default-type intinf), this step only takes
+# 2 minutes (otherwise 2 hours with OpenTheory/PolyML)
+
+# "&> file (aka >& file) is not part of the official POSIX shell spec, but has been
+# added to many Bourne shells as a convenience extension (it originally comes from
+# csh). In a portable shell script (and if you don't need portability, why are you
+# writing a shell script?), use > file 2>&1 only." (from stackoverflow)
 hol4-large-numbers.html: hol4-large-numbers.art
-	opentheory info --inference --document -o $@ $<
+	opentheory info --inference --document -o $@ $< \
+		> hol4-large-numbers.log 2>&1
+
+# A smaller version containing only theorems useful for proving LLN (and LLN itself)
+hol4-large-numbers-only.art: hol4-large-numbers-only.thy large_number.ot.art \
+			$(HOLDIR)/src/opentheory/hol4.int
+	opentheory info --article -o $@ $<
+
+# TODO: how to show ALL theorems in this article?
+hol4-large-numbers-only.html: hol4-large-numbers-only.art
+	opentheory info --inference --document -o $@ $< \
+		> hol4-large-numbers-only.log 2>&1
 
 endif
--- a/examples/probability/Holmakefile
+++ b/examples/probability/Holmakefile
@@ -3,7 +3,8 @@ INCLUDES = $(HOLDIR)/src/probability $(HOLDIR)/src/n-bit $(HOLDIR)/src/real \
 
 EXTRA_CLEANS = heap \
 	$(patsubst %Theory.uo,%Theory.html,$(DEFAULT_TARGETS)) \
-	$(patsubst %Theory.uo,%Theory.lst,$(DEFAULT_TARGETS))
+	$(patsubst %Theory.uo,%Theory.lst,$(DEFAULT_TARGETS)) \
+	hol4-large-numbers.art hol4-large-numbers.html hol4-large-numbers.log
 
 ifdef POLY
 HOLHEAP = heap
@@ -11,7 +12,6 @@ OBJS = real/realLib n-bit/fcpLib res_quan/src/hurdUtils probability/probabilityT
 FULL_OBJPATHS = $(patsubst %,$(HOLDIR)/src/%.uo,$(OBJS)) \
                 $(HOLDIR)/sigobj/posetTheory.uo
 HEAP0 = $(HOLDIR)/src/probability/heap
-
 
 all: $(HOLHEAP)
 
@@ -22,3 +22,40 @@ endif
 all: $(DEFAULT_TARGETS)
 
 .PHONY: all
+
+ifeq ($(KERNELID),otknl)
+ARTFILES = $(patsubst %Script.sml,%.ot.art,$(wildcard *Script.sml))
+
+all: $(ARTFILES)
+
+# NOTE: the following targets are not included in "all". OpenTheory binary built
+# by Poly/ML (with Int.maxInt = NONE, i.e., --enable-intinf-as-int) must be used
+# here, since OpenTheory built by the Mlton (otherwise recommended!) will report
+# (integer?) "Overflow".  With PolyML built OT, this step takes (on 2019 MacBook
+# Pro) about 6 hours with 8 CPU cores (the total CPU time is ~8 hours) and ~52GB
+# memory to complete.  -- Chun Tian, Apr 8, 2022
+#
+# To easily switch between OpenTheory built by different SML platforms, below is
+# my recommended softlinks in my $(HOME)/bin:
+#
+# opentheory -> opentheory.mlton
+# opentheory.mlton -> ../ML/opentheory/bin/mlton/opentheory
+# opentheory.mosml -> ../ML/opentheory/bin/mosml/opentheory
+# opentheory.polyml -> ../ML/opentheory/bin/polyml/opentheory
+#
+# UPDATE: another (better) way is to rebuild OpenTheory with MLton by adding the
+# following MLton command-line option:
+#
+#   MLTON_OPTS += -default-type intinf
+#
+# into OpenTheory's Makefile (right after the line with "MLTON_OPTS = ..."), and
+# this may enable computers with 32GB memory to finish this build.
+hol4-large-numbers.art: hol4-large-numbers.thy large_number.ot.art \
+			$(HOLDIR)/src/opentheory/hol4.int
+	opentheory info --article -o $@ $<
+
+# This optional step takes about ~52GB memory but (only) two hours to complete.
+hol4-large-numbers.html: hol4-large-numbers.art
+	opentheory info --inference --document -o $@ $<
+
+endif
--- a/examples/probability/hol4-large-numbers-only.thy
+++ b/examples/probability/hol4-large-numbers-only.thy
@@ -1,0 +1,115 @@
+name: hol-large-numbers-only
+version: 1.0
+description: HOL and OT theories towards to the Law of Large Numbers
+author: Chun Tian <binghe.lisp@gmail.com>
+license: MIT
+show: "HOL4"
+show: "Data.Bool"
+show: "Data.Option"
+show: "Data.Unit"
+show: "Data.Sum"
+show: "Data.Pair"
+show: "Data.List"
+show: "Function"
+show: "Relation"
+show: "Number.Natural"
+show: "Number.Real"
+main {
+  import: base
+  import: hol-base
+  import: hol-set
+  import: hol-real
+  import: hol-extreal
+  import: hol-analysis
+  import: hol-probability
+  article: "large_number.ot.art"
+  interpretation: "../../src/opentheory/hol4.int"
+}
+base {
+  package: base-1.221
+}
+hol-base {
+  import: base
+  package: hol-base-1.3
+}
+hol-sort {
+  import: base
+  import: hol-base
+  package: hol-sort-1.1
+}
+hol-string {
+  import: base
+  import: hol-base
+  import: hol-sort
+  package: hol-string-1.2
+}
+hol-words {
+  import: base
+  import: hol-base
+  import: hol-string
+  package: hol-words-1.3
+}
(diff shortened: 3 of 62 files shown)


binghe:
# Changes to OT theory definitions

One goal here is to make sure that all HOL theories needed for proving Probability are exported as OT articles. Historically `ordinalTheory` and `cardinalTheory` (i.e. the whole "pred_set/src/more_theories" folder) are never touched by OT exporting, but now they are even needed by theories in `src/real` (`metricTheory` depends on `topologyTheory` which now depends on `cardinalTheory`. So I create the `hol-set` OT package as "HOL4's set and topology theories".

Other OT packages are either created (`hol-bag`, e.g.), or modified for their previous missing dependencies. Note that, missing OT dependencies are very serious, they directly cause leaking assumptions.  For the most important OT package, `hol-base` (in "src/boss"), now I completely rewrite its dependency map (in "src/boss/hol4-base-unint.thy") by following the Theory Graph generated as part of the help documents.   One result of correct dependencies is that, many proofs of assumptions are not needed in "src/boss/prove_base_assumsScript.sml", which is vulnerable and sensitive to core theory changes.

I also updated the versions of HOL's OT packages, basically to make their version exactly one minor version higher than those in the OpenTheory repo, so that there will be no confusing on versioning when building the sample OT article in `examples/probability`.  (Currently it seems impossible to upload new OT packages to OpenTheory official repo, but this is not a big issue, only meaning OT's native packages have stopped their developments.)

To install these OT packages after a full build, one should go into each relevant directory and execute `Holmake -j1 install`, e.g. in "src/boss".  OT packages have dependencies, thus they must be installed with certain order, which can be easily found out by reading `opentheory` command's error messages.

After installation, the following OT packages can be found in your local "$HOME/.opentheory/packages":

```
hol-analysis-1.1
hol-bag-1.0
hol-base-1.3
hol-coalgebras-1.1
hol-extreal-1.0
hol-finite-maps-1.0
hol-floating-point-1.2
hol-integer-1.2
hol-probability-1.1
hol-quotient-1.2
hol-real-1.2
hol-res-quan-1.2
hol-ring-1.2
hol-set-1.0
hol-sort-1.1
hol-string-1.2
hol-words-1.3
```

binghe:
# Changes in "pred_set/src/more_theories" (in particular, `ordinalTheory`)

It's known that some HOL theories cannot be correctly exported to OT articles (one is "src/coalgebras/lbtreeScript.sml", no idea why). It's not clear if the issues is at OpenTheory side, or the OT exporting code inside HOL.  In `ordinalScript.sml`, luckly I found that the following definitions caused the exporting issue:

```
val ordDIVISION = new_specification(
   "ordDIVISION", ["ordDIV", "ordMOD"],
   SIMP_RULE (srw_ss()) [SKOLEM_THM, GSYM RIGHT_EXISTS_IMP_THM] ordDIVISION0)
```
I conjecture that OT exporting code (or OT itself) has issues in supporting more than one new logical constants created by `new_specification()`.   And I don't see such a way happens in any other HOL core theory.  But anyway, there's a obvious workaround: just create one logical constant instead:

```
Theorem ordDIVISION1[local] :
     !a b:'a ordinal. 0 < b ==> ?c. a = b * FST c + SND c /\ SND c < b
 Proof
     ...
 QED

 (* The next 3 theorems are skipped in ordinal.otd *)
 val ordDIVMOD = new_specification(
    "ordDIVMOD", ["ordDIVMOD"],
     SIMP_RULE (srw_ss()) [SKOLEM_THM, GSYM RIGHT_EXISTS_IMP_THM] ordDIVISION1);

 Definition ordDIV :
     ordDIV a b = FST (ordDIVMOD a b)
 End

 Definition ordMOD :
     ordMOD a b = SND (ordDIVMOD a b)
 End

 (* |- !a b. 0 < b ==> a = b * ordDIV a b + ordMOD a b /\ ordMOD a b < b *)
 Theorem ordDIVISION =
     REWRITE_RULE [GSYM ordDIV, GSYM ordMOD] ordDIVMOD
```

Eventually the theorem `ordDIVISION` has exactly the same statements as before, and it is this theorem who captures the full characteristics of `ordDIV` and `ordMOD`, thus didn't break any future proof.   I suggest we keep this solution until we fixed OpenTheory.

In some other scripts (`cardinal` and `wellorder`) in the folder I explicitly opened `relationTheory` and removed many prefixes of relation theorems in the proofs.  These changes do not fix anything.  This was due to an attempt of fixing a leaked assumptions, but eventually the root cause was not here.  I kept these changes because I think it's in general a direction to explicit open all dependent theories at the beginning of a proof script.

binghe:
# Changes in prove_base_assumsScript.sml

(Some backgrounds of this file has been described in previous closed PR #1016.)

It is painful to maintain this file, because it contains the (bridging) proofs between OT `base` package and HOL's `hol-base`, for theorems defined in HOL's `boolTheory` but not at OT side.  Core theory changes (or changes of theori dependency) may easily cause changes or shifts of those "assumptions" (named by `th` + an increasing number).  In theory, each theorem in HOL's `boolTheory` but not in OT's `bool` or `base` package should have a clone in this file. But unfortunately, so far we never try to make an exhausting list of these extra Boolean theorems. The current list is based on all exported OT packages, by watching if they report any missing assumptions on installation.  Thus, if, in the future, a new theorem were added in HOL while it used a `boolTheory` theorem which was never used by other proofs, the correspond OT package will be broken again by leaking that `boolTheory` theorem into "unsatisfied assumptions".   (I plan to systematically resolve this issue in the future.)   Not to mention, if `boolTheory` itself were changed, even in the order of theorems in it, `prove_base_assumsScript.sml` will be broken too.

## Newly added assumptions are: (reported as "unsatisfied assumptions" by OT packages beyond `hol-base`)
- `|- (y ==> x) ==> ~x ==> ~y` ("MONO_NOT")
- `|- y ==> x <=> ~x ==> ~y` ("MONO_NOT_EQ")
- `|- !b t1 t2. (if b then t1 else t2) <=> (b ==> t1) /\ (~b ==> t2)` ("COND_EXPAND_IMP", forward proof copied from `boolScript.sml`)
- `|- !P P' Q Q'. (~Q ==> (P <=> P')) /\ (~P' ==> (Q <=> Q')) ==>
                   (P \/ Q <=> P' \/ Q')` ("OR_CONG", forward proof copied from `boolScript.sml`)
- `|- (?x::P. F) <=> F` ("RES_EXISTS_FALSE")
- `|- (!x::P. T) <=> T` ("RES_FORALL_TRUE", the above two were reported by `hol-res-quan`)

## Removed assumptions: (most of them are theorems in `satTheory` due to wrong dependencies)
- `|- !A. A ==> ~A ==> F` (previous "th42")
- `|- ~(A \/ B) ==> F <=> (A ==> F) ==> ~B ==> F` (previous "th93")
- `|- ~(~A \/ B) ==> F <=> A ==> ~B ==> F` (previous "th94")
- `|- (p <=> if q then r else s) <=>
       (p \/ q \/ ~s) /\ (p \/ ~r \/ ~q) /\ (p \/ ~r \/ ~s) /\
       (~q \/ r \/ ~p) /\ (q \/ s \/ ~p)` (previous "th95")
- `|- (p <=> (q <=> r)) <=>
       (p \/ q \/ r) /\ (p \/ ~r \/ ~q) /\ (q \/ ~r \/ ~p) /\ (r \/ ~q \/ ~p)` (previous "th96")
- `|- (p <=> q /\ r) <=> (p \/ ~q \/ ~r) /\ (q \/ ~p) /\ (r \/ ~p) *)` (previous "th97")
- `|- (p <=> q \/ r) <=> (p \/ ~q) /\ (p \/ ~r) /\ (q \/ r \/ ~p)` (previous "th98")
- `|- (p <=> q ==> r) <=> (p \/ q) /\ (p \/ ~r) /\ (~q \/ r \/ ~p)` (previous "th99")
- `|- (p <=> ~q) <=> (p \/ q) /\ (~q \/ ~p)` (previous "th100")
- `|- !x. Data_Option_none <> Data_Option_some x` (previous "th29", this is due to the fix in "src/option/option.otd" by removing the line `delproof NOT_NONE_SOME`, which is wrong because OT's `option` package didn't provide this theorem at all. The other one, `NOT_SOME_NONE`, is indeed provided.)

## Fixed assumptons:
- `|- !(e :'b). ?(f :'a prove_base_assums$itself -> 'b).
         f (the_value :'a prove_base_assums$itself) = e` ("itself_Axiom", previously the type variable of `e` was `:'a` and this has caused leaking assumptions reported by `hol-words`, `hol-floating-point`, etc.

## New theorems in `optionTheory`
- `|- option_CASE x v f <=> (x = NONE ==> v) /\ !x'. x = SOME x' ==> f x'` ("option_imp_elim", it was in `cardinalTheory`, now moved here.  This doesn't fix anything, only due to the fixing of relevant issues.)
- `|- (option_CASE (x:'a option) v f = v') <=>
    ((x = NONE) /\ (v = v') \/ ?a. (x = SOME a) /\ (f a = v'))` ("option_case_eq'"), hol-set reports it as unsatisfied assumption but I found everywhere and couldn't find this theorem in HOL core theories.   Literally it's like the existing theorem "option_case_eq" but with different names of variables.   By simply adding this theorem (as a workaround), now `hol-set` is happy.

binghe:
## Remaining issues

As last, the following remaining issues should be investigated in the future:

1. `lbtree` in "src/corealgebras" still cannot be exported as an OT theory. So far I didn't try to investigate which definition (or theorem, but unlikely) has caused the issue.
2. `ordinalTheory` now contain workarounds for the definition of `ordDIV` and `ordMOD` as described already. If OpenTheory (or the exporting code in HOL) gets fixed in the future, my current workaround should be reverted.
3. The newly added theorem `option_case_eq'` in `optionTheory` must have been already proved somewhere but I just cannot find it in HOL's core theory proofs. Perhaps it's actually inside a library?
4. The OpenTheory importing facility is totally broken. In `src/opentheory/compat` there are some sample HOL theories to try to import `bool` and other packages from OT and turn it into a valid HOL theory with proofs.  So far I only fixed obvious mistakes due to SML structure name changes.

That's all.

binghe:
# Some statistics

The generated `hol4-large-numbers.art` (in "examples/probability") is a combination of OT `base`, HOL `hol-base`, `hol-sort`, `hol-string`, `hol-words`, `hol-ring`, `hol-res-quan`, `hol-quotient`, `hol-set`, `hol-integer`, `hol-real`, `hol-extreal`, `hol-analysis`, `hol-probability` and `examples/probability/large_numberTheory`. I believe all theorems in these packages are included in the article. In total, 20,167,185 inference rules are involved:
```
Inference rules:
eqMp ............... 6,408,556
deductAntisym ...... 5,260,856
subst .............. 5,156,259
appThm ............. 1,369,814
betaConv ............. 623,024
absThm ............... 419,171
trans ................ 395,037
refl ................. 197,745
sym .................. 178,196
assume ............... 142,838
proveHyp .............. 14,509
defineConst ............ 1,082
defineConstList ........... 72
defineTypeOp .............. 23
axiom ...................... 3
Total ............. 20,167,185
```

It's also interesting to know, among all the theorems, how many of them are directly or indirectly involved in proving LLN (Law of Large Numbers). For this purpose, in the additional Git commit I added `hol4-large-numbers-only.thy`, which generates a smaller OT article which only includes those theorems (and definitions) that are useful for proving theorems in `large_numberTheory`. The generated new OT article has 223MB, involving only 5,046,695 inference rules (25% of all):
```
Inference rules:
eqMp .............. 1,599,689
subst ............. 1,347,681
deductAntisym ..... 1,304,139
appThm .............. 321,798
betaConv ............ 161,162
absThm ............... 99,491
trans ................ 82,290
refl ................. 47,188
sym .................. 45,253
assume ............... 33,414
proveHyp .............. 4,127
defineConst ............. 419
defineConstList .......... 27
defineTypeOp ............. 14
axiom ..................... 3
Total ............. 5,046,695
```

However, unfortunately, currently there seems no way to see a list of these involved definitions and theorems from the HTML document generated from this smaller `hol4-large-numbers-only.art`.  I feel this is like all those intermediate theorems are now treated as `save_thm` instead of `store_thm`, i.e. not exported.   So far I don't know how OpenTheory article internally treat the difference between exported and unexported theorems.  I plan to take a look at OpenTheory source code, to see if there's a way to fix.

mn200:
This definitely qualifies as a labour of love!  Thanks for the work here.  It seems clear that there's a great deal more that could be done to improve the general HOL4/OpenTheory workflow.  Have you reported the possible OpenTheory issue with new_specification and multiple constants?  (Of course, it's quite possible the problem is entirely ours.)

binghe:
> Have you reported the possible OpenTheory issue with new_specification and multiple constants? (Of course, it's quite possible the problem is entirely ours.)

Not yet. I feel the problem may be at our side: OpenTheory doesn't have any primitive corresponding to `new_specification`, and it's quite possible that the HOL-generated `ordinal.art` is already wrong, which then caused `opentheory` reporting errors when generating `ordinal.ot.art`.   I hope @xrchz can take a look at this problem, otherwise I will do it myself by slowly reading HOL's OT-specific code first.

I have updated #805 with remaining issues mentioned in this PR.  (P.S. I still think `itself` related stuff should be moved to `oneTheory`, and this would slightly reduce the size of `src/boss/prove_base_assumsScript.sml`, although this is a very minor benefit.)

PR #1023, opened 2022-04-10
[holwrap.py] 2 spaces indentation and improvements

dwRchyngqxs:
Removed the use of temporary files to configure hol
Do not log the session by default because log files tend to be big
  so it should not be created unless the user is aware of the log file and wants it
Add timeout to Ctrl+c to ask for prompt when no prompt show up (in case of desync)
Improved handling of open by first trying to load the theories (to be consistent with emacs and Holmake behaviour)
Improved handling of multiple semi-colons scattered in a multiline input

### DIFF
--- a/tools/holwrap.py
+++ b/tools/holwrap.py
@@ -42,7 +42,8 @@ class Tee:
         self.file1.flush()
         self.file2.flush()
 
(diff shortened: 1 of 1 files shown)


mn200:
Thanks for working on this.  It'd also be great if this could get some documentation somewhere; perhaps in `Manual/Interaction-X` for an appropriate value of `X`.

PR #1026, opened 2022-04-11
Add unionWithKey, filterWithKey and filter to balanced_map

oskarabrahamsson:
This PR adds three operations to the balanced_map theory that are present in the Haskell library on which it is based. The operations are quite standard:
- filterWithKey filters map entries using a predicate that looks at both the key and the value;
- filter does the same, but looks only at the value;
- mapWithKey is as map but the function also takes the key as an argument;
- unionWithKey is as unionWith but the joining function also considers the key (unionWithKey was essentially already implemented, but not exposed as an operation).

Adds three new theorems about functional correctness for the above operations.


mn200:
Thanks!

PR #1028, opened 2022-04-12
"orthogonal edge routing" style of core theory hierarchy graph

binghe:
Hi,

I found the DOT-generated theory hierarchy graph hard to track the parent and children nodes when multiple edge curves are crossing together. This small PR makes some changes in the generated DOT file, to have the following new style, which perhaps is slightly better:

Schermata 2022-04-12 alle 21 18 08

--Chun


mn200:
Thanks!  That does look cool.

PR #1032, opened 2022-04-28
Teach Holmake and Parse.sml about the tmux terminal

oskarabrahamsson:
This makes Holmake and Parse.sml look for TERM-variable contents that start with "tmux" and use terminal colors if so.


mn200:
Awesome; thanks!

PR #1033, opened 2022-05-01
Merging old and new theories of derivative and gauge integrals

binghe:
Hi,

currently there are two formalizations of (univariate) Henstock–Kurzweil integration (and derivative) in HOL4. One is at `limTheory` (`diffl`) and `transcTheory`+`integralTheory` (`Dint` and `integral`), inherited from HOL88 due to John Harrison (and then enhanced by some Chinese authors as mentioned at the beginning of `integralScript.sml`.) The other is at `derivativeTheory` and `integrationTheory`,  originally from HOL Light's Multivariate analysis (also by John Harrison), ported to HOL4 by HVG Concordia people (for probability), and then fixed and ported to latest HOL4 by me (#746). (Below let's call them "old" and "new" formalizations.)

In this PR, I have proven that they are basically equivalent and derivative/integration theorems between them can be easily converted.  No theorem is deleted, thus the compatibility to old user scripts (if exist) is maintained.

The old formalization only supports integration on intervals (the basic and most common case), and related definitions are tightly aligned with textbooks.  In particular, the derivative part (`diffl`) is better, because there's a decision procedure `Diff.DIFF_CONV` for computing derivatives of elementary and transcendental fucntions, and certain advanced results like `transcTheory.MCLAURIN` is currently not available w.r.t. the new `derivativeTheory`.

On the other hand, the new one supports integration on more type of sets (essentially they should be Lebesgue measurable), but the related definitions are more complicated, because they were designed to support multivariate calculus. Previously HOL's probability theory depends on `integrationTheory`, but after I moved the construction of `lebesgue` to `$(HOLDIR)/examples/probability/lebesgue_measureScript.sml` now actually probability scripts do not depend on `integrationTheory` any more (it's still opened in `borelTheory`, etc. and in this PR I completely dropped the dependency.)

This theory merging is done by the following steps, roughly:

  1. create a new toplevel folder "src/analysis" (named after `$(HOLDIR)/src/real/analysis.ml`) and move in `seq`, `lim`, `powser`, `poly` and `transc` from `src/real`, and `real_topology`, `derivative` and `integration` from `src/probability`.   All these theories now depends on `real_topology` (and the remaining basic real theories in `src/real`).

  2. In `limTheory`, the following bridging theorem is added, which shows that the old and new derivative results can be easily converted to each other:
```
   [diffl_has_derivative]  Theorem (limTheory)
      ⊢ ∀f l x. (f diffl l) x ⇔ (f has_derivative (λx. x * l)) (at x)

   [diffl_has_vector_derivative]  Theorem      
      ⊢ ∀f l x. (f diffl l) x ⇔ (f has_vector_derivative l) (at x)
```

Also, in `Diff.sml` (now also moved to "src/analysis"), I have added a new decision procedure `HAS_VECTOR_DERIVATIVE_CONV`, which generates derivative theorems based on the new `has_vector_derivative`.

  3. In `integralTheory`, the following (hard) bridging theorem is added, which shows that the old gauge integral is equivalent to the new gauge integral on closed intervals:
```
   [Dint_has_integral]  Theorem (integralTheory)
      ⊢ ∀f a b k.
          a ≤ b ⇒ (Dint (a,b) f k ⇔ (f has_integral k) (interval [(a,b)]))

   [integrable_eq_integrable_on]  Theorem      
      ⊢ ∀f a b.
          a ≤ b ⇒ (integrable (a,b) f ⇔ f integrable_on interval [(a,b)])
   
   [integral_new_to_old]  Theorem      
      ⊢ ∀f a b.
          a ≤ b ∧ f integrable_on interval [(a,b)] ⇒
          integration$integral (interval [(a,b)]) f = integral (a,b) f
   
   [integral_old_to_new]  Theorem      
      ⊢ ∀f a b.
          a ≤ b ∧ integrable (a,b) f ⇒
          integral (a,b) f = integration$integral (interval [(a,b)]) f
```

(The proof of `Dint_has_integral` is quite hard due to the very different definitions of gauge and tagged divisions between the old and new formalizations, and there's no textbook proofs to learn. I managed to think as mathematicians  to work out this proofs. It takes about 1000 lines of code at the end of `integralScript.sml`.)

Some existing theorems are moved from one theory to another, I believe they are all reasonable. Besides small movements and additions, the following big movements are explained:

  1. Theorems about `content` and `interval_lowerbound`/`interval_upperbound` are moved from `integrationTheory` to `real_topologyTheory` (they really belongs to it), so that probability scripts do not need to open `integrationTheory` at all.

  2. The old gauge integration has a lot of definitions (e.g. `Dint`) and theorems (e.g. `FTC1`) in `transcTheory` but they have no contributions in proving properties of transcendental functions, and some theorems have duplications in `integralTheory`.  Now I have moved all of them to `integralTheory`, which now becomes a middle-size proof script of 2700 lines.

  3. Previously `real_sigmaTheory` depends on `transcTheory` but now the dependency is opposite (`real_sigmaTheory` is still in `src/real`). The AGM and Young's inequalities (#971) previous added into `real_sigmaTheory` now has to be moved to `transcTheory`.

From user point of view, moving core theories from `src/real` to `src/analysis` does not break the compatibility (although some official examples now need to include `src/analysis` beside `src/real`.) The only notable (minor) incompatiblity is due to the movements of `Diff.sml`, as now it cannot be part of `realLib`.  User code which relies on `DIFF_CONV` must now explicitly open `Diff` (and `transcTheory` to have supports on derivative of transcendental functions). I have added this information into the release notes).

OpenTheory theory definitions are also updated, and the OT build is confirmed being not broken.

P. S. It remains to update «HOL Description» in the future, and now I feel ready to establish the equivalence between Henstock–Kurzweil integration and Lebesgue integration (on integrable functions).

Regards,

Chun Tian


binghe:
I was... overconfident about the OT build status, please squash all commits when possibly merging this PR.

mn200:
I'm not so keen on having too many directories under `src` (hence moving all of `sum`, `pair` and `option` under `core types` recently).  I like the `analysis` name, so could you keep that directory, but move the construction of the reals in there too, possibly under a sub-directory if you like?

binghe:
> I'm not so keen on having too many directories under `src` (hence moving all of `sum`, `pair` and `option` under `core types` recently). I like the `analysis` name, so could you keep that directory, but move the construction of the reals in there too, possibly under a sub-directory if you like?

If this is your concern, I’d rather keep the famous `src/real` but put the newly added “analysis” folder under `src/real` as `src/real/more_theories`, a temporal solution until many year later HOL4 got more systematic core math theories which can be put as `src/algebra`, `src/geometry`, `src/analysis`, `src/topology`, etc..

What do you think?

mn200:
I'm no fan of `more_theories` as a name, and if the material you'd put there is all analysis; then `src/real/analysis` seems the right choice.

binghe:
> I'm no fan of `more_theories` as a name, and if the material you'd put there is all analysis; then `src/real/analysis` seems the right choice.

Perfect, thanks. I was thinking "src/real_analysis" and now I have "src/real/analysis", the same good. I will make the changes and it will take another day to make sure OT build is really not broken (some relative paths in `*.thy` files need to change).

binghe:
I think the OT build is fine now. (I’m using low-memory building machine and it may takes another 20 hours to get another huge OT article to see if there’s any leaking assumption.) Normal non-OT builds are definitely ok.

mn200:
Thanks for your work on this!

PR #1034, opened 2022-05-05
Various updates to probablity scripts

binghe:
Hi,

Below are some various updates to probability scripts.

1. In `borelTheory`, during the construction of one-dimension Borel measure space (`lborel`), the proof of an intermediate theorem "lborel0_finite_additive" was using `sortingTheory.QSORT` to sort a list converted from a set. This dependency is a little weird, after all pure math proofs shouldn't depend on the correctness of quicksort algorithm.  Recent I found in `iterateTheory` there's the following topological sort theorem which can be used as alternative (`TOPOLOGICAL_SORT'` is an easy corollary added in #1033 using `count` instead of `numseg`):
```
   [TOPOLOGICAL_SORT]  Theorem (iterateTheory)
      ⊢ ∀ $<<.
          (∀x y. x << y ∧ y << x ⇒ x = y) ∧
          (∀x y z. x << y ∧ y << z ⇒ x << z) ⇒
          ∀n s.
            s HAS_SIZE n ⇒
            ∃f. s = IMAGE f {1 .. n} ∧
                ∀j k. j ∈ {1 .. n} ∧ k ∈ {1 .. n} ∧ j < k ⇒ ¬(f k << f j)
   
   [TOPOLOGICAL_SORT']  Theorem      
      ⊢ ∀R s n.
          transitive R ∧ antisymmetric R ∧ s HAS_SIZE n ⇒
          ∃f. s = IMAGE f (count n) ∧
              ∀j k. j < n ∧ k < n ∧ j < k ⇒ ¬R (f k) (f j)
```
With the above `TOPOLOGICAL_SORT'`, now probability proofs do not depend on `sortingTheory` any more.  NOTE: to apply the above theorem, the supplied relation must be both transivive and antisymmetric, the later property requires a more careful construction of sorting relation `R` than those needed by `sortingTheory.QSORT` (transitive + total).

2. In `measureTheory`, an intermediate theorem `OUTER_MEASURE_CONSTRUCTION` (needed by `CARATHEODORY_SEMIRING`) was using a concrete bijiection between one- and two-dimensional number sets (`npair`, `nfst`, `nsnd`) provided by `numpairTheory`.  But actually only the existence of such bijiections would be sufficient. When I was working on these proofs in 2018, I wasn't aware that, in `util_probTheory`, Hurd has already proved a number of such existence theorems, e.g. the following one:
```
   [NUM_2D_BIJ_INV]  Theorem (util_probTheory)      
      ⊢ ∃f. BIJ f 𝕌(:num) (𝕌(:num) × 𝕌(:num))
```
Now, with the above `NUM_2D_BIJ_INV` the proof of `OUTER_MEASURE_CONSTRUCTION` can also be done. Thus the need of `numpairTheory` gets removed from probability proofs.

(Below are other minor changes)

3. In `martingaleTheory`, the recently added `function_space` (#972) has been renamed to `lp_space`, to be aligned with Lean mathlib.  (I was thinking how to best name $L^p$ spaces in HOL, but never reached a satisfied one.)

4. In `metricTheory` and also `extrealTheory`, the recently added `reduced_metric` has been renamed to `bounded_metric`, an old (good) name but then changed in the last minute before sending the PR (#993).  There's no other formalization or textbook as references, but I personally think "bounded" is a better character than "reduced" which is a little ambiguous.

5. In `lebesgueTheory`, one existing theorem `integral_add3` has been improved with much less antecedents, after `integral_add'` added in #1017.
```
   [integral_add3]  Theorem (lebesgueTheory)      
      ⊢ ∀m f g h.
          measure_space m ∧ integrable m f ∧ integrable m g ∧
          integrable m h ⇒
          ∫ m (λx. f x + g x + h x) = ∫ m f + ∫ m g + ∫ m h   
```

6. In `hurdUtils`, I added some comments near `STRONG_DISJ_TAC`, with connections to `LEFT_DISJ_TAC` and `RIGHT_DISJ_TAC` provided by `schneiderUtils`.  All these tactics are sometimes more useful than the existing standard tactics.

--Chun

### DIFF
--- a/Manual/Description/math.stex
+++ b/Manual/Description/math.stex
@@ -852,7 +852,7 @@ More precisely, we are interested in the sets:
 Or formally:
 \begin{hol}
 \begin{alltt}
-##thm function_space_alt_finite
+##thm lp_space_alt_finite
 \end{alltt}
 \end{hol}
 
@@ -871,7 +871,7 @@ Clearly, $u\in\mathcal{L}^p(\mu)$ if, and only if, $u\in\mathcal{M}(\mathscr{A})
 Furthermore, $\|u\|_p$ implies that $u(x) = 0$ for almost every $x$.
 \begin{hol}
 \begin{alltt}
-##thm function_space_alt_seminorm
+##thm lp_space_alt_seminorm
 
 ##thm seminorm_eq_0
 \end{alltt}
--- a/examples/probability/lebesgue_measureScript.sml
+++ b/examples/probability/lebesgue_measureScript.sml
@@ -64,7 +64,7 @@ QED
 val integral_indicator_UNIV = store_thm ("integral_indicator_UNIV",
   ``!s A. integral UNIV (indicator (s INTER A)) =
           integral A (indicator s)``,
-  REWRITE_TAC [integral] THEN REPEAT STRIP_TAC THEN AP_TERM_TAC THEN
+  REWRITE_TAC [integral_def] THEN REPEAT STRIP_TAC THEN AP_TERM_TAC THEN
   ABS_TAC THEN METIS_TAC [has_integral_indicator_UNIV]);
 
 val integrable_indicator_UNIV = store_thm ("integrable_indicator_UNIV",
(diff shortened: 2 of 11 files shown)


binghe:
The last commit c5674ef contains a fix of `examples/lebesgue_measureScript.sml` due to the name change of `integrationTheory.integral` (to `integral_def`) in #1033.

mn200:
Thanks!

PR #1035, opened 2022-05-08
Migrated all theorems from the old analysis.ml and transc.ml (and delete them)

binghe:
Hi,

The two old HOL88 scripts `src/real/analysis.ml` and `src/real/transc.ml` were added 23 years ago (as part of the initial commit of the repository). Most of the theorems in these two scripts now can be found in various HOL core theories but there are still a few of them not ported/migrated yet.   Now I have checked each theorem in these two files, and if there's any theorem who cannot be found in existing HOL core theories, I port the old proofs and put them into suitable core theories.

In summary, the following new theorems are added: (in a few cases the old proofs involved non-exist old real and arithmetic theorems, I guess their statements, confirm with hol-light, and recovered them)

```
   [LE_LT]  Theorem (arithmeticTheory)
      ⊢ ∀m n. m ≤ n ⇔ m < n ∨ m = n

   [MULT_DIV_2]  Theorem (arithmeticTheory)
      ⊢ ∀n. 2 * n DIV 2 = n

   [EVEN_DIV_2]  Theorem (arithmeticTheory)
      ⊢ ∀n. ¬EVEN n ⇒ SUC n DIV 2 = SUC ((n − 1) DIV 2)

   [INTERVAL_LEMMA_LT]  Theorem (limTheory)
      ⊢ ∀a b x.
          a < x ∧ x < b ⇒ ∃d. 0 < d ∧ ∀y. abs (x − y) < d ⇒ a < y ∧ y < b

   [IVT_DERIVATIVE_0]  Theorem (limTheory)
      ⊢ ∀f f' a b.
          a ≤ b ∧ (∀x. a ≤ x ∧ x ≤ b ⇒ (f diffl f' x) x) ∧ f' a > 0 ∧
          f' b < 0 ⇒
          ∃z. a < z ∧ z < b ∧ (f' z = 0)
   
   [IVT_DERIVATIVE_NEG]  Theorem (limTheory)
      ⊢ ∀f f' a b y.
          a ≤ b ∧ (∀x. a ≤ x ∧ x ≤ b ⇒ (f diffl f' x) x) ∧ f' a < y ∧
          f' b > y ⇒
          ∃z. a < z ∧ z < b ∧ (f' z = y)

   [IVT_DERIVATIVE_POS]  Theorem (limTheory)
      ⊢ ∀f f' a b y.
          a ≤ b ∧ (∀x. a ≤ x ∧ x ≤ b ⇒ (f diffl f' x) x) ∧ f' a > y ∧
          f' b < y ⇒
          ∃z. a < z ∧ z < b ∧ (f' z = y)

   [REAL_EXP_BOUND_LEMMA]  Theorem (transcTheory)      
      ⊢ ∀x. 0 ≤ x ∧ x ≤ 2⁻¹ ⇒ exp x ≤ 1 + 2 * x

   [REAL_LE_LCANCEL_IMP]  Theorem (realTheory)
      ⊢ ∀x y z. 0 < x ∧ x * y ≤ x * z ⇒ y ≤ z

   [REAL_POW_LE]  Theorem (realTheory)
      ⊢ ∀x n. 0 ≤ x ⇒ 0 ≤ x pow n
```

P. S. In addition, I added a Makefile target "install-all" in `src/boss/Holmakefile`. It installs all HOL OpenTheory theories in a correct order, after the full OT build. This target is useful, because otherwise manually installing them with correct orders takes unnecessary extra time.

Chun Tian


binghe:
It seems that there was a powerful `REAL_RAT_REDUCE_CONV` in the time of HOL88.  For example, if we had the following proof goal:
```
        2⁻¹ < 1
   ------------------------------------
 ...
```
Then the tactic `CONV_TAC REAL_RAT_REDUCE_CONV` will simply finish the proof.  But now I can't find anything similar, despite there are some `rat` related lower case theorems still in `realTheory`.  Any idea?

binghe:
> It seems that there was a powerful `REAL_RAT_REDUCE_CONV` in the time of HOL88. For example, if we had the following proof goal:
> 
> ```
>         2⁻¹ < 1
>    ------------------------------------
>  ...
> ```
> 
> Then the tactic `CONV_TAC REAL_RAT_REDUCE_CONV` will simply finish the proof. But now I can't find anything similar, despite there are some `rat` related lower case theorems still in `realTheory`. Any idea?

A possible answer to my own question:

`REAL_RAT_REDUCE_CONV` can be found in HOL-Light's `calc_rat.ml` (https://github.com/jrh13/hol-light/blob/master/calc_rat.ml) but I believe its functionalities are now part of HOL-Light's (powerful) `REAL_ARITH` decision procedure (in the same code file), e.g. in HOL-Light,

```
# REAL_ARITH `&1 / &2 < &1`;;
val it : thm = |- &1 / &2 < &1
# REAL_ARITH `inv(&2) < &1`;;   
val it : thm = |- inv (&2) < &1
```

while in HOL4, the existing `REAL_ARITH` is quite weak (or incapable) in proving each simplest real inequalities: (note also that the exception message is very strange)

```
> REAL_ARITH ``&1 / &2 < &1``;
Exception-
   HOL_ERR
     {message = "original exn. not a HOL_ERR", origin_function = "trye", origin_structure =
      "Lib"} raised
> REAL_ARITH ``inv(&2) < &1``;
Exception-
   HOL_ERR
     {message = "original exn. not a HOL_ERR", origin_function = "trye", origin_structure =
      "Lib"} raised
```

Previously, when I was porting `integrationTheory`, I also found that HOL4's `REAL_ARITH` cannot handle `min` and `max` at all, while HOL-Light's `REAL_ARITH` can.  In many such cases, I had to first rewrite `min` and `max` by their definitions and then call `REAL_ARITH_TAC` as usual.

It seems, from the header of `src/RealArith.sml`, that the current version of `REAL_ARITH` was ported by Joe Hurd from HOL Light in 1998.  By the header of HOL-Light's `calc_rat.ml`, which I copy here:
```
(* ========================================================================= *)
(* Calculation with rational-valued reals.                                   *)
(*                                                                           *)
(*       John Harrison, University of Cambridge Computer Laboratory          *)
(*                                                                           *)
(*            (c) Copyright, University of Cambridge 1998                    *)
(*              (c) Copyright, John Harrison 1998-2007                       *)
(* ========================================================================= *)
```
It seems that John Harrison has made many further changes in 1998-2007, providing the extra features mentioned above.  Thus it seems (to me at least) quite *urgent* for HOL4 to have the same powerfulness in `REAL_ARITH_TAC`, among other possible contributions, because if `REAL_ARITH_TAC` gets improved, many existing proofs involving real arithmetics can be simplified, while future proofs will benefit more by having shorter development time.

mn200:
Normalisation of expressions like this is done by the conversions in `realSimps`.  Just loading `realSimps` adds machinery to `srw_ss` to handle lots of this sort of thing.  Thus
```
> load "realSimps";
val it = () : unit
> SCONV [] ``inv 2 < 1``
val it = ⊢ 2⁻¹ < 1 ⇔ T: thm
```
Many more examples of this code’s capabilities are in `src/real/selftest.sml`. 

For more precise uses, there are the underlying conversions `RMULCANON` and `mulrelnorm` as well as simpset fragments. 

There is more that can and should be done.  In particular, there is no `addrelnorm` yet.

src/boss/Holmakefile
```diff
@@ -79,6 +79,25 @@ hol4-base.art: hol4-base.thy hol4-base-unint.art ../opentheory/hol4.int \
 install: hol4-base.thy hol4-base.art 
 	opentheory install --reinstall $<
 
+# This installs all HOL-OT theories (after building) in correct order
```
mn200, line 82: This is not good style.  In particular, it assumes that `Holmake` is in the PATH.  Also, we know that recursive make is not a good idea. I'd much rather see explicit targets listed for the sub-directories.  For example, 
```
install-all: $(patsubst %,%/install,$(dirs))
```
If you need the linearisation implied by the `-j1`s or for logical dependency reasons, that dependency information should be in the relevant directories’ `Holmakefile`s.

mn200: This is also the wrong place for this.  I don't want (implicit) forward pointers from `src/boss` into material (such as `real` and `probability`) that comes later in the build sequence.

binghe: I'm going to completely revert these changes in `src/boss/Holmakefile`. Actually the best thing I was having in mind, is to have something at toplevel like `bin/build install-ot` (after a full HOL build) but I don't know how to implement it...

binghe:
OT changes reverted. Please squash all commits when (possibly) merging this PR.

binghe:
> Normalisation of expressions like this is done by the conversions in `realSimps`. Just loading `realSimps` adds machinery to `srw_ss` to handle lots of this sort of thing. Thus
> 
> ```
> > load "realSimps";
> val it = () : unit
> > SCONV [] ``inv 2 < 1``
> val it = ⊢ 2⁻¹ < 1 ⇔ T: thm
> ```
> 
> Many more examples of this code’s capabilities are in `src/real/selftest.sml`.
> 
> For more precise uses, there are the underlying conversions `RMULCANON` and `mulrelnorm` as well as simpset fragments.
> 
> There is more that can and should be done. In particular, there is no `addrelnorm` yet.

I took a look again at the proof (in `transcTheory` where both `RealArith` and `realSimps` are opened), actually just `rw []` will resolve the goals where previously using `REAL_RAT_REDUCE_CONV`.  So I misjudged the abilities of real arithmetic simplifications currently in HOL4.  (I still think `REAL_ARITH` should be improved (to match what HOL-Light has), so that porting HOL-Light proofs could be easier.)

binghe:
In another working branch, I managed to move a small number of real theorems from `realTheory` to `realaxTheory` (together with some real operator definitions). This then allows the existing `RealArith` package to not depend on `realTheory` at all, and then `realTheory` can now open `RealArith` and use `REAL_ARITH_TAC` to simplify many existing proofs in `realScript.sml`, which has been reduced for about 10% of file size.

I consider this work the "stage 1" as the next PR (once the present one gets merged).  The "stage 2" would be update Joe Hurd's port of `REAL_ARITH_TAC` with latest code from HOL Light, to add support for `abs`, `min` and `max`, etc.   In the "stage 3", I plan to add `RealArith2.sml` which ports code from HOL-Light's `calc_rat.ml` as a new tactic `REAL_ARITH2_TAC` for use in proof scripts after `realScript.sml`.  The new `RealArith2.sml` will depend on the existing (once updated) `RealArith.sml`, solving a strict super set of real arithmetic problems.

mn200:
Thanks for your work on this.  

I like the idea of having a `REAL_ARITH` available in `realScript.sml`, though that theory file is probably relatively fixed: I don't imagine having `REAL_ARITH` there will make a huge difference to future work in that file.  

Note that `REAL_ARITH` already handles at least some occurrences of `abs`:
```
> REAL_ARITH ``abs x < 10 ==> -12 <  x - 2 ``;
val it = ⊢ abs x < 10 ⇒ -12 < x − 2: thm
```

I'm less keen on the having a sequence of entry-points that just "do more".  How is the extra functionality in the updated code implemented?  I can imagine that the handling of `min`, `max` and `abs` just rewrites with the definitions and is then smart with the guards attached to the conditional expressions.  I totally agree that this is great functionality to have, but my instinct would be to implement it by having the same underlying code parameterised by normalisation rewrites. Then when later theories are loaded, the code will use the additional theorems without further thought required. 

I'd do this by using a `ThmSetData` declaration with name something like `real-arith-norm-rwts`.  (This is how the technology in `analysis/Diff` works.)  

There is already an attempt to handle conditional expressions in `RealArith` *via* the code in `Sub_and_cond.COND_ELIM_CONV`, but this does nothing on real-valued conditional terms, at least since a5b09774c9070e5ca7d3e1 I guess.

binghe:
Thanks for merging this PR (and left the honor of deleting two big old files to me).

HOL-Light implements an early bootstrapping `REAL_ARITH` in its `calc_int.ml` and `realarith.ml`, for use of proofs in `real.ml` (HOL's `realTheory`). This is what Hurd ported (or perhaps at that file there's only this one), but Hurd's code now looks quite different with the latest version of HOL Light.  Then HOL-Light's has `calc_rat.ml` which contains a stronger version of `REAL_ARITH` handling not just more real operators but essentially for rational arithmetics (well, we can think it as a support of division and `realinv` but the code is quite different), which uses a lot of it owns theorems and some in `real.ml`. So far I don't see any repeated patterns, because the new `REAL_ARITH` simply calls the ML functions used for old `REAL_ARITH` to do some jobs.

But anyway, I plan to first obtain a stronger `REAL_ARITH`, and then think how to make the old and new version live in the same `RealArith.sml` and behave like `analysis/Diff`, i.e. after loading `realTheory` the same solver directly becomes stronger.

mn200:
Sounds good.

PR #1036, opened 2022-05-08
Use relative paths in the generated "HOL Reference Page"

binghe:
Hi,

This PR fixes a small but very annoying issue in the generated HTML "HOL Reference Page": previously all links in that page are `file://`-like absolute paths. As a (bad) result, HOL's "help" folder cannot be directly hosted by a Web server. Even not consider web hosting, if you pack the "help" folder as a zip and send to others, now they can just decompress it and click the entry HTML file to see the contents of each theories (previously this is impossible).

After the changes in this PR, now it's easy to create a soft link (by `ln -s`) from your local HOL working directory (the `help` sub-folder only) to the root directory of your local HTTP server (e.g. apache2) and view all HOL core theorems from a (remote) HTTP client, e.g. the Web browser of your tablet.  The HTML file name `HOLindex.html` has been also renamed to just `index.html`, which is the default entry HTML file name accepted by Apache and many other Web servers, so that there's no need to explicit type "HOLindex.html" as part of the URL.

Note that the links to proof scripts files (`*Script.sml`) are not touched. They are still in absolute paths.

P. S. The official online "HOL Reference Page" [1] (of the most recent HOL release) has already used relative paths. I was wondering how it was done, but now I think it should be manually modified after the HTML generation, by simply removing the absolute path prefix.

[1] https://hol-theorem-prover.org/kananaskis-14-helpdocs/help/HOLindex.html

Regards,

Chun Tian

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,6 +69,7 @@ tools/build-logs
 help/Docfiles/*.txt
 help/Docfiles/HTML/*
 help/HOLindex.html
+help/index.html
 help/HOL.Help
 help/src-sml/htmlsigs
 help/src-sml/index.tex
@@ -130,3 +131,4 @@ examples/l3-machine-code/**/*-heap
 
 *Script_ttt.sml
 *.tttsave
+/TAGS
--- a/help/src-sml/HOLPage.sml
+++ b/help/src-sml/HOLPage.sml
@@ -123,9 +123,11 @@ fun printHOLPage version bgcolor HOLpath idIndex TheoryIndex (dbfile, outfile)
         out "<P>";
 
         out"<DT><STRONG>";
-        href "IDENTIFIERS" idIndex;
+        href "IDENTIFIERS"
+             (normPath["src-sml","htmlsigs","idIndex.html"]); (* was: idIndex *)
         out "&nbsp;&nbsp;&nbsp;&nbsp;";
-        href "THEORY BINDINGS" TheoryIndex;
+        href "THEORY BINDINGS"
+             (normPath["src-sml","htmlsigs","TheoryIndex.html"]); (* was: TheoryIndex *)
         out "</STRONG>";
         out "<P>";
 	out "</DL>\n";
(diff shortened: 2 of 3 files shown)


binghe:
The 2nd commit fixed the two links "IDENTIFIERS" and "THEORY BINDINGS" in the bottom of the "HOL Reference Page", but I don't like this solution: the two HTML file names `idIndex.html` and `TheoryIndex.html` are defined in `makebase.sml` by global variables but in full absolute paths:
```
(* Default filename for the HTML format database for identifiers: *)
val htmlIndexDef = normPath[HOLpath,"help","src-sml",htmlDirDef,"idIndex.html"]

(* Default filename for the HTML format database for theories: *)
val htmlTheoryIndexDef =
   normPath[HOLpath,"help","src-sml",htmlDirDef,"TheoryIndex.html"]
```
But when these variables are passed to structure HOLPage, I can't think out a way to remote their prefix (to retrieve the relative paths).

binghe:
Not only web server hosting but also HOL builds on Windows (Linux in WSL2) can benefit from this PR. On Windows WSL2, one can access HOL working directory by, e.g. `\\wsl$\Ubuntu-20.04\home\...`, then it's possible to open the new `help\index.html` and click in the links without problems. Previously such browsing is impossible because in the generated HTML file(s) all file paths are absolute Linux paths invalid for Windows.

binghe:
Sorry, the commit [02b7e5d](https://github.com/HOL-Theorem-Prover/HOL/pull/1036/commits/02b7e5d23d6b37b47ba3d93a3861c3ab7ffbae2d) was trying to remove some code duplications but actually the code is not compilable (somehow I failed to see it when rebuilding HOL4 without any theory changes). This bad commit is now reverted...

Please squash all commits when possibly merging this PR, and I have to leave the rest of possible code improvements to you, since currently I'm still not very similar with list manipulation functions (or the whole standard library) of SML...

mn200:
Thanks for your work on this: I agree the absolute paths are a mistake.

binghe:
> Thanks for your work on this: I agree the absolute paths are a mistake.

You're welcome.  But I'm not going to make further commits in the branch of this PR, as I don't have a better solution beyond its current status, which from the user view is just fine (but internally has some duplications).

PR #1038, opened 2022-05-13
Report workaround on size_eq proof failure

talsewell:
Catch errors in proving the size_eq results in DataSize and remind the user
that they can skip that step if necessary with DataSize.prove_size_eqs. At
least one situation where this is necessary was recently reported by @agomezl.

### DIFF
--- a/src/datatype/DataSize.sml
+++ b/src/datatype/DataSize.sml
@@ -242,6 +242,10 @@ fun define_size {induction, recursion} db = case define_size_rec recursion db of
     NONE => NONE
   | SOME r => if ! prove_size_eqs then let
     val comb_eqs = size_def_to_comb db (SOME induction) (#def r)
+      handle HOL_ERR err =>
+        let in
+        print "Error proving size_eqs, consider DataSize.prove_size_eqs := false\n\n";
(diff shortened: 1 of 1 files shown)


binghe:
By the way, following discussions on Slack, I'm going to put the following line in `extrealScript.sml`:
```
val _ = DataSize.prove_size_eqs := false;
```
to disable the exporting of `extreal_size_eq` which is not very useful (`|- T`) although also not harmful.

src/datatype/DataSize.sml
```diff
@@ -242,6 +242,10 @@ fun define_size {induction, recursion} db = case define_size_rec recursion db of
     NONE => NONE
   | SOME r => if ! prove_size_eqs then let
     val comb_eqs = size_def_to_comb db (SOME induction) (#def r)
+      handle HOL_ERR err =>
+        let in
+        print "Error proving size_eqs, consider DataSize.prove_size_eqs := false\n\n";
```
mn200, line 247: This should be a `Feedback.HOL_MESG`

mn200 (changes requested):

mn200:
Closing, as this is subsumed by #1039

PR #1039, opened 2022-05-16
Sync type sizes in DataSize and TypeBasePure

talsewell:
This patch generalises the type-size builder in TypeBasePure so that code in
DataSize can share it rather than (as previously) defining its own.

The previous situation was observed by @agomezl to result in problems when
a datatype contained the word type. The two implementations (in DataSize and
TypeBasePure) query the type database differently and so see the word type
(which is a specialised pun of the cartesian product type) differently and assign
it different sizes. This new version should pick the standard size.

One of the goals of earlier work on size_eqs was to avoid silly situations where
TypeBasePure.type_size was the wrong choice of size. This happened, for instance
multiple names for a size existed (e.g. tree1_size vs list_size tree_size), and so
the system proves they are equal. It would be good to remove further discrepancies.

I expect that tests will fail, I'm putting this up anyway in case it attracts discussion.

### DIFF
--- a/src/1/TypeBasePure.sig
+++ b/src/1/TypeBasePure.sig
@@ -121,6 +121,7 @@ sig
                          (hol_type -> term option) *
                          (hol_type -> term) -> hol_type -> term
 
+   val type_size_pre   : (hol_type -> term option) -> typeBase -> hol_type -> term
    val type_size       : typeBase -> hol_type -> term
    val type_encode     : typeBase -> hol_type -> term
    val type_lift       : typeBase -> hol_type -> term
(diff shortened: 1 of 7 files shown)


mn200:
Happy to merge this for communal testing etc on `develop`.  Will close the other PR that this one subsumes.

talsewell:
I'd report on how I'm doing with tests, but something confusing has happened. My HOL directory has
got into a state where some selftests report errors in `./bin/build -t2` which then don't appear running
those scripts interactively. I've tried a git clean, so I'm still trying to get to the bottom of that.

talsewell:
OK, I've debugged my previous confusion.

If you load Datatype but not basicSize, and then define a datatype that uses a basic
type and would have a size eq, then the code will fail. I should probably just skip
the size eqs and print a warning in a case like this. It's an obscure scenario, but some
of the datatype selftests set up exactly this kind of environment.

mn200:
Thanks for your work on this!

talsewell:
No worries. A run of `build -t2` went fairly far on my machine until a TeX test that tries to create an executable hit an issue with libpolyml. Give me a heads-up if anything breaks in an interesting way.

PR #1041, opened 2022-05-23
Add a size for the "itself" type in basicSize

talsewell:
Add "itself" to the collection of types with manual size setup
in basicSize. This doesn't seem to cause any problems. I don't
have any examples to point at where it solves any problems
either.


mn200:
If it's clearly the "right thing", we should do it anyway!  :)  Thanks for putting the time in.

PR #1043, opened 2022-06-03
The new REAL_ARITH(_TAC) (and REAL_FIELD) ported from HOL-Light

binghe:
Hi,

Domain-specific automatic provers are very handy in the practice of proof engineering. For instance, the `REAL_ARITH` (and its tactic version, `REAL_ARITH_TAC`) are used ~2,200 times in HOL4's core math theory library. With it, many simple real arithmetics lemmas can be directly proved by the tool without writing manual proof tactics (so that the proof engineer can focus on the "key" steps of the proof, for which there's no hope to have automatic program to do.)

`REAL_ARITH` (in `src/real/RealArith.sml`) was ported from HOL-Light (to hol98) by Joe Hurd in 1998. But since then HOL-Light has developed a lot, the author (John Harrison) added support of rational-valued coefficients and support of real operators like `abs` (enhanced), `max` and `min`.  When I was porting proofs from HOL-Light, I constantly feel difficulties  when proof steps calling HOL-Light's more powerful `REAL_ARITH_TAC` cannot directly run in HOL4, and usually I need to insert some more steps (sometimes non-trivial) before calling `REAL_ARITH_TAC`.

For example, in a almost "worst" case (`real_topologyTheory.INTER_BALLS_EQ_EMPTY`), before calling `REAL_ASM_ARITH_TAC` (a variant) I had to eliminate all occurrences of `abs`, `max` and `min` by their definitions and then do case analysis, this has caused 1024 (instead of 4 as the proof in HOL-Light) subgoals, each needs to call `REAL_ASM_ARITH_TAC`, and eventually the proof of this small lemma has taken more than 10 seconds to complete.)

In this huge PR, I have brought HOL-Light's latest version of `REAL_ARITH` (real arithmetic prover) and also another related tool called `REAL_FIELD` into HOL4.  The original OCaml source code were from HOL-Light's several files: [1]

- `calc_int.ml` and `realarith.ml`: the integer-valued real expression reduce conversions and a bootstrapping version of `REAL_ARITH` (only supports integer coefficients) for use of proving theorems in `real.ml` (corresponding to HOL4's `realTheory`). The bootstrapping version is a specialization of the general `GEN_REAL_ARITH` defined also here. These code has been ported into the existing `src/real/RealArith.sml`.
- `calc_rat.ml`: the rational-valued real expression reduce conversions and the final version of `REAL_ARITH` (by specializing `GEN_REAL_ARITH` using rational-aware conversions).   And a more advanced prover `REAL_FIELD` based on real polynomials and Gröbner Basis. These code has been ported to a new file `src/real/RealField.sml`.
- `normalizer.ml`: a general semi-ring normalizer, which generates many conversions in `cal_int.ml` and `calc_rat.ml`. This work is general for any (semi)ring number systems and is ported to a new file `Normalizer.sml` in `src/num/reduce/src`, since it uses a lot of things from `reduceLib`.
- `grobner.ml`: the algorithm for calculating Gröbner Basis. It's also general for any ring system. It's ported and put as `Grobner.sml` also in `src/num/reduce/src`.
- `canon.ml`: I have ported many low-level conversions used in above files but HOL4 do not have. Most of the newly added conversions are put into `src/metis/normalForms.sml`, including a general NNF conversion.

# entry API

The new (final version) `REAL_ARITH`, `REAL_ARITH_TAC` and `REAL_ASM_ARITH_TAC` are exported by the structure `RealField`.   The same 3 functions previously exported by `RealArith` are also based on the new code, but this is an "intermediate" version to be used by `realTheory`, etc. before the proof scripts can `open realLib`.  (`RealArith` now only depends on `realaxTheory`)

The old API (and all its code) are still kept in `RealArith` but are exported with changed names: `OLD_REAL_ARITH`, `OLD_REAL_ARITH_TAC` and `OLD_REAL_ASM_ARITH_TAC`.   There's also a `PURE_REAL_ARITH_TAC` (not exists in HOL-Light) without name change.

For user code, it's recommended to open `realLib` instead of `RealArith` (low-level structure).  In `realLib`, it's possible to switch between the old and new tools, by a simple line like `val REAL_ARITH = RealArith.OLD_REAL_ARITH;`   Currently the old tools are still used if user code actually opens `realLib`, and HOL core theories are still built with the old `REAL_ARITH` (with several exceptions), simply because the old tool is 2-3x faster. (In the future, the new code should be further analyzed and optimized.) Users having needs of more features or HOL-Light compatible `REAL_ARITH` can directly open `RealField` in their proof scripts, or have a line such as `val REAL_ARITH = RealField.REAL_ARITH;` to explicitly switch to the new tool.

However, I do have tested the builds of entire HOL with only the new `REAL_ARITH`, and it works (although slower, currently).  In fact, I did many fixes to make sure the new code can solve some (less-interesting) problems that even the HOL-Light version of `REAL_ARITH` cannot solve.     Backwards compatibilities and HOL-Light compatibilities are both maintained.  But in very rare cases where user proofs may have `TRY(REAL_ARITH_TAC)` followed by several subgoals, the number of subgoals may get reduced and the proofs could be broken. (To fix it, just delete some dead proof branch) This is similar with the cases when HOL4's global simplifier changed (or improved), but it's very rare to have `REAL_ARITH_TAC` not at the last proof step of a proof branch.

(See, e.g. `src/real/selftests.ml` for some test cases that the new `REAL_ARITH` can solve.)

It is hard to describe all, even major code changes in more than 6,000 lines of code additions (and 1000+ lines deletions). Please allow me to do this slowly in the next days, and in forms of comments of this PR.

Regards,

Chun Tian

[1] https://github.com/jrh13/hol-light


binghe:
# Changes in `realaxTheory` and `realTheory`

For long time there was no way to add any new theorem into `realTheory` if the proof involves `REAL_ARITH` (or `REAL_ARITH_TAC`, hereafter let's only mention one of them), because currently `RealArith` depends on `realTheory`.

But clearly `RealArith` does not need all theorems in `realTheory`, instead, just a small set of "primitive" real theorems would be enough.  On the other hand, `realaxTheory` is relatively small and for long time there's no need to change it at all.   Thus, the first step of this project is to identify a small set of "primitive" `realTheory` theorems which can be moved into `realaxTheory`, so that `RealArith` only needs to depend on `realaxTheory`, and then inside `realScript.sml` we can `open RealArith` (for a basic version of `REAL_ARITH`) to easily prove other simple real theorems.  This work is essentially irrelevant with the new `REAL_ARITH` ported from HOL Light but is done as the first step here.

To maintain the backward compatibility, all theorems moved from `realTheory` to `realaxTheory` must still be exported in `realTheory`, because it's never recommended for user proof scripts to open `realaxTheory`.  This is done here, and in the changed `realScript.sml` now we can see a lot of trivial `save_thm` calls for this purposes.

Beside theorems, what's more important is the definitions of various real operators like `+`, `-`, `*`, `div`, `abs`, `pow`, `max`, `min`. Now they are all defined in `realaxTheory` (`sqrt`, `sup` and `inf` are still in `realTheory` since they are irrelevant with `REAL_ARITH`).  Once this is done, the structure `realSyntax` also does not depend on `realTheory` any more, and can be opened and use in `RealArith`.

The other related consideration is the OpenTheory build. `realaxTheory` is NOT included in the generated OT package `hol-real`. This means that, any theorem about normal real operators (i.e. before the quotation from `treal` or `hreal`) in `realaxTheory`, if ever been used by other proofs in subsequent core theories, will become a leaking assumption in the related OT package.   Thus, I must also make sure that, every new theorem I added in `realaxTheory`, even only used by `RealArith`, must also be exported again in `realTheory`, as in this way these theorems can be available from `hol-real` OT package.

In this PR, to ease the code review, I have modified `realScript.sml` in "minimal" way such that the order of theorems basically did change: all you can see is the previous full theorem definition being replaced by a simple `save_thm`. Besides, there's also something new (not used by `RealArith`), the real sign function, `sgn`.

## Real sign function `sgn` and supporting theorems

The real sign function `sgn(x)` sometimes occurs in mathematical proofs. I ported them from HOL-Light, mainly to demonstrate the new `REAL_ARITH`, almost all proofs about `sgn` are directly done by `REAL_ARITH`. Below is its definition and the list of new theorems added in `realTheory`: (most of the statements about `sgn` look very nice)

```
   [real_sgn]  Definition      
      ⊢ ∀x. sgn x = if 0 < x then 1 else if x < 0 then -1 else 0

   [REAL_SGN]  Theorem      
      ⊢ ∀x. sgn x = x / abs x
   
   [REAL_SGNS_EQ]  Theorem      
      ⊢ ∀x y.
          sgn x = sgn y ⇔
          (x = 0 ⇔ y = 0) ∧ (x > 0 ⇔ y > 0) ∧ (x < 0 ⇔ y < 0)
   
   [REAL_SGNS_EQ_ALT]  Theorem      
      ⊢ ∀x y.
          sgn x = sgn y ⇔
          (x = 0 ⇒ y = 0) ∧ (x > 0 ⇒ y > 0) ∧ (x < 0 ⇒ y < 0)
   
   [REAL_SGN_0]  Theorem      
      ⊢ sgn 0 = 0
   
   [REAL_SGN_ABS]  Theorem      
      ⊢ ∀x. sgn x * abs x = x
   
   [REAL_SGN_ABS_ALT]  Theorem      
      ⊢ ∀x. sgn x * x = abs x
   
   [REAL_SGN_CASES]  Theorem      
      ⊢ ∀x. sgn x = 0 ∨ sgn x = 1 ∨ sgn x = -1
   
   [REAL_SGN_DIV]  Theorem      
      ⊢ ∀x y. sgn (x / y) = sgn x / sgn y
   
   [REAL_SGN_EQ]  Theorem      
      ⊢ (∀x. sgn x = 0 ⇔ x = 0) ∧ (∀x. sgn x = 1 ⇔ x > 0) ∧
        ∀x. sgn x = -1 ⇔ x < 0
   
   [REAL_SGN_EQ_INEQ]  Theorem      
      ⊢ ∀x y. sgn x = sgn y ⇔ x = y ∨ abs (x − y) < max (abs x) (abs y)
   
   [REAL_SGN_INEQS]  Theorem      
      ⊢ (∀x. 0 ≤ sgn x ⇔ 0 ≤ x) ∧ (∀x. 0 < sgn x ⇔ 0 < x) ∧
        (∀x. 0 ≥ sgn x ⇔ 0 ≥ x) ∧ (∀x. 0 > sgn x ⇔ 0 > x) ∧
        (∀x. 0 = sgn x ⇔ 0 = x) ∧ (∀x. sgn x ≤ 0 ⇔ x ≤ 0) ∧
        (∀x. sgn x < 0 ⇔ x < 0) ∧ (∀x. sgn x ≥ 0 ⇔ x ≥ 0) ∧
        (∀x. sgn x > 0 ⇔ x > 0) ∧ ∀x. sgn x = 0 ⇔ x = 0
   
   [REAL_SGN_INV]  Theorem      
      ⊢ ∀x. sgn x⁻¹ = sgn x
   
   [REAL_SGN_MUL]  Theorem      
      ⊢ ∀x y. sgn (x * y) = sgn x * sgn y
   
   [REAL_SGN_NEG]  Theorem      
      ⊢ ∀x. sgn (-x) = -sgn x
   
   [REAL_SGN_POW]  Theorem      
      ⊢ ∀x n. sgn (x pow n) = sgn x pow n
   
   [REAL_SGN_POW_2]  Theorem      
      ⊢ ∀x. sgn x² = sgn (abs x)
   
   [REAL_SGN_REAL_SGN]  Theorem      
      ⊢ ∀x. sgn (sgn x) = sgn x
```

NOTE: although the new `REAL_ARITH` is globally switched off in `realLib`, to get the above proofs `realTheory` must open `RealArith` and use the intermediate version of `REAL_ARITH` to prove the above theorems.   Since most theorems in `realTheory` are quite small and simple, even the new `REAL_ARITH` is slower, the time difference is negligible within the full HOL build.

binghe:
# Internals of `REAL_ARITH`

Surprisingly, there's no magic (or any advanced algorithms) behind `REAL_ARITH`, despite its extremely usefulness. For any real arithmetic statements about Boolean connectives, real ordering and real operators. The following steps are done inside `REAL_ARITH`:

1. The input term is negated (and the whole prover is a "refuter") and normalized into DNF (disjunctive normal forms) of primitive equations or inequations in the form of `P(X) = 0` or `P(X) > 0` or `P(X) >= 0`, where `P(X)` must be a polynomial as a summation of "variables" multiplied with literal coefficients. (Note that `x <> y` will become `x - y >0 \/ y - x > 0`.)   In particular, `REAL_ARITH` does understand multivariate polynomials, thus a term like `1 * x + 2 * (x * y)` will be treated like `1 * x + 2 * z`, losing the information about `x * y`.   This simplifier is exported `REAL_POLY_CONV` (from `RealField`) and can be also used for other purposes.


2. Given a (conjuctive) list of equations and "lt" and "le" inequations. The goal is to find a contradiction. And if all groups of the DNF led to contradictions than the original input term must be true as a proven theorem.  Inside `REAL_ARITH` there's a plug-able `REAL_LINEAR_PROVER` doing the job of finding contradictions. It's only heuristic and not complete. (In HOL-Light's `Examples/sos.ml` there's a more advanced `REAL_NONLINEAR_PROVER` which depends on external "SDP" solvers. See [1] for more details, and thanks Yong Kiam Tan for sharing the paper with me.)

This heuristic linear prover works by eliminating variables from the input inequations by summing them up with positive rationals multiplied.  In this way, the resulting inequations are still in the same shape (Note, in particular, that an `P(X) >= 0` inequation adding up with an `P(X) >0` inequation leads to another `P(X) > 0` ineqaution).    First, each equations are eliminated by adding up with all other inequations, each with an arbitrary chosen variable in them.

When there are only inequations, the algorithm works by choosing the variable with has minimal "blow up": the number of occurrences with positive coefficients, multiplied with the number of occurrences with negative coefficients.  Once such a variable is chosen, the algorithm adds up each pair of inequations where the variable has different sign of coefficients, so that in the resulting inequations the variable gets eliminated. The number of new inequations clear will cause a "blow up" on the total number of inequations at next rounds, and that's why the variable with minimal "blow-up" is chosen first.

Repeating this process, eventually all variables are elimianted and only two types of inequations with literal rational numbers are left, either `n > 0` or `n >= 0`, and if they actually do not hold, this will be a contradiction causing the entire proof branch complete with success.  This approach is guaranteed to terminate but does not guarantee any solvable problem can be actually solved.

The existing `REAL_ARITH` ported by Hurd basically does the same job.

Since `REAL_ARITH` doesn't understand the relationship between `x` and `inv x`, the end user is recommended to first use the global real simplifier (provided in `realSimps`) to eliminate `x * inv x` as much as possible, and then call `REAL_ARITH`.

[1] Harrison, J.: Verifying Nonlinear Real Formulas Via Sums of Squares. In: Schneider, K. and Brandt, J. (eds.) LNCS 4732 - Theorem Proving in Higher Order Logics (TPHOLs 2007). pp. 102–118. Springer, Berlin, Heidelberg (2007).

binghe:
In the last commits, "try the old code first and then fall back to the new code if the old doesn't succeed" is implemented (in `realLib.sml`, which can be easily adjusted in the future). Also, the `--otknl` buid should be fixed now.

binghe:
OpenTheory build is confirmed working, although this was not easy and has brought another 1,000+ lines of changes in `src/boss/prove_base_assumsScript.sml` and `src/real/prove_real_assumsScript.sml`.

binghe:
Good news: some lost performance is back.

In the last commit ([33d29dc](https://github.com/HOL-Theorem-Prover/HOL/pull/1043/commits/33d29dca20145718b89deb2d988675ecdc62b6f8)) I added (back) a conversion called `CACHE_CONV` which existed in the original code from HOL-Light but was omitted when I was porting the code to HOL4.  This `CACHE_CONV` was used in the conversion of input terms into NNF forms, and by this caching the build time of `real_topologyTheory` has been reduced from 384s to 220s (in my laptop), while the original `REAL_ARITH` takes about 110-120s.

P.S. By searching inside the entire HOL-Light repository, this `CACHE_CONV` was only used once in defining `REAL_ARITH`, thus must be a solution to a key bottleneck.

However, currently the default `REAL_ARITH` still first try to old version, and then fallback to the new version.  I think this is still the best strategy currently.

mn200:
Thanks for your continuing work on this!

binghe:
Viewing the comments in #1044 I fixed a porting mistake in `Grobner.sml` in the code for eliminating `PRE` and num subtractions. In summary, the following new arithmetics elimination theorems are ported from HOL-Light, all have single `P` at rhs (thus no term blowup after the elimination):

```
   [PRE_ELIM_THM']  Theorem      
      ⊢ P (PRE n) ⇔ ∀m. n = SUC m ∨ m = 0 ∧ n = 0 ⇒ P m

   [PRE_ELIM_THM_EXISTS]  Theorem      
      ⊢ P (PRE n) ⇔ ∃m. (n = SUC m ∨ m = 0 ∧ n = 0) ∧ P m

   [SUB_ELIM_THM']  Theorem      
      ⊢ P (a − b) ⇔ ∀d. a = b + d ∨ a < b ∧ d = 0 ⇒ P d

   [SUB_ELIM_THM_EXISTS']  Theorem      
      ⊢ P (a − b) ⇔ ∃d. (a = b + d ∨ a < b ∧ d = 0) ∧ P d

   [DIVMOD_ELIM_THM]  Theorem      
      ⊢ ∀P m n.
          0 < n ⇒
          (P (m DIV n) (m MOD n) ⇔ ∀q r. m = q * n + r ∧ r < n ⇒ P q r)
   
   [DIVMOD_ELIM_THM']  Theorem      
      ⊢ ∀P m n.
          0 < n ⇒
          (P (m DIV n) (m MOD n) ⇔ ∃q r. m = q * n + r ∧ r < n ∧ P q r)
```
Perhaps a better `SUB_COND_ELIM_CONV` can be worked out from the above `SUB_ELIM_THM_EXISTS'`.

mn200:
In the context of the arithmetic decision procedure, these terms are going to be converted to DNF, and so the P's will end up being duplicated anyway.  But if the code works with these versions, I have no objection.

mn200:
And even if the `P`s do end up being duplicated later, it's nicer not to have the doubling happen while nat-subtractions and conditionals are being removed.  It'd be interesting to compare performance of a revised `SUB_CONV_ELIM_CONV` on the example in #6.

binghe:
> And even if the `P`s do end up being duplicated later, it's nicer not to have the doubling happen while nat-subtractions and conditionals are being removed. It'd be interesting to compare performance of a revised `SUB_CONV_ELIM_CONV` on the example in #6.

sure, I will take a look at the example in #6 and see what I can do.

PR #1046, opened 2022-06-28
Prove a roundoff error bound for denormal floating-points

HeikoBecker:
This PR adds a theorem similar to `relative_error` in `lift_ieeeScript`. The purpose is to provide an upper bound on the maximum roundoff error from rounding to a denormal number in floating-point arithmetic.

The theorem is a existentially quantified version of `error_bound_norm_strong`.


mn200:
Thanks for this!

Issue #1045, opened 2022-06-24
OpenTheory issues after moving some definitions and theorems into realaxTheory


binghe:
Hi,

I'm sorry to report that, the OpenTheory packages (in particular, `hol-real`) built after #1043 (the new `REAL_ARITH_TAC`) now have a lot of leaking assumptions not covered by OpenTheory base package and `hol-base` et al.. Note that this was only caused by my willing of using `REAL_ARITH` in `realScript.sml`, in which many definitions and theorems have to be moved to `realaxScript.sml` which is (by design) not included in OT packages.  Actually I was trying to fix this issue when I saw the OT building failure of `$(HOLDIR)/src/real/prove_real_assumsScript.sml` but apparently my fixes are not correct: indeed I have proved all leaking assumptions, but it seems that the logical constants (mostly real operators) in these assumptions are not the correct one. Below is the error output when installing the current `hol-real`:

```
opentheory install --reinstall hol4-real.thy
uninstalled package hol-real-1.2
WARNING: 7 axioms:
  |- realax.real_0 = 0
  |- realax.real_1 = 1
  |- realax.inv 0 = 0
  |- !x y. realax./ x y = x * realax.inv y
  |- prove_real_assums.real_of_num 0 = realax.real_0 /\
     !n.
       prove_real_assums.real_of_num (suc n) =
       prove_real_assums.real_of_num n + realax.real_1
  |- 0 = realax.real_0 /\
     !n. fromNatural (suc n) = fromNatural n + realax.real_1
  |- !x. ~(x = 0) ==> realax.inv x * x = 1
WARNING: redundant show: "Data.List"
WARNING: 2 ungrounded external constants: prove_real_assums.real_of_num realax./
WARNING: 51 unsatisfied assumptions:
  |- !x. ~(x < x)
  |- 0 < 1
  |- 0 <= 1
  |- !n. 0 <= fromNatural n
  |- ~(1 = 0)
  |- !n. 0 < fromNatural (suc n)
  |- !n. abs (fromNatural n) = fromNatural n
  |- !x. 0 <= x * x
  |- !x. x + 0 = x
  |- !x. prove_real_assums.real_of_num 0 + x = x
  |- !x. x ^ 0 = prove_real_assums.real_of_num 1
  |- !x. x * 0 = 0
  |- !x. x + ~x = 0
  |- !x. 0 * x = 0
  |- !x. ~x + x = prove_real_assums.real_of_num 0
[...]

PR #1051, opened 2022-07-07
Fixed OpenTheory build issues (#1045) introduced in #1043

binghe:
Hi,

this PR fixes the OpenTheory build issues described in #1045, i.e. a lot of leaking assumptions when installing `hol-real`. As I mentioned in #1045, these issues were introduced when I was moving some real theorems from `realTheory` to `realaxTheory` but `realaxTheory` is not included in the "hol-real" OT package.

The key of my current solution is to build `realax.ot.art` and include it in "hol-real" OT package. Only in this way, I don't need to copy a large piece of code from `realaxScript.sml` to `prove_real_assumsScript.sml` and the number of real assumptions can also be minimized (from 76 back to 21, the same as before #1043).

Let's recall why previously `realaxTheory` was excluded from `hol-real`: that's because `realaxTheory` contains a lot of `treal` and `hreal` theorems and the higher order quotient from `treal/hreal` to the final `real` type, and these theorems shouldn't be included in OT articles because HOL4 wants to use OpenTheory's own `real` package as the basis of real numbers, and the _bridge_ is done by `$(HOLDIR)/src/prove_real_assumsScript.sml`.

So, I did two things to `realaxScript.sml`: first, I move all `HREAL` and `TREAL` theorems to the upstream `hrealTheory`; this should be fine as nobody except `realax` itself should use any of these theorems.  Second, I created a big `realax.otd` where I carefully eliminated all theorems irrelevant with the OT exports. In another word, all theorems using `hreal` and `treal` stuff are either skipped or with proofs deleted in the final `realax.ot.art`. (the movements of `HREAL` and `TREAL` theorems actually makes `realax.otd` shorter, because otherwise also these theorems must be explicitly excluded.)

Once `realax.ot.art` is ready, in `$(HOLDIR)/src/prove_real_assumsScript.sml` I recovered the old, shorter contents of this file (before #1043) but still keep my other improvements to this file (i.e. reading axioms from OT `real-1.61` and only use them to prove other assumptions.).  Now `prove_real_assumsScript.sml` is as short as before, and more importantly, in the future, if more (or less) theorems were moved from `realTheory` to `realaxTheory` (because `RealAirth` needs it, maybe), *no further changes* are required in `prove_real_assumsScript.sml`.

P. S. in `ternaryComparisonsTheory` I added the following two new theorems to handle the leaking assumptions when installing `hol-ring`, since I failed to find where these two theorems were actually used:
```
   [ordering_distinct1]  Theorem      
      ⊢ EQUAL ≠ LESS
   
   [ordering_distinct2]  Theorem      
      ⊢ GREATER ≠ EQUAL
```

Regards,

Chun Tian


binghe:
For the first time, `prove_real_assumsScript.sml` are fully commented and I'm glad to say that I finally understood every line in this file and the idea (of Ramana Kumar @xrchz) behind them. For example, why at the beginning we have the following "strange" definitions:
```
val REAL_0 = new_definition("REAL_0",concl realaxTheory.REAL_0);
val REAL_1 = new_definition("REAL_1",concl realaxTheory.REAL_1);
```

Now I have left the following comments for future maintainers:
```
(* NOTE: the purpose of the following two definitions is to replace HOL4's
   "invalid" definitions of “real_0” and “real_1” (in realaxTheory), from

   |- real_0 = real_ABS treal_0
   |- real_1 = real_ABS treal_1

   to

   |- real_0 = real_of_num 0
   |- real_1 = real_of_num 1

   where “real_of_num” (Number.Real.fromNatural) is provided by OT's real-1.61.
  (In HOL4, “real_0|1” are more primitive and “real_of_num” is defined upon
  “real_0|1”. In OpenTheory, on the other hand, there's no “real_0|1” at all,
   while Number.Real.fromNatural is defined in a completely different way.)

   For this purpose, the existing constants “real_0” and “real_1” defined in
   realaxTheory must be deleted when exporting OT articles (by the following
   two lines in realax.otd):

   delconst real_0
   delconst real_1

   Furthermore, when reading OT articles generated by HOL4, all occurrences of
  “realax$real_0” and “realax$real_1” must be redirected to the new definitions
   here. This is done by the ML function "const_name" (below), having the
   following two lines:

   | const_name (["HOL4","realax"],"real_0") = {Thy=Thy,Name="real_0"}
   | const_name (["HOL4","realax"],"real_1") = {Thy=Thy,Name="real_1"}

   Finally, all occurrences of “real_of_num” in the present theory will be
   also replaced by Number.Real.fromNatural. This is done by the following
   line in $(HOLDIR)/src/opentheory/hol4.int:

   const "HOL4.realax.real_of_num" as "Number.Real.fromNatural"

 *)
```

xrchz:
Bravo. Thanks for doing all this and sorry my contributions weren't more easily decipherable.

binghe:
> Bravo. Thanks for doing all this and sorry my contributions weren't more easily decipherable.

You're welcome (and thanks for your answers to all my related questions, mostly in private).

mn200:
Sorry to have left this so long!

binghe:
> Sorry to have left this so long!

It's OK (except that I was a little worry about your security (and my future) ^_^)

Issue #1020, opened 2022-04-05
Neater filing of `Holmake` artefacts


hrutvik:
`Holmake` generates `*Theory.{uo,ui,sml,sig}` files of little use to the user. To avoid clutter, these could be stored in the existing `.HOLMK` directory produced by `Holmake`.

There is an argument for keeping `*Theory.sig` files around for inspection and so `ls` easily reveals whether files have been built in the current directory.

As suggested by @myreen in PureCake meeting.

mn200:
See 32bbfddd as a start on this.

binghe:
@mn200 Will those generated OpenTheory articles stay inside `.holobjs` or outside, by your design? Besides, in some `Holmakefile`s I explicitly wrote `touch *.uo` and `touch *.ui` to let OT articles join the dependency checking, so I'm not sure if `--otknl` is still OK or broken after your these (good) changes. Whatever possible issues there, should be identified after PR #1051 get merged.

mn200:
The things moved into the `.holobjs` directory are chosen by the logic in https://github.com/HOL-Theorem-Prover/HOL/blob/83b8c3c331f2b49f8c9ab8622aca9b3cecaf7dc8/tools/Holmake/poly/HFS_NameMunge.sml#L43-L51
If the `changep` variable is true, the remapping occurs.

mn200:
If the wildcard-ing is really necessary, then instead of `*.uo` _etc_, you can use `./$(HOLOBJDIR)/*.uo`.

PR #1055, opened 2022-09-02
Fix --otknl build due to neater filtering of Holmake artifacts (#1020)

binghe:
This small patch further fixed OpenTheory building (`--otknl`) due to movements of `Holmake` artifacts (`.ui`, `.uo` files).


mn200:
Thanks!

PR #1056, opened 2022-09-08
Add bitArithLib to src/real

HeikoBecker:
This PR adds a new library bitArithLib.sml to the folder `src/real`. The library implements efficient Karatsuba-style multiplications for large numbers. It can be used to evaluate multiplications of type `:real` and type `:num`.

Key to the library is a translation from natural numbers to bitstrings which are then multiplied using the Karatsuba multiplication algorithm ([https://en.wikipedia.org/wiki/Karatsuba_algorithm](https://en.wikipedia.org/wiki/Karatsuba_algorithm)).

The library was used for some benchmarks of the Dandelion tool ([github.com/HeikoBecker/Dandelion/](https://github.com/HeikoBecker/Dandelion/)).

The code of the library was developed with the help of @mn200 and @myreen.


mn200:
Please put test-cases into the `src/real/selftest.sml`.  The `convtest` entry-point is probably the right thing to start with.

mn200:
Thanks for your work on this!

binghe:
Hi, if one wants to use this `bitArithLib` to accelerate `RealArith` (or `RealField`) when dealing with multiplications of large numerals, is it _too late_ to do like this in user code?
```
open bitArithLib
open RealArith
```

HeikoBecker:
@binghe It depends. First of all, the numbers need to be really large (i.e. ~100 digits) for bitArithLib to be really useful. But other then that, this should work the way you are describing it.
One catch is that you **have to** load the library before you define the object that you want to evaluate, i.e. if you run 
```
EVAL  ``myFun largeNum1 largeNum2``
```
Your theory file defining `myFun` has to look like:
```
open bitArithLib;
val _ = new_theory "myTheory"
Definition myFun_def: ...
```
This is because `computeLib` keeps track of the conversions used for evaluation when a function is defined. I do not know the exact details of how this works, but this is what I learned from asking around.

binghe:
> @binghe It depends. First of all, the numbers need to be really large (i.e. ~100 digits) for bitArithLib to be really useful. But other then that, this should work the way you are describing it. One catch is that you **have to** load the library before you define the object that you want to evaluate, i.e. if you run
> 
> ```
> EVAL  ``myFun largeNum1 largeNum2``
> ```
> 
> Your theory file defining `myFun` has to look like:
> 
> ```
> open bitArithLib;
> val _ = new_theory "myTheory"
> Definition myFun_def: ...
> ```
> 
> This is because `computeLib` keeps track of the conversions used for evaluation when a function is defined. I do not know the exact details of how this works, but this is what I learned from asking around.
> 
> Addendum: If you only evaluate terms with `*:real->real->real` your suggested setup should work fine too btw.

Now I think I understood what API you have added: changes in the global `compset` (for `EVAL`) and two dedicated conversions, one for `:num` the other for `:real`.   And from your words they have no advantage over small numbers.   The `REAL_ARITH_TAC` (and related facilities) basically use the old existing conversions in `Arithconv` for arithmetics simplifications. To actually use your procedure, in theory one can define a new `REAL_ARITH` by calling `RealArith.GEN_REAL_ARITH` with customized conversions, e.g. the ones you added.

P. S. do you still have your large-number example (which you showed last time on Slack), which I cannot even input it directly from Terminal command-line or Emacs)?  Perhaps adding it into `selftest.sml` (or a dedicated test file) is a good direction to show the real power of your new additions?

HeikoBecker:
I still have the example, it is here: https://gist.github.com/HeikoBecker/8b54d4e30fc3a62a2161fdedbd76feab
While the `bitarithLib` can evaluate the term, it still takes very long, so I think it would just unreasonable prolong the testing sequence by being added to the `selftest.sml` file.

binghe:
HOL building on Moscow ML seems broken after this PR:

```
Theory "bitArith" took 27.6s to build
Holmake: Compiling bitArithTheory.sig
Holmake: Compiling bitArithTheory.sml
Holmake: Compiling bitArithLib.sml
! File bitArithLib.sig exists, but there is no signature constraint in bitArithLib.sml
*** FATAL: Build failed in directory /Users/binghe/ML/HOL.mosml/src/parallel_builds/core
```

I don't quite understand the error message ("no signature constraint ")...

binghe:
> HOL building on Moscow ML seems broken after this PR:
> 
> ```
> Theory "bitArith" took 27.6s to build
> Holmake: Compiling bitArithTheory.sig
> Holmake: Compiling bitArithTheory.sml
> Holmake: Compiling bitArithLib.sml
> ! File bitArithLib.sig exists, but there is no signature constraint in bitArithLib.sml
> *** FATAL: Build failed in directory /Users/binghe/ML/HOL.mosml/src/parallel_builds/core
> ```
> 
> I don't quite understand the error message ("no signature constraint ")...

OK, it seems that changing the first code line in `bitArithLib.sml` in the following way, will fix the compilation issue:
```
structure bitArithLib :> bitArithLib =
```

P. S. It looks like, in the theory of Standard ML, a structure containing functions and variables merely "implements" a signature, which usually has the same name with the structure but _not always_.

binghe:
Together with other fixes irrelevant with this PR, I will send another PR for cumulative changes fixing the Moscow ML builds.

mn200:
The convention of insisting on the same name for signatures and structures is a Moscow ML thing.  

In general, signatures are a way to control the degree to which implementation details and names escape from structures. 

Structures can definitely exist independent of signatures.  You can even play "silly games": 
```
structure S = struct val x = 10 val y = 10 end;

signature One = sig val x : int end;
signature Two = sig val y : int end;

structure T1 :> One = S;
structure T2 :> Two = S;
```

After doing this you can write `S.x`, `S.y`, `T1.x` and `T2.y`, but _not_ `T2.x`.  If the values were references, they'd be shared and could be updated _via_ one name, and have that change visible _via_ another name.

PR #1064, opened 2022-10-25
Generalized definition of measurable mappings in measure theory

binghe:
Hi,

Measurable functions or mappings is a fundamental concept in Measure Theory. In particular, random variables are nothing but measurable mappings from a probability space to a σ-algebra (usually Borel σ-algebra).

But currently its definition (`sigma_algebra$measurable`) is not very flexible:
```
   [measurable_def]  Definition
      ⊢ ∀a b.
          measurable a b =
          {f |
           sigma_algebra a ∧ sigma_algebra b ∧ f ∈ (space a → space b) ∧
           ∀s. s ∈ subsets b ⇒ PREIMAGE f s ∩ space a ∈ subsets a}
```
The problem is at the parts `sigma_algebra a ∧ sigma_algebra b` which requires that, when stating `measurable a b` (the set of measurable mappings from a to b), both `a` and `b` must be σ-algebras.  But sometimes one must work on generators of these σ-algebras, and a very useful lemma says "measurable over generators is equivalent to measurable over the σ-algebra generated". (This σ-algebra is the smallest one containing that generator as a system of sets.)  The hard part of this lemma is supposed to be provided by an existing theorem with correct proof:
```
   [MEASURABLE_LIFT]  Theorem      
      ⊢ ∀f a b.
          f ∈ measurable a b ⇒
          f ∈ measurable a (sigma (space b) (subsets b))
```
But if `measurable a b` implicitly requires that `sigma_algebra b`, we can see that `(sigma (space b) (subsets b))` will be nothing but `b` itself (cf. `SIGMA_STABLE`), then this theorem becomes meaningless (a set is subset of itself).

On the other hand, when `f ∈ measurable a b` occurs as a proof goal, for all the existing cases found in core HOL theories, the proofs (and their textbook versions) only focus on showing `∀s. s ∈ subsets b ⇒ PREIMAGE f s ∩ space a ∈ subsets a`. For the part `sigma_algebra a ∧ sigma_algebra b`, it's either directly provided as antecedents of the related theorem, or is trivial to prove.

In this PR, I have modified the above definition of `measurable` by removing the σ-algebra requirements:
```
   [measurable_def]  Definition      
      ⊢ ∀a b.
          measurable a b =
          {f |
           f ∈ (space a → space b) ∧
           ∀s. s ∈ subsets b ⇒ PREIMAGE f s ∩ space a ∈ subsets a}
```
With this changes, now the above `MEASURABLE_LIFT` is meaningful, by adding just one easy antecedent (making sure that `b` is indeed a system of sets satisfying the `subset_class` relation):
```
   [MEASURABLE_LIFT]  Theorem      
      ⊢ ∀f a b.
          sigma_algebra a ∧ subset_class (space b) (subsets b) ∧
          f ∈ measurable a b ⇒
          f ∈ measurable a (sigma (space b) (subsets b))
```

Many other theorems must be fixed (not hard) due to the above changes. Sometimes they require more antecedents (of σ-algebras), other times some trivial proof branches (after expanding `measurable_def`) cease to exist.  In particular, many theorems in `miller` and `diningcryptos` examples, previous no requirements of `prob_space p` when a probability space `p` is involved, now `prob_space p` must be added. But eventually the final toplevel theorems in these examples have no change in their statements at all. In other words, changed theorem statements only happen at intermediate levels.

Only with this PR, certain more advanced probability theorems can be proved. For example, to show a (finite) sequence of mutually independent random variables, after dividing into two disjoint sets, the functional composition of each sets as new random variables, are still independent.  This independent proof must be done on measurable mappings of carefully chosen  generators, which themselves are not σ-algebras.

Regards,

Chun Tian

### DIFF
--- a/doc/next-release.md
+++ b/doc/next-release.md
@@ -138,6 +138,14 @@ Incompatibilities:
     This change makes the naming consistent across all of HOL’s fragments.
     These names are used when referring to fragments in calls to `diminish_srw_ss`, when using `ExclSF` (see above), and in printing the values in the REPL.
 
+*   In `sigma_algebraTheory`, the definition of `measurable` has been generalized without
+    requiring that the involved systems of sets must be σ-algebras. This change allows the user to
+    express measurable mappings over generators of σ-algebras. (cf. `MEASURABLE_LIFT` for a related
+    important lemma.)  Existing proofs may break in two ways (both are easy to fix): 1. The need of
+    extra antecedents (usually easily available) when applying some existing measure and probability
+    theorems. 2. When proving `f IN measurable a b`, some proof branches regarding σ-algebras no
+    longer exists (thus the related proof scripts must be eliminated).
+
 * * * * *
 
 <div class="footer">
--- a/examples/diningcryptos/informationScript.sml
+++ b/examples/diningcryptos/informationScript.sml
@@ -65,9 +65,9 @@ val conditional_mutual_information_def = Define
 (* Reductions                                                                *)
 (* ************************************************************************* *)
 
-val finite_mutual_information_reduce = store_thm
-  ("finite_mutual_information_reduce",
-   ``!b p s1 s2 X Y. (POW (p_space p) = events p) /\
+(* NOTE: added ‘prob_space p’ into antecedents due to changes of ‘measurable’ *)
+Theorem finite_mutual_information_reduce :
+    !b p s1 s2 X Y. prob_space p /\ (POW (p_space p) = events p) /\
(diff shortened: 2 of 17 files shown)


binghe:
Please squash all commits (when possibly merge this PR) into just one, since many individual commit messages are meaningless.

jdyeager:
Looks good.
I look forward to having to actually declare some things to be sigma algebras instead of cheating my way out of mentioning that precondition.

mn200:
Thanks for your work on this!

PR #1068, opened 2022-10-27
Changed whitespace in order to satisfy -t2 flag.

jdyeager:
I needed to change some whitespace in order to pass the scrutiny of building with the `-t2` flag.


mn200:
Thanks!

PR #1069, opened 2022-10-31
Port of Viktor Vafeiadis's Armv8 memory model and proofs

acjf3:


acjf3:
I've moved it, together with ARM_security_properties, to be under "arm"

mn200:
Awesome, thanks!

PR #1037, opened 2022-05-12
HOL-Light's permutations.ml and vectors.ml (partial) ported to HOL4

binghe:
Hi,

more than 4 years ago (Jan 2018), a Chinese author @LiLiming ever sent a PR (#513) here about their porting work of HOL-Light's vector and matrix theories. The PR was closed due to missing necessary preliminary work, but actually there are many small mistakes in the code to be fixed, it definitely cannot run Kananaskis-6 although it may be "developed" there.

However, I was considering this work important, because HOL-Light's real-valued vector formalization is the basis of its `Multivariate` calculus, and matrices may have applications in discrete-time Markov Chains (DTMC) that one day I may also touch. Thus I saved their code in a branch of my fork of HOL repository, and recently I started to fix issues in these old code.

There were four ML scripts:

1. `permutation.sml` is a port of HOL-Light's `Library/permutations.ml`, a theory of permutations of set elements. I will explain soon the difference between HOL-Light's permutation and HOL4's existing `PERMUTES` (in `pred_setTheory`, as an overload of `BIJ f s s`, i.e. bijections between the same set)
2. `iterate.sml` is another port of `iterateTheory`. I checked this file and found no theorems beyond what we already have in src/real/iterateScript.sml`. Thus this file can be just thrown away.
3. `Vectors.ml` is a port of HOL-Lights' `Multivariate/vectors.ml` (up to at most 2018). It is this theory containing the formalization of vector and matrix arithmetics.
4. `Determinant.sml`, a theory of "determinants" (special application of square matrix) of Linear Algebra.

In this PR, I fully fixed `permutation.sml` (now `permutationTheory`) and partially (~50%) fixed `Vectors.ml` (now `vectorScript.sml`).   I didn't continue the work and stopped here (for now), because I found it's more urgent to improve HOL4's `REAL_ARITH_TAC` (see some comments in #1035).

# `permutationTheory`

In `permutationTheory` we have HOL-Light's version of permutations defined in this way:
```
permutes
⊢ ∀p s. p permutes s ⇔ (∀x. x ∉ s ⇒ p x = x) ∧ ∀y. ∃!x. p x = y
```
I added the following new theorem to show the relationship between this `permutes` and HOL4's existing `PERMUTES` in `pred_setTheory`:
```
permutes_alt
⊢ ∀f s. f permutes s ⇔ f PERMUTES s ∧ ∀x. x ∉ s ⇒ f x = x

permutes_alt_univ
[...]

PR #1067, opened 2022-10-27
add syntax support for relationTheory

acjf3:

PR #1071, opened 2022-11-02
Matrices in Euclidean space, and elementary linear algebra

binghe:
Hi,

this PR continues #1037 and ported the remaining theorems of HOL-Light's `vectors.ml` from old code in PR #513.  I made a separation between vector- and matrix-related theorems and now put all matrix theorems into a new `matrixTheory` in addition of the existing `vectorTheory`, in case the potential users may find the name "vector" misleading when they were looking for formalization of matrix.

For example, beside other basic matrix operations, the inversion of matrix is defined in this way:
```
MATRIX_INV_DEF
⊢ ∀A. MATRIX_INV A = @A'. A ** A' = 1m ∧ A' ** A = 1m
```
Not every matrix is invertible. Those invertible must satisfy the following predicate:
```
invertible_def
⊢ ∀A. invertible A ⇔ ∃A'. A ** A' = 1m ∧ A' ** A = 1m
```
Then there's a theorem saying that, for each invertible matrix, applying inversion operation twice you get the original matrix:
```
MATRIX_INV_INV
⊢ ∀A B. invertible A ⇒ MATRIX_INV (MATRIX_INV A) = A
```

(It would be nice to have an actual algorithm as a conversion for calculating matrix inversions, returning a theorem `|- MATRIX_INV A = ...` for any concrete invertible matrix A given as input.)

In `pred_setTheory`, beside minor cleanups, I added several theorems for HOL-Light compatibility. In this way many original HOL-Light proofs of matrix theorems basically need only grammar changes: the same tactics just work for HOL4 (also thanks to many previously compatibility fixes, including the new `REAL_ARITH_TAC`, well, mostly done by myself).

It remains to port the theory of determinants (a special application of `matrixTheory` to NxN square matrices).

Chun Tian


binghe:
I must say, that this whole matrix formalization (even the original work in HOL Light) is far from satisfaction, because "Linear algebra works over any field." [1], but currently this HOL theory is specified to real numbers.   If the field is rational numbers (the half-broken `ratTheory`, e.g.) or rational functions, in theory many existing proofs should still work.  One solution is link with `example/algebra` and parameterize all theorems and then specialize a version for reals. Another (better) solution is (perhaps) to define the `vectorTheory` and `matrixTheory` as ML functors and instantiate the specific theories for linear algebra over different fields (like the floating numbers of different bits).  The `REAL_ARITH_TAC` has similar possible extensions, as it's implicitly used in some vector and matrix proofs.

[1] D. Cox, J. Little, D. O'Shea, Ideals, Varieties, and Algorithms, Third Edition,
      Springer Science & Business Media, 2013. doi:10.1007/978-1-4757-2693-0.

binghe:
The last commit is a fix for the remaining compilation issues caused by #1067 on `develop` under Moscow ML.

binghe:
The last commits added something in the next release notes for the present vector and matrix theories, and also for the new `REAL_ARITH` (#1043).

mn200:
I agree that a nice algebraic presentation of this would be nice.  I'm not sure that the functor-based approach is particularly good.  We did use to do our theory of _n_-bit words that way, and it was rather awkward.  Note that though there is a functor in `src/floating-point` it is only used to generate `..Syntax` structures, not the actual theories, which all just build on the existing words development and are polymorphic over word sizes that way.

mn200:
Anyway, thanks for getting all this done!

PR #1075, opened 2022-12-09
Various itree goodies

IlmariReissumies:
Monad operators + monad laws, loop combinators, weak bisimulation, and a proof that tauless itrees are isomorphic to itrees up-to weak bisimulation.


mn200:
Supervisor just couldn't resist doing the work that the student was going to attempt? :)

PR #1076, opened 2022-12-09
Add `Q.LIST_REFINE_EXISTS_TAC`

### LINKED ISSUE #1021: List version of `Q.REFINE_EXISTS_TAC`
This issue is about implementing a version of `Q.REFINE_EXISTS_TAC` called `qrefinel : term quotation list -> tactic`, which:
- Attempts to refine a goal consisting of nested top-level existentials with the quotations passed to it
- If the quotation is an underscore, leaves the existential unchanged

In this way `qrefinel` is similar to `qexistsl_tac`, but additionally provides the ability to choose exactly which quantifiers to instantiate.

hrutvik:
And `bossLib` shorthand, `qrefinel`. Fixes #1021.

There are quite a few potential variable shadowing problems with this kind of tactic - hopefully I've accounted for all of them...

Also update the cheatsheet - fix some typos and add:
- `qrefinel` / `qexistsl`
- `SCONV` / `SRULE`
- `Cong`
- `MK_COMB_TAC`
- `ASSUME_NAMED_TAC` / `LABEL_{,X_}ASSUM`

### DIFF
--- a/Manual/Cheatsheet/cheatsheet.md
+++ b/Manual/Cheatsheet/cheatsheet.md
@@ -115,7 +115,7 @@ If these are still too blunt, we can use *conversions* to carry out surgical rew
 Conversions are functions of type `term -> thm`, such that applying a conversion to a term `t` produces a theorem `⊢ t = t'`.
 These can be converted into tactics using <code>CONV_TAC <i>conversion</i></code>.
 There are many conversions and conversion combinators - an exhaustive list can be found in the HOL4 documentation, mostly listed under the `Conv` structure.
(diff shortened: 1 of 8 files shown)


mn200:
The "easier to achieve" comment above assumes you package up that tech. with some sort of useful helper function.  Perhaps you have a `refine_expat` that does the work, given a refined pattern that occurs in the goal.  Not that `refine_existsl` wouldn't also be useful in circumstances where the conjunct-patterns are too much work to write out.

mn200 (approved):

PR #1078, opened 2022-12-16
Improved `LLN` definition, fixing some related proofs

binghe:
Hi,

For the statements of various version of the "Law of Large Numbers", I ever introduced the constant `LLN` whose definition was:
```
Definition LLN_def :
    LLN p X convergence_mode =
    let Z n x = SIGMA (\i. X i x) (count n) in
      ((\n x. (Z (SUC n) x - expectation p (Z (SUC n))) / &SUC n) --> (\x. 0))
       (convergence_mode p)
End
```
In some actual LLN proofs, the above `let Z n x` will become an abbreviation `Z` representing the partial sums of random variable X. Usually one expects the "first" of such partial sum should be X itself, but unfortunately this is not true: `Z 0 x` is always zero, since `SIGMA (\i. X i x) (count 0) = SIGMA (\i. X i x) {} = 0`, and I found this fact has caused some mis-alignments between textbook proofs and their formalizations.

In this PR, I changed above definition of `LLN` into the following equivalence form:
```
Definition LLN_def :
    LLN p X convergence_mode =
    let Z n x = SIGMA (\i. X i x) (count (SUC n)) in
      ((\n x. (Z n x - expectation p (Z n)) / &SUC n) --> (\x. 0))
       (convergence_mode p)
End
```
One can see that, after expanding `Z` the two definitions are literally the same. However, in the actual proofs one now have `Z` as the "correct" partial sums: `Z(0) = X(0)`, `Z(1) = X(0) + X(1)` and so on.    I took the hard way to go over all LLN proofs again and fixed them. At the end, I found the overall proof size shrunk a little after this change, and some previously hard constant estimations when doing `Q.EXISTS_TAC`, now becomes more reasonable and straightforward (see the ending part of `SLLN_uncorrelated`.)

I also did some minor cleanups at various places `src/probability`. In particular, some lemmas in `util_probTheory` are generalized and merged with each other.   In `extrealTheory`, I added several new extreal lemmas (not all of them are immediately used), e.g. `pow_pow: |- !x m n. (x pow m) pow n = x pow (m * n)`.

--Chun

### DIFF
--- a/examples/probability/large_numberScript.sml
+++ b/examples/probability/large_numberScript.sml
@@ -1,7 +1,7 @@
 (* ========================================================================= *)
 (* The Laws of Large Numbers (for Uncorrelated and I.I.D. Random Variables)  *)
 (*                                                                           *)
-(* Author: Chun Tian (binghe) <binghe.lisp@gmail.com> (2020 - 2021)          *)
+(* Author: Chun Tian (binghe) <binghe.lisp@gmail.com> (2020 - 2022)          *)
 (* Fondazione Bruno Kessler and University of Trento, Italy                  *)
 (* ========================================================================= *)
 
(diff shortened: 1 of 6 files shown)


mn200:
Thanks for your work on this!

PR #1079, opened 2022-12-20
Docker-based CI for GitHub Actions

binghe:
Hi,

This PR proposes a possible CI solution, which added a GitHub Action workflow called `docker-ci.yml` (under `.github/workflows/`). First of all, please note that, if other persons propose other CI solutions, there's no conflict, as long as they put their workflow as a different file name under `.github/workflows/`.

After some investigations, I found the Docker-based workflow has some advantages:

1. To build HOL in a cloud-based environment (usually a Linux virtual machine), beside basic packages, the PolyML (optionally with Mlton) or Moscow ML must also be installed.   Mlton has officially downloadable Linux packages, while PolyML and Moscow ML should be built from the official source code.   Using docker, we can first build a "base" image which contains all necessary packages and these ML installations.  This base image is defined in `developers/docker-ci/base/Dockerfile` and can be pulled to any user's Docker Hub repository.

2. Using docker, the workflow definition itself (for GitHub Actions) is basically some boilerplate code: all application-specific building details are provided by another docker image which actually build HOL, possibly with various options. This docker image is defined in `developers/docker-ci/Dockerfile` whose `FROM` is based on the above "base" docker image, currently at my personal Docker Hub repository [`binghelisp/hol-dev:latest`](https://hub.docker.com/repository/docker/binghelisp/hol-dev) but can be easily switched to any other.

3. Using docker, after the CI build, all the artifacts are packed into a new docker image and uploaded to the above mentioned Docker Hub repository, tagged with Git branch name + commit hash. User can easily pull the docker image into their local machine and run it or check the artifact files.

Some sample CI build can be found here: https://github.com/binghe/HOL/actions

To fully setup the CI and see it working from the "Actions" tab, after merging the code, HOL maintainer (@mn200) should create (if not having already) a (free) Docker account and create a new public Docker Hub repository, named `hol-dev` (in this way minimal changes are need). Then do the following steps:

1. In Docker Hub, create a new "Access Tokens" (from User Settings -> Security). What's being created is a password-like string, which must be copied and later pasted into GitHub. The name of the token is not important, can be "hol", for instance.

2. In GitHub, go to "Settings" of HOL, then "Security -> Secrets -> Actions", then click "New repository secrets" twice, to create the following named secrets: 1) `DOCKER_HUB_USERNAME`, whose value is your docker ID.  2) `DOCKER_HUB_ACCESS_TOKEN`, whose value is the token string given by the previous step in Docker Hub.

This should be enough for getting the CI working up.  If the Docker base image is changed, then there's no need to depend on my personal image, and the line "FROM binghelisp/hol-dev:latest" in one of the `Dockerfile` can be updated to HOL maintainer's own docker images.

Hope this helps,

Chun Tian

### DIFF
--- a/.github/workflows/docker-ci.yml
+++ b/.github/workflows/docker-ci.yml
@@ -41,3 +41,7 @@ jobs:
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+          build-args: |
+            SML=poly
+            # NOTE: the arg-value cannot be quoted here:
+            BUILDOPTS=--expk -j2
--- a/.github/workflows/docker-ci.yml
+++ b/.github/workflows/docker-ci.yml
@@ -1,0 +1,43 @@
+name: docker-ci
+
+on:
+  pull_request:
+    branches: [ develop ]
+  push:
+    branches:
+      - "master"
+      - "develop"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: ${{ secrets.DOCKER_HUB_USERNAME }}/hol-dev
+          tags: |
+            # dynamically set the branch name as a prefix
+            type=sha,prefix={{branch}}-
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_HUB_USERNAME }}
+          password: ${{ secrets.DOCKER_HUB_ACCESS_TOKEN }}
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Build and push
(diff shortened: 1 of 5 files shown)


binghe:
With latest updates, both `--stdknl` and `--expk` builds are triggered (in parallel) on new commits or pull requests.

mn200:
Thanks for putting in the time to do this.  I've followed instructions and hope to see it working shortly!

binghe:
I see the CI actions already working. You may consider adding `-t` to the `BUILDOPTS=` lines in the `docker-ci.yml` file. This will at least make sure no Unicode and Tab break on new commits.

binghe:
Also, currently HOL building with Moscow ML is too slow: it goes beyond the 6-hour limit on GitHub Actions (I estimate the total time is about 7-8 hours using the same hardware.)   This is even slower than HOL builds for Linux ARM64 using QEMU emulator on AMD64 Linux, which is about 5 hours on GitHub Actions, at least 10 times slower than native builds.

If Moscow ML builds can also benefit multi-processing like PolyML, i.e. supporting `-jN`, then the building time will shorten to 3.5-4 hours, then becomes possible for CI builds using GitHub Actions.

PR #1080, opened 2022-12-24
Alternative definitions of `indep_vars` (totoal independence of r.v.'s)

binghe:
Hi,

This PR is independent of the other existing one and improves the definitions of (total) independent random variables.

Recall that a number of independent events `A n` where `n IN J` (an possibly infinite index set) is defined below:
```
   [indep_sets_def]  Definition      
      ⊢ ∀p A J.
          indep_sets p A J ⇔
          ∀N. N ⊆ J ∧ N ≠ ∅ ∧ FINITE N ⇒
              ∀E. E ∈ N --> A ⇒
                  prob p (BIGINTER (IMAGE E N)) = ∏ (prob p ∘ E) N
```
Note that, the independence of an infinite number of events is defined by choosing an arbitrary *finite* subset of them.

Furthermore, a number of (total) independent random variables was defined below, on top of the above `indep_sets`:
```
   [indep_vars_def]  Definition
      ⊢ ∀p X A J.
          indep_vars p X A J ⇔
          ∀E. E ∈ J --> subsets ∘ A ⇒
              indep_events p (λn. PREIMAGE (X n) (E n) ∩ p_space p) J
```
If we expand `indep_events` from the above definition, we get this:
```
val it =
   |- !p X A J.
        indep_vars p X A J <=>
        !E. E IN J --> subsets o A ==>
            !N. N SUBSET J /\ N <> {} /\ FINITE N ==>
                prob p (BIGINTER (IMAGE (\n. PREIMAGE (X n) (E n) INTER p_space p) N)) =
                PI (prob p o (\n. PREIMAGE (X n) (E n) INTER p_space p)) N:
```
This definition is NOT wrong, only that the requirement `E IN J --> subsets o A` (here `-->` is `DFUNSET`, meaning `!n. n IN J ==> E n IN subsets (A n)`) is a bit too strong: for the conclusion part `prob p = PI ...` only `E IN N --> subsets o A` (where `N` is a subset of `J`). _Somehow_, with the above definition it's not easy (or straightforward) to prove an easy fact that, **"if a set of random variables is totally independent, so is any subset of them."**:
```
   [indep_vars_subset]  Theorem      
      ⊢ ∀p X A s t. indep_vars p X A t ∧ s ⊆ t ⇒ indep_vars p X A s
```

In this PR, I changed the definition of `indep_vars` into the following one without using `indep_sets`:
```
   [indep_vars_def]  Definition      
      ⊢ ∀p X A J.
          indep_vars p X A J ⇔
          ∀E N.
            N ⊆ J ∧ N ≠ ∅ ∧ FINITE N ∧ E ∈ N --> subsets ∘ A ⇒
            prob p
              (BIGINTER (IMAGE (λn. PREIMAGE (X n) (E n) ∩ p_space p) N)) =
            ∏ (prob p ∘ (λn. PREIMAGE (X n) (E n) ∩ p_space p)) N
```
The idea is to move the clause `E IN J --> subsets o A` into deeper level, after the quantification of `N`, and then change it to `E IN N --> subsets o A`. With this new definition, now the theorem "indep_vars_subset" can be easily proved:
```
Theorem indep_vars_subset :
    !p X A (s :'index set) (t :'index set).
       indep_vars p X A t /\ s SUBSET t ==> indep_vars p X A s
Proof
    RW_TAC std_ss [indep_vars_def, IN_DFUNSET, indep_events_def]
 >> FIRST_X_ASSUM irule >> simp []
 >> PROVE_TAC [SUBSET_TRANS]
QED
```

Then, with more proof steps, it turns out that, the new and old definitions are equivalent, as shown by the following new theorem: (thus the intuitive connection with `indep_sets` is still maintained)
```
   [indep_vars_alt_indep_events]  Theorem      
      ⊢ ∀p X A J.
          (∀n. n ∈ J ⇒ sigma_algebra (A n)) ⇒
          (indep_vars p X A J ⇔
           ∀E. E ∈ J --> subsets ∘ A ⇒
               indep_events p (λn. PREIMAGE (X n) (E n) ∩ p_space p) J)
```

Furthermore, when `indep_vars p X A J` itself is a proof goal, it's not easy to prove it with an arbitrary subset `N` of `J`.  In case the index set is `univ(:num)`, the set of all integers, a much easier alternative definition exists:
```
   [indep_vars_alt_univ]  Theorem      
      ⊢ ∀p X A.
          prob_space p ∧ (∀n. sigma_algebra (A n)) ∧
          (∀n. random_variable (X n) p (A n)) ⇒
          (indep_vars p X A 𝕌(:num) ⇔
           ∀E n.
             E ∈ count1 n --> subsets ∘ A ⇒
             prob p
               (BIGINTER
                  (IMAGE (λn. PREIMAGE (X n) (E n) ∩ p_space p) (count1 n))) =
             ∏ (prob p ∘ (λn. PREIMAGE (X n) (E n) ∩ p_space p)) (count1 n))
```
The above theorem is invaluable when a countable sequence of r.v.'s is to be constructed: one can first construct the first `n` independent r.v.'s (from `X(0)` to `X(n)`), and then construct `X (n+1)`, etc.

The `count1` is an "inclusive" version of `count` as an overload. I found it appearing frequently in probability proofs. `count1 n` is never an empty set, as `count1 0 = {0}`. Thus now it's in `util_probTheory`:
```
Overload count1 = “\n. count (SUC n)”;
```
(NOTE: It's no helpful to define it as `count o SUC` because after any `rw []` the `o` gets eliminated.)

To help potential users who wants to guess the "definition" of `count1` by searching `count1_def`, the following "fake" definition is added:
```
   [count1_def]  Theorem      
      ⊢ ∀n. count1 n = {m | m ≤ n}
```

P.S. There could be minor incompatibilities with existing user code but I believe the definition of `indep_vars` is not frequently used. There's no change in any statement of any existing theorems about `indep_vars`.

--Chun

### DIFF
--- a/src/probability/probabilityScript.sml
+++ b/src/probability/probabilityScript.sml
@@ -3100,12 +3100,177 @@ Proof
  >> METIS_TAC [SUBSET_DEF]
 QED
 
-(* 9. extension of `indep-rv`: total independent random variables. *)
-val indep_vars_def = Define
-   `indep_vars p X A (J :'index set) <=>
-      !E. E IN (J --> (subsets o A)) ==>
-          indep_events p (\n. (PREIMAGE (X n) (E n)) INTER p_space p) J`;
+(* 9. extension of `indep-rv`: totally/mutually independent r.v.'s
 
+  cf. indep_vars_alt_generator for a weaker equivalent condition for testing
+      independence.
+
+  NOTE: ‘indep_vars’ has been modified to make sure [indep_vars_subset] holds.
+
+  old definition:
+
+Definition old_indep_vars_def :
+    old_indep_vars p X A (J :'index set) =
+      !E. E IN (J --> (subsets o A)) ==>
+          indep_events p (\n. (PREIMAGE (X n) (E n)) INTER p_space p) J
+End
+
+> REWRITE_RULE [indep_events_def] old_indep_vars_def;
+val it =
+   |- !p X A J.
+        old_indep_vars p X A J <=>
+        !E. E IN J --> subsets o A ==>
+            !N. N SUBSET J /\ N <> {} /\ FINITE N ==>
+                prob p (BIGINTER (IMAGE (\n. PREIMAGE (X n) (E n) INTER p_space p) N)) =
+                PI (prob p o (\n. PREIMAGE (X n) (E n) INTER p_space p)) N:
+
+  new definition:
+ *)
+Definition indep_vars_def :
+    indep_vars p X A (J :'index set) =
+      !E N. N SUBSET J /\ N <> {} /\ FINITE N /\
+            E IN (N --> subsets o A) ==>
+            prob p (BIGINTER (IMAGE (\n. PREIMAGE (X n) (E n) INTER p_space p) N)) =
+            PI (prob p o (\n. PREIMAGE (X n) (E n) INTER p_space p)) N
+End
+
+(* NOTE: If a set of r.v.'s is (totally) independent, so is any subset of them.
+         With the new definition of ‘indep_vars’, this proof is very easy now.
+ *)
+Theorem indep_vars_subset :
+    !p X A (s :'index set) (t :'index set).
+       indep_vars p X A t /\ s SUBSET t ==> indep_vars p X A s
+Proof
+    RW_TAC std_ss [indep_vars_def, IN_DFUNSET, indep_events_def]
+ >> FIRST_X_ASSUM irule >> simp []
+ >> PROVE_TAC [SUBSET_TRANS]
+QED
+
+(* NOTE: the old and new definitions are actually equivalent, given ‘A n’ is indeed
+   a sigma-algebra (which can be actually weakened to ‘?x. x IN subsets (A n)’), or
+   ring, semiring, algebra, etc.
+ *)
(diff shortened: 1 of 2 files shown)


binghe:
I've fixed the code conflicts by merging with `develop` which now contains the CI workflow, which is then automatically triggered. But strange, the workflows failed at Docker Login action saying "Username and password required"?

binghe:
If you didn't delete `DOCKER_HUB_ACCESS_TOKEN` and `DOCKER_HUB_USERNAME`, which is needed by the workflow, then my conjecture of the above failed actions is the following: CI on pull requests only works when the branch is forked from a branch where CI is already configured and working.   Normally the workflow definitions should be directly edited on GitHub web page (instead of merging from PRs).

mn200:
I expect you're right about the CI failures.  Thanks for your work on this!

mn200:
Grr.  The history now has two merges in it.  Nevermind.

binghe:
> I expect you're right about the CI failures. Thanks for your work on this!

There are indeed some other issues. Honestly speaking I never tested CI builds on pull requests, nor I have any way to test building pull requests from other repositories. I've sent a new PR for the possible fix of these issues.

PR #1081, opened 2022-12-28
The new Redblackmap (aka HOLdict) structure with richer APIs

binghe:
Hi,

This PR adds a new red-black map data structure (ported from SML/NJ) with richer APIs than the existing `Redblackmap`. In  particular, when inserting a pre-sorted list of elements, its `fromOrderList` API can provide linear-time performance. [1]  I named this new SML structure (with signature) as `Redblackmap2` and also exported it as `HOLdict` (following `HOLset`), hopefully it will subsume all other map-like structures currently found in HOL code base in the future.

  [1] Hinze, R., 1999, May. Constructing red-black trees. In Proceedings of
      the Workshop on Algorithmic Aspects of Advanced Programming Languages,
      WAAAPL (Vol. 99, pp. 89-99).

This work wasn't done recently, instead, it was done when I was working on the new `REAL_ARITH` decision procedure (ported from latest code of HOL-Light, see PR #1043). In that work, I was in need of merging two maps, whose keys and values represents linear expressions with real-valued coefficients. For instead, the linear expressions `x + y + 1` and `-x + y - 1` can be seen as two maps: `[x |-> 1, y |-> 1, 1 |-> 1]` and `[x |-> -1, y |-> 1, 1 -> -1]`, where the special key `1` represents the constant item of the linear expressions.   If we want to "add" the two expressions, to get `2 * y` as the result, this is like merging two maps, adding the values for the same key: `[x |-> 0, y |-> 2, 1 |-> 0]` but then all keys with zero values must be removed, to get `[y |-> 2]` as the final result.

In `RealArith.sml` now there are totally three different solutions. The first one is to use `Termtab` (a specialization of `Table`, 2-3 trees), and the above "merge" operation is implemented by the following (long) code:
```
fun mergeWithoutZero f (m1 :linear_type) (m2 :linear_type) :linear_type = let
    fun add (key,y) tab =
        case lookup tab key of
           NONE   => update_new (key,y) tab
         | SOME x => let val sum = f (x,y) in
                         if sum = Arbrat.zero then
                             delete key tab
                         else
                             update (key,sum) tab
                     end;
in
    if is_empty m1 then m2
    else fold add m2 m1
end; (* mergeWithoutZero *)
```

The second solution is use the existing `Redblackmap`, and with the same idea as above (folding one map with the other, and remove the key when values summing to zero):
```
fun mergeWithoutZero f (m1 :linear_type) (m2 :linear_type) :linear_type = let
    fun add (key,y,tab) =
        case peek (tab,key) of
            NONE   => insert(tab,key,y)
          | SOME x => let val sum = f (x,y) in
                          if sum = Arbrat.zero then
                              fst(remove (tab,key))
                          else
                              insert (tab,key,sum)
                      end;
in
    if isEmpty m1 then m2
    else foldl add m2 m1
end
```

Only with the third solution, using the newly added `Redblackmap2` (aka `HOLdict`) I got a much nicer way to implement this special merge operation:
```
fun mergeWithoutZero f (m1 :linear_type) (m2 :linear_type) :linear_type =
let
    fun add (SOME x, SOME y) = let val z = Arbrat.+ (x,y) in
                                   if z = Arbrat.zero then NONE
                                   else SOME z
                               end
      | add (SOME x, NONE  ) = SOME x
      | add (NONE,   SOME y) = SOME y
      | add (NONE,   NONE  ) = NONE
in
    mergeWith add (m1,m2)
end
```
Here, the new API `mergeWith` provided by `Redblackmap2` takes option values. If after merging a value is zero, all user have to do is to return `NONE` instead, so that underlying `mergeWith` the key is automatically deleted.

The new `Redblackmap2` is completely compatible with the old `Redblackmap`, but with the following new APIs:
```
   [singleton ordr (i, v)] returns the specified singleton map.

   [insert' ((i, v), m)] extends (or modifies) map m to map i to v.
    Like insert, except that type of inputs are different. insert' is
    useful when the entry (i, v) to be inserted is given by a single
    variable.

   [insertWith comb (m, k, v)] inserts an item with a combining function
    to resolve collisions. The first argument to the combining function is
    the existing value, and the second argument is the value being inserted
    into the map.

   [insertWithi comb' (m, k, v)] is similar with insertWith, except that the
    combining function also takes the key as an argument.

   [inDomain (m, k)] returns true, if the key k is in the domain of the map m

   [first m] returns the first item in the map (or NONE if it is empty)

   [firsti m] is like first, except that the key is also returned.

   [listItems' m] returns an ordered list of the items in the map m.

   [listKeys m] returns an ordered list of the keys in the map.

   [collate ordr] is given an ordering on the map's range, it returns
    an ordering on the map (lexcographic).

   [unionWith f (m1, m2)] returns a map whose domain is the union of the
    domains of the two input maps, using the supplied function to define the
    map on elements that are in both domains.

   [unionWithi f (m1, m2)] is like unionWith, except that the supplied
    function also takes the key as an argument.

   [intersectWith f (m1, m2)] returns a map whose domain is the intersection of
    the domains of the two input maps, using the supplied function to define
    the range.

   [intersectWithi f (m1, m2)] is like intersectWith, except that the supplied
    function also takes the key as an argument.

   [mergeWith f (m1, m2)] merges two maps using the given function to control
    the merge. For each key k in the union of the two maps domains, the function
    is applied to the image of the key under the map.  If the function returns
    SOME y, then (k, y) is added to the resulting map.

   [mergeWithi f (m1, m2)] is like mergeWith, except that the function f also
    takes the key as an argument.

   [app' f m] applies the function f to the ranges of the map m in map order.

   [foldl' f e m] applies the folding function f to the ranges of the map m in
    increasing map order.

   [foldr' f e m] applies the folding function f to the ranges of the map m in
    decreasing map order.

   [filter p m] filters out those elements of the map m that do not satisfy the
    predicate p.  The filtering is done in increasing map order.

   [filteri p m] is like filter, except that the predicate p also takes the key
    as an argument.

   [mapPartial f m] maps a partial function f over the elements of a map in
    increasing map order.

   [mapPartiali f m] is like mapPartial, except that the function f also takes
    the key as an argument.

   [exists p m] checks the elements of a map with a predicate and return true
    if any element satisfies the predicate. Return false otherwise.
    Elements are checked in key order.

   [existsi p m] is like exists, except that the predicate p also takes the key
    as an argument.

   [all p m] checks the elements of a map m with a predicate p and return true
    if they all satisfy the predicate. Return false otherwise.  Elements
    are checked in key order.

   [alli p m] is like all, except that the predicate p also takes the key as an
    argument.

   [fromOrderedList ordr xs] returns a map that contains the (index, value)
    pairs in xs, whose keys have ordering ordr.  The list xs must be sorted by
    the keys w.r.t. the ordr and has no key duplications.

    NOTE: This function constructs the map in linear-time, but raises no error
    if the supplied xs is not ordered (the resulting map may lack some pairs).
```

There's no performance changes (for `REAL_ARITH`) after switching to this `HOLdict`, only the code is nicer. To keep small code changes, I didn't replace every usage of `Redblackmap` to the new `HOLdict` but only touched a few places (loading compiled HOL theories, e.g.) to show that frequent use cases of `Redblackmap` didn't break after switching to new code.

For the case of code under `src/HolQbf`, this code break was found by Moscow ML. That's because, in `Lib.sml` there's the following function `dict_topsort` being switched to `HOLdict` and `HolQbf` code calls that function:
```
local open HOLdict in
type ('a, 'b) dict = ('a, 'b) dict

fun dict_topsort deps =
   let
      val deps = transform (fn ls => ref (SOME ls)) deps
      fun visit (n, ls) =
         let
            val r = find (deps, n)
         in
            case !r of
               NONE => ls
             | SOME dl => let
                             val _ = r := NONE
                             val ls = List.foldl visit ls dl
                          in
                             n :: ls
                          end
         end
      fun v (n, _, ls) = visit (n, ls)
   in
      foldl v [] deps
   end
end (* local *)
```
The idea was to eliminate all usage of the old `Redblackmap` from `Lib.sml` so that whenever a dictionary is implicitly used, it automatically goes to the new `HOLdict`.   I searched all SML code in `src/`, only code in `src/HolQbf` calls once the `dict_topsort` function.

Overall speaking, a slightly better (in the sense of richer APIs) red-black map data structure has been added into HOL code base, but currently its benefits are quite limited - so far I haven't found any code place where the faster `fromOrderList` can be applied, i.e. the list of elements to be inserted is already sorted somehow.   All I can hope is two things: 1) future HOL developers may found those newly added APIs useful, 2) other duplicated implementations of map-like data structures can be removed from HOL code base. I let HOL maintainer to decide if this side work is worth to keep or not.

Chun Tian


mn200:
How does this code compare to the `Table` functor? If there's a clear winner among the four dictionary implementations we now have (`Table`, `Redblackmap`, `Binarymap` and this new `Redblackmap2`), I'd be keen to see us use just one.

binghe:
> How does this code compare to the `Table` functor? If there's a clear winner among the four dictionary implementations we now have (`Table`, `Redblackmap`, `Binarymap` and this new `Redblackmap2`), I'd be keen to see us use just one.

I believe all these map-like structures have similar performance. (for those 3 different solutions used in `RealArith`, I ever checked the theory building time on `real_topologyTheory`, no visible difference.). I don’t know if there’s some selftest code for `Redblackmap` and others, otherwise we can have a test.    In particular, `Table` is based on 2-3 trees but red-black trees, but I checked on Internet, it seems they have similar ideas.    The new `Redblackmap2` wins mostly because of its richer APIs, including the linear-time element insertion algorithm (which only works when the elements to be inserted is already sorted).

To have a unique piece of code, for the first step I think we can just delete the old `Redblackmap` code and let the new one have its name. This shouldn’t break anything and can be easily done in this PR.  For the rest parts, I can try to emulate their APIs using the ones of `HOLdict`, elinanating their the algorithm parts, shorten the related code files, later.

binghe:
To reduce code duplications, now I have completely removed the old `Redblackmap`.

mn200 (approved):

mn200:
Thanks for starting this process off.  I must say, I prefer the curried style of the Table interface.  I also like the movement of the accumulator argument in the fold function to be last.  This means that one can do things like 
```
  val newvalue = oldvalue |> fold f dict |> ...
```
because the partial application gives you something of type `'acc -> 'acc`.  Changing all of the existing folds to be this way around would be a lot of busy-work though.

I don't know quite what naming convention you could use to add these variants to the existing signature, given that the primed versions are already there (for versions without keys).  The `it`-prefix (standing for iterate) is possible, and would be consistent with our existing `itlist` and `rev_itlist` (though I've always hated the name `rev_itlist`).

That would make for entry-points called `itdict` _etc_ (and you could imagine adding `itdict'`...).  Despite the fact that `itlist` is effectively `foldr`, I'd make `itdict` do an in-order ("left to right") traversal.

PR #1087, opened 2023-01-10
Updated Docker-based CI workflow for more platforms

binghe:
First of all, I updated my docker base image `binghelisp/hol-dev` with more platforms (`linux/386` and `linux/arm64`) supported.  This is based on Docker's `buildx` technology: a single tag with multiple platforms behind it.

Then, the `self-runner` workflow can be manually triggered with platform options: `linux/386`, `linux/amd64` and `linux/arm64`.   HOL building on Linux/ARM64 is relatively slow on GitHub Actions' x86-based hardware, it took about 3 hours 33 minutes (without `-t`) [1].

[1] https://github.com/binghe/HOL/actions/runs/3881807957

As a side effect, on M1 Mac with Docker Desktop installed, now HOL users may pull the above docker image to obtain a HOL building environment with Moscow ML, PolyML and OpenTheory installed.

### DIFF
--- a/.github/workflows/docker-ci-for-pr.yml
+++ b/.github/workflows/docker-ci-for-pr.yml
@@ -1,0 +1,46 @@
+# A dedicated workflow for building pull requests
+name: docker-ci-pull_request
+
+on:
+  pull_request:
+    branches:
+      - "develop"
+
+jobs:
+  build-stdknl:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          context: .
+          file: developers/docker-ci/Dockerfile
+          push: false
+          build-args: |
+            SML=poly
+            BUILDOPTS=--stdknl -j2 -t
+
+  build-expk:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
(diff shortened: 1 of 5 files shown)


binghe:
Hmmm... CI triggered by pull requests still cannot build due to secret passing issues ("Error: Username and password required")

binghe:
OK, I will have to design another separate workflow just for the pull requests, which do not use Docker at all.

binghe:
Now I have created a dedicated simple workflow (`docker-ci-for-pr`) for building pull requests. The idea is to use the same `Dockerfile` for the actual building commands while there's no need to login to Docker Hub, because the resulting docker image is not uploaded to anywhere - the contributors should look at the building logs to see where the code breaks, just like in the days of Travis CI.  Meanwhile, the existing `docker-ci` workflow has been modified to only build new commits on `master` and `develop`.

As we can see, now even without merging the code, GitHub has already started using the new workflow to build the present PR. (This sounds like a security hole where arbitrary contributor may trigger heavy computations on GitHub's servers. However, we can assume HOL contributors won't do that, and essentially there's no way to stop such "attacks", for now.)

binghe:
> Curiously, neither my nor the HOL-Theorem-Prover organisation have recorded any VM minutes against their respective accounts. Perhaps numbers will appear (this in the "billing and plans" section of the accounts’ setting pages) eventually.

There won’t be any bill: “ GitHub Actions usage is free for standard GitHub-hosted runners in **public** repositories” [2]

[2] https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration

mn200:
Ah; I thought it was free up to some limit.  Good to know, thanks!

PR #1088, opened 2023-01-11
Proof of an implementation of float_is_nan

michael-roe:


mn200:
Thanks!

I’ll make the fixes myself after the merge, but a couple of things:
- we now prefer the "modern" syntax, so that `val foo = Q.store_thm("foo", term, tactic)` should instead be `Theorem foo: term Proof tactic QED` (with the keywords in column 0)
- `simp` should be idempotent, so that `tac \\ simp[] \\ simp[] \\ simp[]` should be the same as `tac \\ simp[]`; I guess you have three `simp`s there because the previous step generated three subgoals.

mn200:
See 21546e9f5 (which includes a commit message that typos the number of this PR, duh).

PR #1091, opened 2023-01-16
Proof of an implementation of float_is_zero

michael-roe:


mn200:
Can you try reducing `simp[] \\ simp[] \\ simp[]` into `simp[]` please, and see if it still works?

michael-roe:
I think most of the calls to simp[] are necessary, because they're applying to different branches of the case split. "rep simp[]" would probably work.

mn200:
_Au contraire_: in `tac1 \\ tac2` the second argument (`tac2`) is applied to all the sub goals generated by `tac1`. Interactively, you end up applying three separate tactics because of the generated subgoals.  If you want to apply three different tactics you have to do 
```
tac1 
>- (tac2a)
>- (tac2b)
>- (tac2c)
```

michael-roe:
Ok, thanks for that.

I'll revise the pull request. (It may take me a day or two to better understand how HOL tactical work).

mn200:
@binghe : there doesn't seem to be any suggestion that this PR has attracted any CI attention...

binghe:
I have no idea... the previous two PRs from @michael-roe were built correctly, and nothing has changed since then...

mn200:
If this PR was off a baseline HOL that didn't have the CI stuff merged in would that do it? Not a big deal.

binghe:
> If this PR was off a baseline HOL that didn't have the CI stuff merged in would that do it? Not a big deal.

The branch was indeed forked from very recent `develop`, and that whole idea was from my guess and was actually not true (a PR can indeed modify existing actions or creating new one, just like those I submitted for improving the CI itself).

I could be wrong but perhaps this is a (temporary) system-wide (GitHub) issue only affecting certain features (like pull-request trigger) when the related code (on GitHub's own side) were still using Node.js v12, which is long-obsoleted but only became unavailable from NPM very recently.

mn200:
Thanks for your work on this!

PR #1095, opened 2023-03-09
Speed up Prim_rec.prove_case_elim_thm

acjf3:
This was a significant bottleneck when defining large enumerated types (e.g. > 200 values).


mn200:
Thanks for this!

PR #1100, opened 2023-04-05
Vimhol respect `lastmaker`

arolle:
The vimhol plugin (with (n)vim terminal feature) now dynamically chooses
the first `hol` executable in the following order, when starting a new
interactive hol session:

1. hol relative to the lastmaker
  If `Holmake` was run on the edited theory before, its path resides in
  `.HOLMK/lastmaker`. vimhol probes that path with `hol` substituted for
  `Holmake`.
2. `$HOLDIR` environment variable
3. `hol` if in scope, e.g. from `PATH`

This feature removes the earlier requirement of setting a `$HOLDIR`
environment variable.

The path of the chosen hol is printed as a message for the user (via
`echomsg`), which is as well accessible through `:messages`. The message
has the shape `hol repl cmd(<id>): <cmd>` containing the related
terminal buffer id and the hol command.

Hope this is a reasonable somewhat analogous behaviour to [`hol-mode.el`](https://github.com/HOL-Theorem-Prover/HOL/blob/59508de8d4c907351c9ffa2bc4a7c5acc8a54088/tools/hol-mode.src#L844-L884)

### DIFF
--- a/tools/vim/README.md
+++ b/tools/vim/README.md
@@ -32,9 +32,10 @@ All files are located under tools/vim
 
 1. Add the contents of `hol-config.sml` to your `~/.hol-config.sml` file
 2. Add the contents of `filetype.vim` to your `~/.vim/filetype.vim`
-3. Run hol to start the HOL session
+3. Run hol to start the HOL session (only necessary when using vim without
+   support for terminal buffers)
 4. Run vim and open a HOL script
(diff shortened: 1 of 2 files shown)


mn200:
Thanks for this; it seems a great change (the emacs mode doesn't use this information for starting _its_ hol process, but it totally could/should).  To date, this has just driven `Holmake`’s behaviour.

PR #1104, opened 2023-04-19
fix last change: an overlooked `let` was used to set a global var in hol-input-translations

chenzhawyang:
ed.: i don't why it is trying to merge four commits instead of one, only the latest commit ([; use setq-local instead of let to bind global var](https://github.com/HOL-Theorem-Prover/HOL/pull/1104/commits/95d87f175a7ab4ddb5a57c74815a28cd100577e5)) matters, it is a one-liner fix


mn200:
It's because I rebased the previous PR and those commits are now not seen as the same as your originals.  You should rebase your latest single commit against `origin/develop` and then do a fresh PR for that change.

PR #1105, opened 2023-04-19
fix last change: use `setq-local` instead of `let` to mutate global var

chenzhawyang:
re-pr of https://github.com/HOL-Theorem-Prover/HOL/pull/1104 and a fix of https://github.com/HOL-Theorem-Prover/HOL/pull/1102


mn200:
Thanks!

PR #1106, opened 2023-05-14
HOL4 port of Candle's  compute primitive

oskarabrahamsson:
This PR contributes a HOL4 port of [Candle](https://github.com/CakeML/candle)'s compute primitive. The compute primitive is an interpreter for first-order lisp-style pairs of numbers with arithmetic. The implementation comes with a low-level library which instantiates the primitive, and a few examples of basic usage. The intention is to build a high-level interface around the compute primitive later, which would do proof-producing translation from more general HOL terms into the language accepted by the compute interpreter.

Disclaimer: The new compute primitive conjures up theorems, and needs to live inside the HOL4 kernel. Although Candle's compute primitive has been verified sound, this port was made by hand. The original Candle implementation and its mechanized verification proof can be found [here](https://github.com/cakeml/cakeml/tree/master/candle/prover/compute). This port attempts to stay as faithful as possible, but some things are a bit different in HOL4:
  * HOL numerals use a slightly different representation from Candle/HOL Light.
  * Division by zero is defined in Candle/HOL Light.
  * In Candle, the characteristic equations used by the compute primitive are defined using HOL4 overloads. We use recognizers that should resemble the overloads as closely as possible to check equations.


mn200:
The failure on the experimental kernel is a real one I'm afraid.

oskarabrahamsson:
The latest failure (in Fact) seems a bit random. It occurs here: https://github.com/HOL-Theorem-Prover/HOL/commit/35ad91b434cd807833d3fdd04e50cad6fbcc8731, but not in the next commit https://github.com/HOL-Theorem-Prover/HOL/commit/54ba85165ecfd958bac2da9d491e6314a3fe68fa, which contains a seemingly unrelated change. It also occurs several times earlier, but not all of the time. Is there a way to trigger a new test manually?

binghe:
> The latest failure (in Fact) seems a bit random. It occurs here: [35ad91b](https://github.com/HOL-Theorem-Prover/HOL/commit/35ad91b434cd807833d3fdd04e50cad6fbcc8731), but not in the next commit [54ba851](https://github.com/HOL-Theorem-Prover/HOL/commit/54ba85165ecfd958bac2da9d491e6314a3fe68fa), which contains a seemingly unrelated change. It also occurs several times earlier, but not all of the time. Is there a way to trigger a new test manually?
> 
> (The whole test (`bin/build --stdknl -t`) finishes without error on my machine.)

The best way to re-test your branch is to create a (free) Docker Hub account and setup GitHub Actions for your forked HOL repository (instructions are here [1], and there's `self-runner` action for testing any (recent) branch. This is one-time task, then you can `docker pull` the building images (pushed into your Docker Hub repository) and go inside the container to get the precise status and all intermediate files when the building process is broken.

[1] https://github.com/HOL-Theorem-Prover/HOL/pull/1079

mn200:
That error in Fact seems to happen non-deterministically in both kernels.  I’ll ignore it.

src/cv_compute/cvScript.sml
```diff
@@ -0,0 +1,254 @@
...
+ * :cv type, using the functions below (e.g., cv_add, cv_if, ...). See the
+ * files in examples/ for examples showcasing usage of the primitive.
+ *
+ *)
+
+open HolKernel boolLib bossLib arithmeticTheory prim_recTheory;
+
+val _ = new_theory "cv";
+
+val op \\ = op THEN;
```
mn200, line 14: Please delete these: they're built-in with these definitions already.

src/cv_compute/examples/examplesScript.sml
```diff
@@ -0,0 +1,216 @@
...
+ * satisfy.
+ * ------------------------------------------------------------------------- *)
+
+open HolKernel boolLib bossLib cv_computeLib;
+open arithmeticTheory listTheory prim_recTheory cvTheory;
+
+val _ = new_theory "examples";
+
+val op \\ = op THEN;
+val op >- = op THEN1;
```
mn200, line 48: Delete these also.

src/cv_compute/examples/examplesScript.sml
```diff
@@ -0,0 +1,216 @@
...
+          (Num 1)
+          (cv_mul n (factc (cv_sub n (Num 1))))
+Termination
+  WF_REL_TAC `measure cv_size_alt`
+  \\ Cases \\ simp [cv_size_alt_def, CaseEq "bool"]
+End
+
+Definition fact_def:
+  fact n = if n < 1 then 1 else n * fact (n - 1)
+End
```
mn200, line 70: Please use the existing `FACT` constant from arithmeticTheory.

src/cv_compute/examples/examplesScript.sml
```diff
@@ -0,0 +1,216 @@
...
+  isprime_aux dvs n =
+    if dvs * 2 < n then
+      if n SAFEMOD dvs <> 0 then
+        isprime_aux (dvs + 2) n
+      else F
+    else T
+Termination
+  WF_REL_TAC `measure (\(dvs,n). n - dvs)`
+End
+
```
mn200, line 159: Would be nice if this could point back to the core theories’ definition of `prime`.

src/cv_compute/examples/examplesScript.sml
oskarabrahamsson: Done

src/cv_compute/examples/examplesScript.sml
oskarabrahamsson: Done

src/cv_compute/examples/examplesScript.sml
oskarabrahamsson: Done

src/cv_compute/cvScript.sml
oskarabrahamsson: Done

mn200:
If you possibly can, it'd be great to have documentation for this. In particular: 
- the release notes (edit `doc/next-release.md`);  and
- the DESCRIPTION manual, where the API could get a nice explanation.

Let me know if you'll be able to do this.

mn200:
Also: please move the examples out from under `src` into the top-level `examples` directory and then make sure the example gets built at test-level 1 by making the appropriate edit to the `Holmakefile` in `src/parallel_builds/core`.

oskarabrahamsson:
> If you possibly can, it'd be great to have documentation for this. In particular:
> 
> * the release notes (edit `doc/next-release.md`);  and
> * the DESCRIPTION manual, where the API could get a nice explanation.
> 
> Let me know if you'll be able to do this.

Sounds good. I’ll find some time this weekend and start writing a summary for the DESCRIPTION manual.

oskarabrahamsson:
There's a point in the new tools list in `docs/next-release.md`, and a new section in the DESCRIPTION manual right after the `computeLib` section.

mn200:
Thanks for this!

PR #1107, opened 2023-05-17
Extend finite_mapSyntax to cover more constants

acjf3:
Note:
- commit does some renaming so as to adhere to standard naming conventions, e.g., fdom_t -> fdom_tm
- some exceptions may change as a result of updating old code to use HolKernel.syntax_fns


mn200:
Thanks for this!

PR #1109, opened 2023-06-02
hol_parser

HaoyangMains:
hol_parser is a tool designed for visually parsing HOL4 files in an interactive manner. Its primary functionality revolves around generating graphs that depict the theories' dependencies and theorems' dependencies. It has also some convenient features such as checking incompatible timestamps, theorem tags and unused theorems. For more details, please check the readme.
[readme.txt](https://github.com/HOL-Theorem-Prover/HOL/files/11633498/readme.txt)


barakeel:
Nice work. 

There seems to be a lot of auto-generated files in that pull request (e.g. binaries and maybe vs files). 
It's preferable to tell people how to compile your software from the source files and only include those.

For your information, there are some limitations to the dependency tracking in the ".dat" file. 
When a theorem is erased or overwritten before exporting the theory (happens when a theorem is replaced by a more general theorem with the same name), this creates a broken dependencies in the graph (e.g. a dependency identifier with no corresponding theorem in the theory).

HaoyangMains:
> There seems to be a lot of auto-generated files in that pull request (e.g. binaries and maybe vs files). It's preferable to tell people how to compile your software from the source files and only include those.

I am new to github, thanks a lot for your advice!

> For your information, there are some limitations to the dependency tracking in the ".dat" file. When a theorem is erased or overwritten before exporting the theory (happens when a theorem is replaced by a more general theorem with the same name), this creates a broken dependencies in the graph (e.g. a dependency identifier with no corresponding theorem in the theory).

Yeah, I hadn't considered this risk before. If that case happends in different theories, users can find it by the incompatible timestamps report. However, if the theorems 'x' depends on 'y' and they are both in the same theory, I can't even remind users the changes of 'y' by the timestamps when 'y' was erasered or overwritten by a different theorem. I don't have a good idea at this stage unless the theorems also have timestamps just like theories.

barakeel:
>> I don't have a good idea at this stage unless the theorems also have timestamps just like theories.

I don't think there is any simple solution since HOL4 allow erasing theorems. 
Propagating the dependencies of erased theorems could be a solution but this is in practice really complicated.
I was just telling you this so that you do not worry if you observed this behavior in some theories.

mn200:
I think I will move this code into a separate sub-repository under HOL-Theorem-Prover.

PR #1111, opened 2023-06-15
Limits of extreal-valued functions (ext_tendsto) and related changes

binghe:
Hi,

This PR follows #993. Currently there's a lack of formalisation on the limits of functions of type `'a -> extreal`, especially `num -> extreal`. Other important analysis properties of extreal-valued functions such as continuity, can only be properly defined after the limits is defined.    The solution is not obvious, because a straightforward imitation of limits of real-valued functions (see below), setting `dist (x,y:real) = abs (x - y)` doesn't work, because this "distance" is not a valid real number if infinities are involved (recall that metrics for whatever systems of "numbers", e.g. vectors, must be real numbers.)
```
   [LIM_SEQUENTIALLY]  Theorem (real_topologyTheory)
      ⊢ ∀s l.
          (s ⟶ l) sequentially ⇔
          ∀e. 0 < e ⇒ ∃N. ∀n. N ≤ n ⇒ dist (s n,l) < e
```

My solution is based on the metric space (`extreal_mr1`) defined in #993 and the Moore-Smith covergence nets from `real_toplogyTheory` (instead of the old `netsTheory`).  With two definitions `ext_tendsto` and `extreal_lim`, and then (nicely) overloading them on existing symbols for real-valued functions:
```
   [ext_tendsto_def]  Definition (extrealTheory)      
      ⊢ ∀f l net.
          (f --> l) net ⇔
          ∀e. 0 < e ⇒ eventually (λx. dist extreal_mr1 (f x,l) < e) net

   [extreal_lim_def]  Definition      
      ⊢ ∀net f. lim net f = @l. (f --> l) net
```

Then I proved the following important bridge theorems for the connection between limits of extreal and real functions. The first one says, if a real-valued function sequence `f(n)` tends to `l`, then the extreal-valued function `Normal o f` tends to `Normal l`:
```
   [lim_sequentially_imp_extreal_lim]  Theorem      
      ⊢ ∀f l.
          (f --> l) sequentially ⇒ (Normal ∘ f --> Normal l) sequentially
```
The second theorem (and its simple variant) is important for converting many existing practice of "finite" extreal-valued function limits to the new concepts defined here. If the extreal-valued function `f` has finite values after certain index, and the sequential limit `l` is also finite, then the real-valued function `real o f` tends to `real o l`, so is the vice versa.
```
   [extreal_lim_sequentially_eq]  Theorem      
      ⊢ ∀f l.
          (∃N. ∀n. N ≤ n ⇒ f n ≠ +∞ ∧ f n ≠ −∞) ∧ l ≠ +∞ ∧ l ≠ −∞ ⇒
          ((f --> l) sequentially ⇔ (real ∘ f --> real l) sequentially)
   
   [extreal_lim_sequentially_eq']  Theorem      
      ⊢ ∀f r.
          (∃N. ∀n. N ≤ n ⇒ f n ≠ +∞ ∧ f n ≠ −∞) ⇒
          ((f --> Normal r) sequentially ⇔ (real ∘ f --> r) sequentially)
```

## Current practices without `ext_tendsto`

If an extreal-valued function `f` (of type `:num->extreal`) is monotonic, either increasing or descreasing, then its sequential limit coincide with either `sup (IMAGE f univ(:num))` or `inf (IMAGE f univ(:num))`. This is exactly the workaround currently found in all proofs which monotone function limits are needed.   I should provide bridging theorems to help converting these `sup` and `inf` forms back to `ext_tendsto` but this task is not urgent and will be done in the future when it's needed.

If an extreal-valued function `f` but the values are always finite (i.e. not `PosInf` not `NegInf`) and the limit is also finite, currently the limits of the corresponding real-valued function `real o f` is discussed in various proofs.  In particular, the convergence concepts of random variables are all based on limits of `real o X` where `X` is given by `real_random_variable X p` (taking only finite values). See the following applications for some improvements on this situation.

## Applications

As an application of the newly defined concepts (NOTE: `extreal_lim` is not yet used, it's useful only after we have unique limits theorem for extreal functions, to be done in the future when needed), I have proven the equivalent alternative definitions for the convergence concepts of finite-valued random variables, namely `converge_AE` and `converge_LP`: (their original definitions are based on limits of real-valued functions involving ugly form `real o f`)
```
   [converge_AE]  Theorem (probabilityTheory)      
      ⊢ ∀p X Y.
          (∀n. real_random_variable (X n) p) ∧ real_random_variable Y p ⇒
          ((X --> Y) (almost_everywhere p) ⇔
           AE x::p. ((λn. X n x) --> Y x) sequentially)

   [converge_PR]  Theorem      
      ⊢ ∀p X Y.
          real_random_variable Y p ⇒
          ((X --> Y) (in_probability p) ⇔
           ∀e. 0 < e ∧ e ≠ +∞ ⇒
               ((λn. prob p {x | x ∈ p_space p ∧ e < abs (X n x − Y x)}) -->
                0) sequentially)
```

## Related changes

Previously `extrealTheory` doesn't depend on `real_toplogyTheory`, but once `real_topologyTheory` is open, it's immediately found that there are two "sums" (and also "summable") theorem in both `seqTheory` and `real_topologyTheory` breaking some proofs in `extrealTheory`.   And the overloading the `ext_tendsto` (f --> l) also cause some problems in, e.g. `borelTheory` due to the ambiguity without typing information.

My solution is to rename the theorem "sums" of `real_topologyTheory` (the newer one) to "sums_def", "summable" to "summable_def", and fixed all broken theroems in various places in "src/analysis".   The overloading issue also caused a bit changes in various places in "src/probability".

Moreover, I moved several small HOL-Light compatible arithmetic theorems from various places (`cardinalTheory` and `iterateTheory`) to `arithmeticTheory`.

## Other changes

The theorem `EXTREAL_SUP_SEQ` and the related definition `sup_seq` seems dedicated to the proof of Radon-Nikodym theorem in `lebesgueTheory`. Now I moved them back to `lebesgueTheory`, for a more clean `extrealTheory` of independent interests to non-probability users.


mn200:
Thanks for this! Could you write a brief summary of the “incompatibilities” caused by theorem renaming in the release notes please?

binghe:
I have done an alternative definition of LP convergence using `lp_space` and the new extreal function limits:
```
   [converge_LP]  Theorem      
      ⊢ ∀p X Y r.
          real_random_variable Y p ∧ 0 < r ∧ r ≠ +∞ ⇒
          ((X --> Y) (in_lebesgue r p) ⇔
           (∀n. X n ∈ lp_space r p) ∧ Y ∈ lp_space r p ∧
           ((λn. expectation p (λx. abs (X n x − Y x) powr r)) --> 0)
             sequentially)
```
But this result depends on a further enriched theory of `lp_space` and `seminorm` (in `martingaleTheory`), which consists of about 1000 lines of additions and can only be submitted as another PR after the present one were merged.

mn200:
Thanks for your work on this!

PR #1112, opened 2023-06-18
Fix ASL equivalence proofs

hrutvik:
These proofs broke due to changes in automatic rewrites, causing CakeML regression to fail in turn (https://cakeml.org/regression.cgi/job/2152). This commit should fix them.


mn200:
Thanks for this!

PR #1113, opened 2023-06-26
Avoid issues with evaluation of `n2l` and `l2n`.

acjf3:
For example, restricts evaluation of `n2l` to terms with standard base sizes and `NUMERAL` values. Previously, EVAL ``n2l 10 a`` and EVAL ``n2l a 10` would both loop.


mn200:
Could you add regression tests in a `selftest.sml` file please?

PR #1114, opened 2023-06-27
[extreal] more natural definition of extreal_sub (x - y = x + -y), etc.

binghe:
Hi,

This PR slightly modifies the definition of `extreal_sub` (subtraction of extended reals) and simplified many existing proofs with more natural versions.

First of all, recall that the subtraction of real numbers is defined by their additions and arithmetic negations:
```
   [real_sub]  Definition (realaxTheory)      
      ⊢ ∀x y. x − y = x + -y
```
While the subtraction of extended reals is currently defined below, following the idea of `extreal_add`:
```
   (extreal_sub (Normal x) (Normal y) = Normal (x - y)) /\
   (extreal_sub PosInf (Normal x) = PosInf) /\
   (extreal_sub NegInf (Normal x) = NegInf) /\
   (extreal_sub (Normal x) NegInf = PosInf) /\
   (extreal_sub (Normal x) PosInf = NegInf) /\
   (extreal_sub NegInf PosInf = NegInf) /\
   (extreal_sub PosInf NegInf = PosInf)
```
With the above definition, two exreals `x - y` and `x + -y` are equal if and only if `x - y` is not unspecified, i.e. when `PosInf - PosInf` and `NegInf - NegInf` do not happen:
```
   [extreal_sub_add]  Theorem      
      ⊢ ∀x y. x ≠ −∞ ∧ y ≠ +∞ ∨ x ≠ +∞ ∧ y ≠ −∞ ⇒ x − y = x + -y
```
This is rather inconvenient (see below) and unnatural. But if we take `x − y = x + -y` as the *definition* of extreal subtraction:
```
   [extreal_sub]  Definition (extrealTheory)      
      ⊢ ∀x y. x − y = x + -y
```
Then the above original definition would become a theorem, and no existing proof is broken by this change, meanwhile many proofs involving the property of `\x. f x - g x` can be directly derived from the related properties for `\x. f x + g x` and `\x. -x`.  The biggest saving is at `borelTheory` where we had had two almost identical long proofs for `IN_MEASUSURABLE_BOREL_ADD'` and `IN_MEASUSURABLE_BOREL_SUB'` (~1000 lines for each of them) but one cannot be derived from the other. Now, with the new definition of `extreal_sub`, the "sub / SUB" versions are now trivial!

# Other additions: full versions of `abs_triangle`, etc.

This is not due to the changes of `extreal_sub`. It turns out that, the long standing triangle properties of extreals, such as:
```
   [abs_triangle]  Theorem      
      ⊢ ∀x y.
          x ≠ +∞ ∧ x ≠ −∞ ∧ y ≠ +∞ ∧ y ≠ −∞ ⇒ abs (x + y) ≤ abs x + abs y

   [abs_triangle_neg]  Theorem      
      ⊢ ∀x y.
          x ≠ +∞ ∧ x ≠ −∞ ∧ y ≠ +∞ ∧ y ≠ −∞ ⇒ abs (x − y) ≤ abs x + abs y

   [abs_triangle_sub]  Theorem      
      ⊢ ∀x y.
          x ≠ +∞ ∧ x ≠ −∞ ∧ y ≠ +∞ ∧ y ≠ −∞ ⇒ abs x ≤ abs y + abs (x − y)
```

All antecedents are actually unnecessary and can be removed. The idea is that, e.g. in `abs_triangle`, although `PosInf + NegInf` is unspecific, such an unspecific value is always less or equal than PosInf. Thus, when `x = PosInf` and `y = NegInf`, the `abs_triangle` formula actually holds. To prevent breaking many existing proofs which call these theorems by `MATCH_MP_TAC`, now I have added the following so-called "full versions":
```
   [abs_triangle_full]  Theorem      
      ⊢ ∀x y. abs (x + y) ≤ abs x + abs y

   [abs_triangle_neg_full]  Theorem      
      ⊢ ∀x y. abs (x − y) ≤ abs x + abs y

   [abs_triangle_sub_full]  Theorem      
      ⊢ ∀x y. abs x ≤ abs y + abs (x − y)
```
I have started using these new theorems in my new proofs, sometimes they save a lot of efforts in proving the finiteness of involved extreal values.

Several other extreal theorems are also simplified with less antecedents. The broken proofs caused by these changes are minimal (just one in `probabilityTheory`, which is also simplified.)

--Chun Tian


binghe:
The lesson I learnt here: if there are several different ways to define the same concept [1], and these different alternative definitions are not unconditionally equal, by choosing the most _appropriate_ one as the very definition, we also get an  unconditional rewriting rule from the definition, while other alternative definitions (as derived theorems) may require extra antecedents, which is sometimes hard to satisfy.  It's good to see that, sometimes (also in this case), the most natural, better understandable, minimal (in term size) choice also leads to the minimal overall proof size.

[1] For example the `inf` of real sets has an alternative definition in `iterateTheory` (derived from HOL Ligh). `inf s` is only specified when `s` has a lower bound. But in this case it's not clear which `inf` is _better_.

mn200:
Fantastic! I totally agree that defining subtraction this way is appropriate and now we see it leads to superior outcomes as well.

PR #1115, opened 2023-06-27
Define k DIV 0 = 0 and k MOD 0 = k

myreen:


binghe:
Hi...I think this "total function" change may break the `--otknl` build, because HOL4's `DIV` and `MOD` are currently mapped directly to OpenTheory's `Number.Natural.div` and `Number.Natural.mod` [1]. Perhaps the build is not immediately broken, because currently there's no proof in the core library using the extended definitions, but once any proof (or user code) starts to relay on the new definitions, the OT exporting of these proof will fail.

Yes, it's possible to fix the OpenTheory builds by making some changes in `src/boss/prove_base_assums.otd` (and also `src/boss/prove_base_assumsScript.sml`), but I don't think this is the right direction, which is to converge to OT's (reasonable) definitions w/o division-by-zero issues (for both `:num` and `:real`).   --Chun

[1] https://opentheory.gilith.com/opentheory/packages/natural-div-def-1.41/natural-div-def-1.41.html

binghe:
Well, it is currently indeed broken (at `src/boss/prove_base_assumsScript.sml`) in this branch. The error message seems at an irrelevant theorem because the order of theorems (assumptions) has shifted due to (perhaps the current) changes in `arithmeticTheory`:

```
Saved ORACLE thm ____ "th32"
Saved ORACLE thm ____ "th33"
Saved ORACLE thm ____ "th34"
Saved ORACLE thm ____ "th35"
Saved theorem _______ "th36"
Proof of 

∀P. P Data_List_nil ∧ (∀t. P t ⇒ ∀h. P (Data_List_cons h t)) ⇒ ∀l. P l

failed.
Failed to prove theorem th37.

Exception raised at Thm_cont.DISCH_THEN:
 (THEN1 on line 648)
error in quse /Users/binghe/ML/HOL.otknl/src/boss/prove_base_assums.artScript.sml : HOL_ERR {message = " (THEN1 on line 648)", origin_function = "DISCH_THEN", origin_structure = "Thm_cont"}
error in load /Users/binghe/ML/HOL.otknl/src/boss/prove_base_assums.artScript : HOL_ERR {message = " (THEN1 on line 648)", origin_function = "DISCH_THEN", origin_structure = "Thm_cont"}
Uncaught exception: HOL_ERR {message = " (THEN1 on line 648)", origin_function = "DISCH_THEN", origin_structure = "Thm_cont"}
```

myreen:
My opinions:

1. We ought to consider what to prioritise here. The new equations about `DIV` and `MOD` align with the new compute primitive, which is a tool that might get used quite a lot in HOL4 soon. On the other hand, it's very uncommon to export anything to OpenTheory and these new equations about `DIV` and `MOD` only hinder OpenTheory exports if the to-be-exported proofs rely on these new equations.

2. HOL Light is consistent with `k DIV 0 = 0 and k MOD 0 = k`. I think OpenTheory is the one that ought to change.

3. Something seems overly fragile if my change to `arithmeticTheory` broke something. I mean I didn't change the statement of any theorem. I just added two new theorems. How can that break something?

binghe:
I quickly went through the paper but don't quite understand why the existing `DIV` and `MOD` have to be changed, to accomplish the goal of using the new compute primitive extensively in the future.   Can you elaborate a bit more?

Let's say, as a user I don't mind `SAFEDIV` and `SAFEMOD` (perhaps they could have a better prefix other than "SAFE", e.g. "FAST") being moved to `arithmeticTheory` with some extra supporting theorems, and if anyone wants to use them when they needs `DIV` and `MOD`, they can just use them.

myreen:
To me, `SAFEDIV` and `SAFEMOD` are just corrected version of `DIV` and `MOD`. By updating `DIV` and `MOD` as this PR does, I make it easier to use the compute primitive.

If this PR is rejected, then the compute primitive can still be used, but it requires either users using `SAFEDIV`/`SAFEMOD` instead of `DIV`/`MOD` or potentially slightly fragile rewriting from `DIV` to `SAFEDIV`. To illustrate why rewriting might be fragile, consider `m DIV (f x)` which needs to be rewritten to `m SAFEDIV (f x)`, but that requires that `f x` is non-zero.

I'm happy that the new definitions of `DIV` and `MOD` are exactly those of HOL Light. These definitions should be the standard.

One could do the opposite of this:

binghe:
If this must happen, I hope the original `DIV` and `MOD` will be renamed to `SAFEDIV` and `SAFEMOD` (here "safe" means "strange things like division by zero won't happen").

And I think the code changes can be further optimized. In `arithmeticTheory`, we don't really need to touch the proof of `MOD_DIV_exist`. Instead, when calling `new_specification` we give it new constant names:
```
val DIVISION = new_specification ("DIVISION", ["SAFEMOD", "SAFEDIV"], MOD_DIV_exist);
```
And then the original definitions (of `SAFEDIV` and `SAFEMOD`) in `cvScript.sml` can be directly moved in, with only name changes:
```
Definition DIV_def:
  DIV m n = if n = 0 then 0 else m SAFEDIV n
End
Definition MOD_def:
  MOD m n = if n = 0 then m else m SAFEMOD n
End
```
Then, with small changes to some other proofs of existing theorems, I believe their statements should remain exactly the same as before, meanwhile you can add new theorems as much as possible.  Since your existing proofs already depends on the above definitions, the rest of changes (comparing with its current status in `develop`) should be smaller than the present PR.  But what's more important is that `SAFEMOD` and `SAFEDIV` will be later mapped to OpenTheory's corresponding constants.

Is this plan acceptable by everyone?

myreen:
I don't quite buy that functions that **don't** specify what happens for one argument are "SAFE", but the name is not so important. 

Since you write "their statements should remain exactly the same as before", I take it that all theorem would be about `DIV` and `MOD` rather than `SAFEDIV` and `SAFEMOD`. I think one would need to change:

```
val DIVISION = new_specification ("DIVISION", ["SAFEMOD", "SAFEDIV"], MOD_DIV_exist);
```

because this theorem is currently about `DIV` and `MOD`, but this line would make it about `SAFEDIV` and `SAFEMOD`. 

```
val SAFE_DIVISION = new_specification ("SAFE_DIVISION", ["SAFEMOD", "SAFEDIV"], MOD_DIV_exist);
```

and there there ought to be a new proof of `DIVISION` based on `SAFE_DIVISION`.

We need a better name than "SAFE".

binghe:
Yes, you are right, the new `DIVISION` (the bridging theorem) should be proved by `SAFE_DIVISION` (directly returned by `new_specification`) and the new `DIV_def` and `MOD_def`.

The names of the old constants are not so important. To this extent, `OT_DIV` and `OT_MOD` now seem reasonable to me.

binghe:
This version of `arithmeticScript.sml` changes should be enough:

```
(* |- !n. 0 < n ==> !k. k = OT_DIV k n * n + OT_MOD k n /\ OT_MOD k n < n *)
val OT_DIVISION = new_specification ("OT_DIVISION", ["OT_MOD", "OT_DIV"],
                                     MOD_DIV_exist);

(* HOL4 now switches to HOL-Light compatible version of DIV and MOD *)
val DIV_def = new_definition
  ("DIV_def", “DIV m n = if n = 0 then 0 else m OT_DIV n”);

val MOD_def = new_definition
  ("MOD_def", “MOD m n = if n = 0 then m else m OT_MOD n”);

val _ = set_fixity "MOD" (Infixl 650);
val _ = set_fixity "DIV" (Infixl 600);

Theorem DIVISION :
    !n. 0 < n ==> !k. k = k DIV n * n + k MOD n /\ k MOD n < n
Proof
    NTAC 3 STRIP_TAC
 >> ‘n <> 0’ by ASM_REWRITE_TAC [NOT_ZERO_LT_ZERO]
 >> RW_TAC bool_ss [DIV_def, MOD_def, OT_DIVISION]
QED
```

I think there's no need to also set `OT_MOD` and `OT_DIV` as infix operators (they look ugly as infix operators). I will fix the OpenTheory builds once this PR gets merged.

myreen:
@binghe My latest commit defines `OT_DIV` and `OT_MOD` as you wanted them.

mn200 (approved):

PR #1119, opened 2023-06-29
Equivalence between ‘lim$differentiable’ and ‘derivative$differentiable’

binghe:
Hi,

This small PR continues #1033 bridging the old and new derivative theories of the calculus. The concept of differentiable functions in `limTheory` and the more general version from `real_topologyTheory` (when focusing on single point `at x`) are actually equivalent:
```
   [differentiable_alt]  Theorem (limTheory)      
      ⊢ ∀f x. f differentiable x ⇔ derivative$differentiable f (at x)
```

This proof wasn't obvious to me, until the following theorem is proven (after some discussions on Slack), which shows that any one-dimension linear function has an explicit form (NOTE: this result doesn't hold in higher dimensions, i.e. vector functions):
```
   [linear_repr]  Theorem (real_topologyTheory)
      ⊢ ∀f. linear f ⇔ ∃l. f = (λx. l * x)

   [linear_alt_cmul]  Theorem      
      ⊢ ∀f. linear f ⇔ ∀c x. f (c * x) = c * f x
```
Where `linear` is defined in this way:
```
   [linear]  Definition
      ⊢ ∀f. linear f ⇔
            (∀x y. f (x + y) = f x + f y) ∧ ∀c x. f (c * x) = c * f x
```

# Future directions

The possible next step in this direction is to update «HOL Description», expanding the part related to calculus. I'm near the end of this documentation work.

But what's more important is to prove the equivalence between Lebesgue Integral and Gauge Integral for functions integrable in both integrals. Otherwise it's almost impossible to calculate any concrete Lebesgue integrals (over intervals) involving higher-order polynomials, and transcendental functions. In other words, under certain "integrable" conditions [2], something like the following theorem should be proven:
```
|- !f a b. f absolutely_integrable_on (interval[a,b]) ==>
          (lebesgue$integral lebesgue (\x. Normal (f x) * indicator_fn (interval[a,b]) x) =
           Normal (integration$integral (interval (a,b)) f))
```
where `lebesgue` is the measure space of all (real-valued) Lebesgue measurable sets defined in `examples/probability/lebesgue_measureScript.sml`.  This proof is not easy but I know a version of this proof in a book [1]

[1] R. G. Bartle, A Modern Theory of Integration. American Mathematical Soc., 2001.
[2] "A function is Lebesgue integrable if and only if the function and its absolute value are Henstock–Kurzweil integrable." somewhere from [1]

### DIFF
--- a/src/real/analysis/limScript.sml
+++ b/src/real/analysis/limScript.sml
@@ -278,6 +278,20 @@ Proof
     rw [differentiable, diffl_has_vector_derivative]
 QED
 
+(* The equivalence between ‘differentiable’ and ‘derivative$differentiable’ *)
+Theorem differentiable_alt :
+    !f x. f differentiable x <=> derivative$differentiable f (at x)
+Proof
+    rw [differentiable, diffl_has_derivative, derivativeTheory.differentiable]
+ >> EQ_TAC
+ >- (STRIP_TAC \\
+     Q.EXISTS_TAC ‘\x. l * x’ >> rw [])
+ >> DISCH_THEN (Q.X_CHOOSE_THEN ‘g’ ASSUME_TAC)
+ >> ‘linear g’ by PROVE_TAC [has_derivative]
+ >> ‘?l. g = \x. l * x’ by METIS_TAC [linear_repr]
+ >> Q.EXISTS_TAC ‘l’ >> rw []
+QED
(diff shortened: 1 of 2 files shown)


mn200:
Thanks for this!

PR #1123, opened 2023-07-12
Remove the restrictive licensing of the pair library.

jimgrundy:
When I created the pair library back in 1992 I reserved for myself rights for commercial use. Odly enough, in past 30 years, nobody has every asked for a license. It was early days and open source wasn't such a thing.

I no longer wish to reserve those rights for myself, and so I now allow this code to be covered by the same COPYRIGHT terms as cover the rest of the HOL distrubution.

Jim Grundy (jim.grundy@acm.org)


mn200:
Thanks for the contribution!

PR #1124, opened 2023-07-13
The Data Encryption Standard (DES) in HOL

binghe:
Hi,

I notice that, in `examples/Crypto` there's a good collection of encryption algorithms including AES.  Let's do a small exercise (for me, first time using `wordsTheory`), to formalize the Data Encryption Standard (DES) in HOL. For other related works on DES formalization, to my knowledge this was only seen with Mizar [3]. If we compare the two work, we can immediately see how excellent HOL's `wordsTheory` (and also `fcpTheory`) are.

There's no academic value here. But it has been said that "DES trained a generation of cryptographers" [1, p.13]. DES is a block cipher working on 64-bit blocks. The key is 56-bit but the key interface is usually in 64-bit with some parity bits to be ignored. The whole encryption process, beside the initial and final permutations, has 16 rounds (each round consumes one 48-bit round key generated from the 56-bit user key), as shown below:
```
    +-----+                           +-----+
    | KS  | <--- KEY     MESSAGE ---> | IP  |
    +--+--+    (56-bit)  (64-bit)     +--+--+
       |                                \|/
       |   u_0      (32-bit)       +-----+-----+       (32-bit)      v_0
       |    +----------------------+   split   +----------------------+
       |    |                      +-----+-----+                      |
       +----|------------------------------------+ k_1                |
       |   \|/      +---+          +---+        \|/          +---+    |
       |   (+) <--- | P | <------- | S | <===== (+) <======= | E | <--+
       |    |       +---+ (32-bit) +---+            (48-bit) +---+    |
       |     \                                                       /
       |  v_1 +-------------------------+  +------------------------+ u_1
       |                                 \/
       |                                 /\
       |  u_1 +-------------------------+  +------------------------+ v_1
       |     /                                                       \
       +----|------------------------------------+ k_1                |
       |   \|/      +---+          +---+        \|/          +---+    |
       |   (+) <--- | P | <------- | S | <===== (+) <======= | E | <--+
       |    |       +---+          +---+                     +---+    |
       |     \                                                       /
       |  v_2 +--------------------------+  +-----------------------+ u_2
       |                                  \/
       |                                  /\
       |      +--------------------------+  +-----------------------+
       |     /                                                       \
       .    .                                                         .
       .    .                                                         .
       |    |                                                         |
       +----|------------------------------------+ k_r                |
           \|/      +---+          +---+        \|/          +---+    |
           (+) <--- | P | <------- | S | <===== (+) <======= | E | <--+
            |       +---+          +---+                     +---+    |
            |                      +-----------+                      |
            +--------------------> |   join    | <--------------------+
           u_r       (32-bit)      +-----+-----+      (32-bit)       v_r
                                        \|/
                                      +--+--+
                                      | IIP | ---> CIPHERTEXT
                                      +-----+       (64-bit)
```

At toplevel, the core DES operations (without key schedule) but, parameterized by the number of rounds is defined (nicely) below:
```
[DESCore_def]
⊢ ∀r ks. DESCore r ks = IIP ∘ Join ∘ desRound r r ks ∘ Split ∘ IP
```
Where `desRound` is an overload on `Feistel RoundOp`, where `Feistel` is the general device (possibly) of Feistel networks (`RoundOp` is DES-specific):
```
[Feistel_def]
⊢ (∀f r ks w. Feistel f 0 r ks w = w) ∧
  ∀f n r ks w.
    Feistel f (SUC n) r ks w =
    (let
       (u,v) = Feistel f n r ks w;
       k = EL n ks
     in
       if SUC n = r then (u ⊕ f v k,v) else (v,u ⊕ f v k))

[RoundOp_def]
⊢ ∀w k. RoundOp w k = P (S (E w ⊕ k))
```
Note that, the encryption/decryption correctness of DES (the last theorem in this PR) depends only on the definition of `Feistel`! Other functions such as `RoundOp` and the sub-functions inside it are totally irrelevant. On the other side, the security of DES depends on S-Box function `S_def`, which is the *only* non-linear transformation in the cipher:
```
[S_def]
⊢ ∀w. S w =
      concat_word_list
        [S8 ((5 >< 0) w); S7 ((11 >< 6) w); S6 ((17 >< 12) w);
         S5 ((23 >< 18) w); S4 ((29 >< 24) w); S3 ((35 >< 30) w);
         S2 ((41 >< 36) w); S1 ((47 >< 42) w)]

Overload S1 = “SBox S1_data”
...

[SBox_def]
⊢ ∀box w.
    SBox box w =
    (let
       row = w2n ((5 >< 5) w @@ (0 >< 0) w);
       col = w2n ((4 >< 1) w)
     in
       n2w (EL col (EL row box)))
```

Now, let `KS` be the key scheduling whose details are not given here. The encryption and decryption of DES (elegantly) use the same core but different key orders: (`DES r key` returns a pair of functions, one is for encryption, the other is for decryption)
```
[DES_def]
  ⊢ ∀r key.
    DES r key =
    (let keys = KS key r in (DESCore r keys,DESCore r (REVERSE keys)))
```
Here I call the `DES r` as "DES reduced to r rounds" (usually r < 16), while the overloaded `FullDES` means `DES 16`.

Thanks to the `computeLib`, this is both a formalization and algorithm implementation. For the correctness, I have proved that encryption and decryption of the same plaintext restores the plaintext (proof idea is from [2]):
```
[DESCore_CORRECT]
⊢ ∀keys r plaintext.
    LENGTH keys = r ⇒
    DESCore r (REVERSE keys) (DESCore r keys plaintext) = plaintext

[FullDES_CORRECT]
⊢ ∀key plaintext.
    (encrypt,decrypt) = FullDES key ⇒ decrypt (encrypt plaintext) = plaintext
```

We can also compute concrete values. In `des_testTheory` I have added many test cases in which all literal values are taken from DES papers and articles. For example, if both plaintext and key are all zeros, the encryption result is `0x8CA64DE9C1B123A7w`, which is correct (w.r.t. other DES implementations). Actually `EVAL` works so well for all these computations. And all these test case theorems can be proved by one-line `EVAL_TAC`.
```
[Test_FullDES_0]
⊢ let
    (encrypt,decrypt) = FullDES 0x0000000000000000w
  in
    encrypt 0x0000000000000000w = 0x8CA64DE9C1B123A7w

[Test_FullDES_0']
⊢ let
    (encrypt,decrypt) = FullDES 0x0000000000000000w
  in
    decrypt 0x8CA64DE9C1B123A7w = 0x0000000000000000w
```

DES requires a large mount of data tables, such as the eight S-boxes. The first one is shown below:
```
[S1_data]
⊢ S1_data =
  [[14; 4; 13; 1; 2; 15; 11; 8; 3; 10; 6; 12; 5; 9; 0; 7];
   [0; 15; 7; 4; 14; 2; 13; 1; 10; 6; 12; 11; 9; 5; 3; 8];
   [4; 1; 14; 8; 13; 6; 2; 11; 15; 12; 9; 7; 3; 10; 5; 0];
   [15; 12; 8; 2; 4; 9; 1; 7; 5; 11; 3; 14; 10; 0; 6; 13]]
```
To guarantee the correctness, all the data tables were copied directly from PDF pages describing DES, without any handwriting, then adding necessary delimiters to make them parsable as HOL definitions.

Now I have gained some experiences on `wordsTheory`, in which I added the support of `:word56` for DES.

Regards,

Chun Tian

P. S. Thanks many people on Slack for my blocking issue on the proof goal `!n. n < 64. P n` during this work. I've learnt `numLib.BOUNDED_FORALL_CONV` from Konrad Slind.

[1] Knudsen, L.R., Robshaw, M.J.B.: The Block Cipher Companion. Springer Publishing Company, Incorporated, Berlin, Heidelberg (2011).
[2] Coopersmith, D.: The Data Encryption Standard (DES) and its strength against attacks. IBM J. 38, 243–250 (1994).
[3] Okazaki, Hiroyuki, Kenichi Arai, and Yasunari Shidama. "Formal Verification of DES Using the Mizar Proof Checker." Proceedings of the International Conference on Foundations of Computer Science (FCS). The Steering Committee of The World Congress in Computer Science, Computer Engineering and Applied Computing (WorldComp), 2011.


mn200:
Thanks for this!  It looks like it was a fun exercise.

PR #1117, opened 2023-06-28
Hölder's and Minkowski's inequalities in the infinity cases

binghe:
Hi,

This PR follows #972 and #1009. The previously added Hölder's inequality and Minkowski's inequality for L^p spaces (part of Lebesgue Integration Theory) have missed the extreme case which one of the two parameters is (positive) infinity. This is now fixed by finishing the proof for the missing cases. As a result, the following key theorems now have their `p <> PosInf /\ q <> PosInf` removed from their antecedents:
```
   [Hoelder_inequality]  Theorem (martingaleTheory)      
      ⊢ ∀m u v p q.
          measure_space m ∧ 0 < p ∧ 0 < q ∧ p⁻¹ + q⁻¹ = 1 ∧
          u ∈ lp_space p m ∧ v ∈ lp_space q m ⇒
          integrable m (λx. u x * v x) ∧
          ∫ m (λx. abs (u x * v x)) ≤ seminorm p m u * seminorm q m v

   [Minkowski_inequality]  Theorem      
      ⊢ ∀p m u v.
          measure_space m ∧ 1 ≤ p ∧ u ∈ lp_space p m ∧ v ∈ lp_space p m ⇒
          (λx. u x + v x) ∈ lp_space p m ∧
          seminorm p m (λx. u x + v x) ≤ seminorm p m u + seminorm p m v
```

Unfortunately, there was one mistake in the previous definition of `lp_space` in the branch of `p = PosInf`:
```
   [lp_space_def]  Definition
      ⊢ ∀p m.
          lp_space p m =
          {f | f ∈ Borel_measurable (measurable_space m) ∧
           if p = +∞ then
             ∃c. 0 < c ∧ c ≠ +∞ ∧
                 measure m {x | x ∈ m_space m ∧ c ≤ abs (f x)} = 0
           else ∫⁺ m (λx. abs (f x) powr p) ≠ +∞}
```
The term `c ≠ +∞` was added here.  This is because, the whole idea of a (measurable) function belongs to L^∞ space is that there's a upper bound `c` such that the function is almost-everywhere bounded by it. But if we allow `c` can be `PosInf`, then the function is just finite-valued not bounded.

Then I have added a chain of new theorems concerning to the basic properties of `lp_space` and `seminorm`. The major part and end of this chain are the following theorems about the "linearity" of L^p spaces (including `p = PosInf`):
```
   [seminorm_cmul]  Theorem      
      ⊢ ∀p m u r.
          measure_space m ∧ 0 < p ∧
[...]

PR #1126, opened 2023-07-16
Increased parallel build levels of core theories

binghe:
This PR moves some theorems between core theories. IAt the end, the following theory dependencies are achieved:

- derivativeTheory no more depends on seqTheory
- integrationTheory no more depends on transcTheory

This increased the parallel build levels but having more independent theories being built together. Before:

Schermata 2023-07-16 alle 22 21 28

After:

Schermata 2023-07-16 alle 22 20 49

On computers with typical 4 CPU cores, I believe this saves at least 1 minute of total building time.

Besides, the `util_probTheory` is cleaned up by moving many theorems into upstream theories. In particular, the following two theorems are moved to `pairTheory` and `combinTheory` (I think they are general enough):
```
   [K_PARTIAL]  Theorem (combinTheory)      
      ⊢ ∀x. K x = (λz. x)

   [PAIRED_BETA_THM]  Theorem (pairTheory)
      ⊢ ∀f z. UNCURRY f z = f (FST z) (SND z)
```
There are a series of useful `NUM_2D_BIJ` theorems (by Joe Hurd) in `util_probTheory` that I wanted to move to `pred_setTheory` for long time but they need new proofs. This is done now. The first two are shown below:
```
   [NUM_2D_BIJ]  Theorem      
      ⊢ ∃f. BIJ f (𝕌(:num) × 𝕌(:num)) 𝕌(:num)

   [NUM_2D_BIJ_NZ]  Theorem      
      ⊢ ∃f. BIJ f (𝕌(:num) × (𝕌(:num) DIFF {0})) 𝕌(:num)
```


mn200:
I definitely welcome the parallelisation here.  Note that `PAIRED_BETA_THM` is simply the definition of `UNCURRY` (available under that name), so please replace all uses of `PAIRED_BETA_THM` with `UNCURRY`.  As for the bijection proofs, I see you've used `pair_to_num`, which is reasonable, but that constant seems like it should be an overload at best (it's a perfect copy of the pairing function in `numpairTheory`.  But that's something I can attack independently.

binghe:
> I definitely welcome the parallelisation here. Note that `PAIRED_BETA_THM` is simply the definition of `UNCURRY` (available under that name), so please replace all uses of `PAIRED_BETA_THM` with `UNCURRY`. As for the bijection proofs, I see you've used `pair_to_num`, which is reasonable, but that constant seems like it should be an overload at best (it's a perfect copy of the pairing function in `numpairTheory`. But that's something I can attack independently.

Thanks for your comments. I didn't see there are `num_to_pair` and `pair_to_num` already in `pred_setTheory`.  I will make use of them but then hide the theorems of concrete NUM_2D_BIJ lemmas locally (I think probability scripts only needs the existence theorems like `NUM_2D_BIJ`, etc.) Please hold until this change is done and tested.

mn200:
Sorry, you've done it the "wrong way round" with `npair`: I'd prefer to see `npair` kept (because it's older) and `pair_to_num` removed (it's redundant).

binghe:
> Sorry, you've done it the "wrong way round" with `npair`: I'd prefer to see `npair` kept (because it's older) and `pair_to_num` removed (it's redundant).

But `pair_to_num` and `num_to_pair` were already defined on top of `npairTheory` (`npair`, `nfst`, `nsnd`) and their related proofs in `pred_setTheory` have no duplications with the core proofs in `npairTheory`.
```
   [num_to_pair_def]  Definition      
      ⊢ ∀n. num_to_pair n = (nfst n,nsnd n)
   
   [pair_to_num_def]  Definition      
      ⊢ ∀m n. pair_to_num (m,n) = m ⊗ n
```
I'd that they are easier to use (than `npair`, `nfst`, `nsnd`). I wouldn't have used `UNCURRY npair` if I knew `pair_to_num`.  Changing them to overloads may break something existing proofs. My last changes have the minimal changes to `pred_setTheory` adding smallest number of new theorems.

What's really redundant, is the following definitions in `ind_typeTheory` (seems part of the Datatype package):
```
   [NUMPAIR]  Definition
      ⊢ ∀x y. ind_type$NUMPAIR x y = 2 ** x * (2 * y + 1)
   
   [NUMPAIR_DEST]  Definition      
      ⊢ ∀x y.
          NUMFST (ind_type$NUMPAIR x y) = x ∧
          NUMSND (ind_type$NUMPAIR x y) = y
```
They and many proofs in this theory can be simplified by using `npair`, `nfst` and `nsnd`.

binghe:
Hmmm... `ind_type$NUMPAIR` and `numpair$npai` have different definitions. They are not equivalent, just both of them have the desired one-one mapping character between one- and two-dimensional natural numbers, which is usually the only thing needed. (Thus it's not a good idea to merge them as certain proofs may break.)

binghe:
I made some further fixes to eliminate the bottleneck of `limTheory` in the theory dependency graph.

Schermata 2023-07-17 alle 16 34 05

Furthermore, in the previous commits I almost succeed in makeing `extrealTheory` completely independent of Probability scripts (i.e. not depending on `util_probTheory`). The only lefover was the definition `q_set` and related theorems on the rational subset of real numbers:
```
   [real_rat_set_def]  Definition      
      ⊢ q_set =
```
The main characteristics of this set are two: 1) this set is countable. 2) there exist one rational number between any two real numbers:
```
   [QSET_COUNTABLE]  Theorem      
      ⊢ countable q_set
   
   [Q_DENSE_IN_REAL]  Theorem      
      ⊢ ∀x y. x < y ⇒ ∃r. r ∈ q_set ∧ x < r ∧ r < y
```
These results are directly used in `real_borelTheory` and `extrealTheory`. If not `util_probTheory`, it must be in somewhere  near `realTheory` (but not in it). After some thoughts, I consider the relatively small `real_sigmaTheory` a suitable place.

mn200:
I take your point about the `pair_to_num` and `num_to_pair` constants.  Happy to merge this now.  Incidentally the pairing in ind_types is not onto (doesn't hit `0`) and I probably will do away with it at some point (those constants are not used elsewhere).

mn200:
It does also seem as if the core `extreal` definitions (basic arithmetic and orderings) could be moved much earlier, to be an independent child of `real`. This could be `extreal_base` or similar. The more sophisticated stuff that depends on limits and transcendentals could then live in a smaller `extreal` that stays where it currently is.

binghe:
> It does also seem as if the core `extreal` definitions (basic arithmetic and orderings) could be moved much earlier, to be an independent child of `real`. This could be `extreal_base` or similar. The more sophisticated stuff that depends on limits and transcendentals could then live in a smaller `extreal` that stays where it currently is.

This work is done (now I have a `src/real/extreal_baseScript.sml` of 3,000 lines) but before #1117 gets merged I can't create the new PR, because #1117 has some code changes in `extrealScript.sml` which is to be moved to the new `extreal_baseTheory`.

Issue #1110, opened 2023-06-12
Proof failed but Holmake doesn't report it


binghe:
Hi,

currently `examples/dev` (I don't know the underlying theories at all) fails to build *sometimes* with experimental kernel [1]. When I try to build locally, I found the building log of `FactScript.sml` contains some failed proofs:
```
Saved theorem _______ "FactIter_TOTAL"
Saved theorem _______ "FACT_dev"
Proof of 

(COMB (λt. (t,1)) (inp,f_1' <> f_2') ⇔ COMB (λt. t) (inp,f_1')) ∧
COMB (λt. 1) (inp,f_2')

failed.
Proof of 

(COMB (λ(p1,t). (p1,t)) (f_1'³' <> f_2'³',f_1' <> f_2') ⇔
 COMB (λ(p1,t). p1) (f_1'³' <> f_2'³',f_1')) ∧
COMB (λ(p1,t). t) (f_1'³' <> f_2'³',f_2')

failed.
Proof of 

(COMB (λ(p1,t). (p1,t,0)) (f_1'³' <> f_2'³',f_1 <> f_2 <> f_3') ⇔
 COMB (λ(p1,t). p1) (f_1'³' <> f_2'³',f_1)) ∧
COMB (λ(p1,t). (t,0)) (f_1'³' <> f_2'³',f_2 <> f_3')

failed.
Proof of 

(COMB (λ(p1,p1',t). (0,p1',t))
   (f_1'³' <> f_2'³' <> f_3'³',f_1 <> f_2 <> f_3') ⇔
 COMB (λ(p1,p1',t). 0) (f_1'³' <> f_2'³' <> f_3'³',f_1)) ∧
COMB (λ(p1,p1',t). (p1',t)) (f_1'³' <> f_2'³' <> f_3'³',f_2 <> f_3')

failed.
Proof of 

(COMB (λ(p1,p1',t). (p1 − 1,p1',p1' + t))
   (f_1'³' <> f_2'³' <> f_3'³',f_1 <> f_2 <> f_3) ⇔
 COMB (λ(p1,p1',t). p1 − 1) (f_1'³' <> f_2'³' <> f_3'³',f_1)) ∧
COMB (λ(p1,p1',t). (p1',p1' + t)) (f_1'³' <> f_2'³' <> f_3'³',f_2 <> f_3)

failed.
runtime: 0.05062s,    gctime: 0.00000s,     systime: 0.00014s.
[...]

PR #1127, opened 2023-07-17
Updated Dockerfile for CI builds

binghe:
Thanks for the hints. It seems that the remaining issue in #1110 is the PolyML itself - the latest release (5.9) has some issues at exception handling. After switching to latest PolyML `master` code for Docker-based CI builds, so far I haven't observed another building failure.

This PR contains the definition of the current version of Docker Hub image `binghelisp/hol-dev:latest` [1]. (NOTE: it's based on Debian "stable", which got updated to version 12 recently at June 10 as a major Debian release [2].)

There's also a Dockerfile for PolyML-5.4 (useful for very old builds such as HOL-Omega).

NOTE: The docker image here also contains the Unicode fixes ever discussed on Slack when running Emacs IDE in it, thus is also suitable as a "reference development environment" to some HOL users (In this case, the volume `ML` must be created and attached as a persistent personal data store, when running the docker image (`docker run -v ML:/ML ...`)

[1] https://hub.docker.com/r/binghelisp/hol-dev
[2] https://www.debian.org/News/2023/20230610

### DIFF
--- a/developers/docker-ci/base-5.4/Dockerfile
+++ b/developers/docker-ci/base-5.4/Dockerfile
@@ -1,0 +1,50 @@
+# syntax=docker/dockerfile:1
+#
+# HOL4 building environment (Docker), base image for HOL-Omega
+#
+# e.g. docker buildx build --platform linux/386,linux/amd64 -t binghelisp/hol-dev:polyml-5.4 .
+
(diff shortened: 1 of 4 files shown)


binghe:
I was over optimistic... the switching to PolyML master seems indeed increased the success rate of HOL builds, but still not 100%.

PR #1132, opened 2023-07-24
Make Arithconv use cv_compute

myreen:
Thanks to @oskarabrahamsson for spotting and fixing a mistake I made in `Arithconv`!


mn200:
Thanks for this!

PR #1134, opened 2023-07-25
Promote `examples/vector/permutationTheory` to the core library

binghe:
Hi,

I propose we promote the theory of permutation (from HOL-Light) in `examples/vector/permutationScript.sml` to the core library. This theory is more advanced than the `PERMUTES` (a simple overload to `BIJ`) and in practice it's easy to satisfy the extra total-function requirements of it.

The issue was that the the core definition `permutes` is actually already defined in `real_topologyTheory` (and is also used there), and to solve it we shall move the entire theory of permutation to somewhere before `real_topology`.
```
   [permutes]  Definition (real_topologyTheory, old)      
      ⊢ ∀p s. p permutes s ⇔ (∀x. x ∉ s ⇒ p x = x) ∧ ∀y. ∃!x. p x = y
```
Since the entire concept is irrelevant with real numbers but for general sets.  I propose we move this `permutationTheory` into `pred_set/src/more_theories`, just after `cardinalTheory` (where several set-theoretic lemmas are needed).  But the name `permutation` is not very good: it's already used in another file in the examples and is perhaps also used in many unknown user scripts. So I renamed it a little, to `permutes`.   Then, the duplication in `real_topologyTheory` gets removed (and it now opens the new `permutesTheory`). For the previous theory parts involving real numbers and iterations, I appended them at the end of `iterateTheory`.

Other changes: for the building parallelism, I also removed the heap in `src/probability` so that theories like `util_prob` can be built much earlier (than `transcTheory`). I remember this heap was added before the multi-directory build system was introduced.

Chun

### DIFF
--- a/src/real/analysis/real_topologyScript.sml
+++ b/src/real/analysis/real_topologyScript.sml
@@ -18,13 +18,13 @@
 
 open HolKernel Parse boolLib bossLib;
 
-open numTheory numLib unwindLib tautLib Arith prim_recTheory
-     combinTheory quotientTheory arithmeticTheory realTheory real_sigmaTheory
+open numTheory numLib unwindLib tautLib prim_recTheory
+     combinTheory quotientTheory arithmeticTheory realTheory
      jrhUtils pairTheory boolTheory pred_setTheory optionTheory
-     sumTheory InductiveDefinition ind_typeTheory listTheory mesonLib
+     sumTheory InductiveDefinition listTheory mesonLib
      realLib topologyTheory metricTheory netsTheory;
 
-open wellorderTheory cardinalTheory iterateTheory hurdUtils;
+open wellorderTheory cardinalTheory permutesTheory iterateTheory hurdUtils;
 
 val _ = new_theory "real_topology";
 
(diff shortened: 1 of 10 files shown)


mn200:
Excellent!  Thanks for this.

src/real/analysis/real_topologyScript.sml
```diff
@@ -1428,7 +1412,7 @@ val EXCHANGE_LEMMA = store_thm ("EXCHANGE_LEMMA",
       FINITE t /\ independent s /\ s SUBSET span t ==>
       ?t'. t' HAS_SIZE CARD t /\ s SUBSET t' /\ t' SUBSET s UNION t /\
         s SUBSET span t') ==>
-    (!t'':real->bool s':real->bool'. (CARD (t'' DIFF s') < CARD (t DIFF s)) ==>
+    (!t'':real->bool s':real->bool. (CARD (t'' DIFF s') < CARD (t DIFF s)) ==>
```
binghe, line 1415: This PR contains a strange fix that I don't know how to explain. Here at line 1431 (old) there was an extra apostrophe after the type `bool` but for long time HOL parser didn't report any error.   Now, perhaps it's because I removed a few `open`s at the beginning, HOL starts to report parsing error on this line, then I had to remove this extra char.    Do you have any explaination? @mn200

mn200: The code with the extra apostrophe actually continues to work on my most recent checkout of HOL.  What's going on is apparent if you do 
```
> “!b:bool'. b /\ p”;
<<HOL message: inventing new type variable names: 'a>>
val it = “∀b '. b ∧ p”: term
```

We see that the apostrophe is interpreted as an extra variable at the level of the quantifiers, not as part of the `bool` token.  Of course, it's vacuous, so in many contexts you may never realise that this is happening.  I'll change the way that types are lexed to fix this.

PR #1136, opened 2023-08-10
pairwiseN (pred_set) and pairwiseD (topology)

binghe:
Hi,

This is a re-worked PR of #1135 without breaking any compatibility, and without any irrelevant code changes. Since one of the goals is to ease the porting work of HOL-Light proofs using its `pairwise`, I think it cannot be re-defined by overloads. Instead, we can overload HOL-Light's version to `pairwiseD` (and HOL4's version to `pairwiseN` but only locally), and then have their relation as an alternative definition of `pairwiseD`:
```
Overload pairwiseD        = “topology$pairwise”
Overload pairwiseN[local] = “pred_set$pairwise”

[pairwiseD_alt_pairwiseN]  Theorem (topologyTheory)      
    ⊢ ∀R. pairwiseD R = pairwiseN (RC R)
```
I also ported a few more `pairwiseD` theorems directly from HOL-Light.

Another goal is to move the `disjoint` (now as an overload to `pairwiseD DISJOINT`) from `util_probTheory` to `topologyTheory`, so that `disjoint` can be used in theories which do not need `realTheory`, e.g. in some graph theory scripts.

Regards,

Chun Tian


mn200:
If possible, I'd have thought that pairwise-disjointness could be in `pred_set`.

binghe:
> If possible, I'd have thought that pairwise-disjointness could be in `pred_set`.

Are you suggesting the definition and theorems of `pairwiseD` should be moved into `pred_setTheory`? But in this way they cannot both have the name `pairwise`. And if we rename the HOL-Light derived one directly to `pairwiseD`, how can we later run unmodified HOL-Light proofs using just `pairwise`?  (After all, overloading two constants of the same type to the same name seems a bad idea.)

mn200:
No, I meant the results about `pairwiseD DISJOINT`.  But if they need to be expressed with `pairwiseD` then I accept that will be impossible.

binghe:
> No, I meant the results about `pairwiseD DISJOINT`. But if they need to be expressed with `pairwiseD` then I accept that will be impossible.

Yes, currently `disjoint` (`pairwiseD DISJOINT`) are expressed with `pairwiseD`, and many related proofs require small  `DISJOINT_*` lemmas proved by `SET_TAC`. I found it difficult to move all of them to `pred_setTheory` (without using `SET_TAC`).

binghe:
Now I have moved `disjoint` (re-defined as `pairwise (RC DISJOINT)`) to `pred_setTheory` with re-worked proofs.

mn200:
Awesome; thanks for this!

PR #1137, opened 2023-08-10
[fsgraph] fixed walk_def, etc. for trivial (single-vertex) walks

binghe:
Hi,

I think there's a minor problem in the definition of `walk` (and `path`, etc.) in `fsgraphTheory` (finite simple graphs):
```
Definition walk_def:
  walk g vs ⇔ vs ≠ [] ∧ ∀v1 v2. adjacent vs v1 v2 ⇒ adjacent g v1 v2
End
```
The definition is fine when there are two or more vertices (nodes) in the walk, as `adjacent g v1 v2` will guarantee that v1 and v2 are nodes in the graph. But if there's only one vertex in the walk, nothing in the existing definition will make sure this single vertex is part of the involved graph.  As minimal a solution, I think we can add an explicit requirement `(∀v. MEM v vs ⇒ v ∈ nodes g)`:
```
walk_def
⊢ ∀g vs.
    walk g vs ⇔
    vs ≠ [] ∧ (∀v. MEM v vs ⇒ v ∈ nodes g) ∧
    ∀v1 v2. adjacent vs v1 v2 ⇒ adjacent g v1 v2
```
With the new definition, we can prove the following theorems for trivial walks and paths:
```
trivial_path
⊢ ∀g v. v ∈ nodes g ⇒ path g [v]
trivial_walk
⊢ ∀g v. v ∈ nodes g ⇒ walk g [v]
```
while the antecedents `v ∈ nodes g` could be removed under the original definition of `walk`, which seems wrong.

Regards,

Chun Tian

### DIFF
--- a/examples/generic_finite_graphs/fsgraphScript.sml
+++ b/examples/generic_finite_graphs/fsgraphScript.sml
@@ -426,13 +426,33 @@ QED
     Perambulations
    ---------------------------------------------------------------------- *)
 
+(* NOTE: added ‘!v. v IN nodes g’ to make sure vertices of single-vertex
+   walks are indeed vertices in the graph.
+
+   The existing ‘adjacent vs v1 v2 ==> adjacent g v1 v2’ can only guarantee
+   it for walks of two or more vectices.  -- Chun Tian, August 10, 2023.
+ *)
 Definition walk_def:
-  walk g vs ⇔ vs ≠ [] ∧ ∀v1 v2. adjacent vs v1 v2 ⇒ adjacent g v1 v2
+  walk g vs <=> vs <> [] /\ (!v. MEM v vs ==> v IN nodes g) /\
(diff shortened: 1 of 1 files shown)


mn200:
Thanks for spotting this!

binghe:
There's another issue that I may need your help.

Currently the `fsgraphTheory` uses something from `listTheory` but doesn't explicitly open it. I found that, if we actually `open listTheory` explicitly in the beginning of `fsgraphScript.sml`, some proofs will be broken and I don't know how to fix. Could you please have a look on this? The goal is to have `listTheory` explicitly opened, becaue I'm trying to enrich your this theory with proofs using list-related theorems.

mn200:
I just made the obvious change and didn't observe any problems.  Will commit what I have.

binghe:
> I just made the obvious change and didn't observe any problems. Will commit what I have.

hmmm… perhaps my session was broken due to other changes. Thanks anyway.

PR #1142, opened 2023-08-25
Automation for defining from/to functions for cv_compute's cv type

myreen:
The recent cv_compute feature in the HOL4 kernel allows for fast computation, but requires all input to cv_compute to be phrased in terms of the simple Lisp-like `cv` type. 

This PR adds automation for defining mappings between the `cv` type and user-defined datatypes. For each run of the new automation, the tool proves that applying the mapping into the `cv` type and then the mapping out of the `cv` type is the same as applying the identity function.

Future PRs will provide further automation to ease use of the new cv_compute feature.


src/num/theories/cv_compute/automation/Holmakefile
```diff
@@ -0,0 +1,22 @@
...
+.PHONY: all link-to-sigobj
+
+ifdef HOLBUILD
+all: link-to-sigobj
+.PHONY: link-to-sigobj
+link-to-sigobj: $(DEFAULT_TARGETS)
+	$(HOL_LNSIGOBJ)
+endif
+
+selftest.exe: selftest.uo
```
myreen, line 12: @mn200 should this line have more deps?

mn200: As long as selftest.sml clearly depends (via on `open fooTheory` type line) on what's needed, this should be fine.

mn200 (approved):

PR #1144, opened 2023-09-08
Fixed duplicated theorems after the forbidden "rebinds" changes

binghe:
Hi,

There are some (embarrassing) duplicated theorems in core theories about real analysis and measure theory, exposed by commit 705d331db97605557a37c06490763d530ccf9829 . The duplicated theorem in `integrationTheory` is alpha-equivalent with the other, thus can be safely removed. The rest in measure-related theories are renamed in reasonable ways (older theorems keep its name).

Regards,

Chun Tian

### DIFF
--- a/src/probability/measureScript.sml
+++ b/src/probability/measureScript.sml
@@ -5446,7 +5446,7 @@ val _ = reveal "C";
 
 (*** measure_space Theorems ***)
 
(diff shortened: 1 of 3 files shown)


binghe:
NOTE: from the docker builds it seems that there are irrelevant failures at `n-bit-selftest.log` when building HOL with `-t`.

mn200:
Thanks for this!

PR #1145, opened 2023-09-16
itree_bind respects weak bisimulation

Plisp:
feedback appreciated!

### DIFF
--- a/src/coalgebras/itreeTauScript.sml
+++ b/src/coalgebras/itreeTauScript.sml
@@ -638,6 +638,12 @@ Proof
   rw[Once itree_unfold,FUN_EQ_THM]
 QED
 
+Theorem itree_bind_left_identity:
+  itree_bind (Ret x) k = k x
+Proof
+  rw[itree_bind_thm]
+QED
+
 Theorem itree_bind_right_identity:
   itree_bind t Ret = t
 Proof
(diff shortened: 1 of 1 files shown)


src/coalgebras/itreeTauScript.sml
```diff
@@ -638,6 +638,12 @@ Proof
   rw[Once itree_unfold,FUN_EQ_THM]
 QED
 
+Theorem itree_bind_left_identity:
```
mn200, line 641: If not one already, this could be made an automatic rewrite.

src/coalgebras/itreeTauScript.sml
```diff
@@ -898,6 +936,100 @@ Proof
             itree_wbisim_sym]
 QED
 
+Theorem itree_bind_strip_tau_wbisim:
```
mn200, line 939: Maybe also prove the iff result that removes `Tau` constructors?  Something like 
```
wbisim (itree_bind (Tau x) k) y <=> wbisim (itree_bind x k) y
```
And the one with the pattern as the second argument.

src/coalgebras/itreeTauScript.sml
mn200, line 1: Some needless parentheses also.  In many/most cases use the parenthesisation that the pretty-printer gives you.

mn200 (changes requested):

src/coalgebras/itreeTauScript.sml
Plisp: I just removed it since it's the same as itree_bind_thm anyways, which is probably better left manual in case `k x` expands to another bind

src/coalgebras/itreeTauScript.sml
Plisp: this lemma strips any finite number of taus, I'm not sure how to generalize it. The statement you suggested can be proven with metis using `itree_bind_thm` and `itree_wbisim_tau` and transitivity

src/coalgebras/itreeTauScript.sml
Plisp: hopefully gone!

src/coalgebras/itreeTauScript.sml
mn200: I realise it's a trivial consequence, but it's also a great deal more convenient to use for the user as it can apply inside any goal and will be used "for free" by the simplifier with the `simp` annotation.

Plisp: ah I see what you mean and you're right! It wasn't strictly a consequence since `strip_tau` force removes all taus, but it did make the proof of the theorem simpler. Added

mn200:
Thanks for this!

mn200 (approved):

PR #1146, opened 2023-09-21
Fermat

jhlchan:
Rework the hopping algorithm of Fermat's Two Squares Theorem for an extended paper.


mn200:
Sorry, I've just done a little more on this myself and we now have a merge conflict.  If you've handled all duplicates, can you remove the trace-setting line in `jcLib.sml` and check that everything still goes through.

mn200:
Thanks.  My comment about a trace relates to a commit that I now see is only on my local machine.  Sorry for any confusion.

jhlchan:
It's all good, thanks.

PR #1150, opened 2023-10-03
more results of terms having head normal forms (`has_hnf`)

binghe:
Hi,

This is a stage work proving more results of terms having head normal forms (`has_hnf`), independent with solvable terms. The end of the chain of proved theorems is the following theorems with `[simp]` (Proposition 8.3.13 (i) of [1, p.174]):
```
[has_hnf_iff_LAM] (standardisationTheory)
    ⊢ ∀x M. has_hnf (LAM x M) ⇔ has_hnf M
```

Now I have to use `chap3Theory`, i.e. using theorems about (beta) reductions (`beta` and `grandbeta`) to prove results about beta conversions (`lameq`). The following important lemmas are proved:
```
lameq_CR (chap3Theory)
    ⊢ ∀M N. M == N ⇒ ∃Z. M -β->* Z ∧ N -β->* Z

grandbeta_imp_betastar (chap3Theory)
    ⊢ ∀M N. M =β=> N ⇒ M -β->* N

grandbeta_imp_lameq (chap3Theory)
    ⊢ ∀M N. M =β=> N ⇒ M == N

abs_betastar (chap3Theory, statements modified in 2nd commit)
    ⊢ ∀x M Z. LAM x M -β->* Z ⇒ ∃N'. Z = LAM x N' ∧ M -β->* N'
```

P.S. With the above theorem "has_hnf_iff_LAM" and the following theorem (slightly renamed) added by previous PR #1148:
```
solvable_iff_LAM (solvableTheory)
    ⊢ ∀x M. solvable (LAM x M) ⇔ solvable M
```
we are now ready to show `|- solvable M <=> has_hnf M` (a term is solvable iff has head normal forms) by first reducing all involved terms to closed terms (by closing the terms with `LAM`s of free variables.), i.e. `|- closed M ==> solvable M <=> has_hnf M`.

[1] Barendregt, H.P.: The Lambda Calculus, Its Syntax and Semantics. College Publications, London (1984).

### DIFF
--- a/examples/lambda/barendregt/chap2Script.sml
+++ b/examples/lambda/barendregt/chap2Script.sml
@@ -80,15 +80,17 @@ val one_hole_is_normal = store_thm(
                                   MP_TAC (MATCH_MP imp (CONJ cth th))))) THEN
   SIMP_TAC std_ss []);
 
-val (lameq_rules, lameq_indn, lameq_cases) = (* p. 13 *)
-  IndDefLib.xHol_reln "lameq"
-    `(!x M N. (LAM x M) @@ N == [N/x]M) /\
+Inductive lameq :
+     (!x M N. (LAM x M) @@ N == [N/x]M) /\
      (!M. M == M) /\
+[~SYM:]
      (!M N. M == N ==> N == M) /\
+[~TRANS:]
      (!M N L. M == N /\ N == L ==> M == L) /\
      (!M N Z. M == N ==> M @@ Z == N @@ Z) /\
      (!M N Z. M == N ==> Z @@ M == Z @@ N) /\
-     (!M N x. M == N ==> LAM x M == LAM x N)`;
+     (!M N x. M == N ==> LAM x M == LAM x N)
+End
 
 val lameq_refl = Store_thm(
   "lameq_refl",
@@ -104,9 +106,6 @@ val lameq_weaken_cong = store_thm(
   "lameq_weaken_cong",
   ``(M1:term) == M2 ==> N1 == N2 ==> (M1 == N1 <=> M2 == N2)``,
   METIS_TAC [lameq_rules]);
-
-Theorem lameq_SYM = List.nth(CONJUNCTS lameq_rules, 2)
-Theorem lameq_TRANS = List.nth(CONJUNCTS lameq_rules, 3)
 
 val fixed_point_thm = store_thm(  (* p. 14 *)
   "fixed_point_thm",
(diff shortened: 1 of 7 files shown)


examples/lambda/barendregt/chap3Script.sml
```diff
@@ -925,9 +925,11 @@ Proof
  >> MATCH_MP_TAC grandbeta_imp_betastar >> art []
 QED
 
-(* cf. abs_grandbeta, added by Chun Tian *)
+(* |- !R x y z. R^+ x y /\ R^+ y z ==> R^+ x z *)
+Theorem TC_TRANS[local] = REWRITE_RULE [transitive_def] TC_TRANSITIVE
+
 Theorem abs_betastar :
```
mn200, line 931: This could be an iff.

examples/lambda/barendregt/standardisationScript.sml
```diff
@@ -2499,10 +2488,68 @@ Proof
...
  >> MATCH_MP_TAC lameq_appstar_cong >> art []
 QED
 
+Theorem hnf_LAMl[simp] :
+    hnf (LAMl vs M) <=> hnf M
+Proof
+    Induct_on ‘vs’ >> rw []
+QED
+
+Theorem hnf_appstar :
```
mn200, line 2502: If you lift the `Ns <> []` assumption out, I believe the statement could be an iff:
```
Ns <> [] ==> (hnf (M @* Ns) <=> hnf M /\ ~is_abs M)
```

examples/lambda/barendregt/chap2Script.sml
```diff
@@ -106,6 +106,7 @@ val lameq_weaken_cong = store_thm(
   METIS_TAC [lameq_rules]);
 
 Theorem lameq_SYM = List.nth(CONJUNCTS lameq_rules, 2)
```
mn200, line 108: If you use the labelling option in the Inductive definition of `lameq` you should be able to get these for free:
```
...
[~SYM:]
(!M N. M == N ==> N == M) /\
[~TRANS:]
(!M N P. M == N /\ N == P ==> M == P) /\ ...
```
You will need to replace the call to `xHol_reln` with `Inductive lameq:`

binghe:
Thanks for your code review @mn200, I have made all the related changes (my first time using the modern syntax of `Inductive` with tags)

binghe:
If you could take a look at `hnf_cases `:
```
    !M. hnf M ==> ?vs args y. M = LAMl vs (VAR y @* args)
```
I believe this theorem may be also improved as an iff: if `M = LAMl vs (VAR y @* args)` holds, then there should be no way to do one more head reduction as `VAR y` plus `HD args` (if exists) is not a valid head redex. Is this true?

mn200:
Thanks for all this!

I agree that your implication looks to definitely be an iff.  Will look at it.

mn200:
See b1932f570

PR #1152, opened 2023-10-05
itree_iter respects wbisim + strong bisimulation

Plisp:
added itree_bind_thm simp + itree_iter respecting pointwise wbisim proof


src/coalgebras/itreeTauScript.sml
```diff
@@ -1026,6 +1038,113 @@ Proof
   metis_tac[itree_bind_resp_t_wbisim, itree_bind_resp_k_wbisim, itree_wbisim_trans]
 QED
 
+Theorem itree_iter_ret_tau_wbisim:
```
mn200, line 1041: If you can't ever see users wanting to use this result outside of this script file, then may add a `[local]` annotation to this one.

mn200 (changes requested):

src/coalgebras/itreeTauScript.sml
Plisp: in what situation should I use `[local]` vs `Triviality`?

mn200: They’re equivalent.  Use whichever you find most appealing aesthetically.

mn200 (approved):

Issue #1010, opened 2022-03-24
HolSmt tests failed when Yices is installed


binghe:
Hi,

I have my `HOL4_YICES_EXECUTABLE` environment variable set to Yices 1.0.40 since many years ago. I never actually use Yices to do anything:
```
$ echo $HOL4_YICES_EXECUTABLE 
/Users/binghe/ML/yices-1.0.40/bin/yices
```

Usually I run HOL building process with `-t` only to check Unicode violations and I terminate the building process once I saw no violation. Today I occasionally leave the building process running there, and then I suddenly met this HolSmt test failure. To reproduce it, just run `./selftest.exe` in `$(HOLDIR)/src/HolSmt`: (Yices must installed and configured though the above environment variable, of course.)

```
~/ML/HOL.expk/src/HolSmt$ ./selftest.exe 
<<HOL message: Can't replay proofs with Z3 v4.8.1>>
<<HOL message: HolSmtLib: solver Yices is available.>>
<<HOL message: HolSmtLib: solver Z3 (oracle) is available.>>
<<HOL message: HolSmtLib: solver Z3 (with proofs) is available.>>
Testing HolSmtLib
Saved theorem _______ "num2dt1_dt12num"
Saved theorem _______ "dt12num_num2dt1"
Saved theorem _______ "num2dt1_11"
Saved theorem _______ "dt12num_11"
Saved theorem _______ "num2dt1_ONTO"
Saved theorem _______ "dt12num_ONTO"
Saved theorem _______ "num2dt1_thm"
Saved theorem _______ "dt12num_thm"
Saved theorem _______ "dt1_EQ_dt1"
Saved theorem _______ "dt1_case_def"
Saved theorem _______ "datatype_dt1"
Saved theorem _______ "dt1_distinct"
Saved theorem _______ "dt1_nchotomy"
Saved theorem _______ "dt1_Axiom"
Saved theorem _______ "dt1_induction"
Saved theorem _______ "dt1_case_cong"
Saved theorem _______ "dt1_case_eq"
<<HOL message: Defined type: "dt1">>
Saved theorem _______ "person_size_eq"
Saved theorem _______ "datatype_person"
Saved theorem _______ "person_11"
Saved theorem _______ "person_nchotomy"
Saved theorem _______ "person_Axiom"
Saved theorem _______ "person_induction"
Saved theorem _______ "person_case_cong"
Saved theorem _______ "person_case_eq"
[...]

PR #1173, opened 2023-12-12
HolSmt: add support for the cvc5 SMT solver + doc update

someplaceguy:
This PR adds support for the [cvc5 SMT solver](https://cvc5.github.io/). It also updates HolSmt's documentation and enables more tests for the Z3 SMT solver.

It is split into 4 commits:

* The first commit is just a small fix to a couple of HolSmt's tests. These tests seemed to expect the goals to be theorems, but in fact they are false.
  * The first one was demonstrated to be false when cvc5 gave me the `x=-127` and `y=127` counterexample, which I subsequently verified purely in HOL4 just to be sure.
  * The second one was already marked as false and commented out (with a counterexample), but I uncommented it (with an empty list of provers) in preparation for adding a cvc5 satisfiable test (which succeeds when the goal is false) in a subsequent commit.

This minor fix to these two tests was put into a separate commit for two reasons: 1) it's independent of the other improvements/commits, and 2) just to make sure that this reversal in the expected behavior wasn't lost in the noise.

* The second commit adds proper support for the cvc5 SMT solver. Adding support for it was extremely straightforward, exactly as HolSmt's documentation suggested it would be.

The only very minor issue was that cvc5 printed out a warning that performance will not be optimal if a specific logic was not specified. To avoid this warning, I modified HolSmt to prepend a `(set-logic <logic>)` command to the generated file.

Currently, and just like before this PR, no attempt is being made to determine what is the optimal logic for a query. Instead, Z3 passes `NONE` (which instructs `SmtLib.sml` to not generate a `set-logic` command) and cvc5 passes `SOME "ALL"` in order to generate a `(set-logic ALL)` command, as that is what cvc5 suggests should be done in this case.

The `HOL4_CVC_EXECUTABLE` environment variable is used to find the `cvc5` executable, following the same pattern as for the other SMT solvers.

* The third commit updates HolSmt's documentation:
  * URLs were updated and/or changed from http->https.
  * Documentation for the cvc5 solver was added.
  * It is now mentioned that a modern version of Z3 has been tested to work as an oracle (version 4.8.17 specifically, as that's the one I've been testing).
  * It is now clarified that only Z3 2.19 supports proof reconstruction.
  * I've also added a note that Z3 2.19.1 does not work with reals due to an apparent bug in parsing real literals. That said, this note is kind of buried in the download instructions, not sure if it should be mentioned more prominently.
    * As a side note, I've added a workaround to fix this, but then HolSmt failed elsewhere in reconstructing a Z3 2.19.1 proof, and since I'm not familiar with proof reconstruction I decided to scrap the workaround and just document the issue for now.
  * The example session was updated to the newer syntax.

Note: I don't know what are the build commands to generate the documentation PDFs since (ironically) I couldn't find documentation for that. I also have zero LaTeX experience, so I have no idea if the changes I've made to the documentation are correct or how they actually look in rendered form.

* The fourth commit simply enables (or re-enables) tests for Z3 that I've confirmed to be working for both Z3 2.19 (which HolSmt supports for proof reconstruction) and Z3 4.8.17 (a modern version of Z3).

There are a couple of related questions that I haven't dealt with (at least, not yet):

* Does HOL4 run the HolSmt self tests as part of some kind of CI? If so, I wonder which Z3 version is being used.

More importantly, I wonder if support for cvc5 should also be added to the CI pipeline (assuming it exists?). In theory, all that's needed is to install cvc5 and set the `HOL4_CVC_EXECUTABLE` environment variable to point to it.

* SMT solver reliability.

I know that Z3 (and I believe cvc5 as well) support deterministic queries, by setting a specific random seed and by setting the search limit to be a specific number of steps rather than a time limit.

However, HolSmt currently does not attempt to do this at all, so I'm guessing that SMT solvers default to using a random seed and running with a time limit as the criteria for giving up when the query is too difficult.

Given that, I wonder if anyone has experienced HolSmt tests to fail occasionally, due to the non-deterministic nature of these time limits and random seeds or if this hasn't been an issue at all?

### DIFF
--- a/Manual/Description/HolSmt.tex
+++ b/Manual/Description/HolSmt.tex
@@ -35,43 +35,46 @@ be much better).
 
 \subsection{Interface}
 
-The library currently provides three tactics to invoke different SMT
-solvers, namely \ml{YICES\_TAC}, \ml{Z3\_ORACLE\_TAC}, and
-\ml{Z3\_TAC}.  These tactics are defined in the \ml{HolSmtLib}
-structure, which is the library's main entry point.  Given a
-goal~$(\Gamma, \varphi)$ (where $\Gamma$ is a list of assumptions, and
-$\varphi$ is the goal's conclusion), each tactic returns (i)~an empty
-list of new goals, and (ii)~a validation function that returns a
-theorem~$\Gamma' \vdash \varphi$ (with $\Gamma' \subseteq \Gamma$).
-These tactics fail if the SMT solver cannot prove the
-goal.\footnote{Internally, the goal's assumptions and the
-  \emph{negated} conclusion are passed to the SMT solver.  If the SMT
-  solver determines that these formulas are unsatisfiable, then the
-  (unnegated) conclusion must be provable from the assumptions.}  In
+The library currently provides four tactics to invoke different SMT
+solvers, namely \ml{CVC\_ORACLE\_TAC}, \ml{YICES\_TAC},
+\ml{Z3\_ORACLE\_TAC}, and \ml{Z3\_TAC}.  These tactics are defined in
+the \ml{HolSmtLib} structure, which is the library's main entry point.
+Given a goal~$(\Gamma, \varphi)$ (where $\Gamma$ is a list of
+assumptions, and $\varphi$ is the goal's conclusion), each tactic
+returns (i)~an empty list of new goals, and (ii)~a validation function
+that returns a theorem~$\Gamma' \vdash \varphi$ (with $\Gamma'
+\subseteq \Gamma$). These tactics fail if the SMT solver cannot prove
+the goal.\footnote{Internally, the goal's assumptions and the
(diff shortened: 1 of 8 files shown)


mn200:
Thanks for all your work on this.  I'd certainly like to get this tested automatically.  We have a docker image setup for the CI runs done on Github (@binghe set these up). I also run `-t3` tests on a `cron` job every 24 hours.  I'd like to have one or both of these augmented with an appropriate set of SMT installs.

binghe:
In next days, I will modify our CI docker image with Z3, Yices and latest cvc5 solver installed, and make each of them separately switchable. Besides, it's definitely possible to switch between different versions of each solver.

P. S. I ever reported some test issues with a version of Yices (#1010), thus by default Yices should be disabled. Then perhaps it's time to take a deeper look and see if the problem is HOL itself or Yices.

binghe:
See #1179

PR #1177, opened 2024-01-12
[CCS] define "nil" by "I"-combinator (rec X (var X))

binghe:
Hi,

This small PR follows the previous #1176 and future re-defined CCS deadlock process `nil` by the "I"-combinator (`rec X (var X)`, where `X` is arbitrarily chosen).  This change slightly reduced the efforts in defining CCS terms by `generic_termsTheory` and all the proofs involving inductions on CCS terms, as now the induction case for `nil` now disappeared. This change also pave the way to some other ideas in the future.

This settles all the necessary changes before porting CWB to HOL4.

--Chun


mn200:
Thanks!

PR #1180, opened 2024-01-16
Fixed temporal_deep tests

binghe:
Hi,

HOL built by Moscow ML is still working, including the `HolBdd` examples and `temporal_deep` examples with a SMV checker. When `HOL4_SMV_EXECUTABLE` is set to NuSMV, nuXmv or NuRV (all from FBK), the self tests in `examples/temporal_deep/src/model_check/selftest.sml` also still work but there are two unmatched parentheses by last commit (in 2018).

Chun

### DIFF
--- a/examples/HolBdd/MachineTransitionScript.sml
+++ b/examples/HolBdd/MachineTransitionScript.sml
@@ -990,6 +990,7 @@ val TotalMooreTrans =
     THEN Q.EXISTS_TAC `(q',nextfn(q,r))`
     THEN RW_TAC std_ss [MooreTrans_def]);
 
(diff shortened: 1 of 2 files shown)


binghe:
In #1179, the latest updates to Docker image `binghelisp/hol-dev:latest` now contains an installation of NuRV [1], which is compatible with the original SMV checker.  When pulling this image and building HOL with Moscow ML, all tests under `examples/temporal_deep/src/model_check` will work, with help of the SMV checker on LTL formulas.

[1] https://es-static.fbk.eu/tools/nurv/

mn200:
Thanks for this!

PR #1186, opened 2024-01-23
HolSmt improvements

someplaceguy:
This PR includes 8 commits. A brief description of each one follows (please don't miss the one that affects code outside HolSmt):

1. a51b053777340c4ad7c0ef41062dc187c774b4ff Removes `loose_equality ()` in HolSmt self-tests, updating the tests so their meaning doesn't change.
2. 0bdd3da86e622d64d7e52e2bee4dfee56c2a062e Updates HolSmt documentation with regards to tested SMT solver versions.
3. ff36b748b281898eecab13b1378bfa1ed85232ff Improves the error message shown when a string-to-integer conversion fails during parsing of Z3 proofs.
4. 3635ff3579a228a97fa4dd59e8117aca304d66b9 This is the only commit that should affect code outside HolSmt. It adds the function `pair_map` to Lib and Portable. I was surprised such a function wasn't there already, since it's so similar to `apfst` and `apsnd`, so I looked around and saw it was being defined locally in a couple of other theories. In subsequent commits I make use of such a function in HolSmt. However, if you'd like me to drop this change, please let me know and I will define such a function locally in HolSmt as well.
5. 5208b7e86e067e0d3e78e744284c6890b4952acd Fixes HolSmt's SMT-LIB parser to allow symbols when parsing indexed identifiers (introduced in SMT-LIB's 2.0 -> 2.5 changes). However, this commit by itself doesn't change how HolSmt parses anything (this will change in the next commit).
6. cf96a5f0948d8df66723766e808acc35997a0c1c Fixes HolSmt to parse `th-lemma` rules (i.e. theory-specific rules) in Z3 proof replay/reconstruction, now that we can parse indexed identifiers with symbols. Apparently, theory-specific rules weren't being replayed at all anymore due to this parsing issue. It also reenables proof replay in some tests that were broken due to this issue.
7. f443d2a460aeb503fa80b9c4e50d744d23b43d46 Reenables Z3 proof replay in some tests that include reals.
8. bd42c1739c72f317faab3a53d80c33d10bd52a6b Minor code simplification. Also enables cvc5 and Z3 for more tests, including Z3 with proof replay.

In total, this series of improvements enables or reenables Z3 proof reconstruction in 45 more test cases in the HolSmt self-tests.

Also, please note that a more detailed description of each commit is included in the commit messages.

Thanks!

**Edit:** added one more commit enabling more tests, and renamed `pmap` to `pair_map` as per @tjark's suggestion.


someplaceguy:
Regarding `pmap`, I just noticed that I missed removing an identical definition of a `pair_map` function in `examples/HolCheck/commonTools.sml`, which is used in `cearTools.sml` inside the same directory. Additionally, `src/HolSat/satCommonTools.sml` also defines an identical `pair_map` function but it doesn't seem to be used anywhere.

If you'd like I can clean this up as well. I also wonder if `pair_map` is a better name than `pmap`... it's a bit more verbose but also more descriptive. Any thoughts?

tjark:
I had to read the description of your pull request before I realized that pmap did not mean parallel map, so I'd prefer pair_map for the reason that you mentioned: it is more descriptive.

someplaceguy:
Update: renamed `pmap` to `pair_map` as per @tjark's suggestion, and added one more commit which does a minor code simplification and enables 8 more tests for cvc5 and Z3, including Z3 with proof reconstruction. Thanks!

mn200:
Thanks!

PR #1188, opened 2024-01-26
HolSmt: fix Z3 proof replay in real arithmetic with rational coefficients

someplaceguy:
Sorry, while working on the newer Z3 proof reconstruction, I unexpectedly ran into another easy fix, hence this PR.

Namely, due to a comment in the self-tests I was under the impression that Z3 2.19 wasn't printing real constants correctly, but it turns out that it does (I suspect the issue may be with Z3 2.19.1 instead). Hence, it is possible to fix the proof reconstruction for the remaining real-related test cases.

After a quick investigation, it was easy to figure out the problem.

I will reproduce the initial part of the commit message here (see the commit for more details):

>    It turns out that the existing Z3 proof reconstruction code was using
>    RealArith.REAL_ARITH to replay linear arithmetic steps in proofs.
>
>    However, RealArith.REAL_ARITH only handles integral coefficients and
>    therefore proof steps such as `(x: real) / 1r = x` were failing.
>
>    To fix this, we simply switch to RealField.REAL_ARITH, which supports
>    rational coefficients.
>
>    This allows us to enable proof reconstruction for 3 more self-tests,
>    all related to the reals. This also means that all real-related tests
>    in the self-tests now work with proof reconstruction.

Another commit was also added that simply eliminates a few self-tests that were duplicated.

**Edit:** instead of removing the test cases, they were fixed to do what was almost surely intended, which is to test the reals in the same way the nums and ints are being tested.

Thanks!

### DIFF
--- a/src/HolSmt/selftest.sml
+++ b/src/HolSmt/selftest.sml
@@ -548,10 +548,10 @@ in
     (``(x:real) < x``, [sat_CVC, sat_YO, sat_Z3, sat_Z3p]),
     (``(x:real) < y ==> 42 * x < 42 * y``, [thm_AUTO, thm_CVC, thm_YO, thm_Z3, thm_Z3p]),
 
-    (``0n <= 1n``, [thm_AUTO, thm_YO]),
-    (``1n <= 0n``, [sat_CVC, sat_YO, sat_Z3, sat_Z3p]),
-    (``(x:num) <= x``, [thm_AUTO, thm_YO]),
-    (``(x:num) <= y ==> 2 * x <= 2 * y``, [thm_AUTO, thm_YO]),
+    (``0r <= 1r``, [thm_AUTO, thm_CVC, thm_YO, thm_Z3, thm_Z3p]),
+    (``1r <= 0r``, [sat_CVC, sat_YO, sat_Z3, sat_Z3p]),
+    (``(x:real) <= x``, [thm_AUTO, thm_CVC, thm_YO, thm_Z3, thm_Z3p]),
+    (``(x:real) <= y ==> 42 * x <= 42 * y``, [thm_AUTO, thm_CVC, thm_YO, thm_Z3, thm_Z3p]),
 
     (``1r > 0r``, [thm_AUTO, thm_CVC, thm_YO, thm_Z3, thm_Z3p]),
     (``0r > 1r``, [sat_CVC, sat_YO, sat_Z3, sat_Z3p]),
--- a/src/HolSmt/selftest.sml
+++ b/src/HolSmt/selftest.sml
@@ -213,9 +213,6 @@ in
 
     (* real *)
 
-    (* Z3 2.19 prints reals as integers in its proofs; I see no way to
-       reliably distinguish between them. *)
-
     (``0r = 0r``, [thm_AUTO, thm_CVC, thm_YO, thm_Z3, thm_Z3p]),
     (``1r = 1r``, [thm_AUTO, thm_CVC, thm_YO, thm_Z3, thm_Z3p]),
     (``0r = 1r``, [sat_CVC, sat_YO, sat_Z3, sat_Z3p]),
@@ -450,9 +447,9 @@ in
     (``1 * (x:real) = x``, [thm_AUTO, thm_CVC, thm_YO, thm_Z3, thm_Z3p]),
     (``(x:real) * 42 = 42 * x``, [thm_AUTO, thm_CVC, thm_YO, thm_Z3, thm_Z3p]),
 
-    (``(x:real) / 1 = x``, [thm_AUTO, thm_CVC, thm_YO, thm_Z3(*, thm_Z3p*)]),
-    (``x > 0 ==> (x:real) / 42 < x``, [(*thm_AUTO,*) thm_CVC, thm_YO, thm_Z3(*, thm_Z3p*)]),
-    (``x < 0 ==> (x:real) / 42 > x``, [(*thm_AUTO,*) thm_CVC, thm_YO, thm_Z3(*, thm_Z3p*)]),
+    (``(x:real) / 1 = x``, [thm_AUTO, thm_CVC, thm_YO, thm_Z3, thm_Z3p]),
+    (``x > 0 ==> (x:real) / 42 < x``, [(*thm_AUTO,*) thm_CVC, thm_YO, thm_Z3, thm_Z3p]),
+    (``x < 0 ==> (x:real) / 42 > x``, [(*thm_AUTO,*) thm_CVC, thm_YO, thm_Z3, thm_Z3p]),
 
     (``abs (x:real) >= 0``, [thm_AUTO, thm_YO]),
     (``(abs (x:real) = 0) = (x = 0)``, [thm_AUTO, thm_YO]),
--- a/src/HolSmt/Z3_ProofReplay.sml
+++ b/src/HolSmt/Z3_ProofReplay.sml
@@ -798,7 +798,7 @@ local
         handle Feedback.HOL_ERR _ =>
 
         if term_contains_real_ty t then
-          profile "rewrite(11.1)(REAL_ARITH)" RealArith.REAL_ARITH t
+          profile "rewrite(11.1)(REAL_ARITH)" RealField.REAL_ARITH t
         else
           profile "rewrite(11.2)(ARITH_PROVE)" intLib.ARITH_PROVE t
     in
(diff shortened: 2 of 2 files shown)


someplaceguy:
Oh, I just noticed that `realLib.REAL_ARITH` also exists:

```sml
  fun REAL_ARITH tm      = RealArith.OLD_REAL_ARITH tm
                           handle HOL_ERR _ => RealField.REAL_ARITH tm;
```

I'm sure you're aware of this, but for anyone who might not be aware, `RealArith.OLD_REAL_ARITH` is described in `src/real/selftest.sml` as:
> The original version and old port by Hurd

I wonder why `realLib`'s version first tries the old port and only if it fails, it tries the new one. Is it because the old port is faster (or solves more goals)?

If the `realLib` version is preferred, I can switch the code to use `realLib.REAL_ARITH` rather than `RealField.REAL_ARITH`.

binghe:
The reason that realLib's `REAL_ARITH` first tries the old port and then the new one (after #1043), is because the old port can solve some (less interesting) inputs that the new port (same as HOL-Light's current version) cannot. To strictly align with HOL-Light's `REAL_ARITH` (for porting proofs from HOL-Light to HOL4), I chose to not further enhance the new port to cover those (known) extra inputs.

When `HolSmt` was written, HOL4's `REAL_ARITH` was the old port, and the original author of `HolSmt` has only `REAL_ARITH` with integral coefficients. I'm glad to see that the new port of `REAL_ARITH` can also enrich the set of solvable cases of `HolSmt`.

binghe:
Your current choice of using `RealField.REAL_ARITH` is good, I think. It's slightly slower but given the `HolSmt` code has no ability to actively choose different versions (by checking the coefficients), it has no choice but to always use the full version.  

The interface provided by `realLib` is basically for users who doesn't care which version to use.

someplaceguy:
Thank you for the background info, that clarifies things for me!

So it sounds like it'd be better to use `realLib` rather than `RealField`, since that would allow HolSmt to reconstruct more proofs.

I will fix the PR accordingly. Thanks!

someplaceguy:
> is because the old port can solve some (less interesting) inputs that the new port

I just read the PR where you introduced the new port. Are those extra inputs that the old port can solve (but the new can't) those that contain `abs`, `min` and `max`, or is there anything else that the old port can do that the new port can't?

If those are the only inputs, then it will be fine to use `RealField` as you say, because `abs`, `min` and `max` are already being handled in another HolSmt "layer".

binghe:
> Thank you for the background info, that clarifies things for me!
> 
> So it sounds like it'd be better to use `realLib` rather than `RealField`, since that would allow HolSmt to reconstruct more proofs.
> 
> I will fix the PR accordingly. Thanks!
> 
> **Edit:** I just saw your latest comment. Now I am unsure again!

There's no difference (on the number of solvable cases) for your two choices at this moment, because I don't think the extra problems that `OLD_REAL_ARITH` can solve, may appear inside `HolSmt`.  Those "extra" problems solvable by `OLD_REAL_ARITH` are something like `|- f (x + x) = f (2 * x)`, where `f` is an arbitrary uninterpreted function irrelevant to reals.  Basically, by doing some extra rewritings, the new `REAL_ARITH` can also solve them.

someplaceguy:
Ok, sorry, I was really confused for a moment.

> Those "extra" problems solvable by OLD_REAL_ARITH are something like |- f (x + x) = f (2 * x), where f is an arbitrary uninterpreted function irrelevant to reals.

Yes, I think those kind of problems (such as the one you mentioned) are already being handled by other proof rules (like monotonicity, in this case), so I think `RealField` should be enough, as you say. And of course, if it turns out not to be the case, it can always be fixed later if necessary.

So I will leave the PR as it is then.

Thank you very much and sorry for the noise!

binghe:
> Ok, sorry, I was really confused for a moment.
> 
> > Those "extra" problems solvable by OLD_REAL_ARITH are something like |- f (x + x) = f (2 * x), where f is an arbitrary uninterpreted function irrelevant to reals.
> 
> Yes, I think those kind of problems (such as the one you mentioned) are already being handled by other proof rules (like `monotonicity`, in this case), so I think `RealField` should be enough, as you say. And of course, if it turns out not to be the case, it can always be fixed later if necessary.
> 
> So I will leave the PR as it is then.
> 
> Thank you very much and sorry for the noise!

OK, I check the code, the known cases that HOL-Light's latest version cannot solve were something like `|- closed {x | 1 * x = a} <=> closed {x | x = a}’, found when building core library with the new port.  Then I put some extra code into `RealArith.sml` (at line 2361) to make sure such cases can also be solved. But there's no way to know what else problems that the old port can do but the new one cannot, because their implementations are completely separate, no shared code at all.

src/HolSmt/selftest.sml
```diff
@@ -551,11 +548,6 @@ in
     (``(x:real) < x``, [sat_CVC, sat_YO, sat_Z3, sat_Z3p]),
     (``(x:real) < y ==> 42 * x < 42 * y``, [thm_AUTO, thm_CVC, thm_YO, thm_Z3, thm_Z3p]),
 
-    (``0n <= 1n``, [thm_AUTO, thm_YO]),
-    (``1n <= 0n``, [sat_CVC, sat_YO, sat_Z3, sat_Z3p]),
-    (``(x:num) <= x``, [thm_AUTO, thm_YO]),
-    (``(x:num) <= y ==> 2 * x <= 2 * y``, [thm_AUTO, thm_YO]),
```
binghe, line 557: Hi. Is this problem (`(x:num) <= y ==> 2 * x <= 2 * y`) supposed to be solvable only by Yices and not others?   I wonder why you have removed the above 4 cases, although the 2nd one (`1n <= 0m`) seems false.

someplaceguy: Right, so there are a few things going on here:

>  I wonder why you have removed the above 4 cases

So, the 4 test cases are being removed because the same test cases are already being performed above:
* The original test cases: https://github.com/HOL-Theorem-Prover/HOL/blob/5eb0e7b85e5c8789be46b9ff2a67d209d3981349/src/HolSmt/selftest.sml#L486-L489
* The duplicate ones: https://github.com/HOL-Theorem-Prover/HOL/blob/5eb0e7b85e5c8789be46b9ff2a67d209d3981349/src/HolSmt/selftest.sml#L554-L557

The duplicate ones are even in the wrong section (reals instead of nums). I suspect the duplicate ones might have been copy/pasted from the nums for reference, but someone forgot to delete them afterwards.

> Is this problem `((x:num) <= y ==> 2 * x <= 2 * y)` supposed to be solvable only by Yices and not others?

Currently, yes. The HolSmt support for Yices 1 uses a different translation which supports natural numbers. However, SMT-LIB doesn't support natural numbers, only integers, so we currently translate the natural numbers into uninterpreted constants.

It is possible to translate the natural numbers to the integers in an equivalence-preserving way (like it was done in Isabelle or perhaps F*) but we don't do this yet. I intend to fix this in the future (if nobody else does it first), since this is an extremely important use-case for me.

> the 2nd one (1n <= 0m) seems false.

It is indeed false. The HolSmt tests include tests for true statements (i.e. theorems) and false ones. When the SMT solvers can prove that the goal is false, they return `sat` instead of `unsat` (hence why that test case is tested with `sat_CVC`, `sat_Z3` and `sat_Z3p` rather than `thm_CVC`, `thm_Z3`, and `thm_Z3p`). In this case, it's easy for them to prove false statements about the natural numbers because they are allowed to assign whatever they want to uninterpreted constants, thus it's easy to assign something that makes the statement false.

This means that when we are forced to use uninterpreted constants or functions (which still happens a lot due to HolSmt limitations), a `sat` answer may not always imply that the goal is false, but an `unsat` answer always implies that the goal is true.

Hopefully, in the future we won't need to rely on uninterpreted symbols so much, making a `sat` answer a much more reliable indicator that the goal is false.

In my to-do list, there's also a task which consists in clarifying this to the HolSmt user: when the SMT solver replies `sat`, I'd like HolSmt to indicate to the user whether the `sat` result may be spurious or not. And also, I'd like to make it clear that when a non-spurious `sat` result is encountered, it indicates that the goal is false.

someplaceguy:
Wait, I made a mistake with those test cases after all. The test cases are duplicated indeed, but now I think the correct solution is not to remove them, but rather to convert them to use reals instead of nums, so that the same pattern of tests is followed for both nums and reals.

I will fix this PR soon to do that.

Thank you for bringing this to my attention @binghe, I'd probably have missed this issue otherwise. It's easy to get lost in the sea of tests :sweat_smile:

someplaceguy:
Update: instead of eliminating the test cases, I've changed them to cover the reals now, which is almost surely what was originally intended: to test these inequalities in the reals in the same way that they are being tested in the nums and the ints.

Thanks!

mn200:
Thanks both!

PR #1191, opened 2024-02-01
[lambda] permutator and more/improved subterm-related lemmas

binghe:
Hi,

I was stuck on the lambda example and finally made some new progress. This PR is a stage work for code review, it mainly adds a new combinator called "permutator" into `chap2Theory`:
```
[permutator_def]
⊢ ∀n. permutator n =
      (let
         Z = FRESH_list (n + 1) ∅;
         z = LAST Z
       in
         LAMl Z (VAR z ·· MAP VAR (FRONT Z)))
```
with its main property as the following theorem:
```
[permutator_thm]
⊢ ∀n N Ns. LENGTH Ns = n ⇒ permutator n ·· Ns @@ N == N ·· Ns
```
The proof has used the `fromPairs` in the previous closed PR #1181. I was thinking that `alistTheory` may help but actually that's not true, mostly because it's hard to use `ALOOKUP` with `ssub` in a natural way. Thus the code changes for the lambda examples in #1181 are now included in the present PR.

With the `permutator`, the existing proof of the following theorem gets simplified:
```
Boehm_transform_exists_lemma1 [boehmTheory]
⊢ ∀M. ∃pi. Boehm_transform pi ∧ is_ready (apply pi M)
```
meanwhile this `permutator` may solve my blocking issue in proving [Boehm_transform_exists_lemma2] (still in progress). Several other new lemmas about `subterm` are also added in `boehmTheory`.

Chun

### DIFF
--- a/examples/lambda/barendregt/boehmScript.sml
+++ b/examples/lambda/barendregt/boehmScript.sml
@@ -9,7 +9,7 @@ open optionTheory arithmeticTheory pred_setTheory listTheory rich_listTheory
      llistTheory relationTheory ltreeTheory pathTheory posetTheory hurdUtils
      finite_mapTheory;
 
-open binderLib nomsetTheory termTheory appFOLDLTheory chap2Theory chap3Theory
+open binderLib termTheory appFOLDLTheory chap2Theory chap3Theory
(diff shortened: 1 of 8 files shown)


mn200:
Thanks!

PR #1194, opened 2024-02-07
HolSmt: add support for Z3 v4.12.4 proof reconstruction

someplaceguy:
This PR adds support for reconstructing Z3 v4.12.4 proof certificates in HOL4.

It contains 16 commits -- a (more) detailed description is available in each of them.

So far it has only been tested with the HolSmt self-tests. All the self-tests that were passing with Z3 v2.19 proof reconstruction now also pass with Z3 v4.12.4 proof reconstruction.

As a consequence, I'm deprecating HolSmt support for Z3 v2.19 (see motivation in commit b53930656d0fb45188a0e18ba859b9f6a9443458). The HolSmt documentation was changed appropriately. However, for the moment this PR still retains support for parsing and reconstructing Z3 v2.19 proofs, since these are mostly just a subset of Z3 v4.12.4 proofs.

A few other minor, unrelated improvements to HolSmt were also made.

Note that this is only an initial implementation: it is almost guaranteed that HolSmt is still missing support for some Z3 v4.12.4 proof rules, but it is difficult to implement them until I run across examples where these rules appear in the proof certificates. I expect this gap will decrease over time, as I intent to keep fixing more HolSmt self-tests to work with proof reconstruction and add support for more features (to the extent that I can). It's also worth noting that the existing HolSmt support for Z3 v2.19 proof reconstruction is similarly incomplete, in part due to bugs in Z3 v2.19 which cannot be fixed, since this version is not being actively developed anymore and the source code is not available.

I've also attempted to update the Docker CI files to use Z3 v4.12.4 by default, although I'm not sure these changes are correct (help would be appreciated, esp. @binghe as he added this feature :smile:). Related question: @binghe you mentioned in #1179 that for `arm64` only Z3 4.12.4 is available: is this enforced anywhere?

The above change to the Docker CI files should be the only change in this PR that affects anything outside HolSmt.

As a side effect of this PR, it should now be possible to reconstruct Z3 proofs in architectures other than x86 and x86-64 (without using an emulator) :tada: 

cc @tjark


binghe:
> I've also attempted to update the Docker CI files to use Z3 v4.12.4 by default, although I'm not sure these changes are correct (help would be appreciated, esp. @binghe as he added this feature 😄). Related question: @binghe you mentioned in #1179 that for `arm64` only Z3 4.12.4 is available: is this enforced anywhere?

Great work. (I ever spent some time searching inside Z3 4.x documentation to try to figure out how it does proof logging but didn't succeed. Now I think I used wrong keywords.)

For your question, it's not enforced anywhere in HOL4's source code. Instead, I simply meant that, only Z3 4.x provided release downloads for arm64 linux (and it's already installed in the docker CI image). The other two supported SMT solvers (CVC5 and Yices 1.x) seems have no publicly available source code, nor they provides arm64 linux binaries.

mn200:
I'll give @tjark a chance to weigh in before merging, but many thanks for the ongoing work on this!

binghe:
> As a side effect of this PR, it should now be possible to reconstruct Z3 proofs in architectures other than x86 and x86-64 (without using an emulator) 🎉

Totally agree on this. Now on macOS (both x86_64 and arm64), HOL4 users can freely use HolSmt (with Z3 4.x as the backend) in their proofs, and there's no worry about the potential mistakes caused by untrusted oracles.

tjark:
> I'll give @tjark a chance to weigh in before merging, but many thanks for the ongoing work on this!

Please go ahead! I haven't reviewed the code, but from the description of the PR, this sounds amazing.

tjark:
> Note that this is only an initial implementation: it is almost guaranteed that HolSmt is still missing support for some Z3 v4.12.4 proof rules, but it is difficult to implement them until I run across examples where these rules appear in the proof certificates.

A good way to get more examples would be to run Z3 on the benchmarks in (relevant logics of) SMT-LIB. (Depending on how far you want to push this, you could basically re-do the evaluation from the "Fast LCF-Style Proof Reconstruction for Z3" paper with Z3 v4.12.4.)

someplaceguy:
Minor update:
* Fixed some typos and slightly inaccurate terminology.
* Fixed `trans*` handler to perform a union of all the hypotheses of all input theorems, which then becomes the set of assumptions used for proving. Otherwise, if any of the input theorems had any hypotheses, the theorem we are trying to prove would become unprovable. I had already realized that this was necessary (hence the `A1 u A2 u A3` in the example I provided in the comment) but I had forgotten to implement it.
* Added one more argument for deprecating Z3 v2.19.

Also, I should say that I'm not too happy with the implementation of `remove_definitions`. I think it should be very fast in most cases, i.e. if the nesting of the definitions isn't too deep, but otherwise it might cause an exponential term blow-up. The problem is that while proving the theorems necessary to remove those definitions from the set of hypotheses, the tactic I implemented will fully expand all definitions. This can be a problem in cases like this, which seem to be quite tricky to handle:

```ocaml
z1 = x + 1
z2 = y + 1

z3 =   z2    + (y + 1) +   z1
z3 = (y + 1) +    z2   + (x + 1)

z4 =          z3         + (z2 + z2 + z1) +   z1
z4 = (z2 + z2 + (x + 1)) +       z3       + (x + 1)

(...)

z64 =          z63          + (z62 + z62 + z1) + z1
z64 = (z62 + z62 + (x + 1)) +        z63       + (x + 1)
```

In particular, this example would fail with the tactic I implemented since it would require more than 2^64 expansions, due to the deep nesting.

I'm thinking of doing a different implementation, where first I would gather all the definitions of `z64` and would perform term unification between them, to arrive at a minimal expression that defines `z64`:

```ocaml
z64 = z63 + z63 + z1

... assuming the following:

z63 = z62 + z62 + (x + 1)
z63 = z62 + z62 + z1
z1 = x + 1
```
I would then somehow remove all other definitions of `z64` (I guess by proving with substitution, using these assumptions).
Once I'd have a single minimal definition for `z64` it would be trivial to remove it.

Finally, I would perform the exact same thing for `z63`, `z62`, `z61` ... `z1` (in order). It'd be important to do this in order, using a topological sort, to avoid doing a lot of extra work (since e.g. when dealing with `z64` we'd be adding new assumptions for `z63` ... `z1`, so if we process them in that order, they would deduplicate automatically when stored in a set).

I think this new alternative implementation could be a lot slower if Z3 provides us many duplicate definitions that are already fully or almost fully expanded (since we'd be doing a lot of unnecessary term unifications with huge terms), but the advantage is that it'd avoid the exponential blow-up for these worst-case scenarios.

I am also suspecting that these worst-case scenarios might actually show up in practice because I suspect that the entire reason why Z3 is inventing these definitions is indeed to avoid this exponential blow-up in proofs!

someplaceguy:
> A good way to get more examples would be to run Z3 on the benchmarks in (relevant logics of) SMT-LIB. (Depending on how far you want to push this, you could basically re-do the evaluation from the "Fast LCF-Style Proof Reconstruction for Z3" paper with Z3 v4.12.4.)

That's a great idea, thanks! I would love to do this after I try fixing as many self-tests as possible to work with proof reconstruction!

someplaceguy:
> Also, I should say that I'm not too happy with the implementation of `remove_definitions`.

> I'm thinking of doing a different implementation (...)

So, I've been discussing this issue with @tjark, who gave me great suggestions. @tjark's idea is similar to what I was proposing, but @tjark's implementation suggestions lead to a simpler and much faster solution that what I was proposing, I think -- probably as fast as the current implementation in this PR in all cases, while still completely avoiding the exponential blow-up!

I'm hoping I'll be able to do the new implementation within the next few days.

mn200:
I'll leave this open for the "new implementation" so we can catch all the work in one PR.  Thanks!

someplaceguy:
Added two commits:
1. 67a8163b8d2b6172c39f852fec2d1b484127c917 Replaces the old implementation of `remove_definitions` with the new one, which should properly handle the tricky cases that me and @tjark came up with.
2. e72e7e5abf52cd1bd9288fce23c74f685ebaa8f4 Add unit tests for `remove_definitions`. The two tricky unit tests were verified to fail with the old implementation of `remove_definitions` (i.e. if the previous commit is reverted).

That said, @tjark seems to have come up with another tricky case which I'm not sure the new implementation can handle. However, I did not understand the exact specifics. @tjark, can you provide more details about the example that you mentioned for which unification might not save us? I would like to add a unit test for that case as well, if possible.

Thank you!

someplaceguy:
Update: @tjark came up with a new clever scenario which 1) I hadn't anticipated and 2) is leading to circular definitions (due to term unification) even though they weren't circular to begin with. I have added a unit test for this new case (in my local repository) and I've confirmed that the new implementation is failing due to this issue. @tjark has also suggested a straightforward fix for this new issue which I'm planning to implement soon. Thanks!

someplaceguy:
I've fixed this PR to handle a kind of definitions which might appear and which I wasn't originally anticipating. These are definitions of the form `var1 = var2`, where at least one of the variables is defined by Z3.

These definitions were problematic for various reasons, especially the issue that @tjark identified which could cause circular definitions to arise even though they weren't circular to begin with.

To fix all the issues in a principled way,  I've added code to keep track of which variables Z3 has defined in its proof log, so that we can refer to them in the proof replay procedure. I wasn't originally planning to do this as it didn't seem necessary, but in hindsight I believe it makes the code that handles these definitions clearer and more robust, especially when handling definitions of this kind. The commit messages and comments in the code contain more details.

I've also added a couple of unit tests to exercise the issue that @tjark identified with circular definitions.

Thanks!

mn200:
Sounds like the epic story has come to a fittingly successful conclusion!  Thanks for all the work on this one.

PR #1196, opened 2024-02-10
[vim-mode] add support for `Proof` modifiers

hrutvik:
Handle things like `Proof[exclude_simps thms]` as follows:

- `ha` previously selected a goal-statement in between `Theorem`/`Triviality` and `Proof`. Now it also selects the `Proof` keyword and associated `[...]` (including over multiple lines).

- To help `ha` above, modify the `HOLSelect` function so that it opens any folds found at the endpoint of its selection.

- `hG` goal-setting command now invokes `proofManagerLib.new_goalstack` and `BasicProvers.mk_tacmod` as suggested by @mn200. It searches the selected "goal" for the `Proof` keyword to pass the correct bit to `mk_tacmod`. It remains compatible with selections which do not include the `Proof` keyword.

### DIFF
--- a/tools/vim/hol.src
+++ b/tools/vim/hol.src
@@ -111,8 +111,15 @@ fu! HOLGoal()
 endf
 
 fu! HOLUQGoal()
-  silent keepjumps normal pG$a`)
-  silent keepjumps normal gg0iproofManagerLib.g(`
+  silent keepjumps normal pG
+  if search('^Proof','cbW')
+    silent keepjumps normal I``) (BasicProvers.mk_tacmod "
+    silent keepjumps normal VGJ
+    silent keepjumps normal A") I
+  else
+    silent keepjumps normal GA``) (BasicProvers.mk_tacmod "Proof") I
+  endif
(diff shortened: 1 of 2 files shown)


tools/vim/hol.src
```diff
@@ -197,6 +204,7 @@ fu! HOLSelect(l,r)
     return
   endif
   call search(a:r,"ce")
+  normal zv
```
xrchz, line 207: is this change supposed to be here?

hrutvik: I could certainly move it to the `nnoremap` for `<LocalLeader>a` below (line 255), if that's preferable. My thinking was that if any other `HOLSelect` calls end up in the middle of a fold, they should probably do the unfold too.

xrchz: interesting, ok! as-is seems fine then.

xrchz (approved):

mn200:
Discord chat suggests there is one further change required for this PR to get the vim mode able to apply tactics properly in an `Proof`-modified block.  I will merge once that's also in place.

mn200:
Awesome!

PR #1197, opened 2024-02-14
Support cycle detection in topological_sortTheory

Gordon-Sau:
- add `has_cycle` which checks if there are cycles in the graph
- prove that any pair of elements in a partition of the output of the `top_sort` is connected.
- add predicats `TC_depends_on` (a TC version of depends_on) and `TC_depends_on_weak` (which removes the `MEM b (MAP FST graph)` condition from `TC_depends_on`)
- prove the correctness of `has_cycle` (`has_cycle_correct` and `has_cycle_correct2`)

### DIFF
--- a/examples/algorithms/topological_sortScript.sml
+++ b/examples/algorithms/topological_sortScript.sml
@@ -460,6 +460,68 @@ Proof
   simp[trans_clos_TC_closed]
 QED
 
+Definition SCC_def:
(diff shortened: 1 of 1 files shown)


Gordon-Sau:
I

mn200:
Thanks for this.  The `SCC` notion feels like it could be moved to `relationTheory`, but that can be done later.

PR #1201, opened 2024-02-21
Upgraded GitHub actions (Node.js 16 actions are deprecated)

binghe:
Hi,

The following annotations can be seen from Github's "Actions" tab for each recent CI builds:
```
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3,
 docker/setup-buildx-action@v2, docker/build-push-action@v3. For more information see: 
https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

This PR simply upgrades all involved obsoleted GitHub actions to their latest version. Since they are perfectly backward compatible, nothing is broken after the simple version changes. We don't really need the new features provided by the new versions of those actions, but if leaving as is, eventually the old actions won't run.

P. S. I was waiting for the merge of PR #1194, where the same workflow files were modified due to Z3 default version changes.

Chun

### DIFF
--- a/.github/workflows/docker-ci-for-pr.yml
+++ b/.github/workflows/docker-ci-for-pr.yml
@@ -11,13 +11,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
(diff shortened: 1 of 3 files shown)


binghe:
Please squash all commits. I cannot combine the commits by myself, because GitHub doesn't allow file changes under `.github` being pushed from local repository (I had to edit them online, one by one).

mn200:
Thanks!

PR #1202, opened 2024-02-28
add few finite map theorems

rsoeldner:
While working on my formalisation, I encountered these theorems which I believe might be useful.
Not sure about the process, I'm keen to feedback.


mn200:
Looks good; thanks!

PR #1204, opened 2024-03-06
add WF_PULL to relationTheory

Gordon-Sau:
- Add the theorem `WF_PULL` that allow one to pull something that stays unchanged in the relation when proving `WF` of a relation
```
  ∀P f R g R'.
    (∀x. P (f x) ⇒ WF (R (f x))) ∧
    (∀x y. R' x y ⇒ P (f x) ∧ f x = f y ∧ R (f x) (g x) (g y)) ⇒
    WF R'
```
An example would be having `f=FST` and `g=SND`, and we know that the first component of the tuple remains unchanged and always satisfy `P` then we can pull it out as an assumption and only have to prove `WF` of a relation only concerning the `SND` component under that assumption.

### DIFF
--- a/src/relation/relationScript.sml
+++ b/src/relation/relationScript.sml
@@ -1119,6 +1119,35 @@ val WF_antisymmetric = store_thm(
   METIS_TAC [TC_RULES]);
 
 (*---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


mn200:
Thanks for this!

PR #1208, opened 2024-03-11
Polarity search functionality

Eric-C-Hall:
Functionality that allows one to search for theorems such that a certain term specifically appears as a premise or as a conclusion to that theorem, rather than simply testing that the term is present somewhere in the theorem.

### DIFF
--- a/src/parse/Hol_pp.sig
+++ b/src/parse/Hol_pp.sig
@@ -27,6 +27,7 @@ sig
   val print_apropos       : term -> unit
   val print_find          : string -> unit
   val print_match         : string list -> term -> unit
+  val print_polarity_match: bool -> term -> unit
 
   val print_theory_to_file      : string -> string -> unit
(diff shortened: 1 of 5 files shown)


mn200:
How about adjusting the Emacs code so that if the user provides a "prefix argument" (types `C-u` first), the sense of the search is flipped so that negative occurrences are searched for?  See the elisp for `hol-subgoal-tactic` for how to do this: 
- you add a parameter to the argument list `p` say
- you grab the value of the parameter with `(interactive "P")`
- the `p` will be non-nil if there has been a `C-u` typed.

mn200:
BTW: your regression failures are a result of having used Unicode in files committed under `src`.  There's an emacs command to fix this up semi-automatically. The rules allow the lambda and the magic quotes to be used even if everything else is forbidden.

mn200:
Thanks for your work on this!

PR #1211, opened 2024-03-13
Updated HOL Description (more math-related contents)

binghe:
Hi,

I have been trying to update «The HOL System Description» with more math-related contents (calculus, ordinals and cardinal, etc.) but it is too hard to make it perfect. This PR is kind of stage work with many new contents and small fixes here and there.

Now I have promoted math-related contents into a new chapter (_Core Theory--Higher Mathematics_) after the existing chapter "Core Theory". The main reason is to reduce LaTeX section levels: previously I had to use things like "subsubsection" in describing probability theories. A sample build of this new chapter is in attach.
[Pagine da description.pdf](https://github.com/HOL-Theorem-Prover/HOL/files/14587205/Pagine.da.description.pdf)

Chun

P.S. There are two figures generated by MetaPost. The source file is "figure.mp". The generated figure files ("1.mps" and "2.mps") are committed to ease the PDF building process. The makefile target "figures" can be used to update the generated figure files.


mn200:
Thanks for this!

Issue #1074, opened 2022-11-28
Inductive definition should better diagnose syntax problems


mn200:
Using a rule like `p ==> q ==> r` fails with very cryptic message, when it should diagnose that it actually just needs `p /\ q ==> r`.

taimoorzaeem:
@mn200 Could you give an example of this and also what exactly is the cryptic message?

mn200:
```
Inductive fib:
[~zero:]
  fib 0 1 /\
[~one:]
  fib 1 1 /\
[~recursive:]
  !r n s. 2 <= n ==> fib(n-2) r ==> fib (n-1) s ==> fib n (r + s)
End
```

fails.  The correct version would be 

```
Inductive fib:
[~zero:]
  fib 0 1 /\
[~one:]
  fib 1 1 /\
[~recursive:]
  !r n s. 2 <= n /\ fib(n-2) r /\ fib (n-1) s ==> fib n (r + s)
End
```

mn200:
Message is: 
```
Exception raised at IndDefLib.Hol_reln:
at IndDefLib.Hol_mono_reln:
at InductiveDefinition.new_inductive_definition:
at InductiveDefinition.derive_nonschematic_inductive_relations:
at InductiveDefinition.derive_canon_inductive_relations:
at Term.list_mk_binder:
expected list of variables
Exception-
   HOL_ERR
     {message =
      "at IndDefLib.Hol_mono_reln:\nat InductiveDefinition.new_inductive_definition:\nat InductiveDefinition.derive_nonschematic_inductive_relations:\nat InductiveDefinition.derive_canon_inductive_relations:\nat Term.list_mk_binder:\nexpected list of variables",
      origin_function = "Hol_reln", origin_structure = "IndDefLib"} raised
```

mn200:
Closed by #1213

PR #1213, opened 2024-03-15
Better diagnose syntax problems in inductive definitions with double implications

mrichards30:
This PR aims to address issue #1074 by adding an error message when trying to include a double implication in inductive definitions.


mn200:
Thanks for this!

PR #1219, opened 2024-04-09
Move `q_set` from real_sigmaTheory to the new real_of_ratTheory

binghe:
Hi,

The main purpose of this PR is to merge some essentially duplicated proofs about `q_set` - the set of all rational numbers (as a subset of reals) in `real_sigmaTheory` and the new `real_of_ratTheory`.

Backgrounds: some fundamental results about Borel measurable sets (and then Lebesgue integrals) rely on the property of rational numbers (as a subset of reals) that "the set of all rational numbers is countable, and rational numbers are dense" ("dense" means that there's a rational number between any two real numbers).  The existing proofs (mostly done by people from HVG Concordia) didn't make use the `ratTheory` in HOL4. Instead, the following "rational number set" were directly defined by natural numbers:
```
   [q_set_def]  Theorem      
      ⊢ ℚ =
```
Then there are long proofs showing the above set is countable, and having properties such that, if `a` and `b` are in this set, so is `a + b`, etc. These proofs were originally in `util_probTheory`, then moved into `real_sigmaTheory` due to dependency issues.
```
   [countable_real_rat_set]  Theorem      
      ⊢ countable ℚ

   [ADD_IN_QSET]  Theorem      
      ⊢ ∀x y. x ∈ ℚ ∧ y ∈ ℚ ⇒ x + y ∈ ℚ
```

Now, with the new `real_of_ratTheory` (a great work), the above set can also be easily defined like this:
```
   [real_rat_set]  Definition      
      ⊢ ℚ = {r | ∃q. r = real_of_rat q}
```

By showing the above two definitions are equivalent, almost all theorems about the previous `q_set` in `real_sigmaTheory` now becomes trivial, because there's already similar results provided in `real_of_ratTheory`, e.g.
```
  [REAL_OF_RAT_ADD]  Theorem      
      ⊢ real_of_rat r1 + real_of_rat r2 = real_of_rat (r1 + r2)
```

The proofs of [ADD_IN_QSET] and [REAL_OF_RAT_ADD] are essentially the same, having one of them, the other becomes trivial. So is the relationship between [REAL_Q_DENSE] and [REAL_RAT_DENSE]:
```
   [REAL_Q_DENSE]  Theorem      
      ⊢ ∀r1 r2. r1 < r2 ⇒ ∃q. r1 < real_of_rat q ∧ real_of_rat q < r2
   
   [REAL_RAT_DENSE]  Theorem      
      ⊢ ∀x y. x < y ⇒ ∃r. r ∈ ℚ ∧ x < r ∧ r < y
```
Thus all q_set related theorems are now moved out of `real_sigmaTheory` into `real_of_ratTheory`. This is definitely a good direction as now we have same number of theorems but less proofs to maintain.

But once the probability stuff starts to load `real_of_ratTheory`, the number parser by default will put `:int` and `:rat` in front of `:num` and `:real`, causing type mistakes in many quoted terms. Thus I have to put the following two lines in many places to "deprecate" int and rat, which is annoying but seems having no other choice:
```
val _ = intLib.deprecate_int();
val _ = ratLib.deprecate_rat();
```

Furthermore, the code behind `deprecate_int` doesn't handle "flr" (floor) and "clg" (ceiling), which is also overloaded by both `intrealTheory` and `realTheory` (`intLib` is built before `intrealTheory`, thus have no knowledge of these overloads at all). I had to do the following trick in `large_numberTheory`:
```
val _ = bring_to_front_overload "flr" {Name = "NUM_FLOOR",   Thy = "real"};
val _ = bring_to_front_overload "clg" {Name = "NUM_CEILING", Thy = "real"};
```


binghe:
And I forgot to mention that, the first commit in the branch has cherry-picked some theorems in the ongoing `cv_translator` branch, without any modification.  What I actually needed is just the following theorem in `ratTheory` about RATN and RATD:
```
   [RATND_suff_eq]  Theorem      
      ⊢ gcd (Num n) d = 1 ∧ d ≠ 0 ⇒
        RATN (rat_of_int n / &d) = n ∧ RATD (rat_of_int n / &d) = d
```
which shows that `RATN` and `RATD` indeed returns the smallest denominator/numerator is the input rational is already given by these numbers. But the free variables in this theorem are integers. I found the following alternative forms (involving only natural numbers) more convenient to use:
```
   [RATND_of_coprimes]  Theorem      
      ⊢ ∀p q.
          gcd p q = 1 ∧ q ≠ 0 ⇒ RATN (&p / &q) = &p ∧ RATD (&p / &q) = q
   
   [RATND_of_coprimes']  Theorem      
      ⊢ ∀p q.
          gcd p q = 1 ∧ q ≠ 0 ⇒ RATN (-&p / &q) = -&p ∧ RATD (-&p / &q) = q
```

binghe:
Added one more commit. I suddenly realised that there's a even more "primitive" definition of this `q_set`: ⊢ real_rat_set = IMAGE real_of_rat 𝕌(:rat)

mn200:
Thanks for your work on this!

Issue #1226, opened 2024-04-25
Moscow ML builds broken after `cv_translator`


binghe:
```
Holmake: Compiling cv_repTheory.sig
Holmake: Compiling cv_repTheory.sml
Holmake: Compiling cv_miscLib.sml
Holmake: Compiling cv_memLib.sml
File "cv_memLib.sml", line 33, characters 51-52:
!     cv_print Verbose ("Took " ^ Time.fmt 1 (finish - start) ^ " seconds.\n");
!                                                    ^
! Overloaded - cannot be applied to argument(s) of type time
*** FATAL: Build failed in directory /Users/binghe/ML/HOL.mosml/src/parallel_builds/core
```

binghe:
Full Moscow ML builds are too slow to be part of the CI, but perhaps a small build sequence of (selected) core theories could finish in 1-2 hours (so that we can put it into CI)?

PR #1227, opened 2024-04-25
Updated Docker CI workflow

binghe:
Hi,

I updated the Docker CI workflows with the following changes:

- Now HOL manuals will be built after HOL itself (only with PolyML). The related docker images have been updated with some additional TeXlive packages.
- In addition to the existing `stdknl` and `expk` builds using PolyML. Now for each new commit and PR, the CI workflow also contains a basic `expk` build using Moscow ML. The command line option is `--expk --seq=tools/sequences/upto-parallel -t`, which can finish in about 35 minutes.
- I installed new versions of Z3 SMT solver (4.13.0), which is also used by default. (The previous version (4.12.x) is still there.)

P. S. The current building errors with Moscow ML is in cv-related code after `boss`, thus the newly added CI tests cannot expose it. I look forward for an enriched build sequence (no parallelisation) which can extend the related CI tests to 1-1.5 hours with more theories covered.

Chun

### DIFF
--- a/.github/workflows/docker-ci-for-pr.yml
+++ b/.github/workflows/docker-ci-for-pr.yml
@@ -24,7 +24,7 @@ jobs:
           push: false
           build-args: |
             SML=poly
-            Z3_VERSION=4.12.4
+            Z3_VERSION=4.13.0
             CVC_VERSION=5
             BUILDOPTS=--stdknl -j2 -t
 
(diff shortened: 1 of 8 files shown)


binghe:
I've pushed a new commit to fix the build with Moscow ML (#1226).

mn200:
Fantastic; thank you!

PR #1232, opened 2024-05-04
Make `monadsyntax` more aware of print depth

hrutvik:
Limiting print depth has an unusual effect on long monadic definitions, because all monadic assignments on the top-level "spine" of `bind`s are printed at the same depth. This commit attempts to reduce the print depth further down the spine, so that later assignments are elided.

### DIFF
--- a/help/Docfiles/Globals.max_print_length.doc
+++ b/help/Docfiles/Globals.max_print_length.doc
@@ -1,0 +1,22 @@
+\DOC max_print_length
+
+\TYPE {max_print_length : int ref}
+
+\SYNOPSIS
+Sets a length bound on pretty printing for list forms.
+
+\DESCRIBE
+
+Defines the maximum number of elements that the pretty printer will print for
+list forms. This closely mirrors {Globals.max_print_depth}: remaining elements
+are abbreviated by {...}, and the default value of {~1} means `print all
+elements'.
+
+\FAILURE
+Never fails.
+
+\SEEALSO
+Globals.max_print_depth
+
+
+\ENDDOC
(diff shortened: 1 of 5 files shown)


mn200:
If I understand you right, you're complaining about the fact that you might get output like:
```
  do
    v1 <- ... ; 
    v2 <- ... ; 
    v3 <- ... ; 
od
```
where you'd prefer to see, _e.g._:
```
do 
   v1 <- ...; 
   ...
od
```
I'm not so sure about this, and may well have deliberately written the print-depth decrementing code to achieve the first effect. I have certainly found it annoying when list-forms are printed so as to lose the information about how many elements there are in a list. But equally, if the list/do-form is a million elements long, having to see them all when you've gone to the trouble of specifying a depth of 10 is a bit user-hostile.  I also dislike the way that when printing lists you get a “triangular” decay so that later elements, if they are printed, get printed with less detail.  Perhaps the logic could be that you don't get any more than _depth_ elements printed, and that each element is printed with _depth-1_.

binghe:
In the world of Lisp, there's the difference between `*print-length*` and `*print-level*` (depth) [1] when printing lists (s-expressions).

[1] https://www.lispworks.com/documentation/HyperSpec/Body/v_pr_lev.htm#STprint-lengthST

mn200:
@binghe That's interesting!  Perhaps we should adopt something similar.

hrutvik:
Apologies @mn200, yes that example is the kind of thing I'm referring to (specifically the situation with very many elements in the list/do-form). I can see that having both behaviours could be useful - following @binghe's suggestion perhaps I can add a notion of "print length" as a separate reference in `monadsyntax`, which can then be set by the user?

mn200:
Thanks for getting this underway! Can you move the reference to the Globals structure to sit alongside the existing `print_depth` variable?  I can then use it in the system’s printing of other “list forms” (sets, lists, records, let-bindings).

mn200:
Thanks!

PR #1234, opened 2024-05-09
Get Keccak hash function to evaluate fast using cv compute

xrchz:


binghe:
Hi, I wonder how fast is it now, in both std and expk kernels? (it was half a hour in std and 10 hours in expk)

myreen:
On my desktop machine, the entire script now takes just under one minute to build using the stdknl. I suspect the timing will be similar for expk, but haven't tried expk.

myreen:
I've now tested it with expk. With expk, keccakTheory builds in exactly one minute on my now quite old desktop computer.

myreen:
On an Apple M2 Max MacBook Pro it runs in 30 second, i.e., twice as fast!

Note that all of my measurements here are for building the entire theory rather than just doing the evaluation at the end of the theory. The evaluations are quick. On the MacBook, each evaluation takes a second.

PR #1235, opened 2024-05-10
Combined group and ring theories from examples/algebra to core library

binghe:
Hi,

This PR continues #1224 and moved the combined `groupTheory` and `ringTheory`, from `examples/algebra` to `src/algebra/construction`.  There are also small further cleanups to the existing theories under `src/algebra/base`, moving theorems back to their best places (in `bagTheory`, `gcdsetTheory`, etc.).

Besides moving theorems here and there, there's also a new `ringLibTheory` added in `examples/algebra/ring`, mostly to make that directory not empty (at this moment, because otherwise the directory should be deleted, together with `examples/algebra/group`). This is the HOL-Light compatible `ringtheory.ml`, with a new type `:'a Ring`. A related work of porting HOL-Light's ring decision procedure (`RING_RULE` and `RING_TAC`) is still on the way and will arrive at `example/algebra/ring` first.

But this `ringLibTheory` depends on a few more theorems about cardinality of sets (from HOL-Lights' `card.ml`), which is put into `cardinalTheory`.  (NOTE: HOL-Light's syntax of cardinal exponentiation, `s ^_c t`, cannot be precisely duplicated into HOL4, and we actually have a better syntax `s ** t` by overloading the `EXP` operator of `:num`.)

In `relationTheory`, for long time I think the theorem `RTC_TRANSITIVE` (`|- transitive (RTC R)`) is hard to use, and it's better to have something explicit (`|- RTC R x y /\ RTC R y z ==> RTC R x z`), directly.  Fortunately, Joseph the original author of `examples/algebra` also felt the same, so now I moved his `RTC_TRANS` to `relationTheory`.

As mentioned in comments of #1228, I forgot to add the `TRANS_TAC` into `Tactic.sig` and thus it's not yet accessible. This PR also fixed this small issue in a separate commit. But there's another `ANTS_TAC` from HOL-Light, used by some new theorems from HOL-Light. I ported it too with the following documentation:
```
\SYNOPSIS
Split off antecedent of antecedent of goal as a new subgoal.

\DESCRIBE
{
   A ?- (p ==> q) ==> r
  ======================= ANTS_TAC
   A ?- p   A ?- q ==> r
}
```

P. S. I'm not going to further move `fieldTheory` (and the rest of `examples/algebra`) in near future.

Regards,

Chun


mn200:
`ANTS_TAC` already exists in HOL4 under the name `impl_tac`.

binghe:
> `ANTS_TAC` already exists in HOL4 under the name `impl_tac`.

Thanks (I didn't know that). I have removed `ANTS_TAC` from `Tactic` together with the newly added documents, but still I think it's a good idea to have that name in `liteLib` (`val ANTS_TAC = impl_tac`) for compatibility purposes.

mn200:
Thanks!

Issue #1241, opened 2024-05-21
Fix intLib to avoid noisy output


acjf3:
Following commit f6c3ef2, loading intLib produces a lot of output that should be suppressed and/or avoided (see below). Ideally there should not be calls to Meson when loading the library. The overloading/parsing warnings are a bit concerning as well.

```
> load "intLib";
<<HOL message: more than one resolution of overloading was possible>>
<<HOL message: more than one resolution of overloading was possible>>
<<HOL message: more than one resolution of overloading was possible>>
<<HOL message: more than one resolution of overloading was possible>>
<<HOL message: more than one resolution of overloading was possible>>
<<HOL message: more than one resolution of overloading was possible>>
<<HOL message: more than one resolution of overloading was possible>>
<<HOL message: more than one resolution of overloading was possible>>
<<HOL message: more than one resolution of overloading was possible>>
<<HOL message: inventing new type variable names: 'a, 'b>>
Meson search level: ...........
<<HOL message: inventing new type variable names: 'a>>
Meson search level: ..............................................
Meson search level: ........
Meson search level: ........
Meson search level: ...
Meson search level: 
Meson search level: ........
Meson search level: ........
Meson search level: ................................................................................................
Meson search level: ................
<<HOL message: more than one resolution of overloading was possible>>
<<HOL message: inventing new type variable names: 'a>>
Meson search level: ....
```

binghe:
Should be fixed now.

PR #1243, opened 2024-05-22
Fixed intLib to avoid noisy output

binghe:
Hi,

This PR fixes the "noisy output" when loading "intLib" (#1241). Some of these noises are actually caused by loading `hurdUtils`, in which I also did some cleanups.

Although I have learnt from @konrad-slind that the verbose outputs of the Meson prover can be turned off by setting `val _ = mesonLib.chatting := 0` (temporarily, and reset back to 1 at the end of the same library), I found the use of Meson prover in all related cases are waste: either `TAUT` or `DECIDE` can be used instead.

Now loading `intLib` there's no more "noisy" outputs:
```
---------------------------------------------------------------------
     HOL4 [Trindemossen 1 (stdknl, built Wed May 22 13:32:03 2024)]

     For introductory HOL help, type: help "hol";
     To exit type <Control>-D
---------------------------------------------------------------------
[Use-ing configuration file /Users/binghe/.hol-config.sml]
Enter `quit();' to quit.
> load "intLib";
<<HOL message: intLib loaded.  Use intLib.deprecate_int() to turn off integer parsing>>
val it = (): unit
```

--Chun

### DIFF
--- a/src/integer/intReduce.sml
+++ b/src/integer/intReduce.sml
@@ -147,10 +147,10 @@ local
   val NUM_LE_CONV = Arithconv.LE_CONV;
   val INT_LE_NEG2 = INT_LE_NEG;
   val [pth_le1, pth_le2a, pth_le2b, pth_le3] = (CONJUNCTS o prove)
-   (“(-(&m) <= &n <=> T) /\
-     (&m <= &n <=> m <= n) /\
-     (-(&m) <= -(&n) <=> n <= m) /\
-     (&m <= -(&n) <=> (m = 0) /\ (n = 0))”,
+   (“(-(&m) <= (&n :int) <=> T) /\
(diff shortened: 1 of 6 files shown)


src/integer/intReduce.sml
```diff
@@ -319,16 +320,16 @@ end (* local *)
 
 local
   val pth0 = prove
-     (“(&0 * &x = &0) /\
-       (&0 * --(&x) = &0) /\
```
binghe, line 323: This actually fixed a porting bug of `INT_MUL_CONV` on the input of the form `&0 * x`: in HOL-Light the integer subtraction const is `--` instead of the overloaded `-`.

mn200:
Thanks!

PR #1244, opened 2024-05-22
Remove RealArith.OLD_REAL_ARITH_TAC, etc.

binghe:
Hi,

Recently more and more decision procedures are added into HOL4's code base. I consider these "non-proof" code will increase the overall maintenance efforts when bugs are found in them. On the other hand, currently there are many duplicate functionalities here and there, some are ported code from HOL-Light from different time periods. This PR does one such cleanups in `RealArith`.

The important decision procedure for real arithmetics, `REAL_ARITH`, currently has two versions in the code base. The old one was ported from HOL-Light by Joe Hurd in 1998 (available as `RealArith.OLD_REAL_ARITH`, etc.), and the new port was done by me in 2022 (#1043). The new version supports rational coefficients while the old one only supports integral coefficients. But the old version was kept for backward compatibility purposes with the concern that it may solve extra problems that the new one cannot.

There are indeed some use cases such that the new `REAL_ARITH` cannot solve, and the code was patched (on top of HOL-Light's original decision procedure) to handle those extra problems, which has the following shape:
```
?- f (l) = f (r)
```
where `f :real -> 'a` is an uninterpreted function and `l = r` is a solvable real arithmetic equation. Essentially `REAL_ARITH` only does some normalizations to both `l` and `r` and make them literally the same, (and then even `REWRITE_TAC []` can reduce it to `T`).

Obviously these "border" cases are not what `REAL_ARITH` are best for (i.e. solving group of inequalities of multivariate rational-coefficient real polynomials), and it turns out that all those broken cases can be alternatively handle by `rw []` (with a little more helper theorems at most).

What I did in this PR is the following: I completely removed the old port of `REAL_ARITH` (this reduced about 1,000 lines of code in `RealArith.sml`) and also removed all the patches (which slightly slows down the whole process) in the new port, and now HOL4's `REAL_ARITH` will solve exactly the same set of problems as in HOL-Light.   Now the interface from `realLib` only uses the new port.   A small number of existing proofs in the core library and examples are fixed due to the removal of those patches. This small incompatibility is documented in the next release notes.

The CI test (total run time) will show that there's no visible performance degrade after these changes (both the old and new `REAL_ARITH` has some internal caches).

--Chun

### DIFF
--- a/doc/next-release.md
+++ b/doc/next-release.md
@@ -52,6 +52,12 @@ New examples:
 Incompatibilities:
 ------------------
 
+- The old decision procedures for real arithmetics in `RealArith` package has been
+  removed: `OLD_REAL_ARITH`, `OLD_REAL_ARITH_TAC`, `OLD_REAL_ASM_ARITH_TAC` and
+ `PURE_REAL_ARITH_TAC` (user code can switch to `REAL_ASM_ARITH_TAC`).
+  Now `realLib.REAL_ARITH_TAC` (= `RealField.REAL_ARITH_TAC`), etc. solve exactly
+  the same set of problems as in HOL-Light (in rare cases existing proofs break.)
+
 * * * * *
 
 <div class="footer">
--- a/examples/dependability/case_studies/WSNScript.sml
+++ b/examples/dependability/case_studies/WSNScript.sml
@@ -14,23 +14,17 @@
 (*                                                                           *)
 (* ========================================================================= *)
 
-(*loadPath := "/home/waqar/Downloads/RBD" :: !loadPath;*)
+open HolKernel boolLib bossLib Parse;
 
-(*app load ["arithmeticTheory", "realTheory", "prim_recTheory", "seqTheory",
-          "pred_setTheory","res_quanTheory", "res_quanTools", "listTheory", "real_probabilityTheory", "numTheory",
-          "transcTheory", "rich_listTheory", "pairTheory", "extra_pred_setTools",
-          "combinTheory","limTheory","sortingTheory", "realLib", "optionTheory","satTheory",
-          "util_probTheory", "extrealTheory", "real_measureTheory", "real_lebesgueTheory","real_sigmaTheory",
-          "dep_rewrite","RBDTheory","FT_deepTheory","VDCTheory","smart_gridTheory","ASN_gatewayTheory"];*)
-open HolKernel Parse boolLib bossLib limTheory arithmeticTheory realTheory prim_recTheory real_probabilityTheory
+open limTheory arithmeticTheory realTheory prim_recTheory real_probabilityTheory
      seqTheory pred_setTheory res_quanTheory sortingTheory res_quanTools listTheory transcTheory
      rich_listTheory pairTheory combinTheory realLib  optionTheory util_probTheory extrealTheory real_measureTheory
-     real_lebesgueTheory real_sigmaTheory satTheory numTheory dep_rewrite extra_pred_setTools
-      RBDTheory FT_deepTheory VDCTheory smart_gridTheory ASN_gatewayTheory ;
+     real_lebesgueTheory real_sigmaTheory satTheory numTheory dep_rewrite extra_pred_setTools;
+
+open RBDTheory FT_deepTheory VDCTheory smart_gridTheory ASN_gatewayTheory;
 
 fun K_TAC _ = ALL_TAC;
 
-open HolKernel boolLib bossLib Parse;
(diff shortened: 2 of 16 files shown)


binghe:
I did some tests using `Count.apply` to measure the time and number of primitive reasons behind variants of `REAL_ARITH`. It seems the old port is indeed (much) faster (when it works). I've chosen a complicated enough input problem:

```
> val tm = “0 < e ==> N <> 0 ==> 0 < &N ==> realinv (&N) < e ==>
                          -a <= &M ==> &M + &N <= &n ==> &n + a <> 0 ==> &N <= abs (&n + a)”# ;
<<HOL message: more than one resolution of overloading was possible>>
val tm =
   ``0 < e ==>
     N <> 0 ==>
     0 < &N ==>
     realinv (&N) < e ==>
     -a <= &M ==>
     &M + &N <= &n ==>
     &n + a <> 0 ==>
     &N <= abs (&n + a)``: term
> Count.apply RealArith.OLD_REAL_ARITH tm;
runtime: 0.00353s,    gctime: 0.00000s,     systime: 0.00296s.
Axioms: 0, Defs: 0, Disk: 0, Orcl: 0, Prims: 651; Total: 651
val it =
   |- 0 < e ==>
      N <> 0 ==>
      0 < &N ==>
      realinv (&N) < e ==>
      -a <= &M ==>
      &M + &N <= &n ==>
      &n + a <> 0 ==>
      &N <= abs (&n + a): thm
> Count.apply RealArith.REAL_ARITH tm;
positivstellensatz: +9+9+8+9.
positivstellensatz: +9+9+8+10.
runtime: 0.27618s,    gctime: 0.01172s,     systime: 0.14293s.
Axioms: 0, Defs: 0, Disk: 0, Orcl: 0, Prims: 2966; Total: 2966
val it =
   |- 0 < e ==>
      N <> 0 ==>
      0 < &N ==>
      realinv (&N) < e ==>
      -a <= &M ==>
      &M + &N <= &n ==>
      &n + a <> 0 ==>
      &N <= abs (&n + a): thm
> Count.apply RealField.REAL_ARITH tm;
positivstellensatz: +9+9+8+9.
positivstellensatz: +9+9+8+10.
runtime: 0.28541s,    gctime: 0.00967s,     systime: 0.02098s.
Axioms: 0, Defs: 0, Disk: 0, Orcl: 0, Prims: 4409; Total: 4409
val it =
   |- 0 < e ==>
      N <> 0 ==>
      0 < &N ==>
      realinv (&N) < e ==>
      -a <= &M ==>
      &M + &N <= &n ==>
      &n + a <> 0 ==>
      &N <= abs (&n + a): thm
> Count.apply RealField.REAL_FIELD tm;
4 basis elements and 0 critical pairs
Translating certificate to HOL inferences
6 basis elements and 2 critical pairs
7 basis elements and 3 critical pairs
8 basis elements and 4 critical pairs
8 basis elements and 3 critical pairs
8 basis elements and 2 critical pairs
9 basis elements and 6 critical pairs
9 basis elements and 5 critical pairs
10 basis elements and 9 critical pairs
10 basis elements and 8 critical pairs
10 basis elements and 7 critical pairs
10 basis elements and 6 critical pairs
10 basis elements and 5 critical pairs
10 basis elements and 4 critical pairs
10 basis elements and 3 critical pairs
10 basis elements and 2 critical pairs
10 basis elements and 1 critical pairs
10 basis elements and 0 critical pairs
positivstellensatz: -+9+9+8+9.
positivstellensatz: -+11+11.
positivstellensatz: -+12+12.
positivstellensatz: -+10+10+9+10.
positivstellensatz: -+13.
positivstellensatz: -+14.
positivstellensatz: -+13.
positivstellensatz: -+12.
positivstellensatz: -+12.
positivstellensatz: -+13.
positivstellensatz: -+14.
positivstellensatz: -+13.
positivstellensatz: -+10+10+9+11.
positivstellensatz: -+12+12.
positivstellensatz: -+11+11.
positivstellensatz: -+9+9+8+10.
runtime: 1.2s,    gctime: 0.04531s,     systime: 0.03388s.
Axioms: 0, Defs: 0, Disk: 0, Orcl: 0, Prims: 21057; Total: 21057
val it =
   |- 0 < e ==>
      N <> 0 ==>
      0 < &N ==>
      realinv (&N) < e ==>
      -a <= &M ==>
      &M + &N <= &n ==>
      &n + a <> 0 ==>
      &N <= abs (&n + a): thm
```

So, the current strategy in `realLib` seems still optimal: first try the old one and then fallback to the new one.

PR #1249, opened 2024-06-06
Further cleanup of hurdUtils (assert -> simple_assert)

binghe:
Hi,

Currently there's a function `assert` in `hurdUtils` which has name conflict with `Lib.assert` (and this stops user code using `Lib.assert`, to open `hurdUtils` safely.) It has the following signature:
```
val assert : bool -> exn -> unit
```
This `assert` function simply checks if the first argument is true and return `()`, or throw the exception provided by the 2nd argument. The function `assert` (and `assert_exn`) in `Lib` is slightly more advanced, in the sense that they take a predicate and a value and calls the predicate on the value to know the assertion condition:
```
val assert : ('a -> bool) -> 'a -> 'a
val assert_exn : ('a -> bool) -> 'a -> exn -> 'a
```
I think the one in `hurdUtils` could be renamed to `simple_assert`, whose behavior is actually the one found in many other programming languages including C.

All code using `hurdUtils.assert` are under `examples/miller`, where I made minimal fixes by putting `val assert = simple_assert` at the beginning of the related scripts (beside some other minor cleanups).

P.S. With this change, I'm going to `open hurdUtils` in `transcScript.sml` (where `Lib.assert` is used), to upgrade the definition of `rpow`.

Chun

### DIFF
--- a/examples/miller/formalize/extra_pred_setScript.sml
+++ b/examples/miller/formalize/extra_pred_setScript.sml
@@ -11,6 +11,8 @@ open ho_proverTools subtypeTheory subtypeTools boolContext listContext;
 
 val _ = new_theory "extra_pred_set";
 
+val assert = simple_assert;
+
 (* ------------------------------------------------------------------------- *)
 (* Tools.                                                                    *)
 (* ------------------------------------------------------------------------- *)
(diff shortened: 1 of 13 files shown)


mn200:
I'm happy with the other ways in which this PR helps `hurdUtils` get smaller, but please also _move_ `simple_assert` to `Lib`.

binghe:
> I'm happy with the other ways in which this PR helps `hurdUtils` get smaller, but please also _move_ `simple_assert` to `Lib`.

Done.

PR #1251, opened 2024-06-06
Some bug fixes to cv_trans functions

myreen:
This PR also defines HEX and UNHEX so that they are well specified for all inputs.


mn200:
Could you provide regression tests for these please?

mn200:
Thanks!

PR #1253, opened 2024-06-12
itree weak bisimulation up to finite taus

Plisp:
this came up in proofs and I worked it out with @IlmariReissumies

also added two other lemmas which seem to come up fairly often


binghe:
Please re-create your commit branch from the latest `develop`.

Plisp:
oops I forgot HOL is always in develop, let's see if that works

mn200:
Thanks!

Issue #1233, opened 2024-05-06
C++ compilation failure of `src/HolSat/sat_solvers/zc2hs` on macOS


binghe:
I think this issue is new on macOS 14 (Sonoma) with latest Xcode (I'm using version 15.3). Apple's `c++` (`clang++`) cannot compile `src/HolSat/sat_solvers/zc2hs/zc2hs.cpp`:

```
src/HolSat/sat_solvers/zc2hs$ make
ln -fs ../minisat/Proof.o
ln -fs ../minisat/File.o
ln -fs ../minisat/File.h
ln -fs ../minisat/Proof.h
ln -fs ../minisat/Global.h
ln -fs ../minisat/Sort.h
ln -fs ../minisat/SolverTypes.h
In file included from zc2hs.cpp:2:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/iostream:43:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/ios:222:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/__locale:15:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/__memory/shared_ptr.h:30:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/__memory/uninitialized_algorithms.h:613:19: error: use of overloaded operator '!=' is ambiguous (with operand types 'std::reverse_iterator<std::pair<enum_ty, std::vector<int> > *>' and 'std::reverse_iterator<std::pair<enum_ty, std::vector<int> > *>')
  while (__first1 != __last1) {
         ~~~~~~~~ ^  ~~~~~~~
[...]

PR #1254, opened 2024-06-14
Fixed C++ compilation failure of zc2hs on macOS

binghe:
Hi,

Once again, with help of @oskarabrahamsson (#1233, also see #985 for the last time), the failure of C++ compilation of `zc2hs` on macOS with Apple's `clang++` is now fixed. I can only guess that, the following template function is too general to create a conflict with some other (internal) definitions:
```
template <class T> static inline bool operator != (const T& x, const T& y) { return !(x == y); }
```
There are only two places where this overloaded operator `!=` is actually used. By rewriting them to `!(x == y)`, I'm sure the semantics doesn't change, and the compilation error is gone.

Let's see when these C++ code will be broken again:)

--Chun

P.S. In `src/HolSat/sat_solvers/zc2hs/Makefile`, I followed the idea of Minisat's Makefile to use `$(MINISAT_CXX)` (with a default value `c++`) instead of the explicit `g++`.  On potential *unix systems without GCC but with other C/C++ compilers, the command `g++` may not be available while `c++` should exist.  (I have been using the command `Makefile MINISAT_CXX=g++-mp-13` to compile `zc2hs` manually for quite several months.)

### DIFF
--- a/src/HolSat/sat_solvers/minisat/Global.h
+++ b/src/HolSat/sat_solvers/minisat/Global.h
@@ -260,15 +260,17 @@ const lbool l_Undef = toLbool( 0);
 //=================================================================================================
 // Relation operators -- extend definitions from '==' and '<'
 
-
 #ifndef __SGI_STL_INTERNAL_RELOPS   // (be aware of SGI's STL implementation...)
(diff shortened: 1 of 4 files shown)


mn200:
Thanks for this!

PR #1261, opened 2024-06-26
Add NOT_AND' (|- ~(~t /\ t)) to BOOL_ss

binghe:
Hi,

Currently `bool_ss` (`BOOL_ss`) can simplify `P /\ ~P` to `F` but cannot handle the symmetric form `~P /\ P`, e.g.:
```
> SIMP_CONV bool_ss [] ``P /\ ~P``;
val it = |- P /\ ~P <=> F: thm
> SIMP_CONV bool_ss [] ``~P /\ P``;
Exception- UNCHANGED raised
```
This is because `boolTheory.NOT_AND` (`|- ~(t /\ ~t)`) is part of `BOOL_ss` without its symmetric version (which we may call it "NOT_AND'").  I found this issue because I met a proof which could benefit from it.

So I added the missing part into `BOOL_ss` (following the idea of  `EXCLUDED_MIDDLE'` in it).  I would say this is *fixing* the asymmetry of `BOOL_ss`, but anyway it brings a minor incompatibility.  The good news is, among all core theories and examples (part of the CI tests), there are only two proofs to fix, trivially.

--Chun

### DIFF
--- a/examples/lambda/barendregt/standardisationScript.sml
+++ b/examples/lambda/barendregt/standardisationScript.sml
@@ -416,16 +416,12 @@ val residuals_to_right_of_reduction = store_thm(
     PROVE_TAC [lrcc_new_labels'],
     PROVE_TAC [],
     PROVE_TAC [lrcc_new_labels'],
-    PROVE_TAC [],
     FULL_SIMP_TAC (srw_ss()) [n_posns_nlabel] THEN SRW_TAC [][],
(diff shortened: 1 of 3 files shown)


mn200:
Of course, something like `CONJ_ss` should be used to handle this sort of situation more generally, but this is clearly a win for the easy cases.  Thanks!

PR #1265, opened 2024-06-28
Extirpate THENL from the HOL tutorial

IlmariReissumies:
I often send people off to learn HOL by going through the Euclid tutorial as a warmup. This has the unfortunate side effect that they come back with some antiquated habits that you have to disabuse them of. In particular, the tutorial teaches them `>|`, they become enamoured with it, and then you have to step in and break their hearts.

With this PR, we have a unique opportunity to put an end to heartbreak as we know it. It removes all occurrences of `>|`/`THENL`, and instead teaches  `>-`/`THEN1`, which was curiously absent before.


mn200:
At last our long national nightmare comes to an end!  Thanks.

PR #1266, opened 2024-06-29
New example: Dijkstra's shortest path algorithm

myreen:


mn200:
You've spelt Dijkstra as “disjkstra” ...  (fingers are clearly too used to writing "disjunction" etc).

binghe:
Hello @myreen, if this is the algorithm for finding the shortest paths between nodes in a weighted graph, I think I failed to see the "weighted graph" here - what's the type of it?

Also, do you think the algorithm can be described on top of @mn200's work of `genericGraphTheory` (at `examples/generic_finite_graphs/genericGraphScript.sml`)?

mn200:
The graph is a finite map from nodes to lists of node-weight pairs (see `has_path` constant).

konrad-slind:
Similar question: where's the algorithm? I see an inductive defn, but
nothing directly executable. Maybe I missed something?

konrad-slind:
(The above isn't a real complaint: I think it's a nice verification of an
important algorithm, just commenting that the algorithm isn't explicit.)

myreen:
Reply to @konrad-slind:

> Similar question: where's the algorithm? I see an inductive defn, but nothing directly executable. Maybe I missed something?

The algorithm is the inductive relation (now called) `calc`. I modelled the algorithm as an inductive relation because it is non-deterministic. At each iteration, it needs to pick a minimal element `min_elem` from the queue, but there might not be a unique minimal element and that's why I made it an inductive relation. Refinements of this algorithm can then choose how they pick the minimal element. (I intend to prove a refinement down to an executable implementation, but this is not yet done and will be separate from this PR.) 

Reply to @binghe:

> Also, do you think the algorithm can be described on top of @mn200's work of genericGraphTheory (at `examples/generic_finite_graphs/genericGraphScript.sml`)?

Good point! I hope it can. @mn200 explained the idea of that theory to me but I haven't yet had time to look at how to use it. As @mn200 writes, I need a function that, given a node `n`, returns a list of node-weight pairs that describes all edges from `n`.

konrad-slind:
Thanks. I will be interested to see how the refinement goes!

Re: the various representations of graphs. There is also depth-first and
breadth-first search algorithms in src/search that work over graphs
described as 'a -> 'a list thingies. It would be nice to maybe move that
stuff to work over genericGraphTheory, for uniformity's sake.

konrad-slind:
It's not a priority thing for anyone though, so I can get to it after my
current tranche of work on regular languages is in better shape.

mn200:
Thanks!

PR #1267, opened 2024-07-02
Completely revise sptree toSortedAList impl

talsewell:
The original implementation of toSortedAList was overcomplicated, using
auxiliary run-length-encoding functions to carry across the idea of the
algorithm from one that works on a dense structure.
    
After a couple of rounds of adjustment and simplification, the proof is
quite a bit simpler, and the new version does have aux constants but they
should all translate into cv or CakeML AST without any additional fiddling.

This is demonstrated in the new cv translation, which is just a couple of
lines, rather than a huge quantity of definition of a specialised variant.
The CakeML translation should be simpler as well (unfortunately it will
be a breaking change anywhere it has already been translated).

### DIFF
--- a/src/finite_maps/sptreeScript.sml
+++ b/src/finite_maps/sptreeScript.sml
@@ -2152,288 +2152,307 @@ Proof
   \\ fs [EVEN_MULT, EVEN_ADD, ADD_DIV_RWT]
 QED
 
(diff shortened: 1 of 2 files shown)


mn200:
Thanks for your work on this!

PR #1268, opened 2024-07-03
HolSmt: fix mod/division by zero, improve tests

someplaceguy:
Hi! This PR contains some minor improvements for HolSmt as well as a modification to a couple of theorems in `integerTheory`.

It contains the following commits:

1. HolSmt: disable one test case for CVC5

... due to an observed failure when running CVC5 with a timeout. This appears to be a regression in CVC5's performance from v1.1.0 -> v1.1.2, although I haven't confirmed this to be the case.

2. HolSmt: disable Z3 v2 proof replay of exponential tests

Proof replay for Z3 v2 does not work as well as for Z3 v4, so some test cases were failing. In particular, Z3 v2 proof replay for these test cases were erroneously introduced in this commit: 495e1a75b0754cb446be4f541f74a97793d5e01f

3. Strengthen `NUM_INT_EDIV` and `NUM_INT_EMOD` theorems (in `integerTheory`)

This allows SMT solvers to deduce the result of `x DIV 0` and `x MOD 0` operations.

4. HolSmt: fix self-tests that contain `_ MOD 0` or `_ DIV 0`

In the test cases that were fixed by this commit, the goals were expected to be false, but they are in fact true.

5. HolSmt: use new constant for SMT-LIB's real division

The `realax$/` constant is probably not safe to be used directly as the SMT-LIB real division operator because it defines `x / 0r` to be zero, which (as I understand) might not necessarily be true in the SMT-LIB's `Reals` theory.

In the case of Z3, it seems to consider `x / 0r` undefined, which does not present a soundness issue, but my interpretation is that SMT-LIB allows implementations to define the result of this operation -- and if they do, strictly speaking there's no guarantee that it will be defined to be zero.

Consequently, a new `smt_rdiv` constant was defined to properly model SMT-LIB's version of real division in HOL4, which leaves the result of `x / 0r` undefined.

A theorem was also added to the proving context so that SMT solvers can convert `realax$/` operations into `HolSmt$smt_rdiv` operations (and back).

As a result of this, it should now be possible for SMT solvers to prove `x / 0r = 0r`, which is true for all `x` in HOL4.

6. HolSmt: add tests for real division

Real division has been supported by HolSmt for many years, but it's always nice to have some tests for it, corresponding to similar tests that we have for the ints and nums.


someplaceguy:
cc @tjark

mn200:
Thanks for your work on this!

PR #1272, opened 2024-07-14
Fix DB.find slowness when searching with regexes

mrichards30:
After adding basic regex support to DB.find (https://github.com/HOL-Theorem-Prover/HOL/pull/1246), DB.find is unusably slow, which I didn't realise since I didn't have many theories built when testing.

Previously, the DB.find search regexp was compiled into a DFA every time an entry in the database was checked. This compilation step has now been moved so this only happens once per search.

image

Also, added `.` and `*` operators for any character and zero-or-more matches respectively:

image


mn200:
Thanks for this!

PR #1273, opened 2024-07-14
New example for cv translate

myreen:
The new example shows that function values with a finite domain can be used with cv translate. The new example teaches `cv_transLib` about the type `:256 word → num`. Values of this type are represented on the `cv` side as `cv` representations of an `num spt`, where lack of mapping for a given key `k` means that `k` maps to 0. This way the size of the cv representation is manageable as long as most entries map to 0.


mn200:
Thanks for this!

PR #1274, opened 2024-07-17
HolSmt: improve ability to solve goals, misc fixes

someplaceguy:
This PR improves the ability of HolSmt to solve goals. **Note:** it also includes a few (unrelated) changes outside of HolSmt.

It contains the following 10 commits:

1. Fix wrong type in `help/Docfiles/Term.dest_term.doc`

2. Remove `/bin/` path from `ls` and `rm` commands in Holmakefiles

In the NixOS Linux distribution, `/bin/ls` and `/bin/rm` don't exist.

**Note:** the above two commits are not related to HolSmt.

3. HolSmt: add theorems selectively

This commit makes HolSmt add theorems to the proving context selectively.

That is, instead of always adding all theorems that might be needed, HolSmt will now look at all the terms in the goal and only add those theorems that are necessary to solve that particular goal.

As an example, for a goal such as `1i + 2 = 3`, HolSmt used `ASSUME_TAC` to add a bunch of unrelated theorems to the proving context, such as `realaxTheory.real_pow`, `realTheory.abs`, etc. These were not actually needed to solve this particular goal, although they are needed to solve other goals.

After this commit, HolSmt no longer adds all these unnecessary theorems. This allows SMT solvers to solve goals more efficiently, in some cases.

To select which theorems to add, HolSmt now looks at all the constants used in all the terms in the goal, and then, for each constant, it adds the relevant theorems. Furthermore, it iteratively does the same process for all theorems that were just added, since they themselves might contain constants that require other theorems to be added.

Note: this was the main goal of this PR, although further improvements were made. This commit seems to allow SMT solvers to solve more goals than before (because irrelevant theorems are no longer added to the proving context). So far, I haven't seen any regressions due to this commit, so this appears to be a strict improvement.

4. HolSmt: fix parsing of Z3 proof certificates

Apparently, some Z3 proof certificates contain indexed identifiers which are not valid SMT-LIB terms (even in new Z3 versions), such as `-3` and `1/2`.

This commit adds a new parsing function that successfully parses these identifiers (as reals, although they are currently ignored for the purposes of proof reconstruction).

A new `th-lemma arith` identifier called `gomory-cut` (which is also ignored for the purposes of proof reconstruction) was also added so that the parser doesn't error out.

Note: these fixes to the Z3 proof parser were required as a result of the previous commit. Apparently, omitting some unnecessary theorems from the proving context caused Z3 to create proof certificates with some proof terms that I had never seen before, hence why this fix was needed.

5. HolSmt: silence misleading errors in Z3 proof replay

When reconstructing Z3 proofs, we attempt to reconstruct certain proof steps with several methods, one after another, until one of them finally succeeds.

One of these methods was printing a failure / error message under certain conditions, even though the failure was unimportant since a subsequent proof method does succeed to prove the goal without any issues.

The fix was to change `Tactical.prove (term, tactic)` into `Tactical.TAC_PROOF (([], term), tactic)`. The latter doesn't print error messages when `tactic` fails to prove the goal (it just raises an exception, which is what we want).

Note: this was also fallout from the "add theorems selectively" commit, although this was more of a cosmetic and usability issue, rather than a proper bug.

6. HolSmt: add regression test

Adds a regression test for a goal that Z3 wasn't able to prove until the improvements in the "add theorems selectively" commit.

Note: I was quite surprised that HolSmt wasn't able to prove this goal, even though it should have. This is what inspired me to create this PR, once some debugging revealed that some theorems (unrelated to this goal) seemed to be overwhelming Z3 and causing it to be unable to solve the goal.

7. HolSmt: enable tests

Either as a result of recent improvements in HolSmt and/or recent improvements in cvc5 and Z3, we can now enable more HolSmt test cases for these solvers.

8. HolSmt: don't use `--finite-model-find` with cvc5

This option was introduced in the following commit, to avoid regressions in the self-tests after introducing the ability to pass relevant theorems to the SMT solvers: 7f0fff50f078f56baf7787ea58a1a64db0833408

However, currently this option seems to be hurting cvc5's performance more than it is helping. Therefore, this commit removes it.

As a result, there are 2 regressions in the self-tests, but on the bright side, many dozens of disabled self-tests can now be enabled for cvc5.

9. HolSmt: add `cvco_tac[]`, similar to `z3o_tac[]` but for cvc5

Note: I was reluctant to add this tactic, because I didn't want to encourage any potential HolSmt users to use cvc5, since the results used to be quite disappointing when compared to Z3. However, due to (what I assume are) the changes in this PR and perhaps recent improvements in cvc5, it appears to be more competitive now. That said, I think that currently, Z3 still performs better when using HolSmt, generally speaking.

10. HolSmt: update documentation

In particular, a description of the recently-added tactics `cvco_tac`, `z3o_tac` and `z3_tac` was included.

Thanks!

cc @tjark


mn200:
Thanks for all your work on this!

PR #1275, opened 2024-07-18
Make `buildheap --exe` not create a temporary `.ML` file

someplaceguy:
**Note:** This PR follows from the recent discussion in the `#hol` Discord channel about the appearance of some (apparently) mysterious `.ML` files. The creation of these files was traced to the code in `tools-poly/execompile.ML`, which runs as part of the `buildheap` command when using options `-o` and `--exe`.

**Note 2:** It's hard to tell for sure why the original code was creating the temporary `.ML` file in the first place. My guess is that the function to be compiled is only available as a string (since the user provides it as a command-line option) containing the name of the function, while `PolyML.use` expects an entire file. However, I believe the latter is easily worked around by using the Poly/ML compiler directly rather than the `PolyML.use` wrapper, as can be seen in the code of this PR. Hopefully, there is no other reason why the temporary `.ML` file was being used, which presumably might invalidate this PR. This PR contains 1 commit, whose description follows:

In some situations, these temporary `.ML` files would be left laying around the HOL source tree instead of being deleted, even when buildheap's `--dbg` option is not being used.

This could happen for multiple reasons, namely: due to a system crash or abrupt power off, the user or the OS killing the buildheap process, or more commonly, due to Holmake killing buildheap because some other build step that was running in parallel failed.

The result of the above situations is that these `.ML` file(s), which are untracked by git, suddenly appeared in the HOL source tree. More importantly, the user might not notice this happening until much later, at which point he might no longer realize what was the cause of the problem or whether they can be deleted.

This commit removes the creation of the temporary `.ML` files altogether, calling the Poly/ML compiler directly instead to create the compiled object file.

Note that this commit does not remove the similar creation of the temporary `.o` object files, created by `PolyML.export(...)`.

Initially, it was considered that `OS.FileSys.tmpName ()` could be used to create these temporary files (including the object files), but this option was rejected for the following reasons:

1. `tmpName ()` does not allow the calling code to choose the file extension, and indeed, in Poly/ML the resulting files don't have an extension.

2. It does not seem possible to extend or recreate `tmpName ()` securely within SML, in a way that allows choosing an extension (although, it might be possible to do it with the FFI, but the effort does not seem worthwhile).

3. According to the documentation, the `cc` command typically expects object files to have a `.o` extension, to differentiate them from C source files. In practice, both gcc and clang's `cc` command seem to link object files successfully even when they don't have a `.o` extension, but this seems to be undocumented (thus probably not a good idea to rely on this behavior) and it might not work for other compilers.

Fortunately, leftover `.o` files are not as bothersome as leftover `.ML` files, because the former have an entry in `.gitignore`, thus being hidden from most git commands.

### DIFF
--- a/tools-poly/execompile.ML
+++ b/tools-poly/execompile.ML
@@ -1,7 +1,6 @@
 structure BuildHeap_EXE_Compile :
 sig
-  val res : OS.Process.status ref
-  val exe_compile : {fnname: string, output_exe: string, keep_temp_files : bool}
+  val exe_compile : {fnname: string, output_exe: string, keep_temp_files: bool}
                       ->
                     'a
 end =
@@ -9,7 +8,7 @@ struct
 
 fun quietremove f = OS.FileSys.remove f handle SysErr _ => ()
 
-fun withtemp_file {stem, extension, keep_file} (k : string -> unit) =
+fun withtemp_file {stem, extension, keep_file} (k : string -> 'a) : 'a =
   let
     fun gen i =
       let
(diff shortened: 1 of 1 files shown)


someplaceguy:
As a further note, it turns out that `OS.Process.atExit` apparently cannot be used to remove the temporary file(s) in the event that `Holmake` kills `buildheap`, at least according to [the SML Basis documentation](https://smlfamily.github.io/Basis/top-level-chapter.html):

> The functions registered by [OS.Process.atExit](https://smlfamily.github.io/Basis/os-process.html#SIG:OS_PROCESS.atExit:VAL:SPEC) are not evaluated in the event of an abnormal program termination.

> A normal termination is a call to [OS.Process.exit](https://smlfamily.github.io/Basis/os-process.html#SIG:OS_PROCESS.exit:VAL:SPEC), or as defined by the implementation.

mn200:
Very nice!  I don't think there's any particular reason to preserve the `.o` files as they are hard to inspect. When I generated the `.ML` file (not thinking to just call the compiler as in your code), I wanted to be able to check that those source files looked reasonable with the `--dbg` flag.  

That's interesting about `atExit`; I guess people could catch signals and then exit with an error code while giving any `atExit` things a chance to run.  (That'd be great if `Holmake` used catchable signals, but at the moment, it uses `SIGKILL`...)

PR #1286, opened 2024-08-20
Add l3-to-sail refinement proofs for 32bit load and store

IlmariReissumies:


mn200:
Thanks; CakeML regression will have to catch any problems.

PR #1288, opened 2024-08-23
List-based n-dimensional Borel spaces

binghe:
Hi,

N-dimensional Borel spaces (LaTeX: $R^\inf$) is the basis of the theory of stochastic process.  In 2021, I ever submitted #969 for N-dimensional Borel spaces based on `fcpTheory`, which based on my work of general product measure spaces (in `martingaleTheory` of core theories).  `fcpTheory` is very elegant, but it depends on fixed type variables `:'N` where the number `CARD univ(:'N)` is used as the dimension. In probability proofs, however, sometimes the number of random variables comes from either a universal quantified variable or just a number calculated locally, and then the Borel spaces of the same dimension gets involved. In such cases, there's no way to use FCP-based formalizations of Borel spaces. Instead, a list-based version would be more natural: in this case a n-dimensional point of (ext)reals is nothing but a list of (ext)reals of length N, which N is the dimension.

This PR adds the list-based version of n-dimensional Borel space with necessary supporting theorems. The code lies in `examples/probability/stochastic_processScript.sml` where I'm working on a general theory of stochastic processes.

First we have the so-called "rectangle" (now list-based) which is a N-dimensional number set where each dimension is limited by a set:
```
[list_rectangle_def] (stochastic_processTheory)
⊢ ∀h N. rectangle h N = {v | LENGTH v = N ∧ ∀i. i < N ⇒ EL i v ∈ h i}
```
Then, the (general, list-based) sigma-algebra (called `sigma_lists`) is defined on top of the list-based rectangle and the `sigma` generator of functions (see `sigma_algebraTheory.sigma_functions_def`):
```
[sigma_lists_def]
⊢ ∀B N.
    sigma_lists B N = sigma (rectangle (λn. space B) N) (λn. B) EL (count N)
```
Then, the extreal-based N-dimensional Borel sets is defined by the above `sigma_lists` and the extreal-based 1-dimensional `Borel` set:
```
[Borel_lists_def]
⊢ ∀N. Borel_lists N = sigma_lists Borel N
```

Note that, zero-dimension `N = 0` is meaningless, thus many theorems here may require that `0 < N`. Below are some basic properties of `Borel_lists`:
- Alternative definition using the most basic `sigma`-generator:
```
[Borel_lists_alt_sigma]
⊢ ∀N. 0 < N ⇒
      Borel_lists N =
      sigma {v | LENGTH v = N}
        {rectangle h N | h | (∀i. i < N ⇒ h i ∈ subsets Borel)}
```
- Construction (or decomposition) of N+1-dimensional Borel sets by "product sigma-algebra" of 1- and N-dimensional Borel sets:
```
[Borel_lists_decomposition]
⊢ ∀N. 0 < N ⇒ Borel_lists (SUC N) = cons_sigma Borel (Borel_lists N)
```
where `cons_sigma` involves a chain of new list-based definitions:
```
[cons_cross_def]
⊢ ∀A B. cons_cross A B = {a::b | a ∈ A ∧ b ∈ B}
[cons_prod_def]
⊢ ∀a b. cons_prod a b = {cons_cross s t | s ∈ a ∧ t ∈ b}
[cons_sigma_def]
⊢ ∀a b.
    cons_sigma a b =
    sigma (cons_cross (space a) (space b)) (cons_prod (subsets a) (subsets b))
```
- The list access function (`EL`) is Borel-measurable:
```
[IN_MEASURABLE_BOREL_EL]
⊢ ∀N i. i < N ⇒ EL i ∈ Borel_measurable (Borel_lists N)
```
(All other intermediate theorems were used to prove the above major theorems.)

---
Chun

### DIFF
--- a/examples/probability/stochastic_processScript.sml
+++ b/examples/probability/stochastic_processScript.sml
@@ -1,8 +1,7 @@
 (* ========================================================================= *)
-(* The Theory of (General) Stochastic Processes (ongoing)                    *)
+(* stochastic_processScript.sml: Theory of General Stochastic Processes      *)
 (*                                                                           *)
-(* Author: Chun Tian (binghe) <binghe.lisp@gmail.com> (2021)                 *)
-(* Fondazione Bruno Kessler and University of Trento, Italy                  *)
+(* Author: Chun Tian (binghe) <binghe.lisp@gmail.com> (2021 - 2024)          *)
 (* ========================================================================= *)
 
 open HolKernel Parse boolLib bossLib;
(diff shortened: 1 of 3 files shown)


binghe:
I think HOL-Light's FCP-based multivariate calculus also has this problem: If a proof needs to use n-dimensional integration but `n` comes from a local value provided by a term, then FCP-based theorems having type variable `'N` cannot be used at all, because in HOL there's no way to have a function to return a new type whose cardinality is given by a term, i.e. dependent type.

For this reason, if one day HOL4 will have its own multivariate calculus, I suggest we do a list-based formalization instead of fcp-based one.

mn200:
What you might be able to do is to add preconditions along the lines of
```
dimindex(:'N) = n
```
and thereby link the type variable to the local variable.

mn200:
Anyway, thanks for your work on this (the error in `expk` is my faulty `aatmonoid` proof, and I’m willing to guess your code is fine).

PR #1293, opened 2024-08-31
Add lemma prob_div_mul_relf, refactor code in proofs of BAYES_THEOREM, and COND_PROB_ADDITIVE

KaiPhan:


binghe:
@KaiPhan HOL's core theories (proof scripts under `src/`) cannot contain Unicode characters (with a few exceptions like term quotations). Please replace the Unicode characters (`∀`, `≠`, ..) from your changes. In Emacs, this can be done by the menu item _HOL--Misc--Clean up (...)_

mn200:
As _per_ the regression test machinery, you have used "illegal" Unicode characters under `src/`.  See details at

  https://github.com/HOL-Theorem-Prover/HOL/blob/develop/Manual/Developers/developers.md#coding-standardsrequirements

The emacs-mode has a menu entry (bottom of HOL/Misc) that removes common Unicode and replaces it with the correct ASCII equivalent.

KaiPhan:
@binghe @mn200 Thanks for reviewing. I already updated code. Could you please take a look?

binghe:
> @binghe @mn200 Thanks for reviewing. I already updated code. Could you please take a look?

Yes, new tests have passed. Thanks.

src/probability/probabilityScript.sml
```diff
@@ -1,3 +1,4 @@
+(* -*- Mode: Holscript; -*- *)
```
mn200, line 1: Please delete this line; if you're using emacs, you should set up your editor configuration to automatically recognise `*Script.sml` as files that should use `holscript-mode`.

KaiPhan: @mn200 I got this. And already updated code.

mn200:
Thanks!  This enables nice simplifications of proofs.

PR #1297, opened 2024-09-03
Fix bug in L3 spec of RISC-V (JAL and JALR addr alignment)

andreaslindner:
This PR fixes a bug in the L3 spec of RISC-V.

The previous version of the spec tautologically checks that bit0 is 0 after masking it (for JALR). Both JAL and JALR fail to check that bit1 is 0. It seems to simply be a typo.

This relates mainly to this sentence:
"The JAL and JALR instructions will generate a misaligned instruction fetch exception if the target address is not aligned to a four-byte boundary."
page 16 in https://riscv.org/wp-content/uploads/2017/05/riscv-spec-v2.2.pdf

For both instructions the check should include bit0 and bit1, but bit0 is always zero in both cases. So the check only needs to be for bit1.


mn200:
Thanks for this!

PR #1301, opened 2024-09-13
Make Thm.SPEC use lazy_beta_conv instead of beta_conv

### LINKED ISSUE #1300: `Thm.SPEC` appears to hang in certain degenerate conditions
I've observed `Thm.SPEC` taking an unreasonable amount of time to complete in certain conditions experienced during the replay of a Z3 proof certificate which contains terms that, while containing a reasonable number of *unique* sub-terms, do cause a combinatorial explosion in the result of `term_size`.

The issue can be reproduced with the following artificial minimal example:

```SML
val _ = let val v = ``x:bool`` fun f 0 t = t | f x t = f (x-1) (mk_disj(t, t)) in
  Thm.SPEC T (Thm.SPEC (f 256 T) (Thm.ASSUME ``!x y. x \/ y``)) end;
```

(Note: `val _ = <expr>` is used instead of just `<expr>` or `val thm = <expr>` to prevent the REPL from trying to print the resulting theorem, which by itself would also cause an apparent hang due to a similar but orthogonal issue affecting the pretty-printing of such degenerate theorems).

In the above example, it's the outer `Thm.SPEC` which appears to hang, although the remaining expression is required to reproduce the problem.

The underlying slowdown seems to occur due to `Thm.SPEC` calling `beta_conv`, which ends up traversing the entire tree of sub-terms in the conclusion of the theorem. This takes exponentially long when performed on such degenerate theorems, i.e. those which have an exponential `term_size` compared to the number of *unique* sub-terms.

A potential fix for the standard kernel might be to use `lazy_beta_conv` instead of `beta_conv` in the implementation of `Thm.SPEC`. The former implementation of beta conversion, unlike the latter, appears to delay the application of the conversion by creating (what appears to be) a closure in the term representation. 

This potential fix seems to completely solve the observed hang both in the minimal example above and also in the more realistic observed conditions experienced during the replay of the Z3 proof certificate, i.e. after applying the fix, `Thm.SPEC` finishes instantly instead of appearing to hang.

However, I suspect it might not be as easy to fix the experimental kernel because currently it does not seem to support such closures in the underlying term representation...

someplaceguy:
The former, unlike the latter, avoid traversing the entire term, which can take exponentially long in certain degenerate conditions observed while replaying certain Z3 proof certificates.

Fixes #1300

Note: I'm not completely certain that it is safe to perform this change, as I do not have a complete understanding of the HOL kernel design or implementation.

Also, I've only done minimal testing by performing a standard build and observing that it completes successfully. I'd be happy to perform more extensive testing if you could provide pointers as to how to perform such testing.

Thanks!


ordinarymath:
Another possible change would be adding SPECL to the kernel. We already have GENL in the kernel already.

ordinarymath:
Another thing is that this change is just making Thm.SPEC behave like Thm.Specialize

someplaceguy:
> Another thing is that this change is just making Thm.SPEC behave like Thm.Specialize

Good catch! I didn't know about `Thm.Specialize`. The comment for the `Thm.Specialize` Docfile reads: `Specialize behaves identically to SPEC, but is faster.`

So why have both implementations with the same externally-observable behavior and not just the faster version?

ordinarymath:
One downside is that the substitution isn't memoized and you could in end up with bad behavior of code that is constantly/repeatedly doing the substitution.

ordinarymath:
Another downside is that the substitution when applied to the same term results in terms that are not pointer equal. This can result in slow down of code like aconv which ends up traversing the whole term.

mn200:
This and #1653 are similar, but #1653 doesn't touch the kernel and forces the term into a canonical form.  I'd be inclined to go with #1653 as a result, and also ignore #1643.  If the experimental kernel needs speeding up in this situation it could be revised to have explicit substitutions too.

someplaceguy:
> This and #1653 are similar, but #1653 doesn't touch the kernel and forces the term into a canonical form. I'd be inclined to go with #1653 as a result, and also ignore #1643. If the experimental kernel needs speeding up in this situation it could be revised to have explicit substitutions too.

Feel free to go ahead with that course of action.

Unfortunately, I can no longer easily reproduce the real-world conditions which led to the problem described in #1300.
I suspect either the newest version of Z3 is producing a different proof which is no longer hitting #1300, or (more likely) I'm hitting a similar problem in a different part of HOL (I suspect `INST`), before HolSmt even has a chance of hitting #1300 during proof replay.

Either way, if I hit #1300 again I'll try to replace `Thm.SPEC` with `Thm.Specialize`, and/or use the improved `SPECL` once #1653 is merged.

Thanks!

ordinarymath:
I've just looked at #1300 and I'm not sure whether my SPECL variant would help if you're using it to specialize huge terms. You end up with having to traverse the huge terms generated in the final SPEC call

someplaceguy:
> I've just looked at #1300 and I'm not sure whether my SPECL variant would help if you're using it to specialize huge terms. You end up with having to traverse the huge terms generated in the final SPEC call

If I end up needing `Drule.SPECL` for such huge terms, perhaps I can add a variant `Drule.SpecializeL` that uses `Specialize` even for the final call.

someplaceguy:
> Unfortunately, I can no longer easily reproduce the real-world conditions which led to the problem described in #1300.

I found some old work-in-progress code and I was able to reproduce that issue.

As @ordinarymath suspected, I've tested and found out #1653 wouldn't actually fix the issue I ran into, but as he also keenly noticed, HolSmt doesn't really need to use `Drule.SPECL` in most cases, it can simply use `Thm.INST` which is much faster when dealing with such huge terms.

So as it turns out, #1678 ended up fixing the HolSmt issue which led to the problem observed in #1300.

For this specific proof replay which is generating such huge terms, I've fixed a few other issues in HolSmt which were also causing hangs, but the proof replay is still hanging in other places.

But in either case, I think this PR can definitely be closed. As @ordinarymath mentioned, `Thm.Specialize` already does what I was trying to do in this PR.

Thanks!

ordinarymath:
> I found some old work-in-progress code and I was able to reproduce that issue.

Do you have it publicly available anywhere?

someplaceguy:
> > I found some old work-in-progress code and I was able to reproduce that issue.
> 
> Do you have it publicly available anywhere?

I have submitted a cleaned-up version of that code already, as part of #1690, so now that issue is easy to reproduce with an up-to-date HOL checkout.

So if you want to reproduce the (still not completely fixed) issue that originated this PR, here's how to do it:

1. Install a relatively recent version of Z3 (say, >= 4.12.0 perhaps?)
2. Start HOL with `env HOL4_Z3_EXECUTABLE=/path/to/bin/z3 /path/to/bin/hol`, so that HolSmt knows how to find the Z3 binary [1]
3. Run the following HOL commands:

```sml
load "wordsTheory"
load "HolSmtLib"
open wordsTheory
open HolSmtLib
val _ = Feedback.set_trace "HolSmtLib" 3
max_print_depth := 10
```

Then start proving this goal: `x:word8 < 0w /\ y >= 0w ==> (word_rem x y = -word_mod (-x) y)`
Then use this tactic: `z3_tac[]`

This tactic will cause HolSmt to call into Z3 to prove the goal, then Z3 will find the proof, then HolSmt will attempt to replay the Z3 proof within HOL. With the standard kernel, this will currently hang in the simplifier (IIRC), after many steps have been proved. With the experimental kernel it will likely hang much sooner due to its current inability to represent such huge terms as efficiently as the standard kernel. 

[1] Actually, for development purposes, instead of pointing HOL to the Z3 binary directly, I recommend pointing it to a Z3 start-up script such as the following, which sets a timeout of 30 seconds so that Z3 doesn't keep running indefinitely when it can't find a proof:
```bash
#!/usr/bin/env bash

set -euo pipefail

/path/to/bin/z3 -T:30 "$@"
```

PR #1302, opened 2024-09-19
Seq vs List Correspondence

adamstucci:
- Current imperative example uses a tree structure for sequential composition
- This PR adds a similar example language that instead uses lists for sequences of statements
- Proofs to show that any (terminating) program in one language can be written as a program in the other

- Please critique
- Proofs are verbose and could be simplified


adamstucci:
Closing to break up into separate PRs

PR #1303, opened 2024-09-19
Fixed some obvious issues in Emacs related files

binghe:
1. The value of `max-lisp-eval-depth` is not enough for compiling `hol-input.el` due to increased grammar structures.
2. There's a grammar mistake in `forward-smlsymbol` of `holscript-mode.el`, found when trying to compile the el file.


mn200:
Thanks!  I guess that function `forward-smlsymbol` isn't used anywhere...

PR #1304, opened 2024-09-21
Fixes to opentheory build

digama0:
These are referring to files that don't exist, so I can only assume they are not being tested. Magnus tells me that this is being used, but OT has scalability issues so possibly the build isn't getting this far. I observed this issue by replacing `opentheory` by a stub to ensure it can run to completion.


binghe:
The `EVAL_` theories will be moved to `examples/` (and perhaps then renamed back) so I didn't spend time to check and fix these OT related settings.

mn200:
Thanks!

PR #1305, opened 2024-09-21
Remove pauses in configure script

digama0:
When determining the properties of dependencies, it would pause for 1 second after each one, making it look like more work than it actually is and also slowing down the tool. I left the 3 second countdown after all settings are confirmed, although 3 seconds is a bit short for a human to respond to the information.


mn200:
Fair enough!

PR #1306, opened 2024-09-21
Report error locations on unhandled exceptions

digama0:
PolyML keeps track of the location of every exception (the position of the `raise` keyword), and this information is very useful for debugging, so let's show that. In VSCode it also enables the ability to click in the terminal to go directly to the error.

A few other locations had to be adjusted to use `reraise` to avoid clobbering the location information.


mn200:
Awesome!  I didn't know PolyML's exception module was so rich.

PR #1307, opened 2024-09-21
Remove mutability from theorem proof field

digama0:
The original code was using a `proof ref` in every theorem just for deletion, but every use was already in a position to pass a modified theorem back, so doing that saves an indirection and additional state.


mn200:
I assume the static error in `build-otknl` will be an easy fix?

PR #1309, opened 2024-09-21
Add tracing kernel

digama0:
* [ ] Depends on #1308 

Adds a new kernel type, `--trknl`. This is mostly the same as the standard kernel, but it uses the OT kernel for the `Thm` type. There is also one additional step during the `export_theory` command, in which the export data is pickled directly into a byte blob (preserving sharing) and written out. The result is postprocessed with `gzip` to avoid the traces from getting too large.

This depends on my fork of PolyML (https://github.com/digama0/polyml/compare/exportSmall) for the `exportSmall` function, which was added for a very similar use case in Isabelle.


mn200:
Could you also add a description of what this is for and how to build it to the "Developers" manual in `Manuals/Developers` please? See the `## Kernel Selection` subsection of the `developers.md` file there.

mn200:
Your `tracing/{yes, no}` directories should include `Tracing.sig` and `Tracing.sml` files (separating out the `structure` and `signature` declarations), and don't need `Holmakefile`s (what you've declared there is the default behaviour).

Note how the Moscow ML build fails.  You can get Moscow ML to work with complicated mixtures of functors, structures and signatures all in the one file, but it’s unnecessary work in cases like this.

digama0:
Is there a way to make `tracing/yes/Tracing.sig` and `tracing/no/Tracing.sig` be the same file? They have the same signature after all.

digama0:
Hm, I'm not really sure how to do the ifdef in this situation. I would like to have a function which is enabled only for polyml+trknl, and which is not assumed to compile on any other combination. It is unclear to me why references to the `PolyML` module from other files (such as `tracing/yes/Tracing.sml`) does not work.

mn200:
There's an example of this sort of dance in the way `src/portable/Arbint` is able to use PolyML's built-in support for arbitrary ints but the Moscow ML code uses its own implementation.  You could also just dispense with the signature entirely in both cases and just make sure the structure doesn't open anything into its exported namespace: 
```
  structure Tracing = struct
      local open stuff in 
      fun trace ... = ...
      end
  end
```
Moscow ML is happy with this "style".

mn200:
Is there an easy fix for the regression here?  Would be nice to resolve the PR if possible.

digama0:
This project has gone to the backburner for me but I think I might be better equipped to understand the issues and get it to build now.

xrchz:
Shall we maybe go with the no-sig-file approach to get the build to pass?

src/portableML/rawtheory/RawTheory_dtype.sml
mn200, line 1: Why has the `thm_src_location` stuff moved?

xrchz: That's a question for @digama0

digama0: `thminfo` type is used in the type signature in the `Tracing/{no,yes}/Tracing.sig` file, and the build order puts `Tracing/{no,yes}` after `src/thm` and before `src/postkernel`, which means that  `thminfo` has to be moved earlier. We just need the type, so the easiest place to put it was `RawTheory_dtype`

mn200: Got it; thanks.

mn200:
Thanks for the work on this!

PR #1310, opened 2024-09-21
Add proofs about clock for fun-op-sem/imp

adamstucci:


mn200:
Can you keep your lines under 80 columns please?

mn200:
Thanks!

PR #1315, opened 2024-09-30
zero-repl: ensure all commands in block are processed

digama0:
Added another test for this issue. It is possible for the lexer to consume the end of the input stream even though it has not yet yielded all of its tokens to the compiler, so we have to make sure to watch the output and not the input before moving on to the next block, to avoid losing commands.


digama0:
build failure appears to be spurious

mn200:
> build failure appears to be spurious

Agreed.

PR #1327, opened 2024-10-25
Allow passing .sig files to `hol`

digama0:
You can't parse some `sml` files on the command line without passing the `sig` file first, but `hol` (or rather `buildheap`) doesn't let you do that.


mn200:
Thanks!

PR #1328, opened 2024-10-25
mllex: write accepting branches in source order

digama0:
It was previously writing them in random order (actually, ASCII sorted on state number). Writing them in source order makes it easier to diff the results.

PS: Can we agree that mixed tab/space indentation (using a combination that only makes sense when tabs are 8 spaces even though the code is actually 2 space indent) should die in a fire? I would really like to do a search and replace PR for this.

(Why are you diffing generated sml files, I hear you ask? Well I happen to have come across a relatively decent SML editor plugin, but it only works on SML files and not mllex files. So it is easier to do my modifications on the generated file and then pull stuff back to the original once I'm done.)

### DIFF
--- a/tools/mllex/mllex.sml
+++ b/tools/mllex/mllex.sml
@@ -808,8 +808,8 @@ end;
 
 exception ParseError;
 
-fun parse() : (string * (int list * exp) list * ((string,string) dictionary)) =
-	let val Accept = ref (create String.<=) : (string,string) dictionary ref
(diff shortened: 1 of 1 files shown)


xrchz (approved):

mn200:
TABs are indeed evil.  The `unicode-grep` tool forbids them under `src` and `examples`.  Getting rid of them in the `mllex` and `mlyacc` sub-directories would totally have my blessing.

mn200:
Thanks for this!

PR #1336, opened 2024-11-11
quiet load

digama0:
This implements a flag `quiet_load` which makes `load` not print to stdErr (but it still throws errors). The `qload` function is shorthand for running `load` with `quiet_load` set to `true`. This is intended for IDEs that want to preload the `open`s of a file without getting errors for nonexistent or already open modules.


mn200:
Please undo reversion of theory changes.

digama0:
Master is still broken. I will rebase on top of #1338 instead?

mn200:
Thanks for this cleanup!

PR #1337, opened 2024-11-11
fix master

digama0:


mn200:
I hope this is unnecessary given work in #1338

digama0:
I don't see how the first commit can be avoided, the current code seems plainly not well typed?

mn200:
For sake of future GitHub archaeologists: I fixed compile errors in a05c02a6e375865c85f2b489bdaf68d6ece4e942; theory problems in OpenTheory build have been fixed in pr #1338.

PR #1338, opened 2024-11-13
Fix otknl build

binghe:
Hi,

This should fix the otknl build. There's only one small shift of goals in the generated "hol4-assums.art" causing failure when building `src/boss/prove_base_assumsTheory`. The previous `th62` is now `th60` (`th62 |-> th60; th60 |-> th61; th61 |-> th62`), very local changes, perhaps due to modifications in `listTheory`.

Let's see how the CI goes.

--Chun

### DIFF
--- a/src/boss/prove_base_assumsScript.sml
+++ b/src/boss/prove_base_assumsScript.sml
@@ -893,6 +893,19 @@ val th59 = store_thm
   conj_tac >- MATCH_ACCEPT_TAC filter_nil
   \\ MATCH_ACCEPT_TAC filter_cons);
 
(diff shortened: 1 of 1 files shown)


binghe:
The way to fix such issues is to evaluate expressions like `el 60 goals |> concl` at the broken theorem, and check if the output matches the theorem statements in near comments (that I added just for this purpose), and if not the same, the current proof is not for the current goal (but another goal in different place). Only in rare cases we need to add whole new proofs to handle more goals (currently there are 96 of them).

mn200:
Thanks for getting this done!

PR #1341, opened 2024-11-14
back up polyml Thread structure

digama0:
This makes it possible to undo this (I needed this to be able to compile bits of the PolyML compiler on top of the HOL basis)


mn200:
Thanks!

PR #1342, opened 2024-11-15
HOL_ERR source locations

digama0:
This was mentioned as a TODO in `Feedback.sml`, and I need it for some error reporting work I'm doing. Adds a `source_location` field to `error_record`, and migrates most uses of the `HOL_ERR` constructor to either use incomplete pattern matching or `mk_HOL_ERR` instead.


mn200:
Looks good!  Can merge when the CI is also good.

binghe:
> Looks good! Can merge when the CI is also good.

This time only OT build was successful:)

mn200:
Awesome: thanks!

PR #1343, opened 2024-11-15
improve binder preterm positions

digama0:
It was previously associating binders with the body only, which is weird and imprecise, and causes some [funny issues](https://discord.com/channels/1171421764379742258/1201829002776420362/1307065300356825088) when trying to locate the outermost term with a given span.


mn200:
Neat: I don't think the location based code here has been touched meaningfully since 2002.

PR #1344, opened 2024-11-15
print type without colon

digama0:
Like `pp_type` but without the initial colon. I wanted to add a space between the colon and the type.


mn200:
Nice idea.

PR #1345, opened 2024-11-15
don't use Loc_None in typed_parse_in_context

digama0:
This might be debatable; the issue is that using `Loc_None` on the outermost node causes my code to assume that the whole syntax is missing position info and skip it. Ideally one should be able to navigate the tree using spans in strict containment relationship to zoom in on a subterm, but having intermediate nodes with `Loc_None` spoils this. In this particular case there is a natural location to use anyway.


mn200:
Thanks; definitely a bug I think.

PR #1346, opened 2024-11-15
typecheck listener

digama0:
This adds a hook (which is normally set to do nothing) which is called whenever a preterm is elaborated. This is used in the VSCode editor mode so that it can record the locations of subterms together with typing information.


mn200:
Cool!

PR #1347, opened 2024-11-16
don't gitignore Theory.sml

digama0:
`Theory.sml` is a regular SML file which should not be ignored by the `*Theory.sml` grep (which is intended to only find files generated from `*Script.sml`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 *Theory.sig
 *Theory.sml
 *Theory.dat
+!Theory.sig
+!Theory.sml
 *.uo
 *.ui
 *.o


mn200:
I don't mind the commit, but as long as `Theory.{sig,sml}` are "known" files this isn't an issue, right?

binghe:
> I don't mind the commit, but as long as `Theory.{sig,sml}` are "known" files this isn't an issue, right?

Now these generated files are moved into ".holobjs", which itself is ignored. So (I guess) removing the ignores of `Theory.{sig,sml}` shouldn't cause anything visible.

mn200:
> > I don't mind the commit, but as long as `Theory.{sig,sml}` are "known" files this isn't an issue, right?
> 
> Now these generated files are moved into ".holobjs", which itself is ignored. So (I guess) removing the ignores of `Theory.{sig,sml}` shouldn't cause anything visible.

`*Theory.sig` files actually stay in the source directory (they have pretty-printed statements of the theorems by default).

digama0:
I noticed this issue because vscode's find in files and ripgrep ignore `.gitignore`d files, so my searches were silently missing some important hits.

PR #1349, opened 2024-11-19
Updated Dockerfile for CI tests

binghe:
Hi,

This PR contains latest updates to the docker images that HOL office is already using (because the images are under my Docker Hub account). Now the Linux-based images are based on Debian 13 "trixie" (currently testing, to be released in early next year), with 4 architectures (`linux/386,linux/amd64,linux/arm64,linux/riscv64`) under the same tags: (`latest` = `base` + SMT solvers)

```
docker pull binghelisp/hol-dev:latest
docker pull binghelisp/hol-dev:base
```

(NOTE: Currently HOL4 cannot be built normally on Linux riscv64 by Poly/ML in interpreter mode. The blocking issue is actually at `mllex` and `mlyacc` source code. I will find time to look into these issues.)

On Debian 13, Moscow ML cannot be compiled by the default GCC 14. Instead, GCC 13 must be used (I've made a small patch to Moscow ML's `Makefile` to enforce using GCC 13).

I also found that there are system packages of Z3 and CVC5 SMT solvers, and I've installed them and make the CI tests always using them:
```
#13 [ 4/10] RUN if [ "" != "4.13.0" ]; then     echo "Using Z3 solver: `/usr/bin/z3 -version`"; fi
#13 0.061 Using Z3 solver: Z3 version 4.13.3 - 64 bit
#13 DONE 0.1s
#14 [ 5/10] RUN if [ "" != "5" ]; then     echo "Using CVC solver: `/usr/bin/cvc5 --version | head -1`"; fi
#14 0.065 Using CVC solver: This is cvc5 version 1.1.2
#14 DONE 0.1s
```

I think in the future we only need to use these system-installed SMT solvers as part of the CI tests and no need to install more versions (this will shrink the image size).

Chun

### DIFF
--- a/developers/Dockerfile
+++ b/developers/Dockerfile
@@ -21,6 +21,6 @@ COPY . .
 
 RUN poly < tools/smart-configure.sml && ./bin/build
 
-ENV PATH /HOL/bin:$PATH
+ENV PATH=/HOL/bin:$PATH
 
 ENTRYPOINT ["/HOL/bin/hol"]
--- a/developers/docker-ci/Dockerfile
+++ b/developers/docker-ci/Dockerfile
@@ -5,7 +5,7 @@
 # NOTE: this docker image is NOT for HOL developers to use interactively.
 
 # This "base" image is described in "base/Dockerfile"
-FROM --platform=$TARGETPLATFORM binghelisp/hol-dev:latest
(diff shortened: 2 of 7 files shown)


mn200:
Thanks for keeping this all up-to-date and working!

PR #1355, opened 2024-11-28
fix #1353

### LINKED ISSUE #1353: Holmake hangs
At current `master` (i.e. 01016dcf0e02558d270fce008fce672df8cdbbf7)

Running `Holmake` in a directory containing only the file `bugScript.sml` with the following contents
```
open HolKernel boolLib bossLib

val _ = new_theory"bug";

Definition bug_def:
  bug = nothing to see here
Termination
  cheat
QED

val _ = export_theory();
```

seems to go into an infinite loop.

digama0:


mn200:
I believe the failure here is my test-case being unreasonable.  Thanks!

PR #1358, opened 2024-11-30
Define SHA-256 (part of the SHA-2 standard) and translate it for cv compute

xrchz:
As with Keccak (SHA-3), this is one of the cryptographic algorithms used in the Ethereum virtual machine.

### DIFF
--- a/examples/Crypto/SHA-2/Holmakefile
+++ b/examples/Crypto/SHA-2/Holmakefile
@@ -1,0 +1,1 @@
+INCLUDES = $(HOLDIR)/src/num/theories/cv_compute/automation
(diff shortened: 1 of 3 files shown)


mn200 (changes requested):
Please put this directory into the build sequence (by adjusting `src/parallel_builds/core/Holmakefile`)

xrchz:
> Please put this directory into the build sequence (by adjusting `src/parallel_builds/core/Holmakefile`)

This is done

mn200:
It's not building because your Holmakefile needs an `INCLUDES` line pointing at the directory with `cv_transLib` etc.  (The reason in turn is that `build` is doing everything in parallel so needs accurate directory dependency info within post-boss directories.)

xrchz:
Thanks I've added that Holmakefile now.

mn200 (approved):

PR #1360, opened 2024-12-02
[real] sup/inf lemmas using set notations

binghe:
Hi,

It's a little strange that, in `realTheory`, the `sup` and `inf` of set of real numbers, definitions and lemmas are not using set notations at all. Instead of, `P x` is used instead of `x IN P` and `?x. P x` is used instead of `P <> {}`, etc. I found these lemmas hard to use, especially when the involved sets are defined in gspec notations (e.g. `{x | P x}`, the simplifier can rewrite `x IN {x | P x}` but cannot deal with the equivalent `{x | P x} x`.

Therefore now I added a variant for each of these lemmas using set notations (the "primed" version is new):

```
   [REAL_SUP_UBOUND]  Theorem (existing)
      ⊢ ∀P. (∃x. P x) ∧ (∃z. ∀x. P x ⇒ x < z) ⇒ ∀y. P y ⇒ y ≤ sup P
   
   [REAL_SUP_UBOUND']  Theorem (new)
      ⊢ ∀P. P ≠ ∅ ∧ (∃z. ∀x. x ∈ P ⇒ x < z) ⇒ ∀y. y ∈ P ⇒ y ≤ sup P
   
   [REAL_SUP_UBOUND_LE]  Theorem (existing)
      ⊢ ∀P. (∃x. P x) ∧ (∃z. ∀x. P x ⇒ x ≤ z) ⇒ ∀y. P y ⇒ y ≤ sup P
   
   [REAL_SUP_UBOUND_LE']  Theorem (new)
      ⊢ ∀P. P ≠ ∅ ∧ (∃z. ∀x. x ∈ P ⇒ x ≤ z) ⇒ ∀y. y ∈ P ⇒ y ≤ sup P
```

Perhaps the original intent was to prevent explicitly using `pred_setTheory` but this is no more any issue. Also, the versions of set notation is actually more foundamental. If we were having these lemmas with set notations, the original versions can be obtained by a simple rewrite with `IN_APP` and `GSYM MEMBER_NOT_EMPTY`. The other side of rewriting is impossible (or hard), because all lambda-applications `x y` can be interpreted as `y IN x`.

P.S. In `extrealTheory` (or `extreal_baseTheory`), the original sup/inf lemmas are also in similar shapes, and I have already added a lot of such variants using set notations (and then use them in new probability proofs).

Chun

### DIFF
--- a/src/real/realScript.sml
+++ b/src/real/realScript.sml
@@ -1931,6 +1931,13 @@ val REAL_SUP_EXISTS = store_thm("REAL_SUP_EXISTS",
   EXISTS_TAC “s + d” THEN
   MATCH_MP_TAC SUP_LEMMA1 THEN POP_ASSUM ACCEPT_TAC);
 
+Theorem REAL_SUP_EXISTS' :
+    !P. P <> {} /\ (?z. !x. x IN P ==> x < z) ==>
+        (?s. !y. (?x. x IN P /\ y < x) <=> y < s)
+Proof
+    REWRITE_TAC [IN_APP, REAL_SUP_EXISTS, GSYM MEMBER_NOT_EMPTY]
+QED
+
 val sup = new_definition("sup",
    “sup P = @s. !y. (?x. P x /\ y < x) <=> y < s”);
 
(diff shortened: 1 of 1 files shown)


mn200:
Thanks for this!

PR #1362, opened 2024-12-07
more typecheck listeners

digama0:
Continuation of #1346. Adds typecheck listeners for

* pretype elaboration, i.e. `“:type”` and `Type foo = ...`
* `Definition` elaboration
* `Datatype` elaboration
  * This also has a parse listener because simple datatypes skip the typechecker

### DIFF
--- a/src/datatype/Datatype.sig
+++ b/src/datatype/Datatype.sig
@@ -34,6 +34,7 @@ sig
  (* into their tyinfo to do inequality resolution.                            *)
  (*---------------------------------------------------------------------------*)
 
+ val typecheck_listener : (string Symtab.table -> ParseDatatype.pretype -> hol_type -> unit) ref
  val build_tyinfos : typeBase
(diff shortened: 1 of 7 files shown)


mn200:
I think the codebase should make it easy for multiple listeners to coincide in the one "slot".   Could perhaps have
```
signature listener = 
sig
 
   type 'a t
   val add_listener : 'a t -> (string * ('a -> unit)) -> unit
   val remove_listener : 'a t -> string -> unit
   val current_listeners : 'a t -> (string * ('a -> unit)) list
end
```
which would be implemented with a ref to a list of listener functions (of type `'a -> unit`)
You might then use with 

```
signature foo = sig
  ...
  val event_listener : eventtype listener.t

end
```

We avoid exposing the implementation as `ref`s, and stop people from foolishly dumping over existing entries.

mn200:
Happy to merge as is for the moment anyway.

PR #1368, opened 2024-12-10
Faster evaluation for Keccak_256_bytes

xrchz:
Define a version that operates on word64s and cv translate that instead of the one based on bool lists (at least for Keccak_256_bytes -- the other Keccak functions are still using the old spts from bool lists).


mn200 (approved):

mn200:
Sorry, I messed up your merge; can you fix that and go ahead?

PR #1376, opened 2024-12-16
Update documentation in HOL4 examples

jhlchan:
The `examples/algebra` library has changed, so the README file is amended.
A paper based on `examples/fermat/twosq` library is just published, so a reference is added to the README file.


mn200:
You have included automatically generated (but now stale) files under `tools/vim` and the file `tools/hol-mode.el`.

jhlchan:
I didn't realize that, thanks for pointing them out.

mn200:
You shouldn't git-ignore `vim/` ; that causes real files in `tools/editor-modes/vim` to get ignored.

mn200:
I'm afraid your use of a slash in the middle of an ignore pattern (see output of `git help ignore`) will not cause old files in `tools/vim` to be ignored.  Given that no-one will ever see these files if they make a fresh checkout, I think it might be cleanest to just not edit `.gitignore` at all.  The effect you're after would be achieved by using the lines
```
**/vim/file1
**/vim/file2
etc
```

If you did this, you could then replace the existing .gitignore lines that capture the ignores that are actually present.

jhlchan:
I shall leave the modification of .gitignore at your discretion.

mn200:

PR #1379, opened 2024-12-23
Avoid excessive computation for cv_eval word_lsr

xrchz:
When the shift amount is large we know the result will be 0.

We can do even better for lsl using modexp (see [verifereum](https://github.com/verifereum/verifereum/commit/155bda521dde50155591b44c79dc952ee1a72a7d)), so not touching that here now.

### DIFF
--- a/src/num/theories/cv_compute/automation/cv_primScript.sml
+++ b/src/num/theories/cv_compute/automation/cv_primScript.sml
@@ -904,8 +904,9 @@ QED
 Theorem cv_rep_word_lsr[cv_rep]:
   from_word (word_lsr (w:'a word) n)
   =
-  cv_if (cv_lt (Num n) (Num (dimindex (:'a))))
-    (cv_div (from_word w) (cv_exp (Num 2) (Num n)))
+  let k = Num n in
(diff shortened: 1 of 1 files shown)


src/num/theories/cv_compute/automation/cv_primScript.sml
```diff
@@ -904,10 +904,15 @@ QED
 Theorem cv_rep_word_lsr[cv_rep]:
   from_word (word_lsr (w:'a word) n)
   =
-  cv_div (from_word w) (cv_exp (Num 2) (Num n))
+  cv_if (cv_lt (Num n) (Num (dimindex (:'a))))
```
myreen, line 907: The expression corresponding to Num n is duplicated on the right-hand side of this equation meaning it'll be computed twice. You'll need to let bind Num n on the right-hand side to avoid this duplication, ie 

... = let k = Num n in ... k ...

myreen (changes requested):

src/num/theories/cv_compute/automation/cv_primScript.sml
xrchz: Done

myreen (approved):

PR #1378, opened 2024-12-22
Improved subtree_equiv_lemma without subtree_equiv'

binghe:
Hi,

This PR improves the previous #1367. With an improved construction of Böhm transform (beyond textbook), I have removed the needs of `subtree_equiv'` and proved a better version of `subtree_equiv_lemma` (NOTE: In previous PR the theorem name was wrongly written in the code as `subterm_equiv_lemma`, this typo is fixed here.):
```
[subtree_equiv_lemma]
⊢ ∀X Ms p r.
    FINITE X ∧ p ≠ [] ∧ 0 < r ∧ Ms ≠ [] ∧
    BIGUNION (IMAGE FV (set Ms)) ⊆ X ∪ RANK r ∧
    EVERY (λM. subterm X M p r ≠ NONE) Ms ⇒
    ∃pi.
      Boehm_transform pi ∧ EVERY is_ready' (MAP (apply pi) Ms) ∧
      (∀M. MEM M Ms ⇒ p ∈ ltree_paths (BT' X (apply pi M) r)) ∧
      ∀M N q.
        MEM M Ms ∧ MEM N Ms ∧ q ≼ p ⇒
        (subtree_equiv X M N q r ⇔
         subtree_equiv X (apply pi M) (apply pi N) q r)
```

The previous added definitions `subtree_equiv'` and `ltree_equiv'`, etc. are now useless and removed.

NOTE: For the next applications of this lemma, the antecedent `EVERY (λM. subterm X M p r ≠ NONE) Ms` may still need to be weaken to `p ∈ ltree_paths (BT' X M r))` or even completely eliminated, and the involved `is_ready'` (a variant of `is_ready`) can then be recovered to the original `is_ready`. These changes will bring a few more trivial cases and further increases the proof size (now `boehmScript.sml` has more than 10K lines.)

--Chun

PR #1381, opened 2025-01-01
Extract "Boehm_construction" from the proof of subtree_equiv_lemma

binghe:
Hi,

This follows #1378. With a little more efforts, the construction of the Böhm transformation in the big proof of `[subtree_equiv_lemma]` can be extracted as the following definition:

```
Boehm_construction_def
⊢ ∀X Ms p.
    Boehm_construction X Ms p =
    (let
       n_max = MAX_LIST (MAP (λe. subterm_length e p) Ms);
       d_max = MAX_LIST (MAP (λe. subterm_width e p) Ms) + n_max;
       k = LENGTH Ms;
       Y = BIGUNION (IMAGE FV (set Ms));
       vs0 = NEWS (n_max + SUC d_max + k) (X ∪ Y);
       vs = TAKE n_max vs0;
       xs = DROP n_max vs0;
       M i = EL i Ms;
       M0 i = principle_hnf (M i);
       M1 i = principle_hnf (M0 i ·· MAP VAR vs);
       y i = hnf_headvar (M1 i);
       P i = permutator (d_max + i);
       p1 = MAP rightctxt (REVERSE (MAP VAR vs));
       p2 = REVERSE (GENLIST (λi. [P i/y i]) k);
       p3 = MAP rightctxt (REVERSE (MAP VAR xs))
     in
       p3 ++ p2 ++ p1)
```

And we can obtain an "explicit" form of that lemma in the following form without existential quantifier:
```
[subtree_equiv_lemma_explicit]
⊢ ∀X Ms p r.
    (let
       pi = Boehm_construction X Ms p
     in
       EVERY (λM. p ∈ ltree_paths (BT' X M r)) (MAP (apply pi) Ms) ∧
       ∀M N q.
         MEM M Ms ∧ MEM N Ms ∧ q ≼ p ⇒
         (subtree_equiv X M N q r ⇔
          subtree_equiv X (apply pi M) (apply pi N) q r))
```
The previous `[subtree_equiv_lemma]`, of course, can be easily derived from it as a corollary.   The definition `Boehm_construction` is not useful in proving the completeness theorem but can be seen as part of the actual "algorithm" which can separate any two λ-terms, if being further implemented as a library.

--Chun

### DIFF
--- a/examples/lambda/barendregt/boehmScript.sml
+++ b/examples/lambda/barendregt/boehmScript.sml
@@ -50,29 +50,6 @@ Overload FV  = “supp term_pmact”
 Overload VAR = “term$VAR”
 
 (*---------------------------------------------------------------------------*
- *  ltreeTheory extras
- *---------------------------------------------------------------------------*)
-
-(* ltree_subset A B <=> A results from B by "cutting off" some subtrees. Thus,
-
-   1) The paths of A is a subset of paths of B
-   2) The node contents for all paths of A is identical to those of B, but the number
-      of successors at those nodes of B may be different (B may have more successors)
-
-   NOTE: Simply defining ‘ltree_subset A B <=> subtrees A SUBSET subtrees B’ is wrong,
-         as A may appear as a non-root subtree of B, i.e. ‘A IN subtrees B’ but there's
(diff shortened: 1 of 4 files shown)


mn200:
Thanks!

PR #1384, opened 2025-01-10
Hilbert axioms and geometry theorems

lizdoki:
Hello, I'm just creating a PR to add Hilbert's axioms and other theorems in Euclidean geometry as part of my project. I've made a new geometry folder for this - I hope that's okay.


mn200:
Thanks for this.  I will create an `examples/axiomatic-developments` (or something shorter) directory for this and the set-theory stuff we already have.

PR #1387, opened 2025-01-20
fix linebreak in holindex if rule exceeds line length

rsoeldner:
fixes a compile error if the line exceeds and a newline is enforced. Will cause an error within the `\infer` statement.

Worked on in a call with @thtuerk


mn200:
Thanks!

Issue #1364, opened 2024-12-08
`Theory` syntax


digama0:
I would like to implement a new syntax for the `new_theory` command and the prelude. This will ensure more consistent use of the header and also make it more introspectable by HOL tooling like IDEs. Proposal:

```sml
(*
  Module documentation
*)
Import fooLib barTheory
Import[open] bazTheory
Theory bla

Definition foo:
  foo = T
End
```
is equivalent to
```sml
(*
  Module documentation
*)
local open fooLib barTheory in end
open bazTheory
val _ = new_theory "bla";

Definition foo:
  foo = T
End

val _ = export_theory ();
```

Some points of interest:
* I'd like to remove the `Theory` suffix from the theory imports, since HOL itself tends to refer to theories without this suffix in e.g. qualified constant names, but I don't think it can be implemented as long as quote-filter is a preprocessor which runs without any filesystem context (unless we change the structure name itself to drop the `Theory` suffix).
* It is mandatory that there be no non-whitespace text between the beginning of the file and the `Theory` command, except for `Import` declarations. In particular, that means that all opens have to happen at the beginning of the file, and there cannot be additional commands before `new_theory`. This happens occasionally (e.g. `temp_delsimps` is often before `new_theory`), but it doesn't seem to make a difference.
* For now, holdep will continue to work as it does currently, but I would like to transition this to a warning and deprecate mid-file holdeps (e.g. using qualified identifiers for unimported files).
* The `export_theory` at the end of the file is implicit when using `Theory` syntax.

xrchz:
I have a feeling this would make HOL exceptionally more welcoming/non-intimidating to newcomers.

mn200:
I agree with Magnus about both the need to make the common case easy and to include support for `set_grammar_ancestry`. 

I’d also prefer to have the `Theory bla` as the first line rather than buried.  

I don’t mind dropping the `Theory` suffix though I wonder a bit about the way users will still need to remember it when they come to refer to the theorems (on the `nobind` imports). 

Finally, is it going to be easy to have a version that works for pre-bossLib files?

digama0:
> I don’t mind dropping the `Theory` suffix though I wonder a bit about the way users will still need to remember it when they come to refer to the theorems (on the `nobind` imports).

I think we should drop the suffix on the structure name itself. But I think that can be done independently of this proposal, it's a pretty large breaking change.

mn200:
[...]

PR #1395, opened 2025-02-11
disable export_theory in interactive mode

digama0:
~~Also a possibility: using `Globals.interactive` as the default value of the flag. I decided not to because I am not sure if users like to run `export_theory` manually in REPL sessions (they shouldn't).~~ Edit: implemented

This flag is needed to prevent hol4-vscode from generating actual theory outputs when the server reads and runs the file in fast-cheat mode. (Yes, I tried rebinding `export_theory`. It's not reliable because the massive block of opens at the top of most script files has a tendency to re-rebind it.)

### DIFF
--- a/src/datatype/theory_tests/monofldBScript.sml
+++ b/src/datatype/theory_tests/monofldBScript.sml
@@ -30,4 +30,5 @@ val _ = if msg = "" then print "No message\n"
 
 val _ = save_thm("MFB", TRUTH)
 
+val _ = Globals.interactive := false
 val _ = export_theory();
--- a/src/postkernel/Theory.sig
+++ b/src/postkernel/Theory.sig
@@ -57,6 +57,7 @@ sig
   val adjoin_after_completion: (unit -> HOLPP.pretty) -> unit
   val quote_adjoin_to_theory : string quotation -> string quotation -> unit
   val export_theory          : unit -> unit
+  val disable_export_theory  : bool ref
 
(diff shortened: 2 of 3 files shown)


digama0:
Currently the HOL parser doesn't even parse `open` declarations, which makes this annoying to do correctly. Plus mid-file opens are a thing. The syntax of #1364 is precisely what we need in order to do this properly, until then I'm still having to deal with a significant fraction of breakages when using hol4-vscode normally.

mn200:
Mid-file opens should not be a thing.  I'd happily grasp that (and other) nettle first just so that eventually layering over a nicer syntax can be done as a separate concern. 

For the purposes of keeping this particular commit, small please just guard exporting with `Globals.interactive`.

digama0:
I think mid-file `open` *should* be a thing, where I mean specifically `open` and not the load side effect. `open` is a normal operation in SML and it is weird to restrict its use to the top of the file; it is used in let blocks too.

mn200:
Of course, `open` should be freely usable in non-Script files. In script files, it is equally reasonable to restrict top-level occurrences to the top of the files.  (People shouldn't be writing much code with nested `open` in script-files, but it's reasonable to keep it legal.) 

`open`'s shotgun destruction of the namespace is a very bad thing to have happen in the middle of a script.

mn200:
It looks like some test-cases will need to be adjusted.

mn200:
Thanks!

PR #1396, opened 2025-02-11
add theory of coinduction up-to context

Plisp:
- companionTheory has the main development based on "Coinduction All the Way Up", giving a way to compose extensions to coinduction principles and also extend the relation on the fly
  - more examples [here](https://github.com/Plisp/proofs/blob/main/HOL/uptoScript.sml#L1265)
- adapted some coinduction schemas in itreeTau to this new style
- added a couple common theorems to posetTheory

I wonder if there could be some toggle that automatically generates the associated functional from a Coinductive definition and proves its monotonicity / coincidence with the fixpoint.

### DIFF
--- a/examples/lambda/barendregt/lameta_completeScript.sml
+++ b/examples/lambda/barendregt/lameta_completeScript.sml
@@ -1443,6 +1443,24 @@ Proof
  >> Q_TAC (TRANS_TAC SUBSET_TRANS) ‘FV M’ >> art []
 QED
 
+Theorem ltree_finite_BT_has_benf :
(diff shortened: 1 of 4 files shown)


mn200:
In a number of places you are using `UNCURRY` where I suspect you could/should use the (equivalent, but better-named) `rel_to_reln` from `set_relationTheory`.   If you agree, please do make one more change :)

Plisp:
heh it turns out my proof could also be simplified, I still don't have great intuition for set automation

Plisp:
should be good to merge!

mn200:
Thanks for this!

PR #1397, opened 2025-02-13
[balanced_bst] AVL_treeTheory and fibonacciTheory (+fibmod3)

binghe:
Hi,

This PR contains a slightly modified version of the AVL tree formalisation, by our student (Sineeha Kodwani) in the last year.

AVL tree is another balanced bst (binary search tree), perhaps nowadays is less famous than red-black trees. A valid AVL tree is a (finite) binary tree structure (defined by a datatype with constructors `Tip` and `Bin`) whose children heights differ by at most one:
```
[avl_def]
⊢ (avl Tip ⇔ T) ∧
  ∀bf k v l r.
    avl (Bin bf k v l r) ⇔
    (height l = height r ∨ height l = height r + 1 ∨ height r = height l + 1) ∧
    bf = &height r − &height l ∧ avl l ∧ avl r
```
It's not obvious from this definition that the height of AVL trees is logarithm to the number of nodes in it (for optimal searching performance), but if we define a function `N(k)` as the minimal number of nodes of all AVL trees whose height is `k`:
```
[N_def]
⊢ ∀k. N k = MIN_SET (IMAGE node_count {x | height x = k ∧ avl x})
```
Then we can prove the following recursive relation between different `N(k)` values and a elegant connection to the Fibonacci numbers:
```
[N_k]
⊢ ∀k. N (k + 2) = N (k + 1) + N k + 1

[N_fibonacci_relation]
⊢ ∀k. N k = fib (k + 2) − 1
```

It's known (but the proof is not mechanised here) that the Fibonacci numbers has the exponential upper/lower bounds [1,2], and thus the height of AVL trees is indeed logarithm to the number of nodes.

The work also correctly defines insert and delete operations to AVL tree, and have some test theorems showing the correctness on concrete AVL trees, but the correctness of these operations is not fully formalised. (The overall code quality is medium but the finished parts are all correct.)

I extracted from the original work all proofs about the Fibonacci number itself and created a dedicated `fibonacciTheory`, also under `exampels/balanced_bst`. It's not putting into `arithmeticTheory` or `numberTheory` because I also merged the existing small `fibmod3Theory` from `examples/misc` but the proof depends on `intLib`.   I plan to formalise the exponential upper/lower bounds, but then real numbers will also involve.  Therefore it's better (for now) just have a dedicated 
`fibonacciTheory` in examples.

--Chun

[1] https://proofwiki.org/wiki/Fibonacci_Number_greater_than_Golden_Section_to_Power_less_Two
[2] https://proofwiki.org/wiki/Fibonacci_Number_less_than_Golden_Section_to_Power_less_One


binghe:
The `fibonacciTheory` contains some proofs ported from HOL-Light.  The new tactic `ASM_ARITH_TAC` in `intLib` is like `ASM_REAL_ARITH_TAC` vs `REAL_ARITH_TAC`, also learnt from HOL-Light.

The following new theorem `TRANSITIVE_STEPWISE_LE` (in `arithmeticTheory`) is also learnt from HOL-Light (only the name and statements):
```
   [TRANSITIVE_STEPWISE_LE]  Theorem      
      ⊢ ∀R. (∀x. R x x) ∧ (∀x y z. R x y ∧ R y z ⇒ R x z) ∧
            (∀n. R n (SUC n)) ⇒
            ∀m n. m ≤ n ⇒ R m n
```
whose proof is based on the existing `transitive_monotone`.

mn200:
I notice that the ASM_ version of the integer tactic pulls down every assumption that is not a forall.   I wonder if we could instead pull down everything that was presburger.  (Certainly, ARITH_TAC can cope with universally quantified assumptions, though the time-complexity will increase.)

mn200:
I think `examples/balanced_bst` is a weird place to find an example about the Fibonacci sequence.  This should either just move to `src`, or should perhaps stay in `examples/misc` (and then have `examples/balanced_bst` have `../misc` as an `INCLUDES = ...` line in its `Holmakefile`).

binghe:
> I notice that the ASM_ version of the integer tactic pulls down every assumption that is not a forall. I wonder if we could instead pull down everything that was presburger. (Certainly, ARITH_TAC can cope with universally quantified assumptions, though the time-complexity will increase.)

Hmmm, I don't know how to implement the presburger test (feel free to improve it later).  The current way is learnt by HOL-Light and the current HOL4 implementation of `ASM_ASM_ARITH_TAC`:
```
(* This function converts REAL_ARITH (a prover) to the corresponding tactics *)
fun mk_real_arith_tac (prover :term -> thm) :tactic * tactic = let
  val arith_tac = CONV_TAC (EQT_INTRO o prover);
  val asm_arith_tac =
      REPEAT(FIRST_X_ASSUM
              (fn th => if not(is_forall (concl th)) then MP_TAC th
                        else ALL_TAC)) THEN arith_tac
in
    (arith_tac, asm_arith_tac)
end;
```
I think the idea is `ASM_` versions of these tactics is to ease the user efforts of manually popping most assumptions (of the current goal, even some are irrelevant) before calling the actual decision procedure. In more complicated cases, those forall assumptions should be instantiated manually (by `MP_TAC o SPEC []`) to make them useful in decision procedure.

P.S. I'll move `fibonacciTheory` to `examples/misc`. Thanks for your advice.

mn200:
There's `is_presburger` in `IntDP_Munge`. 

But the existing tactic already does the right thing anyway (uses assumptions) thanks to `IntDP_Munge.conv_tac`:
```
> intLib.ARITH_TAC ([``x < 4i``, ``P (x:int):bool``], ``x < 10i``);
val it = ([], fn): goal list * validation
```
so we don't actually need the `ASM_` version.

binghe:
> There's `is_presburger` in `IntDP_Munge`.
> 
> But the existing tactic already does the right thing anyway (uses assumptions) thanks to `IntDP_Munge.conv_tac`:
> 
> ```
> > intLib.ARITH_TAC ([``x < 4i``, ``P (x:int):bool``], ``x < 10i``);
> val it = ([], fn): goal list * validation
> ```
> 
> so we don't actually need the `ASM_` version.

I have moved `fibonacciTheory` to `examples/misc` and now removed the useless `intLib.ASM_ARITH_TAC`.

Please squash all commits before merging.

PR #1398, opened 2025-02-16
remove old HOL parser (QuoteFilter)

digama0:
This removes the old parser, completing the work of #1333. The only thing the old parser is being used for currently is its `--quotefix` mode.


mn200:
Asymptotically we approach perfection!

PR #1399, opened 2025-02-16
parse open statements

digama0:
Adds support for top level `open` commands to `HolLex`. The main purpose of this is so that editors can reliably pick these up and handle them specially, either by preloading the things in the open command or bracketing the command with `toggle_quietdec`, or both. The latter behavior is added as an option and enabled by default in the interactive mode.

* [x] depends on: #1398


mn200:
Can you fix the merge conflicts please?  (I guess the other PR has already done some of what this is also doing.)

mn200:
Thanks!

PR #1401, opened 2025-02-17
[lambda] lameta_complete.equivalent2, etc.

binghe:
Hi,

I mainly did some cleanups to the lambda code base, added some comments (e.g. there are totally 3 proofs of the beta-CR theorem), and (more importantly) I added a few new lemmas which is currently used in the ongoing paper.

In `lameta_completeTheory`, there was the `equivalent M N` relation for two λ-terms invented long time ago. In this definition, the excluded set is fixed to `FV M UNION FV N` and the rank is fixed to 0, and the already proved separability lemmas used this definition.  But the connection between `equivalent` and `subtree_equiv` indicates that it's better (easier for some proofs) to have another version with explicit excluded set and rank. The bridging theorem is proved as the following two:
```
[equivalent2_thm]
⊢ ∀X M N r.
    FINITE X ∧ 0 < r ∧ FV M ⊆ X ∪ RANK r ∧ FV N ⊆ X ∪ RANK r ⇒
    (equivalent2 X M N r ⇔ equivalent M N)

[equivalent_alt_equivalent2]
⊢ ∀M N. equivalent M N ⇔ equivalent2 (FV M ∪ FV N) M N 0
```
In this way, recent proofs and old proofs are connected. Not an elegant solution (to have two definitions for the same concept) but the efforts are minimal in this way.

--Chun

### DIFF
--- a/examples/lambda/barendregt/boehmScript.sml
+++ b/examples/lambda/barendregt/boehmScript.sml
@@ -1,8 +1,8 @@
 (* ========================================================================== *)
 (* FILE    : boehmScript.sml (aka chap10Script.sml)                           *)
-(* TITLE   : (Effective) Böhm Trees (Barendregt 1984 [1], Chapter 10)     UOK *)
+(* TITLE   : (Effective) Böhm Trees (Barendregt 1984 [1], Chapter 10)         *)
 (*                                                                            *)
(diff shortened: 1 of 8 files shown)


mn200:
Thanks!

Issue #1400, opened 2025-02-16
Datatype size error when using finite map


pylasnier:
The following datatype declaration, specifically when using a finite map in the pair:

```
Datatype:
  t1 = T1 (((num |-> num) # t2) list)
;
  t2 = T2 t1
End
```

produces the following error:

```
Proof of 

(∀x. t1_size x = t1_size x) ∧ (∀x. t2_size x = t2_size x) ∧
(∀x. t11_size x =
     list_size
       (pair_size
          ((λksize vsize. fmap_size (λk. 0) (λv. 1 + vsize v)) (λx. x)
             (λx. x)) t2_size) x) ∧
∀x. t12_size x =
    pair_size
      ((λksize vsize. fmap_size (λk. 0) (λv. 1 + vsize v)) (λx. x) (λx. x))
      t2_size x

failed.
First unsolved sub-goal is

1 +
(pair_size ((λvsize. fmap_size (λk. 0) (λv. 1 + vsize v)) (λx. x)) t2_size x +
 list_size
   (pair_size ((λvsize. fmap_size (λk. 0) (λv. 1 + vsize v)) (λx. x)) t2_size)
   x') =
1 +
(pair_size (fmap_size (λk. 0) (λv. 1 + (λx. x) v)) t2_size x +
 list_size (pair_size (fmap_size (λk. 0) (λv. 1 + (λx. x) v)) t2_size) x')

<<HOL message: error in size_eqs, consider DataSize.prove_size_eqs := false; >>

Exception raised at Datatype.Hol_datatype:
at Tactical.TAC_PROOF:
unsolved goals
Exception-
   HOL_ERR
     {message = "at Tactical.TAC_PROOF:\nunsolved goals", origin_function =
      "Hol_datatype", origin_structure = "Datatype", source_location = <??>}
   raised
```

A note that HOL still can export a theory containing a declaration like this.

talsewell:
Yes, this is a minor bug in a tactic in an associated proof.

The key issue here is that the `finite_map` type has a size function with a lambda in it, and a `BETA_TAC` in the tactic picked this one rather than the crucial one.

There is a fix for this example now in https://github.com/HOL-Theorem-Prover/HOL/pull/1402 . We will see, when the regression test runs, whether the fix works for other standard examples.

mn200:
Fixed by 7ac6122dbae7ecaaaa4dd138bbd3592539ddf379

PR #1402, opened 2025-02-17
fix for DataSize error for finite_map

talsewell:
[github issue #1400]

A proof of type-size equivalence fails when the finite_map type is present, because the size installed for finite_map contains a lambda and causes a BETA_TAC to apply to the wrong thing. This simple fix seems to solve the problem.

The breaking example depends on the finite_map type, and probably can't be easily added to the self-test.

### DIFF
--- a/src/datatype/DataSize.sml
+++ b/src/datatype/DataSize.sml
@@ -202,8 +202,8 @@ fun size_def_to_comb (db : TypeBasePure.typeBase) opt_ind_rec size_def =
         val th1 = prove (eqs,
                 ho_match_mp_tac ind
                 \\ REWRITE_TAC (size_def' :: aux_size_eqs @ aux_size_rules @ useful_ths)
+                \\ rpt (CHANGED_TAC BETA_TAC)
                 \\ rpt strip_tac
-                \\ BETA_TAC
                 \\ ASSUM_LIST REWRITE_TAC)
         val th2 = prove (list_mk_conj aux_eqs, REWRITE_TAC [FUN_EQ_THM, th1])
       in SOME th2 end
--- a/src/datatype/selftest.sml
+++ b/src/datatype/selftest.sml
@@ -498,29 +498,34 @@ val _ = OK()
 
 val _ = tprint "Test for lambdas in supplied sizes (github issue #1400)";
 
-Hol_datatype `list_syn = L of (('a # 'b) list)`;
+val _ = quiet_warnings (fn () => let in
+  Hol_datatype `list_syn = L of (('a # 'b) list)`;
+  Defn.Hol_defn "list_syn_size_v1_def"
+    `list_syn_size_v1 (f : ('a # 'b) -> num$num) = (\x : ('a, 'b) list_syn. 0)`;
+  ()
+  end) ()
 
-Hol_defn "list_syn_size_v1_def"
-  `list_syn_size_v1 (f : ('a # 'b) -> num) = (\x : ('a, 'b) list_syn. 0)`;
-
-Hol_defn "list_syn_size_v2_def"
+val size_v2_defn = Defn.Hol_defn "list_syn_size_v2_def"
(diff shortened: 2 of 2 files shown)


talsewell:
Hmm, adding a datatype test to the finite_map tests feels strange to me.

Here is a potential compromise. The error can be replicated like this, using fresh types:

```
Datatype:
  list_syn = L (('a # 'b) list)
End

Definition list_syn_size_v1_def:
  list_syn_size_v1 (f : ('a # 'b) -> num) = (\x : ('a, 'b) list_syn. 0)
End

Definition list_syn_size_v2_def:
  list_syn_size_v2 = (\x_sz y_sz. list_syn_size_v1 (\k. x_sz (FST k) + y_sz (SND k)))
End

val _ = TypeBase.export [
    TypeBasePure.mk_nondatatype_info (
      “: ('a, 'b) list_syn”,
      {encode = NONE,
       size = SOME (``\(xsize : 'a -> num) (ysize : 'b -> num).
            list_syn_size_v2 (\x : 'a. 0) (\y. ysize y + 1)``, list_syn_size_v2_def),
       induction = NONE,
       nchotomy = NONE}
      )
  ]

Datatype:
  t1 = T1 (((num, num) list_syn # t2) list)
;
  t2 = T2 t1
End
```

TODO: the selftest.sml files don't seem to use new-style `Definition`, figure out if that is an issue and how to back-port this test.

talsewell:
Hmm, apparently I'm just blogging, I think I've figured it out and added it to the PR.

mn200:
Thanks for working on this!  Note that the list type you are using in your tests is constructed for the purpose in the same `selftest.sml`.  This is required because `src/datatype` precedes `src/list` in the build sequence.  I expect the slight differences between this construction and what happens in `listScript.sml` are not important.

Given that finite maps are not algebraic and involve an operator with arity=2, using lists is superficially rather different in feel.  (You will know if the list example ultimately hits the same problem/bug of course.) 

If you really wanted a finite map case, you could start to fake this with something along the lines of 
```
val _ = new_type ("finite_map", 2)
```
but to truly reflect the context, you'd have to also do the equivalent of (from the bottom of `finite_mapScript.sml`):
```
val _ = TypeBase.export [
    TypeBasePure.mk_nondatatype_info (
      “:'a |-> 'b”,
      {encode = NONE,
       size = SOME (“λ(ksize:'a -> num) (vsize:'b -> num).
                      fmap_size (λk:'a. 0) (λv. 1 + vsize v)”,
                    fmap_size_def),
       induction = SOME fmap_INDUCT,
       nchotomy = SOME fmap_CASES}
      )
  ]
```
which starts to feel like a lot of work for a test-case.

talsewell:
Yeah, I noted that it's a bit of an accident that `'a list` is an existing type. Maybe that should be done explicitly rather than inherited implicitly from a previous test. The other thing to note, there, is that none of it matters. The problem is the shape of the relevant size term and its definition. You'll note that in the test I added, it's all actually just syntax for `0`.

mn200:
- Earlier in the same file is explicit enough I reckon
- Your errors above are an unqualified reference to `Hol_defn`.

mn200:
Oh yeah: "modern syntax" is only allowed in `*Script.sml` files.

talsewell:
Hmm. I figured out how to run it locally, to test it a bit faster, and I've fixed a few more ML naming issues, but I'm stuck on an error reading this line:
```
``: ('a, 'b) list_syn``
```
It results in `HOL_ERR {message = "on line 512, characters 10-17:\ntuple with no suffix", origin_function = "parse_type", origin_structure = "Parse"}`

I don't get it. The syntax `: ('a, 'b) ty` was used earlier in the file.

talsewell:
Never mind, I think I've figured it all out now. Reminder to self, don't get sucked into fiddling with `selftest` files, it eats up too much of a morning.

mn200:
Thanks again!

PR #1403, opened 2025-02-17
Theory of JSON parsing

didriklundberg:
This small JSON parsing theory, written by @arolle and myself, has been validated against the ["Parsing JSON is a Minefield" test suite](https://github.com/nst/JSONTestSuite). @myreen suggested I make a PR in some previous discussion - please give feedback on whether there is additional work or cleanup needed.

### DIFF
--- a/examples/json/parse_jsonScript.sml
+++ b/examples/json/parse_jsonScript.sml
@@ -106,10 +106,10 @@ Definition json_to_string_def:
   (json_to_string obj =
     case obj of
         | Object mems => ["{"] ++
-                concat_with (MAP mem_to_string mems) ([","]) ++
+                concat_with (MAP mem_to_string mems) [","] ++
                 ["}"]
         | Array obs => ["["] ++
-                concat_with (MAP json_to_string obs) ([","]) ++
+                concat_with (MAP json_to_string obs) [","] ++
                 ["]"]
        | String s => [CONCAT ["\""; encode_str T s; "\""]]
        | Number i => [toString i]
@@ -412,8 +412,6 @@ End
 
 (* TODO prove equality between json_to_tok and json_to_string *)
 
-val json_size_def = fetch "-" "json_size_def";
-val json_size_eq = fetch "-" "json_size_eq";
 
 (*
 Examples:
--- a/examples/json/parse_jsonScript.sml
+++ b/examples/json/parse_jsonScript.sml
@@ -11,14 +11,19 @@ EVAL ``int_lt (-1:int) 0``
 (Some kind of of zero comparison would be critical to the json_to_string_def function)
 
 Therefore, integers are stored using a bool flag for sign (T means negative) and a regular num,
-with a separate converter function to HOL4 integers.
+with a separate converter function to HOL4 integers, reals et.c. as future work.
 *)
+Datatype:
+  sign =
+     Positive
+   | Negative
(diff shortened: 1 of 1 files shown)


didriklundberg:
Converted to draft: I'll add another commit with some minor fixes and cleanup.

mn200:
I'd be happy to take delivery of this more or less as is, but do take your time polishing to your satisfaction.  One thing, I'd particularly like to see, but can do myself if you prefer is to really use `:int` and not `:num`.

didriklundberg:
> I'd be happy to take delivery of this more or less as is, but do take your time polishing to your satisfaction. One thing, I'd particularly like to see, but can do myself if you prefer is to really use `:int` and not `:num`.

Thanks for your feedback. I would also hope for a better number representation than the current one, and I tried rewriting the theory to use `int` instead of `sign # num`. However, this results in failing validation tests due to not keeping the distinction between `0` and `-0`, which are indeed distinct e.g. in the IEEE 754 standard. Even worse, since the fractional component is still separate, `-0.3` gets wrongly represented as `0.3`.

Should I try using `rat` and ignore the `0` vs. `-0` issue, or keep things as they are and just add a comment based on this discussion?

mn200:
This is an interesting question.  Of course, if you're committed to being that accurate, you might use HOL's floating point type.

didriklundberg:
> This is an interesting question. Of course, if you're committed to being that accurate, you might use HOL's floating point type.

Thanks for the tip. I've looked through `floating-point/binary_ieeeScript.sml` and `float/ieeeScript.sml`, with the following conclusions:

* `binary_ieeeScript` defines binary floats in terms of words (for which you commit to a specific precision), so this would not be usable for JSON numbers, which require an arbitrary-precision decimal floating point representation.
* `ieeeScript` does not define a float datatype as such. Rather, the arithmetic defined therein operates on a tuple of `num`s and also assumes a binary floating-point representation as opposed to a decimal one. Representing decimal fractional numbers using binary floating-point numbers necessarily incurs rounding errors, which we can't have. So there seems to be no clear benefit in using the conventions of `ieeeScript`, since the definitions would not be usable anyhow.

A brief search reveals there was an extension of Harrison's formalisation (which `ieeeScript` is based on?) to arbitrary radix in 2015 ([paper](https://core.ac.uk/download/pdf/82081461.pdf), [relevant code](https://github.com/jrh13/hol-light/blob/master/IEEE/float.hl)). Although their generalized representation seems a bit weird (e.g. "zero is not a floating point number"), it might be useful as inspiration for a true arbitrary-precision arbitrary-radix floating point representation. But this is probably too much for this PR.

Therefore, I suggest I document the JSON number representation further, make it a separate datatype, then provide mechanisms for extracting `rat`a and `int`s (as a cast with loss of information, and as a lossless function returning an option type). This would allow accurate serialisation+deserialisation, and some compatibility with existing HOL4 formalizations. What do you think?

mn200:
Thanks for taking the time over this.  I agree with you that the current approach is probably the best match for the full generality of what the RFC requires.

didriklundberg:
Thanks again for your feedback. I have now marked the PR as ready for review and merge.

mn200:
Thanks again!  May move within the `examples` directory, perhaps to be under `examples/formal-languages`.

PR #1406, opened 2025-02-24
Add cv_eval_pat to cv_transLib (and an example)

myreen:


examples/cv_compute/cv_eval_exampleScript.sml
```diff
@@ -0,0 +1,28 @@
+(*
+  Demonstrates use of cv_eval_raw_save
```
xrchz, line 2: Where is `cv_eval_raw_save` used here?

src/num/theories/cv_compute/automation/cv_transLib.sml
```diff
@@ -876,7 +876,13 @@ fun cv_eqs_for tm = let
 
 val _ = computeLib.add_funs [cv_fst_def,cv_snd_def];
 
-fun cv_eval_raw tm = let
+datatype pat = Raw
+             | Eval of conv
+             | Name of string
+             | Tuple of pat list
+             | Some of pat;
```
xrchz, line 883: What's the difference between `Some p` and `Tuple [p]`?

examples/cv_compute/cv_eval_exampleScript.sml
myreen: My latest commit fixes this.

src/num/theories/cv_compute/automation/cv_transLib.sml
myreen: This comment makes me realise that it's unclear what the pattern means. I wrote a long comment in the file, see:

https://github.com/HOL-Theorem-Prover/HOL/blob/e211611d3319151d421bac5caee5265866f6ec7b/examples/cv_compute/cv_eval_exampleScript.sml#L15-L24

xrchz: So is `Tuple [Some (Tuple [])]` just not really supposed to be used? Or does that also make sense?

myreen: `cv_eval_pat` will raise an exception if users write `Tuple []` in their pattern. 

Writing `Tuple [x]` will mean the same as a single-element tuple in HOL, i.e. it's not a tuple. What's the difference between `(2)` and `2` in HOL? No difference.

tanyongkiam: Some additional context: the constructor `Tuple` was originally `Pair` (which avoids some of the above issues). However, the `Pair` constructor would need to be disambiguated because it is also used elsewhere in CV.

mn200:
Thanks for this!

PR #1410, opened 2025-02-27
add `redo()` / `rd()` to proof manager

digama0:
I don't understand how we went 30 years without this.


mn200:
You don't know you need something until someone comes along and gives it to you...   Thanks!

PR #1417, opened 2025-03-03
Insertion and deletion operators on ltrees

binghe:
Hi,

The `ltreeTheory` provides a general coinductive datatype for representing potential infinite depth and width trees. Using `ltree_lookup` and `ltree_el`, one can access any ltree subtree or just single node of any path (as `:num list`) of the ltree. But currently there's no way to insert another ltree into the leaf of an existing ltree as a subtree, or to remove one subtree from an existing ltree and get the remaining part. This PR adds two new function `ltree_insert` and `ltree_delete` for these purposes.

However, these operations have limitations: they only operate at the *right most* children of any ltree node. This is because, if a new subtree were added/deleted at the right most position, all other ltree nodes can still be accessed using their original paths. Otherwise, the related shifting process of paths makes things tricky.  (Fortunately, these operations are enough for me, because the η-reduction process of Böhm trees corresponds to the deletion of rightmost children of involved tree nodes.)

Since the insert and delete operations target at the right most children of certain ltree node, these operations do NOT take the path of the actual insertion/deletion node (as arguments) but their parent node(path):

The function `ltree_delete f t p` deletes the right most child of the ltree `t` at path `p`, and if the deletion succeed (the children list is not infinite, and there's at least one child), the ltree node data `d` at path `p` will be modified to `f(d)`. This is the formal definition:
```
   [ltree_delete_def]  Definition
      ⊢ ∀f t p.
          ltree_delete f t p =
          gen_ltree
            (λns.
                 (let
                    (d,len) = THE (ltree_el t ns);
                    m = THE len
                  in
                    if ns = p ∧ len ≠ NONE ∧ 0 < m then (f d,SOME (m − 1))
                    else (d,len)))
```

On the other hand, `ltree_insert f t p t0` inserts a new subtree `t0` as the right most child into the ltree `t` at the parent path `p`. And if the insertion succeeds (the existing children list must not be infinite), the parent node `d` at `p` is modified by the function `f(d)`:
```
   [ltree_insert_def]  Definition      
      ⊢ ∀f t p t0.
          ltree_insert f t p t0 =
          gen_ltree
            (λns.
                 if ltree_el t ns ≠ NONE then
                   (let
                      (d,len) = THE (ltree_el t ns);
                      m = THE len
                    in
                      if ns = p ∧ len ≠ NONE then (f d,SOME (m + 1))
                      else (d,len))
                 else THE (ltree_el t0 (DROP (LENGTH p + 1) ns)))
```

These definitions involve a lot of cases (I don't see shorter ways), and you can imagine it's hard to prove their properties (the proofs in this PR almost doubled the size of `ltreeScript.sml`).

The following theorems show that the set of ltree paths after deletion is indeed the original tree removing just the deleted subtree (and the deletion point, i.e. the parent path, is still there):
```
   [ltree_delete_path_stable]  Theorem      
      ⊢ ∀f p t. p ∈ ltree_paths t ⇒ p ∈ ltree_paths (ltree_delete f t p)
   
   [ltree_delete_paths]  Theorem      
      ⊢ ∀f p t a n.
          ltree_el t p = SOME (a,SOME (SUC n)) ⇒
          ltree_paths (ltree_delete f t p) =
          ltree_paths t DIFF
          IMAGE (λq. SNOC n p ⧺ q)
            (ltree_paths (THE (ltree_lookup t (SNOC n p))))
```

The following theorem shows that, if you delete a subtree and then insert it back, you can get the original ltree. Note that, `ltree_insert'` and `ltree_delete'` are overloads ignoring the modifier function (let `f = I`):

```
   [ltree_insert_delete']  Theorem      
      ⊢ ∀n p t t0.
          ltree_lookup t (SNOC n p) = SOME t0 ∧
          ltree_branching t p = SOME (SUC n) ⇒
          ltree_insert' (ltree_delete' t p) p t0 = t

Overload ltree_delete' = “ltree_delete I”
Overload ltree_insert' = “ltree_insert I”
```

Furthermore, if you insert a finite ltree into a finite ltree, the resulting ltree is (of course) still finite:
```
   [ltree_finite_insert]  Theorem      
      ⊢ ∀f p t t0 d len.
          ltree_finite t ∧ ltree_el t p = SOME (d,len) ∧ ltree_finite t0 ⇒
          ltree_finite (ltree_insert f t p t0)
```

Finally, using all results above (a bit surprisingly), I have proved the following relationship between `ltree_finite` and `ltree_paths`: (In previous PR only the easy direction was proved)
```
   [ltree_finite_alt_ltree_paths]  Theorem      
      ⊢ ∀t. ltree_finite t ⇔ FINITE (ltree_paths t)
```

Let's hope this `ltreeTheory` will become the foundation of all tree-like structures of formalisations in HOL4....

--Chun

### DIFF
--- a/examples/lambda/barendregt/boehmScript.sml
+++ b/examples/lambda/barendregt/boehmScript.sml
@@ -971,7 +971,7 @@ Theorem BT_ltree_paths_thm :
     !X M p r. FINITE X /\ FV M SUBSET X UNION RANK r ==>
              (p IN ltree_paths (BT' X M r) <=> subterm X M p r <> NONE)
 Proof
-    rw [ltree_paths_alt, BT_ltree_el_eq_none]
+    rw [ltree_paths_alt_ltree_el, BT_ltree_el_eq_none]
 QED
 
 (* NOTE: p <> [] is required as ‘[] IN ltree_paths (BT' X M r)’ always holds. *)
--- a/examples/lambda/barendregt/lameta_completeScript.sml
+++ b/examples/lambda/barendregt/lameta_completeScript.sml
@@ -1249,7 +1249,7 @@ Proof
          MATCH_MP_TAC BT_of_principle_hnf >> simp []) >> Rewr' \\
      Q.PAT_X_ASSUM ‘M1 0 = _’ (REWRITE_TAC o wrap) \\
      simp [BT_def, Once ltree_unfold, BT_generator_def, LMAP_fromList,
-           ltree_paths_alt, ltree_el_def] \\
+           ltree_paths_alt_ltree_el, ltree_el_def] \\
      simp [GSYM BT_def, LNTH_fromList, MAP_MAP_o] \\
      Cases_on ‘h < m’ >> rw [])
  >> DISCH_TAC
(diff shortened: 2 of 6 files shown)


mn200:
Thanks!

PR #1419, opened 2025-03-04
Remove cv_rep_add and use [cv_rep] annotation instead

myreen:
A big thank you to @digama0 for tracking down this bug. The bug made it so that cv_trans_deep_embedding only stored the cv_rep theorems in the local theory and they were not accessible from other theories.


digama0 (approved):
Confirmed that it works in the original setting that prompted the report.

mn200 (approved):

PR #1423, opened 2025-03-10
[cheatsheet] A couple of fixes

hrutvik:
- `Simp{L,R}HS` do not take theorem arguments
- `{drule,dxrule}_then` take a theorem-tactic then a theorem, not the other way round

Thanks to Pascal Lasnier for the bug report


mn200:
Thanks!

PR #1427, opened 2025-03-13
HP_complete and HP_complete_lameta

binghe:
In previous PR, some lemmas have unnecessary forall variables that can be eliminated.

Besides, previously I added `TRANSITIVE_STEPWISE_LE` into `arithmeticTheory` but this theorem is already in `iterateTheory` (with other siblings, with smaller proofs). I revert this change before anyone starts to use it.


binghe:
Hi, with a little more efforts, I defined `HP_complete` (Definition 4.1.22 of Barendregt) in `chap4Theory`:
```
[HP_complete_def]
⊢ ∀thy P.
    HP_complete thy P ⇔
    ∀M N. P M ∧ P N ⇒ thy⁺ M N ∨ inconsistent (thy ∪ {(M,N)})⁺
```
and then proved the following more compact form of the main completeness result:
```
[HP_complete_lameta]
⊢ HP_complete (UNCURRY η) has_benf
```
Note that the extra predicate `P` in the definition of `HP_complete` is to limit the involved λ-terms to those having β(η)-normal forms.   There exists another λ-theory called "K*" whose completeness is universal (where `P = K T`) but this will be a future story.

Now I think the present work is really complete.

mn200:
Nice!

PR #1438, opened 2025-03-26
rm UOK

digama0:
See [unicode in `src/`](https://hol.zulipchat.com/#narrow/channel/486798-.E2.9D.84.EF.B8.8F-HOL4/topic/.5BRefactor.5D.20Only.20use.20Unicode/near/504986550).


mn200:
I thought rerunning the tests would run against the new base `develop`, but it seems to be running against the original merge SHA.  Oh well.   Perhaps there is some github-fu I should be able to exploit here, but in the meantime, I'll just merge. 

Thanks for the work!

binghe:
> I thought rerunning the tests would run against the new base `develop`, but it seems to be running against the original merge SHA. Oh well. Perhaps there is some github-fu I should be able to exploit here, but in the meantime, I'll just merge.

Actually I always thought the CI tests for PRs are done only on the commit SHA in the delivery branch, not affected by the `develop` branch which may have test failures at that moment. But then I was confused by the fact that, when `develop` is indeed in failures and I forked my delivery branch from the "good" position in `develop`, and eventually the CI tests in my PR  still fail.  After searching on Internet I found this post:

"If your workflow triggers on the pull_request event, GITHUB_SHA will be set to the merge commit, and actions/checkout will check out the merge of the PR." [1]

So GitHub is actually indeed doing what you thought (and not what I thought). When `develop` has test failures, all CI tests on PRs should also fail, unless the PR intends to fix the test failures in `develop`, no matter where in `develop` it has forked the delivery branch.

[1] https://github.com/actions/checkout/issues/15

PR #1439, opened 2025-03-27
Upstream MD5 from CakeML

ordinarymath:


mn200:
On the assumption that this is fast to build please put an entry for it in `src/parallel_builds/core/Holmakefile`.  Also refer to it in the `Holmakefile` in `examples/Crypto`.

ordinarymath:
Any idea why the regression is like that @binghe ?

binghe:
> Any idea why the regression is like that @binghe ?

It seems that the building process of your `md5Theory` just hangs there and doesn’t finish. The CI machine has only two CPUs, and one is waiting for the hanging theory, only the other one builds all the rest files, which took almost doubled time. At the end of the log, just the `md5Theory` still didn’t finish.

binghe:
I added a necessary line into your empty `Holmakefile` and let’s see how the newly triggered CI tests will go. Meanwhile please confirm your theory file is indeed buildable…

binghe:
This new theory definitely hangs - either it takes many hours or infinite time to finish.

examples/Crypto/MD5/md5Script.sml
```diff
@@ -0,0 +1,604 @@
...
+
+Definition md5_def:
+  md5 str = toHexString (final (FOLDL update1 init (MAP (n2w o ORD) str)))
+End
+
+(*
+    Some testing
+*)
+
+fun run_test str = let
```
ordinarymath, line 586: These tests seems to take longer than expected. Is there a import i need to pull for cv_compute?

ordinarymath: Actually i just ran Holmake and it only took 17.1s on my computer

ordinarymath: wait i was looking at the wrong time

ordinarymath:
I can reproduce locally @binghe you can stop the regression for now while I figure out why its taking so long.

ordinarymath:
This should fix the problem.

mn200:
Thanks!

PR #1442, opened 2025-04-04
HOL keywords don't stop open unless they are col 0

digama0:
Otherwise `open Overload` breaks.


mn200:
Thanks!  (Incidentally, I was very confused by your commit message. I think it's good `git` style to make your commit line describe the change your commit is achieving in present/imperative tense; your commit message here reads like an observation on the state of the code base, and it's not clear if it's an observation of “before” or “after”.)

PR #1443, opened 2025-04-04
Add TacticParse module

digama0:
See [zulip](https://hol.zulipchat.com/#narrow/channel/486798-.E2.9D.84.EF.B8.8F-HOL4/topic/How.20many.20ways.20are.20there.20to.20spell.20THEN.3F/near/510068930) for context.

PR #1444, opened 2025-04-05
goalFrag goal manager

digama0:
See [zulip](https://hol.zulipchat.com/#narrow/channel/486798-.E2.9D.84.EF.B8.8F-HOL4/topic/How.20many.20ways.20are.20there.20to.20spell.20THEN.3F/near/510068930) for context. Companion to the `tac_frag` type from #1443.

### DIFF
--- a/src/proofman/Manager.sig
+++ b/src/proofman/Manager.sig
@@ -5,6 +5,7 @@ sig
   datatype proof0
        = GOALSTACK of goalStack.gstk History.history
        | GOALTREE of goalTree.gtree History.history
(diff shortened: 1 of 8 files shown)


mn200:
You will need to add a `reraise` binding to the `src/portableML/mosml/Exn.{sig,sml}` code; and yes, its implementation will have to just be
```
fun reraise e = raise e
```

digama0:
I grepped and this seems to already exist though?

mn200:
Thanks!

Issue #69, opened 2012-07-25
Reform "extra" theories


mn200:
Joe’s “extra” theories, now present in both `src/probability` and `examples/miller` should have their good theorems moved to the relevant “home theories” under `src`.  Theorems that aren’t “good” should be checked to see how often they are used in the formalisations.  Only those that are used a number of times deserve to live on in a separate theory (and if they’re used that often, perhaps they’re “good”).  Those only used a few times should just be inlined into the relevant theories.

PR #1445, opened 2025-04-07
Remove util_probTheory (merged into sigma_algebraTheory)

binghe:
Hi,

The `util_probTheory` (see #69) is getting smaller and smaller and after these years many theorems in it have been moved to upstream theories (like `pred_setTheory`, `seqTheory` or `real_sigmaTheory`, etc.), I think the remaining part (mostly set-theoretic results) can be merged with the relatively small `sigma_algebraTheory`. This PR does this change without adding or removing any theorem. All `open util_probTheory` are removed from the official probability-related examples.

--Chun


binghe:
The new `sigma_algebraTheory` does not depend on `realTheory` and in theory should be built much earlier. But strangely the building sequence of it is still after `real_sigmaTheory`. I don't quite understand this phenomenon, for now.

mn200:
Everything in `src/probability` depends on the `real-heap`, and so the dependency graph has a bottleneck in it.  If you wanted to get `sigma_algebra` built earlier and in parallel with things it is not connected to, you would need to move it to another directory.  I'll let you decide if you want to bother. 

Thanks for the content of *this* PR anyway.

binghe:
> Everything in `src/probability` depends on the `real-heap`, and so the dependency graph has a bottleneck in it. If you wanted to get `sigma_algebra` built earlier and in parallel with things it is not connected to, you would need to move it to another directory. I'll let you decide if you want to bother.
> 
> Thanks for the content of _this_ PR anyway.

I think it's better to keep `sigma_algebraTheory` in its current position. Thanks.

PR #1446, opened 2025-04-08
refactor: "Theorem foo :" -> "Theorem foo:"

digama0:
Trying my hand at a simple automated style refactoring. This is the result of applying the two regexes
`Theorem (\w+) :` -> `Theorem \1:` and `Definition (\w+) :` -> `Definition \1:`.


binghe:
What’s wrong of leaving a whitespace after the theorem/definition name and before the colon? I always do this and will keep doing it in all my future code.

If you have time doing this, I think you should do something more meaningful, e.g. replacing all legacy grammar (i.e. explicit calls of `store_thm`) to `Theorem` (and fix the name mismatch between SML variables and exported theorem names).

examples/decidable_separationLogic/src/decidable_separationLogicLib.sig
```diff
@@ -4,7 +4,7 @@ sig
 include Abbrev;
 
 (*Contains theorems that state that the universe of
-  some type is infinite. This theorem for :num is
+  some type is infinite. This Theorem for:num is
```
binghe, line 7: What’s this?!

mn200:
Line about “I know, I'll solve my problem with a regular expression...” comes to mind.  (Which I appreciate is rich coming from the guy that wrote `QuoteFilter`...)

More pertinently, I don't think whitespace code-layout style policing is somewhere we really want to go.  I was more curious about the output from
```
grep --include='*.sml' -r "[A-Za-z0-9_']  [A-Za-z0-9_']" .
```
though.  Some hits are definitely bad style and/or mistakes.

digama0:
> If you have time doing this, I think you should do something more meaningful, e.g. replacing all legacy grammar (i.e. explicit calls of store_thm) to Theorem (and fix the name mismatch between SML variables and exported theorem names).

I was planning to get there, but the whitespace throws a lot of more naive regexing off and I was hoping this would be an MVP test bed before doing more complicated things.

digama0:
> What’s wrong of leaving a whitespace after the theorem/definition name and before the colon? I always do this and will keep doing it in all my future code.

It screws up the highlighting, but I appreciate that this can be solved in other ways. I would just like to have some dang code standards

digama0:
> If you have time doing this, I think you should do something more meaningful, e.g. replacing all legacy grammar (i.e. explicit calls of store_thm) to Theorem (and fix the name mismatch between SML variables and exported theorem names).

This is much more than a regex can do (correctly). And i'm worried that someone else is going to [complain about their workflow](https://xkcd.com/1172/) if I try doing it too. This PR was precisely to find out how much people are actually willing to commit to any kind of style such that e.g. a formatter could implement it.

digama0:
> This sounds analogous to deciding that spaces around colons that separate expressions from types, e.g.,
> 
> ```
> val x : int
> ```
> 
> or
> 
> ```
> fun f (x:int) = ...
> ```
> 
> should have a governing rule. (Or that we should always/never have spaces after the commas in tuples and lists; or, ....) But I really don't think there should be such rules.

Yes I do want such rules. I try to follow the style of the code I'm editing, but it's insane when a single code block uses three different styles at the same time.

digama0:
> [Spaces in Theorem ident declarations are not a 20 year-old problem, and will never be "completely nonstandard".]

Does HOL have standards? If so, can we enumerate them and put them on a todo list? Please?

mn200:
The existing standards are documented in the “Developers” manual (see `Manual/Developers/`) and checked in the (misleadingly named) `unicode-grep` tool.

That looks for:
- illegal Unicode
- TABs
- trailing whitespace
- lines longer than 80 chars

TABs and trailing whitespace are currently checked.  The others are turned off currently (we did use to restrict Unicode under `src`).

digama0:
I'm not asking about existing linters, I'm talking about code style guidelines. Existing linters are extremely limited in scope, in part because there isn't anything that really understands SML code in HOL. But *people* have to write code and they should be told how we expect them to do it.

binghe:
> I'm not asking about existing linters, I'm talking about code style guidelines. Existing linters are extremely limited in scope, in part because there isn't anything that really understands SML code in HOL. But _people_ have to write code and they should be told how we expect them to do it.

I wouldn't jump out on the first place if:
1. you were doing the opposite direction, i.e. from `Theorem foo:` to `Theorem foo :`;
2. I'm not the author of at least 3k of those 5k theorems that you were trying to touch.

Below I explain the reasons why I prefer adding that whitespace:

1. We have the following two grammars to write a `Theorem`: 1) `Theorem foo : ...` (`store_thm`) and 2) `Theorem foo = ...` (`save_thm`).  If you think it makes sense the write `int i = 1;` in C (instead of writing `int i=1;`), then it will also make sense to have a whitespace before `=` in `Theorem foo = ...` and then it will be *consistent* to also have a whitespace before `:` in `Theorem foo : ...`, otherwise it looks inconsistent when two kinds of theorem definitions occur together.
2. For people who are imagining the `Theorem foo : ..` is declaring a symbol `foo` whose "type" is the theorem statements, please take a look at how HOL's term pretty printer print a symbol with type: `(a :'a)` and notice that there's one whitespace before the colon.
3. Open the "HOL Reference Page" and take a look at any generated theory signature, e.g.
```
signature alignmentTheory =
sig
  type thm = Thm.thm
  
  (*  Definitions  *)
    val align_def : thm
    val aligned_def : thm
    val byte_align_def : thm
    val byte_aligned_def : thm
```
please use your eyes to see those whitespaces before the colons.

Of course, you could say that there is no two completely identical things/cases in the world (and therefore any kind of analogy is not convincing), and the "esthetics" principles behind coding is simply non-exist for many people. Even so, can I have at least some *freedom* to write in whatever style I want?

digama0:
Let's continue this discussion on Zulip, I think it's worth talking about `Theorem foo :` specifically.

binghe:
Furthermore, sometimes the theorem/definition names come with a tag like `[simp]`, and other times the theorem/definition names contains a `'`. These special letters shouldn't be mixed with the following colon (the separator).

PR #1450, opened 2025-04-11
Fix more grammars

ordinarymath:
For current mcandidate breakage merge there


mn200:
Thanks!  Clearly that code should be being tested within HOL's sourcetree too.

ordinarymath:
I haven't fixed all of the grep results some are in examples but i don't think they break CakeML so they can go to develop

mn200:
Hmm.  Not sure I really wanted this in `mcandidate` separately from `develop`.  Never mind.

PR #1452, opened 2025-04-11
allow rebind listener in interactive mode

digama0:
This is one reason why rebuilding modules can fail in interactive mode.


digama0:
@mn200 Any ideas on what to do about the mosml failure here? I think the issue is that it adds a dependency on `Globals` which hasn't been compiled yet because `Listener` is in `src/portableML` and `Globals` is in `src/prekernel`.

mn200:
Move all the listener stuff to `prekernel/Lib` and perhaps replace the `DUP` exception with a `HOL_ERR` as well.

mn200:
Awesome, thanks.

PR #1453, opened 2025-04-11
remove `prove` from RealArith.sml (part 1)

digama0:
This moves all of the pro-forma theorems in RealArith.sml to realaxScript.sml, [except one](https://github.com/HOL-Theorem-Prover/HOL/blob/60f5573ec10b0a0beef8b6e76ca7e6c8bedbb555/src/real/RealArith.sml#L2452-L2510), which looks like it will need RealArith.sml to be split since it uses the tactic itself to prove a pro-forma theorem for the final version of the tactic.

### DIFF
--- a/src/real/RealArith.sml
+++ b/src/real/RealArith.sml
@@ -18,17 +18,17 @@
 structure RealArith :> RealArith =
 struct
 
-open HolKernel Parse boolLib liteLib;
+open HolKernel Parse boolLib liteLib
 
 open pairTheory pairLib reduceLib tautLib simpLib BasicProvers Ho_Rewrite
      jrhUtils
-     Canon_Port AC prim_recTheory numTheory numLib numSyntax arithmeticTheory;
+     Canon_Port AC prim_recTheory numTheory numLib numSyntax arithmeticTheory
 
-open normalForms;  (* for HOL-Light's GEN_NNF_CONV, etc. *)
-open Normalizer;   (* for HOL-Light's SEMIRING_NORMALIZERS_CONV *)
+open normalForms  (* for HOL-Light's GEN_NNF_CONV, etc. *)
+open Normalizer   (* for HOL-Light's SEMIRING_NORMALIZERS_CONV *)
 
-open realaxTheory; (* NOTE: cannot open realTheory here *)
-open Arbint;       (* TODO: remove this in the future, using the default Int *)
+open realaxTheory (* NOTE: cannot open realTheory here *)
+open Arbint       (* TODO: remove this in the future, using the default Int *)
 
 (*---------------------------------------------------------------------------*)
 (* Establish the required grammar(s) for executing this file                 *)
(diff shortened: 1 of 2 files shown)


binghe:
It seems that you uses regex to remove all semicolons in the code, even it’s not a trailing semicolon and the text after it is not a comment block.  How do you avoid breaking SML code which really needs a semicolon, e.g. `(print “…”; x)` (debug printing before evaluating an expression)?

There’s no visible performance issue in proving some extra helper theorems when loading `RealArith`, and many similar SML modules do that too. The main reason to NOT put these helper theorems into a theory, is that the statements of these theorems are dirty, tailored just for the purposes in the automatic prover. User proofs or later theories never need to directly call these theorems by name, therefore exporting their names in `realaxTheory` is a bad idea.

I appreciate your time spending on this PR but I don’t like the way you (almost blindly) kill all semicolons (as a minimalist), and the general idea of moving all local helper theorems from non-theory SML modules to exported theories.

digama0:
> It seems that you uses regex to remove all semicolons in the code, even it’s not a trailing semicolon and the text after it is not a comment block. How do you avoid breaking SML code which really needs a semicolon, e.g. (print “…”; x) (debug printing before evaluating an expression)?

I removed all of them and then manually put back the load bearing ones

binghe:
> > It seems that you uses regex to remove all semicolons in the code, even it’s not a trailing semicolon and the text after it is not a comment block. How do you avoid breaking SML code which really needs a semicolon, e.g. (print “…”; x) (debug printing before evaluating an expression)?
> 
> I removed all of them and then manually put back the load bearing ones

Are you going to do this for all HOL4’s SML source code (or files whenever you have chance to modify even a bit in it)?

digama0:
> There’s no visible performance issue in proving some extra helper theorems when loading RealArith, and many similar SML modules do that too.

See [#❄️ HOL4 > Why is startup slow? @ 💬](https://hol.zulipchat.com/#narrow/channel/486798-.E2.9D.84.EF.B8.8F-HOL4/topic/Why.20is.20startup.20slow.3F/near/510936952) - it is *very* visible. This is a data-oriented refactoring. And yes, all other files which do this are on the queue to be refactored.

> The main reason to NOT put these helper theorems into a theory, is that the statements of these theorems are dirty, tailored just for the purposes in the automatic prover. User proofs or later theories never need to directly call these theorems by name, therefore exporting their names in realaxTheory is a bad idea.

I think this is an easily mitigated issue: just have a theory specifically for these lemmas, which is never opened in user code. My understanding was that `realaxTheory` is one of these theories - users normally open `realTheory` instead. But if needed I can move the lemmas to a new theory. (Note that the bottom of the file already contains a number of theorems which are only used as pro-forma lemmas, and the new ones are going into the same section.)

> I appreciate your time spending on this PR but I don’t like the way you (almost blindly) kill all semicolons (as a minimalist), and the general idea of moving all local helper theorems from non-theory SML modules to exported theories.

This is discussed on zulip, see [#❄️ HOL4 > Should we use semi-colons at the end of declarations? @ 💬](https://hol.zulipchat.com/#narrow/channel/486798-.E2.9D.84.EF.B8.8F-HOL4/topic/Should.20we.20use.20semi-colons.20at.20the.20end.20of.20declarations.3F/near/509198727). Semicolons in SML are meaningful, and I would like them to exist only when they are doing something important, so that by seeing one I know it is doing something important. If you disagree with this convention please speak up on the thread.

As for moving helper theorems, this is also discussed on zulip, and Michael has already [given the okay](https://hol.zulipchat.com/#narrow/channel/486798-.E2.9D.84.EF.B8.8F-HOL4/topic/Why.20is.20startup.20slow.3F/near/510320173) for this effort. I know that sometimes change is hard but HOL is currently moribund and there are a huge number of cleanup tasks that need to be done to get it into shape.

> Are you going to do this for all HOL4’s SML source code (or files whenever you have chance to modify even a bit in it)?

Possibly? In this case I'm touching almost every line of the file (with more to come in the next part) so as long as the official stance is "don't care" I see no problems here. Doing this for all structures would be more complex. In any case I want to get to a point where there are clear guidelines people can follow and the code that exists follows the guideline in the majority of cases.

binghe:
PolyML does not have on-disk pre-compiled binaries (except for staged heap) and thus loading a SML module always result in on-the-fly compilation. Even this process is usually fast, for large piece of code as automatic provers it still takes some time. You can’t attribute all these time to the local helper theorems, whose proofs are short with just some rewrites. How many percentage of loading time is saved after your changes? Can you give us a clear metric?

digama0:
I will, as soon as I am done.

binghe:
For your regex-based source code manipulations (and then manually reverting some of them), this behavior may cause new bugs, because we don’t have enough function-level unit tests, and not every functionality of every SML modules in HOL are actually used during the building process. (Historically, e.g., many code were broken and fixed much later, by global changes related to loosed equality.)  Even those changes who claim to be formally verified (e.g. the cv kernel) ever introduces soundness bug (for a short period). How can you be sure that you absolutely cause no harm in these unnecessary code changes?

digama0:
I manually inspect the result and run the tests, same as any other refactoring or code change. If a bug comes up, we fix it.

I've already created and fixed a few bugs in the process here (mostly because of weird name shadowing caused by the overlapping of hol light and hol conventions in this file), so I'm confident that the bugs are being detected. But in general, I am getting tired of this obviously biased cross-examination of my motives in an attempt to reject the PR. Please respect my competence. Or maybe just actually read the PR if you would like to review it.

The timing data is in: previously it was one file (RealArith.sml) taking 9.2s, now it is two files taking 0.448s (`RealArith0.sml`) + 0.635s (`RealArith.sml`) = 1.083s, only 11% of the original time.

binghe:
There's no bias here (because I didn't try to reject all your PRs, and your work on VSCode support and many others contributions were appreciated).

mn200:
Thanks!

PR #1454, opened 2025-04-11
remove `prove` from RealArith.sml (part 2)

digama0:
Continuation of #1453. This splits RealArith into two parts, one which is used to construct the bootstrapping version of REAL_ARITH, and the other one which constructs the final interface. (The original API is unaffected.) Timing data shows a reduction from 9.2 s to 0.448s (`RealArith0.sml`) + 0.635s (`RealArith.sml`) = 1.083s.

### DIFF
--- a/src/real/RealArith.sml
+++ b/src/real/RealArith.sml
@@ -20,14 +20,12 @@ struct
 
 open HolKernel Parse boolLib liteLib
 
-open pairTheory pairLib reduceLib tautLib simpLib BasicProvers Ho_Rewrite
-     jrhUtils
-     Canon_Port AC prim_recTheory numTheory numLib numSyntax arithmeticTheory
+open reduceLib Ho_Rewrite Canon_Port AC RealArith0
 
 open normalForms  (* for HOL-Light's GEN_NNF_CONV, etc. *)
 open Normalizer   (* for HOL-Light's SEMIRING_NORMALIZERS_CONV *)
 
-open realaxTheory (* NOTE: cannot open realTheory here *)
+open realaxTheory real_arithTheory (* NOTE: cannot open realTheory here *)
 open Arbint       (* TODO: remove this in the future, using the default Int *)
 
 (*---------------------------------------------------------------------------*)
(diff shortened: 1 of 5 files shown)


binghe:
Adding new, small exported theories just for accelerating loading performance of other modules, is not the right direction. If without doing this the loading time (in your environment) will go from 1.083s to 1.5s, I think we shouldn't do it.

Breaking a non-theory SML module into two files, however, is fine in general if it's worthy doing that (that's why I didn't put `RealArith.sml` and `RealField.sml` together when porting a single piece of `REAL_ARITH` work from HOL-Light to HOL4.

digama0:
> Adding new, small exported theories just for accelerating loading performance of other modules, is not the right direction. If without doing this the loading time (in your environment) will go from 1.083s to 1.5s, I think we shouldn't do it.

Where did you get those numbers? It's 1.083s vs 9.2s. That's an additional 9 seconds on every single file downstream of real numbers, which is every file in cakeml. That's a very significant speedup.

I'm not sure why you asked for timing info when you clearly don't care about the answer. I believe we should do this consistently for every tactic, and Michael seems to agree. I think it's unfortunate that there is not a nicer way to do this currently where the theorems can be proved inline without every downstream file having to pay the cost, but based on the current design of HOL there is no other way. In addition to wasting time, these theorems also end up orphaned and inlined in every theory, which makes a mess of proof export. That's the actual reason I would like to pursue this, the performance improvement is just an added bonus.

binghe:
> > Adding new, small exported theories just for accelerating loading performance of other modules, is not the right direction. If without doing this the loading time (in your environment) will go from 1.083s to 1.5s, I think we shouldn't do it.
> 
> Where did you get those numbers? It's 1.083s vs 9.2s. That's an additional 8 seconds on every single file downstream of real numbers, which is every file in cakeml. That's a very significant speedup.

I assume your "9.2s" is the overall improvements after both PRs. I'm talking about the performance benefits of just the 2nd one.  In my view, the present PR creates more "damages" because a new user-visible exported theory is created. (The previous PR adds theorems with bad names into a rarely directly opened theory, is less damaging but still is.)

The "standard" way of handling this issue is to create a heap containing `RealArith` (and other needed SML modules). CakeML is a ML-dialect compiler, how is it possible that *every file* in CakeML needs real numbers? For those examples needing `RealArith`, they should create a heap (see `HOLDIR/src/real/analysis/Holmakefile` for an example). I don't believe CakeML people doesn't understand this.

> 
> I'm not sure why you asked for timing info when you clearly don't care about the answer. I believe we should do this consistently for every tactic, and Michael seems to agree. I think it's unfortunate that there is not a nicer way to do this currently where the theorems can be proved inline without every downstream file having to pay the cost, but based on the current design of HOL there is no other way. In addition to wasting time, these theorems also end up orphaned and inlined in every theory, which makes a mess of proof export. That's the actual reason I would like to pursue this, the performance improvement is just an added bonus.

There's no such policy that in non-theory SML modules calling `prove` to get some helper theorems is forbidden. It's common that, inside the complex decision procedure, some initial tactics are defined first, and then some intermediate theorems are proved by these tactics, and then more advanced tactics are defined, based on the intermediate theorems, and then more theorems, then more tactics. This case is more common in code ported from HOL-Light.

digama0:
> There's no such policy that in non-theory SML modules calling prove to get some helper theorems is forbidden.

That is precisely the topic of the [Zulip discussion](https://hol.zulipchat.com/#narrow/channel/486798-.E2.9D.84.EF.B8.8F-HOL4/topic/Why.20is.20startup.20slow.3F/near/510414596). A warning is not "forbidden" but it does indicate an intent to migrate away from the practice.

digama0:
> Are you going to "fix" all these modules by moving all the intermediate theorems into newly created intermediate exported theories?

Yes, but preferably putting them in existing theories if possible. The case in this PR is really exceptional, I think. Most tactics don't need a bootstrap phase.

PR #1455, opened 2025-04-12
remove `prove` from Cooper, Omega, intReduce

digama0:
`prove` refactor continued. This covers everything in `src/integer`.


mn200:
Thanks!

PR #1456, opened 2025-04-12
Case const

ordinarymath:


ordinarymath:
I think that I'll do that instantiates it in TypeBase and makes it simp in another PR

src/1/Prim_rec.sml
```diff
@@ -2039,6 +2039,29 @@ fun prove_case_ho_imp_thm ty_def = let
...
+  val ([x],cases) = strip_forall (concl nchotomy)
+  val x = mk_var("x",type_of x)
+  val y' = mk_var("y'",y_typ)
+  (*subst y to ensure freshness*)
+  fun mk_abs_y xs = list_mk_abs((map (subst [y |-> y']) xs),y)
+  val abs_y = strip_disj cases
+              |> map (mk_abs_y o #1 o strip_exists)
+  val goal = mk_eq (list_mk_comb(mk_comb(case_const,x), abs_y), y)
+in
+  (prove(goal,
```
mn200, line 2060: use `TAC_PROOF` rather than `prove` or change to forward proof.

examples/arm/v4/systemScript.sml
```diff
@@ -711,7 +711,7 @@ val DECODE_ARM_THM = store_thm("DECODE_ARM_THM",
                else
                  if b 4 then data_proc else data_proc`,
   SRW_TAC [boolSimps.LET_ss] [DECODE_ARM_def]
-    \\ FULL_SIMP_TAC (srw_ss()) [bool_case_ID]);
+    \\ FULL_SIMP_TAC (srw_ss()) [bool_case_CONST]);
```
mn200, line 714: This theorem should be in `srw_ss` already, so mentioning it again under either name should not be necessary.

mn200 (changes requested):

mn200:
Thanks!

PR #1437, opened 2025-03-26
Canonical probability spaces (from 1-dimensional Lebesgue measure space)

binghe:
Hi,

Note that most results in probability theory talk about properties of an existing probability space and random variables on it. There are few results letting you to construct a new probability space from more primitive materials. The following theorems give two canonical (infinite) probability spaces, generated by restricting the (household) 1-dimensional Lebesgue measure space to the interval [0,1] (reals or extended reals):
```
   [prob_space_lborel_01]  Theorem (probabilityTheory)
      ⊢ prob_space (restrict_space lborel (interval [(0,1)]))

   [prob_space_ext_lborel_01]  Theorem      
      ⊢ prob_space (restrict_space ext_lborel {x | 0 ≤ x ∧ x ≤ 1})
```
where the constant `restrict_space` is defined below (it keeps the same measure but intersecting every element of an existing measure space with a measurable set in it, resulting into a smaller measure space. It came from HVG people's prior work):
```
   [restrict_space]  Definition      
      ⊢ ∀M sp.
          restrict_space M sp =
          (sp ∩ m_space M,IMAGE (λa. a ∩ sp) (measurable_sets M),measure M)
```

Besides, I added the following theorem (cf. PR #1435) asserting the existence of a new probability from two (different) existing probability spaces:
```
   [existence_of_prod_prob_space]  Theorem      
      ⊢ ∀p1 p2.
          prob_space p1 ∧ prob_space p2 ⇒
          ∃p. prob_space p ∧
              ∀e1 e2.
                e1 ∈ events p1 ∧ e2 ∈ events p2 ⇒
                e1 × e2 ∈ events p ∧
                prob p (e1 × e2) = prob p1 e1 * prob p2 e2
```

Using properly these theorems, one can build a lot of new probability spaces (and independent r.v.'s on them).

--Chun

PR #1458, opened 2025-04-13
Existence of random variable having a given distribution

binghe:
Hi,

Given a probability space `p` and a random variable `X` in it, the distribution of `X` forms another probability space if we treat the distribution as a measure:
```
   [distribution_def]  Definition      
      ⊢ ∀p X. distribution p X = (λs. prob p (PREIMAGE X s ∩ p_space p))

   [distribution_prob_space]  Theorem      
      ⊢ ∀p X s.
          prob_space p ∧ sigma_algebra s ∧ random_variable X p s ⇒
          prob_space (space s,subsets s,distribution p X)
```
In fact, any function (:'a set -> extreal), if it satisfies the conclusion of the above theorem, i.e. `prob_space (space s,subsets s,distribution p X)` (for a σ-algebra `s`), then it must be a valid distribution function (of some r.v.'s).

The following "reverse" problem is hard (almost all new theorems in this PR are needed): _given any distribution, construct a random variable (and a probability space) having this distribution._ It turns out that this probability space can be a fixed one, the canonical one (`restrict_space ext_lborel {x | 0 ≤ x ∧ x ≤ 1}`) given in #1437. Thus it only remains to construct the random variable. The following theorem is now proved (and put at the beginning of `distributionTheory`):

```
existence_of_random_variable (distributionTheory)
⊢ ∀m p.
    prob_space (space Borel,subsets Borel,m) ∧
    p = restrict_space ext_lborel {x | 0 ≤ x ∧ x ≤ 1} ⇒
    ∃X. random_variable X p Borel ∧
        ∀s. s ∈ subsets Borel ⇒ distribution p X s = m s
```
The proof follows the idea found in [1] but many gaps were resolved without further references.

A very related concept (to distribution) is the "distribution function" of type `:extreal -> extreal`, by taking a half interval instead of full Borel-measurable set in the distribution set-function:
```
   [distribution_function_def]  Definition      
      ⊢ ∀p X.
          distribution_function p X =
          (λx. prob p ({w | X w ≤ x} ∩ p_space p))

   [distribution_function]  Theorem      
      ⊢ ∀p X t. distribution_function p X t = distribution p X {x | x ≤ t}
```
There's one-one correspondence between `distribution` and `distribution_function` for any random variable (but this is not yet proved). There's no supporting theorems currently in `probabilityTheory`. As a side effect, I proved that a distribution function is always mono-increasing and "right-continuous":
```
[distribution_function_monotone] (distributionTheory)
⊢ ∀p X f.
    prob_space p ∧ random_variable X p Borel ∧ f = distribution_function p X ⇒
    ∀x y. x ≤ y ⇒ f x ≤ f y

[distribution_function_right_continuous]
⊢ ∀p X f.
    prob_space p ∧ random_variable X p Borel ∧ f = distribution_function p X ⇒
    right_continuous f
```
where `right_continuous` has a simple definition (based on `inf` of extreal sets) below (it includes the mono-increasing, because this version only works for mono-increasing functions):
```
   [right_continuous_def]  Theorem (extrealTheory)
      ⊢ ∀f. right_continuous f ⇔
            (∀x y. x ≤ y ⇒ f x ≤ f y) ∧
            ∀x. inf {f x' | Normal x < x'} = f (Normal x)
```

Furthermore, I proved other analytic properties for general mono-increasing extreal functions. In particular, the set of discontinuous (jumping) points is countable (I thought this result were needed in the main existence proof but it turns out that it's not needed.):
```
   [jumping_point_of]  Definition      
      ⊢ ∀x0 f.
          x0 jumping_point_of f ⇔
          sup {f x | x < Normal x0} ≠ inf {f x | Normal x0 < x}

   [countable_jumping_point_of]  Theorem      
      ⊢ ∀f. (∀x y. x ≤ y ⇒ f x ≤ f y) ⇒
            countable {x | x jumping_point_of f}
```

P. S. The problem of constructing countably many independent random variables having the desired (distinct) distributions, is a fundamental problem in the theory of stochastic processes (Kolmogorov extension theorem [2]), which subsumes the present theorem (but is much harder to prove).

--Chun

[1] https://math.stackexchange.com/questions/1400327/existence-of-a-random-variable
[2] https://en.wikipedia.org/wiki/Kolmogorov_extension_theorem

### DIFF
--- a/examples/probability/distributionScript.sml
+++ b/examples/probability/distributionScript.sml
@@ -174,6 +174,9 @@ QED
 
    The corresponding probability space where X is constructed, is always the
    restricted ext_lborel, i.e. restrict_space ext_lborel [0,1].
+
+   This proof is based on idea from StackExchange [6] (answered by @Alexisz).
+   See also [7, p.61] for the only textbook mentioning a similar proof.
  *)
 Theorem existence_of_random_variable :
     !m p. prob_space (space Borel,subsets Borel,m) /\
@@ -1795,4 +1798,7 @@ val _ = html_theory "distribution";
   [4] Shiryaev, A.N.: Probability-1. Springer-Verlag New York (2016).
   [5] Schilling, R.L.: Measures, Integrals and Martingales (2nd Edition).
       Cambridge University Press (2017).
+  [6] math.stackexchange.com:
+      https://math.stackexchange.com/questions/1400327/existence-of-a-random-variable
+  [7] Resnick, S.: A Probability Path. Springer (2019).
  *)
--- a/examples/probability/distributionScript.sml
+++ b/examples/probability/distributionScript.sml
@@ -14,7 +14,7 @@
 open HolKernel Parse boolLib bossLib;
 
 open combinTheory arithmeticTheory numLib logrootTheory hurdUtils pred_setLib
-     pred_setTheory topologyTheory pairTheory tautLib jrhUtils;
+     pred_setTheory topologyTheory pairTheory tautLib jrhUtils cardinalTheory;
 
 open realTheory realLib seqTheory transcTheory real_sigmaTheory iterateTheory
      real_topologyTheory derivativeTheory;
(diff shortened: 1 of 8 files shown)


mn200:
Thanks for your work on this!

Issue #805, opened 2020-04-06
Remaining issues of HOL's OpenTheory packaging


binghe:
Following PR #803, I found the following remaining issues in OpenTheory packaging:

1. The recently added `itself` type in `boolTheory` introduced some new type, constant and theorems. When these theorems are used in other core theories, they become external constants and extra assumptions in the OpenTheory package, because the `hol-base` package doesn't package `boolTheory` at all, instead it tries to reuse what OpenTheory's own `base` package has, with a minimal fake bool theory defined in `src/boss/bool_defsScript.sml`. This mechanism can be seen in `src/boss/hol4-base.thy`:

```
bool {
  article: "bool_defs.ot.art"
  interpret: const "HOL4.bool_defs.LET" as "HOL4.bool.LET"
  interpret: const "HOL4.bool_defs.literal_case" as "HOL4.bool.literal_case"
  interpret: const "HOL4.bool_defs.IN" as "HOL4.bool.IN"
  interpret: const "HOL4.bool_defs.TYPE_DEFINITION" as "HOL4.bool.TYPE_DEFINITION"
  interpret: const "HOL4.bool_defs.ARB" as "HOL4.bool.ARB"

  interpret: type "HOL4.min.ind" as "ind"
  interpret: const "HOL4.min.@" as "select"
  interpret: const "HOL4.bool.!" as "Data.Bool.!"
  interpret: const "HOL4.bool.?" as "Data.Bool.?"
  interpret: const "HOL4.bool.?!" as "Data.Bool.?!"
  interpret: const "HOL4.bool.T" as "Data.Bool.T"
  interpret: const "HOL4.bool.F" as "Data.Bool.F"
  interpret: const "HOL4.bool./\\" as "Data.Bool./\\"
  interpret: const "HOL4.bool.\\/" as "Data.Bool.\\/"
  interpret: const "HOL4.min.==>" as "Data.Bool.==>"
  interpret: const "HOL4.bool.~" as "Data.Bool.~"
  interpret: const "HOL4.bool.COND" as "Data.Bool.cond"
  interpret: const "HOL4.bool.ONE_ONE" as "Function.injective"
  interpret: const "HOL4.bool.ONTO" as "Function.surjective"
}
```

[...]

PR #1459, opened 2025-04-13
rm `prove` in reduceLib, Normalizer, Grobner

digama0:
* [x] Depends on #1454
* [x] Depends on #1455


binghe:
Feel free to drop OpenTheory support (and delete the related CI building options).

digama0:
That should probably be a separate PR, but I'd be fine making this PR depend on it

mn200:
What is the issue with the OT build?  Is @digama0 aware of the way it has its own copy of (old) `reduce` code so that it doesn't need to use `cv_compute`?

digama0:
My ability to diagnose issues in the OT build is approximately zero. I do not understand how to maintain it at all.

binghe:
> What is the issue with the OT build? Is @digama0 aware of the way it has its own copy of (old) `reduce` code so that it doesn't need to use `cv_compute`?

The issue is that, after so many new small theories were created, there's no easy way to update existing `.thy` files which contains dependencies of all theories which generate OT articles. Previously the `prove` calls generate in-memory theorems whose proofs were written together with OT articles of theories calling them, now some of these theorems are stored as disk theorems in separate new theories, which has currently no OT articles.

The choice you made to merge these "removing prove" PRs, basically have killed the `otknl` build. If I'm the only "user", feel free to drop it because I don't want to spend huge efforts to fix it by updating all changed OT theory dependencies (while new similar PRs were still being created to make new breakages).

mn200:
It's not dropping it lightly; it's dropping a heavy great lump that is thankless and gross.

digama0:
To be clear, I'm happy to fix the issues in the OT build, since I wrote the PR that broke it in the first place. The problem is that it seems completely unmaintainable and there is zero documentation on how it works; the error message seems to be something like "I compiled 50 files and the set of theorems generated is not exactly the same as it used to be, please make unspecified changes somewhere in those files until the unspecified mismatch is fixed"

xrchz:
It's true that this is very poorly documented. There is some guidance to be gleaned from discussions like the comments in https://github.com/HOL-Theorem-Prover/HOL/pull/1375 and some of the past issues like https://github.com/HOL-Theorem-Prover/HOL/pull/1022 or https://github.com/HOL-Theorem-Prover/HOL/issues/805

binghe:
> It's true that this is very poorly documented. There is some guidance to be gleaned from discussions like the comments in #1375 and some of the past issues like #1022 or #805

#1051 is my favorite.

digama0:
For future reference: the issue is that OpenTheory maintains what amounts to a copy of the dependency tree of HOL inside `hol4-base-unint.thy`. This PR adds a new theory `reduce`, which means it needs a new theory section in the `.thy` file, and some (but not all) of the theories that HOL-depend on it must also (OT-)depend on `reduce`. The tool encourages precisely trimming the dependencies such that each theory only lists its exact dependencies, but this is brittle because a later proof change could result in another upstream theorem getting used. If you underestimate the dependencies (which is what happened in this case, since the new theory was not imported into the theories that use it), this is not an error, as one might hope; instead it simply adds the undefined theorem as an axiom and propagates it all the way to the end, where there is a list of 98(!) base axioms which are proved in numeric order(!). As a result, the whole setup ends up being extremely brittle to any change, and all errors are pushed to a final check which has completely lost the plot on why a mystery axiom has appeared. What a terrible design.

mn200:
Thanks for the doing the grunt work here!

PR #1464, opened 2025-04-17
make doDecl partially overridable in mkPushTranslatorCore

digama0:
Because `doDecl` is a function in a giant mutual block, it has a ton of arguments, but this setup lets you override the way a single case is rendered without having to copy the whole thing.


mn200:
Thanks!

PR #1468, opened 2025-04-24
[CI] Docker image returns to Debian stable (bookworm)

binghe:
Hi,

Today I try to update the Docker image for CI builds (`binghelisp/hol-dev:latest`) and found that PolyML 5.9.1 and master are no longer buildable with the default C++ compiler in Debian testing, and therefore I have returned to Debian stable (`bookworm`). But then I found that in Debian stable at least one of `z3` and `cvc5` packages doesn't exist, therefore we need to fallback to user-installed versions of these solvers.

By merging this PR, new CI builds should return to normal, otherwise there will be errors saying `/usr/bin/z3` doesn't exist.

--Chun

### DIFF
--- a/developers/docker-ci/Dockerfile
+++ b/developers/docker-ci/Dockerfile
@@ -26,13 +26,13 @@ ARG CVC_VERSION
 ARG YICES_VERSION
 
 # if --build-arg Z3_VERSION=anything, set HOL4_Z3_EXECUTABLE to system-installed z3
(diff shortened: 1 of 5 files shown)


ordinarymath:
I guess that's why #1469 fails

binghe:
> I guess that's why #1469 fails

Yes.  And someone has reported the compilation issue [1] to PolyML - that's exactly what I encountered (in both PolyML 5.9.1 and master).

[1] https://github.com/polyml/polyml/issues/214

mn200:
Thanks for this work!

PR #1469, opened 2025-04-25
Add implicit dependency to listScript

ordinarymath:


mn200:
If the dependency is really on `pred_setSimps`, then perhaps that should be what you mention in the header block rather than `pred_setLib`.

mn200:
Thanks for this!

PR #1472, opened 2025-04-28
Move concrete Moore-Smith covergence nets into netsTheory

binghe:
Hi,

Note that both `netsTheory` and a part of `real_topologyTheory` claim to implement the so-called "Moore-Smith covergence nets" [1], which unifies various concepts of limiting process of functions of type `:'a -> real` (typically `:'a` is either `:num`, the sequence limit or `:real` the normal function limit.)  I was wondering if the two definitions are equivalent (and thus many duplicated proofs can be merged). (NOTE: perhaps they are both works of John Harrison).

In this PR I moved those concrete Moore-Smith covergence nets (such as `at`, `within`, etc.) from `real_topologyTheory` (thus making this theory a bit smaller) to upstream `netsTheory` and connected the old (HOL4) and new (HOL-Light ported) concepts of "limit". As a side effect, many proofs in `limTheory` now becomes trivial as they duplicate with similar proofs in `real_topologyTheory`.

I proved that, the very general `limit` definition (ported from HOL-Light):
```
   [limit]  Definition      
      ⊢ ∀top f l net.
          limit top f l net ⇔
          l ∈ topspace top ∧
          ∀u. open_in top u ∧ l ∈ u ⇒ eventually (λx. f x ∈ u) net
```
and the old definition `tends` (in `netsTheory`):
```
   [tends]  Definition      
      ⊢ ∀s l top g.
          (s tends l) (top,g) ⇔
          ∀N. neigh top (N,l) ⇒ ∃n. g n n ∧ ∀m. g m n ⇒ N (s m)
```
are not equivalent but have the following relationship:
```
   [limit_alt_tends]  Theorem      
      ⊢ ∀top f l net.
          ¬trivial_limit net ∧ l ∈ topspace top ∧
          (∀x y. netord net x y ⇒ netord net y y) ⇒
          (limit top f l net ⇔ (f tends l) (top,netord net))
```
The antecedent `(∀x y. netord net x y ⇒ netord net y y)` is obtained by proof experiments. This property doesn't hold for all valid nets, but it indeed holds for the most important `at` net, therefore we have (unconditionally) the following important bridging result:
```
   [tendsto_real_alt_tends]  Theorem
      ⊢ ∀f l a. (f ⟶ l) (at a) ⇔ (f tends l) (mtop mr1,tendsto (mr1,a))
```

[1] Moore, E.H., Smith, H.L.: A General Theory of Limits. American Journal of
     Mathematics. 44, 102-121 (1922).


mn200:
Neat!  I'd be very surprised if this wasn't originally all by John Harrison.  His GitHub repo might reveal the history of how it played out in the HOL Light sources.

Issue #1476, opened 2025-04-30
Build failed both with moscowml and polyml


kumom:
I could not complete `poly < tools/smart-configure.sml`. The error message is as follows.
```
Poly/ML 5.9.1 Release

HOL smart configuration.

Determining configuration parameters: holdir OS poly GNUMAKE polymllibdir 
OS:                 macosx
poly:               /opt/homebrew/Cellar/polyml/5.9.1/bin/poly
polyc:              /opt/homebrew/Cellar/polyml/5.9.1/bin/polyc
polymllibdir:       /opt/homebrew/Cellar/polyml/5.9.1/lib
holdir:             /Users/kumom/Library/CloudStorage/Dropbox/projects/HOL
DOT_PATH:           SOME /opt/homebrew/bin/dot
MLTON:              NONE
GNUMAKE:            make

Configuration will begin with above values.  If they are wrong
press Control-C.

Will continue in 1 seconds.

Loading system specific functions
Note: Int.maxInt = SOME 4611686018427387903
Compiling system specific functions (sig sml)
Beginning configuration.
Removing old quotation filter from bin/
Making mllex
/opt/homebrew/Cellar/polyml/5.9.1/bin/polyc: line 45: clang++ -std=gnu++11: command not found
Failed to make mllex (Fail "Command failed")
```

I have `clang++` both installed with `xcode-select` and `brew`, and I have tested that `clang++ -std=gnu++11` works on my machine, so I don't know how to fix this error.

I switched to `moscowml`, and the configuration can be finished. However, when I executed `bin/build`, I got
```
Holmake: Compiling intReduceTheory.sig
Holmake: Compiling intReduceTheory.sml
Holmake: Compiling intReduce.sml
Holmake: Compiling CooperSyntax.sml
Holmake: Compiling CooperMath.sml
File "CooperMath.sml", line 28, characters 20-30:
!   | (h::t) => foldl Arbint.gcd h t
!                     ^^^^^^^^^^
! Unbound value component: Arbint.gcd
*** FATAL: Build failed in directory /Users/kumom/Library/CloudStorage/Dropbox/projects/HOL/src/parallel_builds/core
```

I installed both `moscowml` and `polyml` through `brew`. It seems that I am stuck here. What can I do to proceed?

ordinarymath:
Can you compile a sml file with polyc?

binghe:
[...]

PR #1477, opened 2025-04-30
add new Arbnum/Arbint functions

digama0:
Just the Arbnum/Arbint changes from #1459, which include a bug fix for mosml on develop reported in #1476.

### DIFF
--- a/src/portableML/mosml/Arbintcore.sig
+++ b/src/portableML/mosml/Arbintcore.sig
@@ -18,18 +18,21 @@ sig
   val toNat      : int -> num
   val toReal     : int -> real
 
-  val +          : (int * int) -> int
-  val -          : (int * int) -> int
-  val *          : (int * int) -> int
-  val div        : (int * int) -> int
-  val mod        : (int * int) -> int
-  val quot       : (int * int) -> int
-  val rem        : (int * int) -> int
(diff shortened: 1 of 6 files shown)


mn200:
Nice! Thanks.

binghe:
There are still issues when building HOL in Moscow ML. The next issue is at `src/integer/jrhCore.sml` by the following line:
```
val (Type,Term) = parse_from_grammars DeepSyntaxTheory.DeepSyntax_grammars
```
then the building process stops again at `src/rational/ratScript.sml`, at the call of `export_theory`, which I have no idea how to fix.

binghe:
> This is breakage from the adjoin-removal work in unused code (an experiment in deep embedding / “reflection” I did as an alternative approach to the existing Cooper algorithm which I discussed in the TPHOLs paper from 2003), but easily fixed. The Poly/ML build doesn't compile code unless it is actually loaded and used by a script file (or a selftest) so it snuck through.

I think we should at least open `jrhCore.sml` in `src/integer/selftest.sml` even without calling any function in it, because currently there are more issue in this file.

mn200:
Agreed.  I've done this in 56663e5a0351f7ae4b4dcd9bfa5a963610282b93

PR #1478, opened 2025-05-01
Add more pair simps

ordinarymath:


src/coretypes/pairScript.sml
```diff
@@ -580,6 +648,7 @@ val pair_CASE_def =
                  “pair_CASE (p:('a#'b)) f = f (FST p) (SND p)”)
 val _ = ot0 "pair_case" "case"
 
+
```
mn200, line 651: Don't add stray whitespace

src/coretypes/pairScript.sml
```diff
@@ -507,6 +561,20 @@ REWRITE_TAC [boolTheory.LET_DEF] THEN BETA_TAC
   THEN REWRITE_TAC [UNCURRY_VAR] THEN BETA_TAC
   THEN REWRITE_TAC[]);
 
+Theorem UNCURRY_RAND:
```
mn200, line 564: Please express in point-free style on right using combinators (S, K, etc) with LHSes of
```
UNCURRY N M b = ...
```
and 
```
P (UNCURRY N M)
```
then you get maximum applicability.

Rator one can be
```
Theorem UNCURRY_RATOR:
  UNCURRY N M b = UNCURRY (flip (flip o N) b) M
Proof
  rw[UNCURRY_VAR]
QED
```

mn200 (changes requested):
Looks nice!

src/coretypes/pairScript.sml
ordinarymath: I can't seem to write the RHS clearly in point free style
!f'. f' (UNCURRY f x) = UNCURRY (S (K (S (K f'))) (f)) x
This is what i came up with but looks incomprehensible

``\x y. (f' (f x y))`` is much clearer.
Maybe there's a point free form i'm not seeing

ordinarymath:
@binghe  you might want to kill the checks. Some tests aren't terminating

ordinarymath:
The offending lemmas that were not really good simp rules were removed

src/coretypes/pairScript.sml
mn200: Incomprehensible is fine, but you do need to use function composition:

```
   UNCURRY (\a b. P (f a b)) x 
= UNCURRY (\a. P o f a) x
= UNCURRY (\a. B P (f a)) x  (* B = (o) *)
= UNCURRY (B P o f) x 
= UNCURRY ((o) P o f) x
```

The composition over the abstraction that is presumably present in `f` is then rewritten with `combinTheory.o_ABS_R`.

The advantage of this approach is that it will work for `\(x,y,z). ...` and others; the originals won't.

mn200 (changes requested):

mn200 (approved):

PR #1479, opened 2025-05-02
Move obsolete ring theories out of core theories

binghe:
Hi,

This PR moves the obsolete ring theories (already renamed to `EVAL_*`) out of core theories, from `src/ring` into `examples/rings` where there's an example theory depending on them.

It turns out that the only outcome of those old ring theories is the tactic `INT_RING_TAC`, used by some proofs in `src/rational`.  But I have ported HOL-Light's `INT_RING` into `intLib` (based on `Normalizer.SEMIRING_NORMALIZERS_CONV`). It's easy to define a new tactic `intLib.INT_RING_TAC` from `INT_RING`:
```
val INT_RING_TAC = CONV_TAC (EQT_INTRO o INT_RING);
```
Fortunately, all existing proofs using the old `INT_RING_TAC` still work with the above new `INT_RING_TAC`. That's it.

--Chun


binghe:
The "job" of moving the call(s) of `prove()` out of the updated `intLib.sml`, is left to @digama0

mn200:
I think this has fallen foul of another PR; can you fix the merge conflicts please?

src/integer/intLib.sml
```diff
@@ -64,6 +64,84 @@ val ARITH_CONV = Omega.OMEGA_CONV
 val ARITH_TAC = Omega.OMEGA_TAC
 val ARITH_PROVE = Omega.OMEGA_PROVE
 
+(* ------------------------------------------------------------------------- *)
+(* Instantiate the normalizer (code below are ported from HOL-Light)         *)
+(* ------------------------------------------------------------------------- *)
+
+local
+  val sth = prove
```
digama0, line 72: was this code moved from somewhere else? I believe this was significantly refactored, can you make sure it is the new version that is moved?

digama0:
The PR title says "move" but it's +2321 -259, I don't see where most of the code is moved from

binghe:
> The PR title says "move" but it's +2321 -259, I don't see where most of the code is moved from

I guess this is a counting mistake by Git or GitHub when renamed files with modifications are involved. I mainly moved code from `src/ring` (- 2300 lines) to `examples/rings` (+ 2300 lines) and you haven't started to touch `src/ring` yet.

src/integer/intLib.sml
binghe: Your refactoring was in the just merged PR which I cannot depend on. It's moving theorems into the old `integerRingScript.sml`, but now this file has to move to examples, and the updated `intLib` need to provide `INT_RING_TAC` which depends on `INT_RING` which depends on `INT_POLY_CONV`, which depends on a `sth` theorem aka your `INT_POLY_CONV_sth` moved to examples. Since `intLib` cannot depend on code in `examples`, I found it's reasonable to do this entire work assuming your previous PR doesn't exist.

If you want, you can do this refactoring again. I don't know where I should put `INT_POLY_CONV_sth` into.

binghe:
NOTE: The above PR test failure is due to a bug in the just updated docker image where the path of `cvc5` is wrong (I was trying to update cvc5 versions with the new arm64 version installed). I'm still rebuilding the fixed docker image and will rerun the CI tests later.

binghe:
> The PR title says "move" but it's +2321 -259, I don't see where most of the code is moved from

Well, actually somehow I forgot to delete old files in "src/ring" ... This is done now.  (Please squash all commits to show the overall renaming effects) P.S. Docker image has been fixed.

mn200:
Thanks for this!

PR #1482, opened 2025-05-04
fix_IF_NONE_EQUALS_OPTION

ordinarymath:
This change removes IF_NONE_EQUALS_OPTION from the srw_ss and adds if_option_eq which has the last 2
conjuncts + 2 more other useful case splits.

### DIFF
--- a/examples/lambda/barendregt/boehmScript.sml
+++ b/examples/lambda/barendregt/boehmScript.sml
@@ -1058,7 +1058,8 @@ Proof
  >> reverse (Cases_on ‘solvable M’)
  >- rw [subterm_def, BT_of_unsolvables, ltree_el_def]
  (* stage work *)
- >> rw [subterm_def, BT_def, BT_generator_def, Once ltree_unfold, ltree_el_def]
+ >> rw [IF_NONE_EQUALS_OPTION,
+    subterm_def, BT_def, BT_generator_def, Once ltree_unfold, ltree_el_def]
  >> qabbrev_tac ‘M0 = principle_hnf M’
  >> qabbrev_tac ‘n  = LAMl_size M0’
  >> Q_TAC (RNEWS_TAC (“vs :string list”, “r :num”, “n :num”)) ‘X’
--- a/examples/lambda/barendregt/normal_orderScript.sml
+++ b/examples/lambda/barendregt/normal_orderScript.sml
@@ -188,7 +188,8 @@ val bnf_noposn = store_thm(
   "bnf_noposn",
   ``∀M. bnf M ⇔ (noposn M = NONE)``,
   HO_MATCH_MP_TAC simple_induction THEN
-  SRW_TAC [][] THEN Cases_on `noposn M` THEN SRW_TAC [][])
+  SRW_TAC [][] THEN Cases_on `noposn M` THEN
+  SRW_TAC [][EQ_IMP_THM])
 
 val normorder_noposn = store_thm(
   "normorder_noposn",
--- a/src/coretypes/optionScript.sml
+++ b/src/coretypes/optionScript.sml
@@ -329,6 +329,22 @@ val IF_EQUALS_OPTION = store_thm(
   SRW_TAC [][]);
 val _ = export_rewrites ["IF_EQUALS_OPTION"]
 
+Theorem if_option_eq[simp]:
+    (((if P then X else SOME x) = NONE) <=> P /\ (X = NONE)) /\
+    (((if P then SOME x else X) = NONE) <=> ~P /\ (X = NONE)) /\
+    (((if P then X else NONE) = SOME x) <=> P /\ (X = SOME x)) /\
+    (((if P then NONE else X) = SOME x) <=> ~P /\ (X = SOME x))
+Proof
+  OPTION_CASES_TAC“X:'a option” THEN SRW_TAC [](option_rws)
+QED
+
+Theorem if_option_neq[simp]:
+    (((if P then X else NONE) ≠ NONE) <=> P /\ (X ≠ NONE)) /\
+    (((if P then NONE else X) ≠ NONE) <=> ~P /\ (X ≠ NONE))
+Proof
+  OPTION_CASES_TAC“X:'a option” THEN SRW_TAC [](option_rws)
+QED
+
 val IF_NONE_EQUALS_OPTION = store_thm(
   "IF_NONE_EQUALS_OPTION",
   ``(((if P then X else NONE) = NONE) <=> (P ==> IS_NONE X)) /\
@@ -336,7 +352,6 @@ val IF_NONE_EQUALS_OPTION = store_thm(
     (((if P then X else NONE) = SOME x) <=> P /\ (X = SOME x)) /\
     (((if P then NONE else X) = SOME x) <=> ~P /\ (X = SOME x))``,
   OPTION_CASES_TAC“X:'a option” THEN SRW_TAC [](option_rws));
-val _ = export_rewrites ["IF_NONE_EQUALS_OPTION"]
 
 (* ----------------------------------------------------------------------
     OPTION_MAP theorems
--- a/src/pred_set/src/more_theories/wellorderScript.sml
+++ b/src/pred_set/src/more_theories/wellorderScript.sml
@@ -595,15 +595,15 @@ val wo2wo_EQ_NONE = store_thm(
   ``!x. (wo2wo w1 w2 x = NONE) ==>
         !y. (x,y) WIN w1 ==> (wo2wo w1 w2 y = NONE)``,
   ONCE_REWRITE_TAC [wo2wo_thm] >> rw[LET_THM] >>
-  match_mp_tac SUBSET_ANTISYM >> rw[IMAGE_wo2wo_SUBSET] >>
-  match_mp_tac SUBSET_TRANS >>
-  qexists_tac `IMAGE THE (IMAGE (wo2wo w1 w2) (iseg w1 x) DELETE NONE)` >>
-  conj_tac >- (
-    Cases_on`wleast w2 (IMAGE THE (IMAGE (wo2wo w1 w2) (iseg w1 x) DELETE NONE))` >>
-    imp_res_tac wleast_EQ_NONE >> fs[] ) >>
-  simp_tac (srw_ss() ++ DNF_ss) [SUBSET_DEF] >>
-  qsuff_tac `!a. a IN iseg w1 x ==> a IN iseg w1 y` >- metis_tac[] >>
-  rw[iseg_def] >> metis_tac [WIN_TRANS]);
+  fs[IMAGE_wo2wo_SUBSET,SET_EQ_SUBSET]
+  >- (
+    qsuff_tac `elsOf w2 ⊆ IMAGE THE (IMAGE (wo2wo w1 w2) (iseg w1 y) DELETE NONE)` >- fs[] >>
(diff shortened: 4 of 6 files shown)


ordinarymath:
I'm skeptical of the usefulness of the first 2 rewrites so removing them for now.

ordinarymath:
It seems that otknl is some how not terminating. Can someone with permission to kill jobs kill it. Also @mn200 can I get permissions to kill jobs as it seems like this is happening frequency enough.

ordinarymath:
Oh it finished running.

src/pred_set/src/more_theories/wellorderScript.sml
```diff
@@ -4,6 +4,8 @@ open boolSimps mesonLib numLib InductiveDefinition tautLib
 open relationTheory set_relationTheory pred_setTheory pairTheory
      arithmeticTheory optionTheory;
 
+local open newtypeTools in end
```
mn200, line 7: Please don't put unrelated changes into the PR.

src/coretypes/optionScript.sml
```diff
@@ -329,14 +329,29 @@ val IF_EQUALS_OPTION = store_thm(
...
+Theorem if_option_eq[simp]:
+    (((if P then X else SOME x) = NONE) <=> P /\ (X = NONE)) /\
+    (((if P then SOME x else X) = NONE) <=> ~P /\ (X = NONE)) /\
+    (((if P then X else NONE) = SOME x) <=> P /\ (X = SOME x)) /\
+    (((if P then NONE else X) = SOME x) <=> ~P /\ (X = SOME x))
+Proof
+  OPTION_CASES_TAC“X:'a option” THEN SRW_TAC [](option_rws)
+QED
+
+Theorem if_option_neq[simp]:
```
mn200, line 341: Please replace `P <> NONE` with `?y. P = SOME y`

examples/lambda/barendregt/normal_orderScript.sml
```diff
@@ -188,7 +188,8 @@ val bnf_noposn = store_thm(
   "bnf_noposn",
   ``∀M. bnf M ⇔ (noposn M = NONE)``,
   HO_MATCH_MP_TAC simple_induction THEN
-  SRW_TAC [][] THEN Cases_on `noposn M` THEN SRW_TAC [][])
+  SRW_TAC [][] THEN Cases_on `noposn M` THEN
+  SRW_TAC [][] THEN EQ_TAC THEN SRW_TAC[][])
```
mn200, line 192: Seems bad.  Would `SRW_TAC[][EQ_IMP_THM]` work?

mn200 (changes requested):

src/pred_set/src/more_theories/wellorderScript.sml
ordinarymath: newtypeTools in being used below

ordinarymath: see https://github.com/HOL-Theorem-Prover/HOL/blob/d1ef97c352b4fe055e0e9238469c291f82c93b4c/src/pred_set/src/more_theories/wellorderScript.sml#L78

mn200: Yeah, and this occurrence deeper in the file is legal, though bad style.  You're welcome to make a separate PR for that if you wish.

src/coretypes/optionScript.sml
ordinarymath: This change breaks more stuff. I'll try to find time this weekend to fix the breakage

ordinarymath: The breakage does not seem easy to fix for boehmScript.sml. There's a custom tactics like HNF_TAC breaking working. Also putting temp_delsimps doesn't really work due to tactics using std_ss directly which if_option_neq is part of.

ordinarymath: Oh i found out why HNF_TAC wasn't working.

ordinarymath: It should be fixed now

ordinarymath:
@mn200 ping on this?

examples/lambda/barendregt/boehmScript.sml
```diff
@@ -1565,9 +1568,9 @@ Proof
       by METIS_TAC [subterm_disjoint_lemma']
  (* NOTE: the next two HNF_TAC will refine M1 and M1' *)
  >> Q_TAC (HNF_TAC (“M0 :term”, “vs :string list”,
-                    “y :string”, “args :term list”)) ‘M1’
+                    “y'' :string”, “args :term list”)) ‘M1’
```
mn200, line 1571: This many primes is a sure sign that a call to `rename` should be being used.  Relying on generated names is very bad style.

mn200 (changes requested):

examples/lambda/barendregt/boehmScript.sml
binghe: I think the fact that existing proofs are broken at many places and get longer, is a sure sign that the proposed changes in related simp sets is a bad idea and should not be merged.

binghe: The symbol names changed because he replaced my `simp[]` with `fs [..]`, which causes unexpected changes between assumptions (which are sometimes over 200 in later position of `boehmTheory`).   In general, replacing `simp[]` in other's proof with `fs[]` is wrong, and shouldn't be considered as the correct way of fixing others' proofs.

ordinarymath: The symbol names are from this change
> Please replace `P <> NONE` with `?y. P = SOME y`
where y gets generated.
y' is not generated but a fresh var is needed for HNF_TAC which i used y'

examples/lambda/barendregt/boehmScript.sml
```diff
@@ -1565,9 +1568,9 @@ Proof
...
  >> Q_TAC (HNF_TAC (“M0':term”, “vs :string list”,
-                    “y' :string”, “args':term list”)) ‘M1'’
+                    “y''' :string”, “args':term list”)) ‘M1'’
```
binghe, line 1573: The name `y'` was available here, and it's supposed to correspond to `y`.  In my name design, `y` is the head variable of  `M0`, and `y'` is the head variable of `M0'`, which is perfect. Now this bad `y'''` jump out, because somehow `y'` and `y''` occurred from nowhere but changed simplification rules. Shit.

examples/lambda/barendregt/boehmScript.sml
mn200: Did you also choose `y'''` ?  If so, I guess that's fine, though I am not a fan of so many prime symbols.  

It sounds as if, after the call to `fs` you should rename to fix the generated `y'`, allowing the rest of the proof to go through successfully.

binghe: Then you should use `rename` to rename those internally generated `y`s to something else, immediately after they were generated. There could be other (unknown) user proof using `y` for other purposes, and your change will break those proof too.

examples/lambda/barendregt/boehmScript.sml
ordinarymath: Another slightly less obvious change is Overload subterm' = “\X M p r. FST (THE (subterm X M p r))”
does not print as often as
`subterm X M p r <> NONE` before would be simplified in some cases to.
`subterm X M p r = SOME y`
and i get
`FST y` appearing instead

binghe: Both `NONE` and `SOME` are primitive constants provided by `optionTheory`. I think it's very **rude** to automatically rewrite all `a <> NONE` to `a = SOME y` with a new, fixed name `y` introduced. I keep those `_ <> NONE` terms only because I don't care which `SOME y` it equals to.

binghe:
I actually have a _blacklist_ of simplification rules in front of my `boehmTheory`. Some are specific to the related work to lambda-calculus, some are general for all my proofs:

```
(* These theorems usually give unexpected results, should be applied manually *)
val _ = temp_delsimps [
   "lift_disj_eq", "lift_imp_disj",
   "IN_UNION",     (* |- !s t x. x IN s UNION t <=> x IN s \/ x IN t *)
   "APPEND_ASSOC", (* |- !l1 l2 l3. l1 ++ (l2 ++ l3) = l1 ++ l2 ++ l3 *)
   "SNOC_APPEND"   (* |- !x l. SNOC x l = l ++ [x] *)
];
```

If this PR were merged, I will later revert the changed proofs by putting `if_option_eq` and `if_option_neq` into this list, instead. So I give up, just do whatever you feel like.

mn200:
As *per* [Zulip discussion](https://hol.zulipchat.com/#narrow/channel/486798-.E2.9D.84.EF.B8.8F-HOL4/topic/IF_NONE_EQUALS_OPTION.20contains.20bad.20rewrites.2E/with/515864493), I think the best way forward is to avoid automatic simps that generate (existentially-quantified) names, so that `(?y. x = SOME y)` gets replaced with `x <> NONE` (in the `option_neq` rewrite).

examples/lambda/barendregt/boehmScript.sml
```diff
@@ -4829,7 +4829,7 @@ Proof
  >> DISCH_TAC
  >> Q.PAT_X_ASSUM ‘subterm X M (h::p) r <> NONE’ MP_TAC
  >> Q_TAC (UNBETA_TAC [subterm_alt]) ‘subterm X M (h::p) r’
- >> STRIP_TAC
+ >> STRIP_TAC >> fs[]
```
binghe, line 4832: Please revert all your changes in this file. I don’t see how your changed/added tactics are necessary. Even the proof still work, the proof status may become unreadable after your modifications.

ordinarymath: You're right that the added fs[] isn't needed it must have broke initially when i did not have if_option_neq added when I initially did the PR.

binghe: Fortunately it's NOT in the 4300-line big proof, which is not `fs`-safe (but `rfs`-safe) by design. In this proof, your `fs []` changed assumption 0 from
```
    0.  !M r.
          FINITE X /\ FV M SUBSET X UNION RANK r /\
          subterm X M p r <> NONE /\ v # M /\ v IN X UNION RANK r ==>
          v # subterm' X M p r
```
to
```
    0.  !M' r'.
          FV M' SUBSET X UNION RANK r' /\ subterm X M' p r' <> NONE /\
          v # M' /\ v IN X UNION RANK r' ==>
          v # subterm' X M' p r'
```
where some subterms are eliminated by other assumptions (which is unnecessary because the next tactics `FIRST_X_ASSUM MATCH_MP_TAC >> ASM_REWRITE_TAC []` will handle them too).  However, the changed names of quantifiers (because they are fixed symbols in other assumptions, a strange behaviour of `FULL_SIMP_TAC`) makes the assumption harder to read - unnecessary primed names are created here. I always avoid using `FULL_SIMP_TAC` in the trunk of my proofs when there are potential name conflicts between free and bound variables in assumptions.

In many other proofs of this file, where the custom tactic `HNF_TAC` is used to decompose λ-terms to their head normal forms, many equations and abbreviations are created as assumptions, a call of `FULL_SIMP_TAC` which rewrites older assumptions by newer assumptions, will cause unexpected redundancies when reading those assumptions. `REV_FULL_SIMP_TAC` is safe, however, in these proofs.

PR #1485, opened 2025-05-06
[set_relation] rename short constant names (sc -> symmetric_closure, tc -> transitive_closure)

binghe:
Hi,

The short constant name "sc" added in `set_relationTheory` (by #1473) is now renamed to longer name "symmetric_closure". Similarly, the existing "tc" is renamed to "transitive_closure". To minimise the incompatibilities, I still keep all the existing theorem names unchanged, and use local overloads when building `set_relationTheory`.   Some theorem statements about the newly added "sc" are fixed, following their "tc" siblings.

In theory, existing proofs using `tc` explicitly can simply add the following overload to fix their broken code:
```
Overload tc[local] = “transitive_closure”
```

P. S. I wrongly pushed my branch to official repository, but maybe this is good for people who wants to merge this branch directly to branches other than `develop`.

--Chun

### DIFF
--- a/doc/next-release.md
+++ b/doc/next-release.md
@@ -86,6 +86,8 @@ Incompatibilities:
 
 -   `util_probTheory` has been merged into `sigma_algebraTheory`.
 
+-   In `set_relationTheory`, the constant `tc` has been renamed to `transitive_closure`.
+
 * * * * *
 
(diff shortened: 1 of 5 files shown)


mn200:
Thanks for this quick fix!

PR #1487, opened 2025-05-07
[cheatsheet] Fix a couple of underscores

hrutvik:
These were being converted into `<em>` tags.
Use HTML escape codes instead.


mn200:
This is fine, but I think the right approach in the longer term is to dispense with the backticks as HOL-delimiters, and use the fancy quotes.  Then HOL material can be rendered fine within markdown's backticks:
```
  `qspecl_then [‘a’] mp_tac theorem`
```
will work and markdown won't see the underscores in `qspecl_then` and `mp_tac` as anything special.

PR #1488, opened 2025-05-08
Apply a patch to lem_relation before Holmake runs

myreen:


mn200:
We can merge to `mcandidate` if you really want, but would it be enough to go for `develop` and then just go for a fresh `mcandidate` based there?

(Yes, if `develop` races ahead of `mcandidate` we clearly do want something that is a bit like a release branch, but I hope that most of the time they'll be close enough that we can avoid the split&remerge in the history.)

tanyongkiam:
If you create such an `mcandidate` soon-ish, I can run the regression right away. A couple of PRs on CakeML are blocked due to the `mcandidate` failure

myreen:
I'm opening a new PR forthis going into develop.

PR #1492, opened 2025-05-09
[sorting] Add greater_sorted_eq

dnezam:
This theorem was useful for reasoning about a descending list of numbers, which I use to reason about the keys of a map from variable names to unique integers. Since `less_sorted_eq` exists, having a version for `greater` may be a useful addition. If not, feel free to close.


mn200:
Thanks!

PR #1493, opened 2025-05-10
add fun-op coroutine example

adamstucci:
Simple functional big step semantics language demonstrating coroutines


mn200:
- You need to put a reference to this new directory into the file 
`src/parallel_builds/core/Holmakefile` so that it is built as part of the regression machinery.  Assuming it's pretty quick, it could go into the `EXDIRS` variable.

- Could you add a `README` to the directory explaining the basic approach?  (Or a big comment at the head of the file would be OK too.)

mn200:
Thanks! I think the errors are unrelated to your work.

Issue #1489, opened 2025-05-08
Regression in `qpat_abbrev_tac`


hrutvik:
The following example produces an error:
```
Theorem test:
  (λ(x,_). x) y = foo a
Proof
  pairarg_tac >>
  qpat_abbrev_tac ‘bar = foo _’ >>
  cheat
QED
```
This is related to https://github.com/HOL-Theorem-Prover/HOL/commit/ba1eb9a0ee80cd83cc632d3c2cb1c14668b879c1 (and https://github.com/HOL-Theorem-Prover/HOL/commit/7b64934d6780a55fbd0b70ad34aacd2b3833ea63 linked in the commit message). In particular:
- the underscore used as a bound variable becomes `_0`
- `pairarg_tac` brings free variable `_0` into scope
- parsing of `qpat_abbrev_tac` fails as it names its input underscore as `_0` too, which clashes with the one free in the goal
This causes a type conflict.

Discussion with @mn200 and @konrad-slind on [Zulip](https://hol.zulipchat.com/#narrow/channel/486798-.E2.9D.84.EF.B8.8F-HOL4/topic/Potential.20issue.20with.20parsing.20fresh.20underscores/near/512056833) suggests that this is an issue in `qpat_abbrev_tac`, which is failing to account for the new handing of underscores.

PR #1496, opened 2025-05-11
Try to handle underscores in `Parse.parse_in_context`

hrutvik:
Expose and use the `ctxt_absyn_to_preterm` function created in https://github.com/HOL-Theorem-Prover/HOL/commit/ba1eb9a0ee80cd83cc632d3c2cb1c14668b879c1

Aims to fix https://github.com/HOL-Theorem-Prover/HOL/issues/1489


mn200 (approved):

mn200:
Thanks for this!

mn200:
It strikes me that it would be nice to have a regression test for this...

PR #1499, opened 2025-05-13
[vim] Fix attribute highlighting for `Definition` and `Triviality`

hrutvik:
Thanks to @ordinarymath for the bug report.


mn200:
Thanks!

PR #1501, opened 2025-05-16
More (general) topology ported from HOL-Light

binghe:
Hi,

This PR contains some more general topology concepts and related theorems, ported from HOL-Light. It's an endless job of porting useful stuff from HOL-Light (or, more generally, from other theorem provers) but the new stuff here will be used later to support probability theory (where the properties of weak convergence of measure functions are to be established in general metric space.)

In topology, a **closure** of a set is the set together with all its _limit points_. The actual definition, however, involves only `topspace` and `open_in`, two fundamental concepts in topology:
```
   [closure_of]  Definition      
      ⊢ ∀top s.
          top closure_of s =
          {x |
           x ∈ topspace top ∧ ∀t. x ∈ t ∧ open_in top t ⇒ ∃y. y ∈ s ∧ y ∈ t}
```
The **interior** of a set is the largest open sets it contains:
```
   [interior_of]  Definition      
      ⊢ ∀top s. top interior_of s = {x | ∃t. open_in top t ∧ x ∈ t ∧ t ⊆ s}
```
The **frontier** of a set is its closure subtracts its interior:
```
   [frontier_of]  Definition      
      ⊢ ∀top s. top frontier_of s = top closure_of s DIFF top interior_of s
```
With these general definitions, the existing `closure`, `interior` and `frontier` definitions in `real_toplogy` now becomes an instance of these general definitions in (1-dimension) Euclidean space (and the original definitions now become theorems with the same statements as before, therefore didn't break existing proofs):
```
   [closure_def]  Definition
      ⊢ ∀s. closure s = euclidean closure_of s

   [interior_def]  Definition
      ⊢ ∀s. interior s = euclidean interior_of s

   [frontier_def]  Definition
      ⊢ ∀s. frontier s = euclidean frontier_of s
```
I basically ported all related theorems (not too many, all with easy proofs) from HOL-Light.

Besides, more supporting theorems for the previous ported `continuous_map` (the most general definition of continuous functions) were ported too.

Chun

### DIFF
--- a/src/pred_set/src/more_theories/topologyScript.sml
+++ b/src/pred_set/src/more_theories/topologyScript.sml
@@ -13,8 +13,7 @@
 (*              (c) Copyright, Marco Maggesi 2012-2017                       *)
 (*             (c) Copyright, Andrea Gabrielli 2012-2017                     *)
 (* ========================================================================= *)
-(*  Formalization of general topological and metric spaces in HOL Light      *)
-(*                     (from hol-light's metric.ml)                          *)
+(*  General topological and metric spaces (from hol-light's metric.ml)       *)
 (*                                                                           *)
 (*              (c) Copyright, John Harrison 1998-2017                       *)
 (*                (c) Copyright, Marco Maggesi 2014-2017                     *)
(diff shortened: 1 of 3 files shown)


mn200:
Thanks!

PR #1503, opened 2025-05-16
Upstream theorems from Verifereum

xrchz:
(from Verifereum)

### DIFF
--- a/src/list/src/listScript.sml
+++ b/src/list/src/listScript.sml
@@ -1082,7 +1082,7 @@ val LENGTH_TL = Q.store_thm
   ‘!l. 0 < LENGTH l ==> (LENGTH (TL l) = LENGTH l - 1)’,
   Cases_on ‘l’ THEN SIMP_TAC arith_ss [LENGTH, TL]);
 
-Theorem LENGTH_TL_LESS_EQ:
+Theorem LENGTH_TL_LE:
   !ls. LENGTH (TL ls) <= LENGTH ls
 Proof
   Cases \\ rw[]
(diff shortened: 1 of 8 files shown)


src/list/src/listScript.sml
```diff
@@ -1076,6 +1082,12 @@ val LENGTH_TL = Q.store_thm
   ‘!l. 0 < LENGTH l ==> (LENGTH (TL l) = LENGTH l - 1)’,
   Cases_on ‘l’ THEN SIMP_TAC arith_ss [LENGTH, TL]);
 
+Theorem LENGTH_TL_LESS_EQ:
```
mn200, line 1085: Please use `LE` suffix to stand for less than or equal.

src/list/src/listScript.sml
```diff
@@ -2377,6 +2389,18 @@ Proof
   \\ rw[UNCURRY, arithmeticTheory.ADD1]
 QED
 
+Theorem ALL_DISTINCT_MAP_DROP_LESS:
```
mn200, line 2392: Seems too special-case.

mn200 (changes requested):

src/list/src/listScript.sml
xrchz: Done

src/list/src/listScript.sml
xrchz: Removed.

mn200 (approved):

mn200:
I believe the regression failures are an independent problem, so merging.  Thanks!

PR #1504, opened 2025-05-16
Add selftest for `qpat_abbrev_tac`

hrutvik:
As per @mn200's comment in https://github.com/HOL-Theorem-Prover/HOL/pull/1496, add a selftest aiming to mimic the original bug report (https://github.com/HOL-Theorem-Prover/HOL/issues/1489)

### DIFF
--- a/src/q/selftest.sml
+++ b/src/q/selftest.sml
@@ -435,6 +435,11 @@ val _ = require (check_result (fn _ => true))
                  REWRITE_TAC [])
                 ([], ``p /\ T = p``)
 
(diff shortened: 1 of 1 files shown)


mn200:
I'm afraid you will have to fake your pair type or so something similar when testing in `src/q`.

hrutvik:
I suppose the pair is sort of irrelevant, in that the shortcoming was a failure to account for numbered underscores that are free in the goal. I've adapted the test accordingly.

mn200:
Thanks!

PR #1508, opened 2025-05-25
[lambda] Refactoring for next steps

binghe:
Hi,

This is a big refactoring of the lambda/barendregt for potential next steps. It shows so many changed lines for 3 reasons:

1. The previous wrong name (an embarrassing typo) `principle_hnf` has been corrected to `principal_hnf`.
2. The big proof of 4300 lines (`subtree_equiv_lemma_explicit`) has been moved from `boehmTheory` to `lameta_completeTheory`.
3. The eta-expansion idea (proving `lameta_complete` by a restricted version of `separability_thm` requiring two terms have the same set of paths in their Böhm trees), has been moved from `lameta_completeTheory` to `boehmTheory`.

With these changes, now the new `boehmTheory` stands as a shared upstream theory for the updated `lameta_completeTheory` and a potential new theory in the future which tries to re-prove the `lameta_complete` by a shorter proof without using the `equivalent` relation introduced by C. Böhm.  In another word, we are finding a new shorter proof for this important result.

Almost all theorems remain their statements in this PR, but the following lemma has some minor changes:
```
[ltree_paths_BT_expand]
⊢ ∀X M p r m.
    FINITE X ∧ FV M ⊆ X ∪ RANK r ∧ bnf M ∧ p ∈ ltree_paths (BT' X M r) ∧
    ltree_branching (BT' X M r) p = SOME m ⇒
    ltree_paths (BT_expand X (BT' X M r) p r) =
    ltree_paths (BT' X M r) ∪ {SNOC m p}
```
Previously the final part of the conclusion was `... = SNOC m p INSERT ltree_paths (BT' X M r)`, and the (paper) reviewer says it's hard to read it without parenthesis, e.g. `(SNOC m p) INSERT ...` which is type-unnecessary. Now I use `UNION` (of a singleton set) instead.

--Chun

### DIFF
--- a/examples/lambda/barendregt/README.md
+++ b/examples/lambda/barendregt/README.md
@@ -49,9 +49,16 @@ and
   In "Higher Order and Symbolic Computation", 19:169-195, 2006.
 
 solvableScript.sml :
+      solvability of lambda-terms; Wadsworth's theorem (solvable iff has_hnf)
+    
 boehmScript.sml :
+      Effective Boehm trees
+
 lameta_completeScript.sml :
+      Hilbert-Post Completeness of Lambda-Eta-Calculus
+
 takahashiS3Script.sml :
+      Section 3 of Takahashi's paper (has_benf iff has_bnf)
 
 These files are new work.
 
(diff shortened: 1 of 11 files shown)


binghe:
The PR test failure is due to issues `develop`; The previous fixes due to rule changes in `optionTheory` has been already incorporated.

PR #1514, opened 2025-06-03
[lambda] Initial semi_sensibleTheory

binghe:
Hi,

This PR continues the previous work on the lambda-calculus examples and slightly move forward.

I start with the existing `chap4Theory`, where there's the definition for `lambdathy`:
```
[lambdathy_def] (chap4Theory, existing)
⊢ ∀A. lambdathy A ⇔ consistent A⁺ ∧ UNCURRY A⁺ = A
```
I have proved that both `lameq` (`==`) and `lameta` satisfy the above definition:
```
[lambdathy_lameq]
⊢ lambdathy (UNCURRY $==)
[lambdathy_lameta]
⊢ lambdathy (UNCURRY $===)
```
Based on this `lambdathy`, I have added, in the new `semi_sensibleTheory`, the following `gen_separable` definition for the general concept of separability in a lambda theory (following Barendregt 10.4.4):
```
[gen_separable_def] (semi_sensibleTheory)
⊢ ∀R Ms.
    gen_separable R Ms ⇔
    lambdathy R ∧
    (let
       X = BIGUNION (IMAGE FV (set Ms));
       Fs = MAP (λt. LAMl (SET_TO_LIST X) t) Ms
     in
       ∀Ns.
         LENGTH Ns = LENGTH Ms ⇒
         ∃f. ∀i. i < LENGTH Ms ⇒ R (f @@ EL i Fs,EL i Ns))
```
Note two things here: 1) the list of terms `Fs` are closures of input term `Ms` using the union of all their free variables as the  binding variables (but the order is arbitrary chosen by `SET_TO_LIST`). 2) This general separation definition only asserts a single lambda term `f` instead of Böhm transformation or contexts, which is easier to understand.

If using contexts (`ctxt`) instead, there's the following simpler alternative definition for `gen_separable`:
```
[gen_separable_alt]
⊢ ∀R Ms.
    gen_separable R Ms ⇔
    lambdathy R ∧
    ∀Ns.
      LENGTH Ns = LENGTH Ms ⇒
      ∃c. ctxt c ∧ ∀i. i < LENGTH Ms ⇒ R (c (EL i Ms),EL i Ns)
```

The usual (beta)-separability is just `gen_separable` parameterised with `lameq` and eta-separability is parameterised with `lameta`:
```
Overload separable = “gen_separable (UNCURRY (==))”
Overload eta_separable = “gen_separable (UNCURRY lameta)”
```

With all these new devices, the version of Böhm's separability theorem we obtained previously, can be restated below:
```
[eta_separable_thm]
⊢ ∀M N. has_benf M ∧ has_benf N ∧ ¬(M === N) ⇒ eta_separable [M; N]
```

On the other hand, the usual (beta)separability implies incompatiblity (the other direction is not true):
```
[separable_incompatible]
⊢ ∀M N. separable [M; N] ⇒ M # N
```

Our goal here is to prove `|-? has_benf M ∧ has_benf N ∧ ¬(M === N) ⇒ separable [M; N]` (The original Böhm's separability theorem) or `|-? has_benf M ∧ has_benf N ∧ ¬(M === N) ⇒ M # N`. And this is true by the following unproved results:

```
|-? eta_separable [M; N] <=> separable [M; N]
```
or in general, if a theory T is "semi-sensible" (the definition is already in `chap4Theory`), then T-separability is equivalent to T-eta-separability.

--Chun

### DIFF
--- a/examples/lambda/barendregt/boehmScript.sml
+++ b/examples/lambda/barendregt/boehmScript.sml
@@ -345,7 +345,8 @@ Overload bot = “(NONE, SOME 0) :(BT_node # num option)”
 
 (* Unicode name: "base" *)
 val _ = Unicode.unicode_version {u = UTF8.chr 0x22A5, tmnm = "bot"};
-val _ = TeX_notation {hol = "bot", TeX = ("\\ensuremath{\\bot}", 1)};
+val _ = TeX_notation {hol = "bot", TeX = ("\\HOLTokenBottom}", 1)};
+val _ = TeX_notation {hol = UTF8.chr 0x22A5, TeX = ("\\HOLTokenBottom", 1) };
 
 Theorem BT_of_unsolvables :
     !X M r. unsolvable M ==> BT' X M r = bot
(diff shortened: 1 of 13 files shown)


mn200:
Thanks!  These look like nice abstractions to have in our hands.

PR #1515, opened 2025-06-03
Remove polyc in system path dependency

ordinarymath:


mn200:
Nice catch; thanks!

binghe:
This fix explains why some of my builds failed strangely if I use a `polyc` from a specific old version (installed in specific directory) while I have global `polyc` as the latest version.

PR #1516, opened 2025-06-04
[OT] fix for core dependency changes (of pairTheory) and one leaking assumption

binghe:
This PR should fix the OT builds (up to `src/boss` (`hol-base`)) due to core dependency changes after some termination proofs become automatic.

I guess the reason is that `pair$LEX` is a common tool in proving the termination of recursive functions of multiple arguments, and the auto-generated termination proofs now unconditionally need it, therefore many core theories now implicitly depends on `pairTheory`, and such dependencies must be recorded explicitly in OpenTheory `thy` files.

There's also one leaking assumptions in the generated `hol-base.art` due to moving of theorems in `prove_base_assumsScript.sml`. To avoid confusion, I renamed `IMP_DISJ_THM` to `imp_disj_thm` to indicate that it's a "fake" theorem proved in this file, instead of the one from `boolTheory`.

--Chun

### DIFF
--- a/src/boss/hol4-base-unint.thy
+++ b/src/boss/hol4-base-unint.thy
@@ -129,6 +129,7 @@ numeral {
   import: arithmetic
   import: marker
   import: relation
+  import: pair
   article: "../num/theories/numeral.ot.art"
 }
 basic-size {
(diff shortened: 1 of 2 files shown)


mn200:
Thank you!

PR #1518, opened 2025-06-10
[listSimps] remove SNOC_APPEND from list normalisation

binghe:
Hi,

The theorem `SNOC_APPEND` (`|- !x l. SNOC x l = l ++ [x]`) is part of `listSimps.NORM_CONS_APPEND_NO_EVAL_CONV` (and is therefore part of `list_ss`, `srw_ss()`, `rw`, `simp`, `fs`, etc.)  The problem is that, it _unconditionally_ eliminates all `SNOC x l` occurred in user's proof, making other simplification rules on `SNOC` inapplicable. In general, when designing a simplification rule set, the order of rules in it shouldn't matter (because the simplifier doesn't support priority of rules).

This PR removes `SNOC_APPEND` from the above conversion and from now on users must explicitly apply it on need.  I think this change should be done even with costs of breaking some proofs.

There are totally about 35 broken proofs in the entire code base (tested with `-t3`), which can be roughly classified into several groups:

1. The proof explicitly needs `SNOC_INDUCT` or `SNOC_CASES` which involves `SNOC` but immediately a call of `simp[]` eliminates SNOC without using other SNOC-specific theorems.   In this case `SNOC_APPEND` must be explicitly added into the next simplification tactic, e.g. `simp[SNOC_APPEND]` for the proof to work.
2. The proof systematically uses `SNOC` and related theorems, including explicit calls of `SNOC_APPEND`, and the proof engineer is aware of the present issue and therefore only work with `std_ss` plus explicit list theorems. But sometimes `srw_ss()` is used as the ending tactics for list normalisation, which must now add `SNOC_APPEND`.
3. The proof does not involve `SNOC` in terms but uses some SNOC-theorems like `EL_SNOC` (`|- !n l. n < LENGTH l ==> !x. EL n (SNOC x l) = EL n l`) and immediately eliminate `SNOC` (by implicit `SNOC_APPEND`). These proofs actually should use `EL_APPEND1` (`|- !n l1 l2. n < LENGTH l1 ==> (EL n (l1 ++ l2) = EL n l1)`) instead.
4. In rare cases, the proof involves non-deterministic rewritings where assumptions and simp rules give different results depending on which one is handled first. There seems to be subtle changes on order of rules when `SNOC_APPEND` is removed. Essentially this is a robust issue of concrete proofs.

In general, by calling `augment_srw_ss` with `SNOC_APPEND` added back, some complex long proofs can be fixed. 

Besides, I added `TAKE1_DROP` (`|- !n l. n < LENGTH l ==> (TAKE 1 (DROP n l) = [EL n l])`) into `srw_ss`, so that, when combining with the existing `TAKE_SUC` (`|- !n x. TAKE (SUC n) x = TAKE n x ++ TAKE 1 (DROP n x)`) some broken proofs are fixed with minimal changes.   But this change has broken one proof where the user just don't want to simplify `TAKE 1 (DROP n l)` and use it as atom. I had to remove `TAKE1_DROP` from `srw_ss` in the beginning of the corresponding theory. There's only one such case if I remember correctly.

I have fixed all broken proofs when building with `-t3`, but I'm using `--expk`. Perhaps the CI tests will show a few more.

--Chun

### DIFF
--- a/examples/AKS/machine/countPolyScript.sml
+++ b/examples/AKS/machine/countPolyScript.sml
@@ -730,7 +730,8 @@ val poly_extendM_value = store_thm(
     `k = 1` by decide_tac >>
     rw[PAD_RIGHT],
     `SUC (k - 1) = k` by decide_tac >>
-    `PAD_RIGHT 0 (k - 1) [0] ++ [0] = SNOC 0 (PAD_RIGHT 0 (k - 1) [0])` by rw[] >>
+    `PAD_RIGHT 0 (k - 1) [0] ++ [0] =
+     SNOC 0 (PAD_RIGHT 0 (k - 1) [0])` by rw[SNOC_APPEND] >>
     `_ = PAD_RIGHT 0 (SUC (k - 1)) [0]` by rw[PAD_RIGHT_SNOC] >>
     metis_tac[]
   ]);
--- a/examples/logic/modal-models/chap2_1Script.sml
+++ b/examples/logic/modal-models/chap2_1Script.sml
@@ -398,14 +398,11 @@ val prop_2_15_subgoal_1 = store_thm(
               >- fs[LENGTH]
               >- fs[HD])
            >- (fs[RESTRICT_def] >> rw[] (* 3 *)
-              >- (`w ++ [v'] = SNOC v' w` by fs[] >>
-                 `EL m (w ++ [v']) = EL m w` by rw[EL_SNOC] >>
-                 fs[] >>
+              >- (`EL m (w ++ [v']) = EL m w` by rw[EL_APPEND1] >>
                  Cases_on `m < LENGTH w - 1` (* 2 *)
                  >- (`m + 1 < LENGTH w` by fs[] >>
-                    `w ++ [v'] = SNOC v' w` by fs[] >>
-                    `EL (m + 1) (w ++ [v']) = EL (m + 1) w` by rw[EL_SNOC] >>
-                    fs[])
+                     `EL (m + 1) (w ++ [v']) = EL (m + 1) w` by rw[EL_APPEND1] >>
+                     fs[])
                  >- (`m = LENGTH w - 1` by fs[] >>
                     `LENGTH w <> 0` by (SPOSE_NOT_THEN ASSUME_TAC >> fs[]) >>
                     `w <> []` by fs[] >>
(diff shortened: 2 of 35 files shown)


examples/AKS/machine/countPolyScript.sml
```diff
@@ -730,7 +730,8 @@ val poly_extendM_value = store_thm(
     `k = 1` by decide_tac >>
     rw[PAD_RIGHT],
     `SUC (k - 1) = k` by decide_tac >>
-    `PAD_RIGHT 0 (k - 1) [0] ++ [0] = SNOC 0 (PAD_RIGHT 0 (k - 1) [0])` by rw[] >>
```
ordinarymath, line 733: I think that LIST_EQ_SIMP_CONV should be able to simplify
`xs ++ [x] = SNOC y ys` to `xs = ys /\ x = y`

examples/logic/modal-models/chap2_1Script.sml
```diff
@@ -595,17 +586,17 @@ val prop_2_15_subgoal_4 = store_thm(
            `LENGTH l - 1 < LENGTH l` by fs[] >>
            `EL (LENGTH l - 1) (SNOC x' l) = LAST l /\
            EL ((LENGTH l - 1) + 1) (SNOC x' l) = x'` suffices_by metis_tac[] >> rw[] (* 2 *)
-           >- (`SNOC x' l = l ++ [x']` by fs[] >>
+           >- (`SNOC x' l = l ++ [x']` by fs[SNOC_APPEND] >>
```
ordinarymath, line 589: same here

src/list/src/rich_listScript.sml
```diff
@@ -194,6 +194,13 @@ val NOT_NULL_SNOC = Q.store_thm("NOT_NULL_SNOC",
    BasicProvers.Induct_on `l`
    THEN REWRITE_TAC[SNOC, NULL_DEF]);
 
+(* cf. CONS_ACYCLIC *)
+Theorem SNOC_ACYCLIC[simp] :
```
ordinarymath, line 198: Note that there's no need to add the flipped version as the simplifier does it automatically. I don't mind both so its up to you.

binghe: This is just learnt by the shape of `CONS_ACYCLIC`. I think we need also to consider the possible scenario where this theorem is explicitly called by easier tactics like `REWRITE_TAC` instead of the full powered simplifier.

examples/AKS/machine/countPolyScript.sml
mn200: I think that enhancement to `LIST_EQ_SIMP_CONV` can be left for another PR.

binghe: > I think that enhancement to `LIST_EQ_SIMP_CONV` can be left for another PR.

This work is current in-progress. I will do another PR for it. Besides. the branch has been tested with `--stdknl -t3`, no broken proofs.

mn200:
Thanks!

PR #1520, opened 2025-06-10
rename set_parents to from_parents in AncestryData

### LINKED ISSUE #1519: Rename set_parents to from_parents
In ancestry data API. Feedback in talk at workshop.

rsoeldner:
closes #1519

### DIFF
--- a/src/1/ThmSetData.sml
+++ b/src/1/ThmSetData.sml
@@ -289,7 +289,7 @@ fun export_with_ancestry
       {merge = #merge fullresult, DB = #DB fullresult,
        get_deltas = get_fulldeltas,
        record_delta = #record_delta fullresult o mk_raw,
-       parents = #parents fullresult, set_parents = #set_parents fullresult,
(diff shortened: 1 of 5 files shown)


src/parse/GrammarAncestry.sml
```diff
@@ -26,7 +26,7 @@ fun ancestry {thy} =
                     NONE => raise ERR "ancestry" "Badly encoded data"
                   | SOME sl => sl
 
-fun set_ancestry sl =
+fun from_ancestry sl =
```
xrchz: This one probably shouldn't be changed? Because it actually does a stateful update, I think.

rsoeldner: Indeed,thank you

mn200:
I'm sorry to have induced this work with my comments on Tuesday, but `set_parents` is actually a fine name for this function; it induces a state-change that is recorded in theory files.

PR #1526, opened 2025-06-12
kak hol mode

pylasnier:
This is an editor mode for the editor Kakoune.


ordinarymath:
Is there a way to use this without nix?

pylasnier:
You can yes, I've described the installation in the readme, kak plugins just drop into a folder. The nix stuff was mostly just for my own install but I thought there's no harm in case it's useful to someone

mn200:
Thanks for this!

PR #1528, opened 2025-06-12
Universal

ordinarymath:
Try using universal for the Traverser
Universal type is built in for PolyML and
if it leads to speedups one can emulate it using
exceptions for mosml.

### DIFF
--- a/examples/dev/compile.sml
+++ b/examples/dev/compile.sml
@@ -701,8 +701,8 @@ val hwDefineLib = ref([] : (thm * thm * thm)list);
 
 fun hwDefine defq =
  let open TotalDefn
-     val TC_SIMP = TC_SIMP_TAC()
-     val TC_PROVE = TC_PROVE_TAC()
+     val TC_SIMP = TC_SIMP_TAC (termination_ss()) []
+     val TC_PROVE = TC_SIMP >> NO_TAC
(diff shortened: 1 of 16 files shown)


ordinarymath:
I'll try to test performance tomorrow

mn200:
Please rework so the unrelated (and now duplicate/redundant?) commits are not included.

mn200:
Also, if you want to do this, please modify the existing `src/portableML/UniversalType` implementation.  That can point to Poly/ML's version under Poly/ML, and indeed be done with exceptions (as it is already) under Moscow ML.

ordinarymath:
Using a not very robust method of testing the time taken for bin/build -t3 on a laptop.
the 30 seconds saved might be just be noise.
mcandidate
```
20414.52user 1245.86system 2:12:03elapsed 273%CPU (0avgtext+0avgdata 13575400maxresident)k
1743368inputs+4865936outputs (241018major+766196025minor)pagefaults 0swaps
```
mcandidate + change
```
20351.50user 1212.83system 2:11:25elapsed 273%CPU (0avgtext+0avgdata 13710360maxresident)k
3132128inputs+4866328outputs (446833major+743907279minor)pagefaults 0swaps
```
It doesn't seem worth it to change the API.

PR #1534, opened 2025-06-23
Tidy up balanced_map

myreen:
These commits update `balanced_mapScript.sml` to use modern syntax consistently and speed up some of the slower proofs. The proofs can be made much faster still, but I'm PRing this now since I won't have time to improve this file further for a while.


mn200:
Thanks!  Think of the trees we're saving...

PR #1536, opened 2025-06-25
Make it easy to build the CakeML compiler backend from HOL

myreen:
This PR adds `examples/developers/build_cakeml` which is setup so that a run of Holmake there downloads the master branch of CakeML and builds `compiler/backend/proofs`, i.e. the correctness proof for the CakeML compiler backend.

This can be handy to run in case one wants to check whether changes to HOL break the CakeML compiler proofs.

### DIFF
--- a/examples/developers/build_cakeml/.hol_preexec
+++ b/examples/developers/build_cakeml/.hol_preexec
@@ -1,0 +1,1 @@
+[ -d "cakeml" ] || ./get-cakeml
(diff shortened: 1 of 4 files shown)


mn200:
Why not just set up a `cron` job to do this test as part of general CakeML infrastructure?

myreen:
Feel free to close this PR and delete the branch if you see no use in it. 

I see this PR as a simple way to build CakeML for someone not involved in CakeML. Also, this serves as an example of how to use `.hol_preexec`, but it's basically the same as the one for the ARM ASL<->L3 proofs.

binghe:
If I want to play with CakeML, I can clone it outside of HOL and go ahead to build it. Your script intends to clone the huge CakeML repository inside the HOL working directory, which is the worst idea.

PR #1543, opened 2025-07-01
remove broken interactive mode export_theory_return_hash

digama0:
The file which is read doesn't exist so this was causing errors for me in interactive mode. The new `export_theory_return_hash` function doesn't need to be called in the first place in interactive mode.


mn200:
Yeah, looks plausible to me.  @xrchz may recall what he was trying to do.

xrchz:
I didn't write this with much more intention than trying to minimise semantic changes from what was there before. If this works then all good.

mn200:
Thanks.

PR #1544, opened 2025-07-01
Theory syntax

### LINKED ISSUE #1364: `Theory` syntax

digama0:
Fixes #1364 , implementing the plan from https://github.com/HOL-Theorem-Prover/HOL/issues/1364#issuecomment-2862802929 .

### DIFF
--- a/examples/bootstrap/automationLib.sml
+++ b/examples/bootstrap/automationLib.sml
@@ -97,7 +97,7 @@ local
   val comments = ref ([]:term list)
   val next_comment = ref “""”
   fun print_defun defun = let
-    val str = mk_comb(“v2str”,mk_comb(“dec2v”,defun))
+    val str = mk_comb(“printing$v2str”,mk_comb(“printing$dec2v”,defun))
               |> QCONV EVAL |> concl |> rand |> stringSyntax.fromHOLstring
     in print ("\n\n" ^ str ^ "\n\n\n") end
 in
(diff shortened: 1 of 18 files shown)


mn200:
Please put the columns back in the `#(LINE` pragma.

mn200:
Thanks!

PR #1545, opened 2025-07-05
make all sml files parse

digama0:
See discussion [on Zulip](https://hol.zulipchat.com/#narrow/channel/486798-.E2.9D.84.EF.B8.8F-HOL4/topic/HOL.20dev.20meetings/near/527280328). The only one of these which could have fallout is the change to `winNT-systeml.sml`, since this file is read by some tools to put in real values. As [pointed out](https://hol.zulipchat.com/#narrow/channel/486798-.E2.9D.84.EF.B8.8F-HOL4/topic/HOL.20dev.20meetings/near/527279662) by @ordinarymath , the unix version of this file uses dummy values and can be replaced just fine, so this should not cause any problems.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 *.o
 .hol
 .holobjs
+.hollogs
 *.lex.sml
 *.grm.sig
 *.grm.sml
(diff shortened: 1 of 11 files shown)


mn200:
Thanks!

PR #1547, opened 2025-07-07
simplify unused arguments in holrepl

digama0:
This function is only used in one place and almost all of the arguments are set to a single value. Let's just simplify the code (which was originally borrowed from polyml's repl which actually uses the arguments).


mn200:
Nice!  Thanks.

PR #1548, opened 2025-07-07
LSP server, initial framework

digama0:
Currently it doesn't actually do anything, but this implements:

* a `--lsp` flag to start the server
* a multithreaded backend which dispatches request jobs
* parsing for the LSP protocol for a few commands:
  * hovers
  * go to definition
  * `$/eval`, a custom LSP extension which will be used to run SML code from the client


mn200:
Neat!

PR #1549, opened 2025-07-08
LSP server: compilation

digama0:
This implements the main logic for diagnostic reporting, lifted from the vscode mode and adapted to LSP.

Also, there is a library for dynamic arrays hiding in here because I was sufficiently annoyed about using lists all the time when I just want to push to an array.


mn200:
Maybe make the dynamic array a separate pair of files in `src/portableML`?

mn200:
Thanks!

Issue #1538, opened 2025-06-30
Theorems stored to current segment don't report location


mn200:
As per test-case:
```
> Theorem foo = TRUTH;
val foo = ⊢ T: thm
> DB.lookup {Thy = "-", Name = "foo"};
val it = SOME (⊢ T, {class = Thm, loc = Unknown, private = false}):
   DB_dtype.data_value option
```
where the `loc` is reported as `Unknown`.  (Thanks to @digama0 for the bug report.)

PR #1550, opened 2025-07-08
LSP server: go to definition

digama0:
* [x] Depends on #1549 

This part is rather more complicated, because the LSP server runs code before most of HOL is set up, so it can't refer to a lot of things and we need to have a hook system where `LSPExtension`, an early file, adds some global variables that are later modified in `DefnBase` (Maybe this should be in a better location? :shrug: )


mn200:
Rather than `Universal`, perhaps use our existing `UniversalType` to get things to compile in Moscow ML.  (I appreciate this code will never run with Moscow ML anyway, but it simplifies things if it can be made to compile.  I like PolyML's API so it's tempting to just adopt its signature and point the Poly/ML HOL implementation at Poly/ML's, and use the (portable) Poly/ML implementation as the Moscow ML version.

ordinarymath:
I do also prefer PolyML's `Universal`.

digama0:
the next use of `Universal` is not as easy to eliminate, because it's used in the PolyML API for thread local variables (which I'm guessing don't even exist in mosml). I guess that will need to be another portable API somewhere...

mn200:
Yeah, if you’re already in “interacting with Poly API” works, there’s no need to pretend to be portable.

src/coretypes/DefnBase.sml
```diff
@@ -563,4 +563,167 @@ fun LIST_HALF_MK_ABS th =
...
+  in navigateToLL end
+
+fun getBinding s = let
+  exception Ret of thminfo
+  in
+    (Theory.upd_binding s (fn i => raise Ret i); NONE)
+    handle Ret i => SOME i | HOL_ERR _ => NONE
+  end
+
+fun lookupThmInfo kn =
```
mn200: This has been fixed; DB.lookup works for bindings in the current theory (issue #1538)

mn200:
Exciting times...

PR #1551, opened 2025-07-09
examples/l3-machine-code: add 16-bit load/store lemmas

mktnk3:
Adding 16-bit load/store lemmas to arm8-asl equivalence proofs.

(needed for 16-bit shared memory load/store updates of the arm8_asl proofs in cakeml/compiler/encoders)


mn200:
Thanks!

PR #1553, opened 2025-07-09
LSP server: holdep

digama0:
* [x] Depends on: #1550

Adds support for automatically running `holdep` on the input to `$/eval` and normally elaborated files.


mn200:
Have merged stuff in fairly mindless fashion.  And will wait on CI.

PR #1555, opened 2025-07-11
Remove word size limitations of word_to_bytes_word_of_bytes_*

Liamolucko:


mn200:
Please don't remove existing theorems; others may be using them.  Just create new, more general, theorems in addition.

mktnk3:
The two theorems that might appear removed are re-established after the generalised theorem as its instantiations. So I think this is all good?

mn200:
So they are!  I missed that, thanks.

Thanks for the contribution @Liamolucko

PR #1560, opened 2025-07-14
Add string argument to cv_trans_pre and friends

myreen:
Users of `cv_trans_pre` (and its variants) have lamented that, each time one calls `cv_trans_pre`, the automation generates a different name for the precondition it defines. 

The change in this PR allows (and strongly encourages) users to specify what name the defined constant should have.

The simplest way to port old proofs to this new setting is to search-and-replace:
```
         cv_trans_pre → cv_trans_pre ""
     cv_trans_pre_rec → cv_trans_pre_rec ""
    cv_auto_trans_pre → cv_auto_trans_pre ""
cv_auto_trans_pre_rec → cv_auto_trans_pre_rec ""
```


mn200:
Thanks!

PR #1562, opened 2025-07-14
Theory syntax: allow options on headers

digama0:
This allows the syntax
```
Theory foo
Ancestors[opts]
  A B C
```
as an alternative to
```
Theory foo
Ancestors
  A[opts] B[opts] C[opts]
```
which can be nice for e.g. qualified imports.


mn200:
Nice.

PR #1564, opened 2025-07-15
Add coinductive helper theorems for cyclic itree structures and bisimulation

daole2904:
Occasionally, when doing proofs on itrees, we want to specify cases where, after stepping through some nodes (`Ret` or `Tau`) of an itree, we meet another concrete tree, or a tree structure. This PR add coinductive definitions and helper theorems for:

- `ret_or_reach`: After stepping through some nodes, we arrive at a concrete tree.
- `ret_or_reach_abs`: After stepping through some nodes, we arrive at an initialised tree from the `abs` structure.
- `strong_bisim_upfrom`: Specifying two trees that are strongly bisimulating each other starting from some concrete point up to the whole tree.
- `strong_bisim_upfrom_abs`: Similar to the above, but starting from some initialised tree from the `abs` structure.
- `weak_bisim_upfrom`, `weak_bisim_upfrom_abs`: weakly bisimulation version of the above ones.

**Note**: I think this can be considered as "bisimulation up to _some context_" where the _context_ is the abstracted structure `abs`.

Also added some additional theorems for doing proofs with itrees.

### DIFF
--- a/src/coalgebras/itreeTauScript.sml
+++ b/src/coalgebras/itreeTauScript.sml
@@ -1601,7 +1601,7 @@ Proof
   irule (GSYM spin)
 QED
 
-Theorem itree_wbisim_spin_eq:
+Theorem wbisim_spin_eq:
   itree_wbisim t spin <=> t = spin
 Proof
   rw[EQ_IMP_THM]
@@ -1692,7 +1692,7 @@ Theorem FUNPOW_Tau_abs_cyclic_spin:
 Proof
   iff_tac
   >- (rpt strip_tac
-      \\ irule $ iffLR itree_wbisim_spin_eq
+      \\ irule $ iffLR wbisim_spin_eq
       \\ irule itree_wbisim_coind_upto
       \\ qexists ‘CURRY {(FUNPOW Tau n (abs r), spin) | (n, r) | T }’
       \\ reverse $ rw[UNCURRY]
@@ -1723,7 +1723,7 @@ Proof
   >- (strip_tac
       \\ reverse $ Cases_on ‘?t''. strip_tau t t''’ \\ gvs[]
       >- (drule strip_tau_spin
-          \\ metis_tac[itree_wbisim_spin_eq, itree_wbisim_sym]
+          \\ metis_tac[wbisim_spin_eq, itree_wbisim_sym]
          )
       \\ Cases_on ‘t''’ \\ gvs[]
       >- (drule_all itree_wbisim_strip_tau_Ret
@@ -1733,7 +1733,7 @@ Proof
       \\ metis_tac[]
      )
   \\ rpt strip_tac
-  >- rw[itree_wbisim_spin_eq]
+  >- rw[wbisim_spin_eq]
   \\ metis_tac[itree_wbisim_rules]
 QED
 
@@ -1774,7 +1774,7 @@ Proof
   qid_spec_tac ‘t'’
   \\ qid_spec_tac ‘t’
   \\ ho_match_mp_tac after_taus_strongind
-  \\ gvs[spin, GSYM spin, itree_wbisim_spin_eq]
+  \\ gvs[spin, GSYM spin, wbisim_spin_eq]
   \\ rpt strip_tac
   \\ ‘Tau t' = Tau spin’ by metis_tac[spin]
   \\ gvs[]
@@ -1793,7 +1793,7 @@ Proof
       >- (reverse $ Cases_on ‘?x. strip_tau t x’ \\ gvs[]
           >- (dxrule strip_tau_spin
               \\ rw[]
-              \\ irule $ iffLR itree_wbisim_spin_eq
+              \\ irule $ iffLR wbisim_spin_eq
               \\ rw[itree_wbisim_sym]
              )
           \\ Cases_on ‘x’ \\ gvs[]
@@ -2268,7 +2268,7 @@ Proof
   \\ rw[UNCURRY]
   \\ reverse $ Cases_on ‘?x. strip_tau a0 x’ \\ gvs[]
   >- (drule strip_tau_spin \\ rw[]
-      \\ ‘a1 = spin’ by metis_tac[itree_wbisim_spin_eq, itree_wbisim_sym]
+      \\ ‘a1 = spin’ by metis_tac[wbisim_spin_eq, itree_wbisim_sym]
       \\ metis_tac[spin_FUNPOW_Tau]
      )
   \\ Cases_on ‘x’ \\ gvs[]
@@ -2424,7 +2424,7 @@ Proof
      )
   \\ drule $ iffLR itree_wbisim_strip_tau_cases
   \\ rpt strip_tac
-  \\ metis_tac[spin, itree_wbisim_spin_eq]
+  \\ metis_tac[spin, wbisim_spin_eq]
 QED
 
 (* strong bisimulation from an instance up to full tree of an abstraction *)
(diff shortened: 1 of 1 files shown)


src/coalgebras/itreeTauScript.sml
```diff
@@ -1564,6 +1614,1056 @@ Proof
...
+         )
+      \\ rw[strip_tau_spin]
+     )
+  \\ rw[spin_FUNPOW_Tau]
+QED
+
+Theorem itree_wbisim_FUNPOW_Tau_eqn:
+  (itree_wbisim (FUNPOW Tau n t) t' <=> itree_wbisim t t') /\
+  (itree_wbisim t (FUNPOW Tau n t') <=> itree_wbisim t t')
+Proof
```
mktnk3: This is a duplicate of an existing lemma:
https://github.com/HOL-Theorem-Prover/HOL/blob/53ae092d8eb66d0d4e8172239a3e8e32a95e9180/src/coalgebras/itreeTauScript.sml#L1476

daole2904: My bad, I reinvented the wheel. Removed.

src/coalgebras/itreeTauScript.sml
```diff
@@ -1564,6 +1614,1047 @@ Proof
...
+  qid_spec_tac ‘t'’
+  \\ qid_spec_tac ‘t’
+  \\ ho_match_mp_tac after_taus_strongind
+  \\ gvs[spin, GSYM spin, wbisim_spin_eq]
+  \\ rpt strip_tac
+  \\ ‘Tau t' = Tau spin’ by metis_tac[spin]
+  \\ gvs[]
+QED
+
+Theorem after_taus_itree_wbisim_strip_tau:
```
IlmariReissumies: I think this lemma has a cousin that seems prettier and more generally useful to me:

```
after_taus itree_wbisim t t' <=> itree_wbisim
```

src/coalgebras/itreeTauScript.sml
```diff
@@ -1564,6 +1614,1047 @@ Proof
...
+     )
+  \\ imp_res_tac strip_tau_FUNPOW
+  \\ rw[]
+  \\ irule after_taus_FUNPOW_TauL
+  \\ irule after_taus_FUNPOW_TauR
+  \\ irule after_taus_rel \\ rw[]
+QED
+
+(* return or eventually reach another tree *)
+CoInductive ret_or_reach:
```
IlmariReissumies: I find it a bit odd that `ret_or_reach` uses `strip_tau` style in the first two clauses, but `FUNPOW` style in the third. the two styles are obviously equivalent and interconvertible, but it would be nice to be consistent about it within the one definition.

src/coalgebras/itreeTauScript.sml
```diff
@@ -1564,6 +1614,1047 @@ Proof
...
+      \\ disj1_tac
+      \\ qexistsl [‘e’, ‘(λx. itree_bind (k' x) k)’] \\ rw[strip_tau_FUNPOW_strip_tau]
+      \\ metis_tac[]
+     )
+  \\ drule $ iffLR FUNPOW_Tau_SUC_cyclic_spin
+  \\ metis_tac[spin_bind, spin_FUNPOW_Tau]
+QED
+
+Theorem ret_or_reach_tau:
+  ret_or_reach t'' (Tau t) ==> ret_or_reach t'' t \/ t'' = t
```
IlmariReissumies: `t''` is a variable name in this theorem statement, but not `t'`

src/coalgebras/itreeTauScript.sml
```diff
@@ -1564,6 +1614,1047 @@ Proof
...
+  >- (imp_res_tac strip_tau_FUNPOW
+      \\ imp_res_tac FUNPOW_Tau_Vis_eq
+      \\ metis_tac[]
+     )
+  \\ gvs[FUNPOW_SUC]
+  \\ metis_tac[]
+QED
+
+(* return or reaching an instance of an abstracted function *)
+CoInductive ret_or_reach_abs:
```
IlmariReissumies: I would find this definition more natural if `abs` was a set of trees instead of a `ktree`. The arguments to `abs` are, after all, never relevant for your purposes; it only seems to matter whether a particular tree is in the domain or `abs` or not. Sets are a better fit for this.

IlmariReissumies:
argh I decided to use the list-with-boldfaced-topic-sentence style for easy skimming, and now my review looks like chatgpt output :cry:

daole2904:
how it appears on the phone and in outlook looks amazingly more chatgpt alike :D
thanks for taking a look into this, I will take some time to look at it asap

src/coalgebras/itreeTauScript.sml
daole2904: Right, that is completely easier to understand, look, prove and use it. My mind was lifted from the strong version..

src/coalgebras/itreeTauScript.sml
daole2904: Removed with`ret_or_reach`

src/coalgebras/itreeTauScript.sml
daole2904: Removed with`ret_or_reach`

src/coalgebras/itreeTauScript.sml
daole2904: Removed with`ret_or_reach`, but nice catch, I will give more thinking about the latter. Might be one day it will be back with the new `ret_or_reach`.

daole2904:
> Make the bisim_upfrom definitions inductive instead of coinductive

@IlmariReissumies I thought it would work in the first glance. However, restricting the definition to `Inductive` makes it unable to handle infinite trees (e.g. spin) therefore not distributive over `itree_bind`, and also cannot capture the case of something that "reach" itself after infinte moves (e.g. how `While` would look like if the `body` is `spin`).

Is there a better option to hide away the unwated `coind` thing? Can we keep it local?

daole2904:
> Strengthen the bisim_upfrom definitions to collect known pairs along the way

@IlmariReissumies If we are talking about the known pairs, that should be the concrete version rather than `abs` (it sounds ridiculous to "distil" the `abs` pair out of every concrete tree along the way. So this would be like a non-`abs` `upfrom` reborn, but in another shape. I would not thinking of `abs` and list of concrete pairs coexist in a predicate. Anyway there is plenty of time to discuss about this, since I am likely to add that in another PR in the future.

IlmariReissumies:
> > Make the bisim_upfrom definitions inductive instead of coinductive
> 
> @IlmariReissumies I thought it would work in the first glance. However, restricting the definition to `Inductive` makes it unable to handle infinite trees (e.g. spin) therefore not distributive over `itree_bind`, and also cannot capture the case of something that "reach" itself after infinte moves (e.g. how `While` would look like if the `body` is `spin`).
> 
> Is there a better option to hide away the unwated `coind` thing? Can we keep it local?

Oh ok! In that case, just leaving it is probably fine.

IlmariReissumies:
The change in 90259a1 is essentially an uncurrying transformation; that is, you changed the `abs`-pair-passing from being of the form

  `('a -> ktree) -> ('a -> ktree) -> 'b`

to being instead

  `('a -> ktree) × ('a -> ktree) -> 'b`

I don't think it makes much difference which one of these you use;
what I had in mind was instead something like this:

  `((ktree × ktree) -> bool) -> 'b`

...that is, instead of two separate functions that each return `ktree`s, you would use a set of pairs of `ktree`s. This types makes the intent clearer, and also makes it much easier to do this known-pairs collecting I described, though I don't think it needs to happen in this PR.

daole2904:
Thanks!
I think the adding set-of-pairs definition can be an issue assigned to me to be done in a later stage.

IlmariReissumies:
> Thanks! I think the adding set-of-pairs definition can be an issue assigned to me to be done in a later stage.

Sure that's fine by me.

mn200:
Many thanks for this!

PR #1568, opened 2025-07-18
Fix 1563

dnezam:
It seems the main culprit was a minor mistake in the lexing rules. After `Ancestors` or `Libs` we expect attributes to follow, not an identifier with attributes.

The change to `holstartkeyword` feels a bit hacky; additionally, without those parentheses around the added patterns, things seem to break for some reason.

I've also changed `stringScript` to use the new `Theory` syntax. That should hopefully make it easier to catch regressions.

EDIT: After digama0's review, `holstartkeyword` feels less hacky :)


dnezam:
Sorry for the force push - accidentally swapped the last two digits of the issue I was trying to refer to...

tools/Holmake/HolLex
```diff
@@ -233,6 +233,7 @@ attr_symname = {nonspecial_symbol}+ ;
 defn_attribute =
   ({alphaMLid}|{attr_symname}){optallws} ("="{optallws}{attributeEqnRHS}{optallws})?;
 defn_attribute_list = {defn_attribute} (","({ws}|{newline})*{defn_attribute})*;
+attributes = "["{optallws}{defn_attribute_list}"]";
```
digama0, line 236: you should use this in `id_with_attributes`

tools/Holmake/HolLex
```diff
@@ -250,7 +251,7 @@ HOLconjunction = "/\092" | "\226\136\167";
 DefinitionLabelID = {letter}{alphaMLid_tailchar}* | "~"{alphaMLid_tailchar}+ | {HOLconjunction};
 DefinitionLabel = "["{ws}*{DefinitionLabelID}?("["{alphaMLid_list}"]")?{ws}*":"?{ws}*"]";
 Definitionpfx = "Definition"{ws}+{id_with_attributes}{ws}*":";
-holstartkeyword = "Datatype"|"Type"|"Overload"|"Definition"|"Theorem"|"Triviality"|"Quote"|("Co")?"Inductive"|"Ancestors"|"Libs"|"Theory";
+holstartkeyword = "Datatype"|"Type"|"Overload"|"Definition"|"Theorem"|"Triviality"|"Quote"|("Co")?"Inductive"|("Ancestors"({ws}*{attributes})?)|("Libs"({ws}*{attributes})?)|"Theory";
```
digama0: this change is unnecessary

dnezam: I am pretty sure it is necessary - at least something needs to change. Consider

```
Ancestors
  foo
Ancestors[ignore_grammar]
  bar
Libs
  ...
```

The relevant rules are
```
<opens,headerthy>^{holstartkeyword} => (
  yybufpos := !yybufpos - (yyend - yypos); (* unread token *)
  OpenEnd yypos
);
<headerthy>{id_with_attributes} => (HeaderThy (yypos, yytext));
```

Without the change, `^{holstartkeyword}` will only match `Ancestors` of `Ancestors[ignore_grammar]`, while `{id_with_attributes}` will match `Ancestors[ignore_grammar]` in its entirety. AFAIK, in `mllex` the longest match wins and there isn't really a way to change. It is also what I observed looking at unquote output.

tools/Holmake/HolLex
dnezam: Makes sense. I noticed
```
ProofLine = {newline}"Proof"({ws}*"["{optallws}{defn_attribute_list}{optallws}"]")?;
```
has an `{optallws}` at the end, which the other attribute regexes don't have. Is that on purpose? Should I add `{optallws}` to `attributes` and make `ProofLine` also use it?

digama0: No. `optallws` means that newlines are allowed, and this requires additional handling in the SML printer because the generated SML is supposed to have exactly the same newlines as the input so that line numbers match up. The new parser we're working on will not have this constraint and will track positions properly, but for the current version I think we should leave it alone.

tools/Holmake/HolLex
digama0: Maybe a clearer way to do this is to put `({holstartkeyword}|(("Ancestors"|"Libs")({ws}*{attributes})?))` in the first rule so that `holstartkeyword` can remain simple.

Actually, this issue applies just as well for other keywords which may also have a form overlapping `id_with_attributes`, so how about just `{holstartkeyword}({ws}*{attributes})?)`?

dnezam: > Maybe a clearer way to do this is to put `({holstartkeyword}|(("Ancestors"|"Libs")({ws}*{attributes})?))` in the first rule so that `holstartkeyword` can remain simple.

To me, it would make more sense if it stays part of `holstartkeyword`. I could change what I wrote to use `(("Ancestors"|"Libs")({ws}*{attributes})?))` instead of the expanded form, though.

> Actually, this issue applies just as well for other keywords which may also have a form overlapping `id_with_attributes`, so how about just `{holstartkeyword}({ws}*{attributes})?)`?

Do you have any other keywords in mind? From briefly looking into it, it seems that attributes are put on some identifier that follows, not the keyword itself. 

Either way, I leave the final decision to you.

digama0: This is more a future proofing thing; `holstartkeyword` is used in other places as well and it's not intended to be a full specification of the the grammar of the respective keywords, just the keyword list itself. I think it is better to have this exception added to the `<headerthy>` production because that's where the conflict actually arises and if there is a grammar change that's where it would be reflected. The idea here is that `holstartkeyword`s are not identifiers, and so the `id_with_attributes` arm shouldn't trigger on them, including `Theorem` or other keywords which don't actually expect an identifier list there; they should nevertheless terminate the header.

src/string/stringScript.sml
```diff
@@ -6,19 +6,13 @@
...
-(* ---------------------------------------------------------------------*)
-(* Create the new theory                                                *)
-(* ---------------------------------------------------------------------*)
-
-val _ = new_theory "string";
-val _ = set_grammar_ancestry ["rich_list", "ternaryComparisons"]
+Theory string
+Ancestors
+  rich_list ternaryComparisons
+Ancestors[ignore_grammar]
```
mn200: This is unnecessary.  Those theories are behind `rich_list` anyway so "ignoring" their grammars will have no effect; I think you should be able to just put these into the ancestors list.

mn200:
Thanks!

PR #1570, opened 2025-07-21
Use Theory in src/string

dnezam:
A bit of detail on how I approach this:

- If there is no `set_grammar_ancestry`, no `ignore_grammar` attributes are inserted
- If there is `set_grammar_ancestry`, by default anything that is not mentioned in `set_grammar_ancestry` will have `ignore_grammar`. However, I try my best to manually remove unnecessary cases (for example, if the theories that are part of the grammar already depend on those not mentioned).

The first point might break things just by the virtue of adding `set_grammar_ancestry`. The second point should only break things if I make a mistake.


mn200:
Thanks!

PR #1573, opened 2025-07-22
LSP server: hover, doc->markdown

digama0:
This moves `ParseDoc` to portableML so it can be used by the LSP hover plugin and also the `Doc2*` tools.


mn200:
Given that we want rid of the `.doc` files anyway, please just convert them all to Markdown in one pass, commit that, and then change your code to display those files.

digama0:
That seems like an entirely separate PR? That will break all the existing .doc tooling, what will be done with that?

mn200:
I have a replacement for generation of the Reference manual (using `pandoc`) and the HTML files can also be generated with `pandoc`.  The help files served up by `help` in the REPL can just be raw Markdown.  So there's not much impact.  Certainly, it seems unnecessary to work on and commit code that will imminently get deleted.  Unless there is some other use of `.doc` files that I'm forgetting.

mn200:
This will need updating to cope with the disappearance of the `.doc` files and their replacement with markdown.

mn200:
Thanks!

PR #1576, opened 2025-07-23
[HolParser] Don't call set_grammar_ancestry with empty list

### LINKED ISSUE #1572: New theory syntax with empty ancestors causes exception
A file that mentions no explicit ancestors causes `set_grammar_ancestry` to throw an exception (it doesn't like the empty list argument).

dnezam:
This previously happened when the Theory syntax was used without any Ancestors. To avoid this, we first check whether there are any Ancestors before generating set_grammar_ancestry.

Fixes #1572


mn200:
Thanks.   This is a good interim fix.  I think we can change this to always include a Main/prelude/hol base later.

PR #1577, opened 2025-07-24
Use Theory in bool

dnezam:
As `bool` depends on `Unicode,` which is a substructure of `Parse` since commit https://github.com/HOL-Theorem-Prover/HOL/commit/721dd13f6c30158f542e9f28e0b66870374ae92d, we want to be able to mention `Parse.Unicode` in `Libs`, since by the time we hit `Unicode`, I assume that the relevant bindings from `Parse` are not yet available.

Thanks to Irvin (@ordinarymath) for finding the relevant commit and mentioning how splitting `open` could make a difference.

Also worth mentioning:
- Technically, the parser does not error out now when someone uses qualified identifiers in `Ancestors`. I doubt that's a big deal, since nothing stopped people from doing the same using `open` (Indeed, one can still write fun things like `open foo.barTheory` today :D)
- I also took the liberty to move some better hidden `open`s in `bool` into the header.

### DIFF
--- a/src/bool/boolScript.sml
+++ b/src/bool/boolScript.sml
@@ -9,10 +9,10 @@
 (*                 mind.                                                 *)
 (* ===================================================================== *)
 
-open HolKernel Parse
-open Unicode TexTokenMap
-
-val _ = new_theory "bool";
+Theory bool[bare]
+Libs
+  HolKernel Parse Parse.Unicode TexTokenMap Portable
+  GrammarSpecials[qualified] boolpp[qualified]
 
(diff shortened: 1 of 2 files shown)


dnezam:
This isn't the right way to do it.

We want to stop treating open as implicit loads, motivating the Theory syntax. That is, anything that is in the theory header should be loaded.

Since Unicode is supposed to be an open, and not a load, the correct way to deal with the issue in bool is to put Unicode into a separate `open` instead of Libs.

PR #1579, opened 2025-07-25
make ASSUME_NAMED_TAC smarter

ordinarymath:
The idea is that named assumptions should be at the top but the addition of new named assumptions should not be higher than existing named assumptions.
This two examples show the difference.
ASSUME_NAMED_TAC "hi" TRUTH ([``bar``],``foo``)
ASSUME_NAMED_TAC "hi" TRUTH ([``baz :- T``],``foo``) Note that this code is designed to ignore when there's a named example in the middle eg. in this case bar is ignored
ASSUME_NAMED_TAC "hi" TRUTH ([``foo``,``bar :- T``,``baz``],``foo``) Note that this is implementation is fairly inefficient for code with huge number of assumptions but should not be encountered in practice.


myreen:
Thanks!

PR #1583, opened 2025-07-27
Add holTheory and use it in Theory syntax

dnezam:
To my understanding, a normal hol session has at least indexedLists and patternMatches as parents:
```
> parents "-";
val it = ["indexedLists", "patternMatches"]: string list
```

It also appears that some theories depend on this. For example, if one were to take countable_types from pred_set, and set its grammar ancestry to the theories it explicitly opens (bool and pred_set), the proofs will break. This is problematic, as this is exactly what the Theory syntax does.

This PR/commit deals with this issue by
- creating a new theory called "hol", which represents the standard context of something starting in the hol heap
- changes the Theory syntax to load/open holTheory, and include hol in set_grammar_ancestry calls

### DIFF
--- a/examples/lambda/examples/lambda_holScript.sml
+++ b/examples/lambda/examples/lambda_holScript.sml
@@ -4,7 +4,7 @@ open pred_setTheory
 
 open basic_swapTheory nomsetTheory NEWLib
 
-val _ = new_theory "hol"
+val _ = new_theory "lambda_hol"
 
 (* a theory of higher order logic terms, as modelled in systems such as
    HOL4, Isabelle/HOL and HOL Light *)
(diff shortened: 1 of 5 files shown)


mn200:
Please rename the theory to be `holTheory` (`holScript.sml` etc).

dnezam:
It seems like it conflicts with `hol` defined in [examples/lambda](https://github.com/HOL-Theorem-Prover/HOL/blob/398c89b3508ed5c897d9cdf67bb35ce02955e275/examples/lambda/examples/holScript.sml).

Since that is the only instance (and it does not appear to be used anywhere), I am renaming it to `lambda_hol`.

mn200:
Thanks!

PR #1584, opened 2025-07-27
kak hol mode updates

pylasnier:
To overcome a selection size limit issue, `hol.kak` now uses a FIFO and opens a new terminal window running HOL. Dependencies autoload on start. Added syntax highlighting and indentation.


mn200:
Thanks for your work on this!

PR #1585, opened 2025-07-29
Use Theory syntax in src

dnezam:


dnezam:
Also running `bin/build -t3` currently — let's hope it works.

EDIT: Seems like it works.

mn200:
Thanks!

PR #1590, opened 2025-08-01
speedup CARD_dc_set_cross in dining_cryptosTheory

ordinarymath:
This slightly speed ups dining_cryptos. This lemma isn't one of the long ones but medium slowness in the file.


mn200:
Thanks!

PR #1592, opened 2025-08-01
Add every_drule and variants

ordinarymath:
every_drule is like drule but it is attempted on every assumption.
every_drule fails if it does not succeed on any assumption.

### DIFF
--- a/src/1/Tactic.sig
+++ b/src/1/Tactic.sig
@@ -120,21 +120,25 @@ sig
   val dxrule               : thm_tactic
   val rev_drule            : thm_tactic
   val rev_dxrule           : thm_tactic
+  val every_drule          : thm_tactic
 
   val drule_at             : match_position -> thm_tactic
   val dxrule_at            : match_position -> thm_tactic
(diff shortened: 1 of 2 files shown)


mn200:
The `every_assum` you've also added seems disconnected / independent.  

I can see the internal `every_assum_try` being useful though; perhaps *it* should be what `every_assum` does.  Also, why isn't the implementation of `every_assum_try` just
```
fun eat ttac = ASSUM_LIST (MAP_EVERY (TRY o ttac))
```

mn200:
Ah, I see.  I think you'd need to wrap `CHANGED_TAC` around that to get the right failure behaviour.  But that's gross, so your version is probably better.

ordinarymath:
The first commit is an independent change that I can put in a separate PR for adding a lower case version of `EVERY_ASSUM`. I ended up defining an separate every_assum_try as like you noticed, it's meant to try the ttac on every assum, but fail if it doesn't succeed at least once.

mn200:
Thanks!

PR #1593, opened 2025-08-02
add Parse.grammar_ancestry

digama0:
The setter is exposed but not the getter. This is useful for writing `set_grammar_ancestry` calls to know what is there by default.


mn200:
Good catch.  I'd prefer to be consistent with the other places where we use the `{thyname:string}` record though.

PR #1594, opened 2025-08-04
Vim mode: use holdep to determine loads

xrchz:
Remove the hL command and change the hl command to process the selected region with holdeptool.exe and load its results, instead of trying to figure out what to load within vim.

A temporary file is written for this purpose since it seems that holdeptool.exe does not work correctly if used as a filter (i.e. reading from stdin) and needs a filename instead.


mn200:
Thanks!

PR #1596, opened 2025-08-05
Fix a typo

myreen:


mn200:
Now merged into `t2-release-prep` at fc4f923830d4

PR #1607, opened 2025-08-12
[docker-ci] Updated CI images to Debian 13 with Poly/ML 5.9.2 release

binghe:
Hi,

Recently Debian 13 (trixie) is finally released. It provides GCC 14 as default C/C++ compiler (but note that Moscow ML's C source code requires GCC 13) and updated versions of many packages (But the version of Pandoc is still not enough.) And Debian now officially supports RISC-V architecture. Almost at the same time (yesterday), Poly/ML 5.9.2 is finally released with important bugfix for recent versions of macOS (and Linux).

The updated Docker CI images (hosted in my personal Docker Hub repository, `binghelisp/hol-dev`) now provides 4 architectures (`i386`, `amd64`, `arm64`, `riscv64`), Poly/ML 5.9.2 release, Moscow ML, MLton, and many external solvers.  More importantly, now PolyML is installed at non-system location (`/ML/polyml/`) which can be excluded from the `PATH` environment variable (after this PR is merged). This means that, calling `poly` will not work if it's still inside any HOL's makefiles.   This feature was requested by some users and I think is to also prevent issues with user environments having multiple PolyML versions installed.

The new Docker images are already in use, but the HOL building process only starts to not see Poly/ML in `PATH` after merging the file `developers/docker-ci/Dockerfile` in this PR.

--Chun

### DIFF
--- a/developers/docker-ci/Dockerfile
+++ b/developers/docker-ci/Dockerfile
@@ -25,6 +25,9 @@ ARG Z3_VERSION
 ARG CVC_VERSION
 ARG YICES_VERSION
 
+# this overrides the PATH environment from upstream images
+ENV PATH=/ML/HOL/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+
 # if --build-arg Z3_VERSION=anything, set HOL4_Z3_EXECUTABLE to system-installed z3
 ENV HOL4_Z3_EXECUTABLE=${Z3_VERSION:+/ML/z3-${Z3_VERSION}/bin/z3}
 RUN if [ "" != "${Z3_VERSION}" ]; then \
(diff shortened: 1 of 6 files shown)


binghe:
Well, the first bug with non-systemwise Poly/ML has been exposed by the CI tests:
```
 > [ 9/10] RUN bin/build --expk -j4 -t:
47.11  testTheory                                                          (0s)     OK
47.11  /bin/sh: 1: poly: not found
47.11  Full log: /ML/HOL/tools/Holmake/tests/time-logging/.hol/logs/goodlog
```

binghe:
I've fixed code conflicts by merging from `develop`, and now the CI test failures are the same as on `develop`.

mn200:
Thanks!

binghe:
Note there's one bare call of `mosml` in `tools/Holmake/tests/time-logging/Holmakefile`. I don't know how to "fix" it.

PR #1608, opened 2025-08-13
[kak] fixed holscript detection, added close command and hook

pylasnier:
Fixes a filetype detection issue, so that kak now should always be able to detect script files automatically. Can close HOL instance with command, which also runs on closing kak.


mn200:
Thanks!

PR #1609, opened 2025-08-14
Add more EXTREAL_SUM/PROD theorems

KaiPhan:
Hi,

This PR adds several new lemmas to the EXTREAL_SUM and EXTREAL_PROD libraries, extending support for finite sums and products of the extreal type. These are mostly routine results, covering constant factor extraction, absolute value inequalities, conversion to real sums, and product simplifications over support sets. They will be useful later in measure theory and probability proofs.

Kai


mn200:
Thanks!

PR #1616, opened 2025-08-18
Deduplicate itree_el -> itree_rep

FieryIceStickie:
Since it is possible to prove `itree_el = itree_rep`, `itree_el` shouldn't be necessary, so removed the definition for it and fixed the proofs.
Also adds an overload for itree_el at the end in case itree_el was used outside of the package.


mn200:
Thanks!

PR #1619, opened 2025-08-21
fix breakage caused by #1451

ordinarymath:
The syntax upgrades did causes changes
Making w2v_def and v2w_def nocompute

### DIFF
--- a/src/n-bit/bitstringScript.sml
+++ b/src/n-bit/bitstringScript.sml
@@ -86,7 +86,7 @@ Definition w2v_def:
     GENLIST (\i. w ' (dimindex(:'a) - 1 - i)) (dimindex(:'a))
 End
 
-Definition v2w_def:
+Definition v2w_def[nocompute]:
   v2w v : 'a word = FCP i. testbit i v
 End
 


ordinarymath:
Note that #1451 also makes `extend` ,`boolify`,`bitfity`,`s2v`,`v2s` compute but I think those should be ok

mn200:
`master` has
```
val w2v_def = Define`
  w2v (w : 'a word) =
    GENLIST (\i. w ' (dimindex(:'a) - 1 - i)) (dimindex(:'a))`
```
suggesting that `w2v` should stay as `compute`.

mn200:
Please check `examples/l3-machine-code/riscv/decompiler` with this change in place and report back.

ordinarymath:
I'll make it ready for review once I've tested with example `examples/l3-machine-code/riscv/decompiler`

ordinarymath:
running Holmake on that directory seems to work

ordinarymath:
Note: time from docker-ci stdknl 3153.1s

mn200:
Thanks!

PR #1620, opened 2025-08-21
Speedup free vars

ordinarymath:
depends on 
1.  #1619

Speedup up free_vars
FV was not recursive due to continuation calling FV.
results from before and after change using `time bin/build`
real    11m7.854s
user    22m42.042s
sys 3m33.477s

real    10m55.458s
user    22m24.558s
sys 3m24.907s

### DIFF
--- a/src/0/Term.sml
+++ b/src/0/Term.sml
@@ -141,13 +141,14 @@ end;
  * The free variables of a lambda term. Tail recursive (from Ken Larsen).    *
  *---------------------------------------------------------------------------*)
 
-local fun FV (v as Fv _) A k   = k (Lib.insert v A)
-        | FV (Comb(f,x)) A k   = FV f A (fn q => FV x q k)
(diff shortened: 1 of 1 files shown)


ordinarymath:
It's highly suspicious how often `free_vars` is being called that such a change leads to speedups. Code should be using `FVL` instead.

ordinarymath:
Does anyone know why these tests are failing? It doesn't even touch the expk kernel?

ordinarymath:
Note time from docker-ci stdknl 3123.1s

ordinarymath:
@mn200 let me know how to deal with this commits which are kinda stacked on each other the original intention was to keep them separate so the speedups from every change are obvious.

mn200:
I think the slow and steady way of proceeding will keep things simpler: if a base change speeds things up, wait for it to get merged, and then do the next experiment.

mn200 (approved):

mn200:
Thanks for this!

PR #1623, opened 2025-08-22
make is_{select,forall,exists,exists1} faster for stdknl

ordinarymath:
Make is_{select,forall,exists,exists1} faster for stdknl
Code now takes constant time
results from before and after change using `time bin/build`
real    10m55.458s
user    22m24.558s
sys 3m24.907s

real	10m37.510s
user	21m43.869s
sys	3m12.949s


ordinarymath:
note: depends on #1620

ordinarymath:
Note time from docker-ci stdknl 3122.6s

PR #1624, opened 2025-08-22
Use exec in the generated bare HOL script

wildptr:
While trying to implement `^C` (interrupt HOL process) for my custom editor, I noticed that sending SIGINT to the `hol` wrapper process has no effect on the "real" HOL process that actually runs ML code because `sh` seems to be intercepting the signal. Prepending `exec ` to the line that invokes `buildheap` in the `hol` wrapper script appears to solve the problem.

This PR modifies `tools-poly/Holmake/unix-systeml.sml` so that the generated `hol` script actually `exec`s `buildheap`.


mn200:
Thanks for this!  Most of the time the `emacs` mode seems to bypass this issue (maybe it signals the parent and all of its children?), so I've never noticed a problem.

PR #1629, opened 2025-08-24
Faster error backtraces

ordinarymath:
Creating a pull request so that work is viewable. I don't know whether the location should be recursively added too.


ordinarymath:
Will conflict with https://github.com/HOL-Theorem-Prover/HOL/pull/1656

PR #1630, opened 2025-08-30
Minor miscellaneous HolSmt improvements

someplaceguy:
I noticed I had some old HolSmt commits lying around which I had forgotten to submit as a PR.

These are just minor improvements, which may be helpful especially when debugging proof replay failures.


mn200:
Thanks!

PR #1632, opened 2025-08-31
Change the parser to use HOLdict instead of Binarymap

ordinarymath:
Experiments revealed that this speeds the time of
bin/build by around around 30 seconds.
This change is observable if you only
change the use Binarymap to HOLdict in parse_term but
I decided to just make the most of parse use HOLdict anyway.

### DIFF
--- a/doc/next-release.md
+++ b/doc/next-release.md
@@ -34,6 +34,8 @@ New examples:
 
 Incompatibilities:
 ------------------
+-   The return types of `parse_term.mk_prec_matrix`, `type_grammar.parse_map`, `type_grammar.privileged_abbrevs` 
+    have been changed to return maps of type HOLdict instead of Binarymap.
 
 * * * * *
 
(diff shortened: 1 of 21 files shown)


mn200:
If `HOLdict` is so much better, perhaps we should just never use `Binarymap` at all.  

Anyway, thanks for checking.  (Is the 30s improvement apparent over multiple runs?  `build` times can be pretty variable.)

mn200 (approved):

mn200:
It might also be nice to compare against the Isabelle `Table`-functor dictionaries.  Unfortunately that takes more work as it's an API change as well (and we can't have one polymorphic HOLdict based on Table because the key type becomes monomorphic).

ordinarymath:
Yes the 30 seconds improvement is over multiple runs. I've tested it 3 times

ordinarymath:
Another data point is that the current develop took 3116.8s on (docker ci,stdknl) and now takes 3018.9s

PR #1633, opened 2025-08-31
[HolParser] Add no_sig_docs tag for Theory

### LINKED ISSUE #1597: Add Theory tag so Theory.sig files can be exported without theorem list
Introduced in f77c366fd190d8e02356278719b9b57013dcb102, one can use `Feedback.set_trace "TheoryPP.include_docs" 0` to exclude the theorem list from the generated .sig file. This issue is about adding syntactic sugar using a new tag (`no_sig_docs`) for the `Theory` syntax. 

For example:
```
Theory foo[no_sig_docs] 
```
should result in the generated `export_theory` call to look something like
```
Feedback.set_trace "TheoryPP.include_docs" 0 before Theory.export_theory()
```

In other words, with the tag `no_sig_docs`, `unquote` should add `Feedback.set_trace "TheoryPP.include_docs" 0 before` as a prefix to the `export_theory` call.

dnezam:
Closes #1597

### DIFF
--- a/Manual/Description/modern-syntax.smd
+++ b/Manual/Description/modern-syntax.smd
@@ -72,7 +72,11 @@ See also Table~\ref{tab:script-file-syntax}.
 
 The three keywords (`Theory`, `Ancestors` and `Libs`) must all appear in column 0 of the input file.
 The `Ancestors` and `Libs` components may be omitted.
+
+\index{Theory keyword@\ml{Theory} keyword}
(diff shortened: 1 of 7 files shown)


mn200:
Please document this in the Description manual; see file `Manual/Description/modern-syntax.smd`.

dnezam:
Thanks, I missed that.

mn200:
Thanks!

PR #1634, opened 2025-09-01
HolSet asm bound ho net

ordinarymath:
This changes HO_net.enter to take the free vars in the assumptions as a HOLset instead of a list. This also results in the simplifier convdata key to be (term HOLset.set * term) option. ~~This change speeds up bin/build by around 30 seconds(*tested twice*)~~ Compared against wrong timing doesn't actually speed up.

#### TODO
- [ ] Test bin/build -t3
- [ ] Fix documentation


ordinarymath:
Wait I've compared it

ordinarymath:
After sorting out my logs this seems to show like 4 seconds improvement but not enough for it to be worth it.

PR #1635, opened 2025-09-02
Portemanteau theorem for weak convergence of (sub)probability measure

binghe:
Hi,

This PR is in preparation of the proof of (a version of) Central Limit Theorem. Here I formalised (partially) the so-called "Portemanteau" theorem (see, e.g, [1, Theorem 13.16] and [2]) for weak convergence of (sub)probability measure:

Screenshot 2025-09-02 alle 16 09 45

This proof itself is in `examples/probability/distributionTheory` of about 1,000 lines long, but it requires a lot of infrastructures in measure theory (+ general Borel theory), general topology (+ metric space) and topologies for reals and extreals that have to be established first.

Recall that the concept of "convergence in distribution" requires that for each "bounded and continuous" function f, the integration on this function, composed with a sequence of random variables, converges (sequentially) into the same functino composition but with another single random variable. Essentially this is the "weak convergence" of distribution functions of involved random variables:

```
   [converge_in_dist_def]  Theorem [probabilityTheory, existing]      
      ⊢ ∀X Y p.
          (X ⟶ Y) (in_distribution p) ⇔
          ∀f. bounded (IMAGE f 𝕌(:real)) ∧ f continuous_on 𝕌(:real) ⇒
               expectation p (Normal ∘ f ∘ real ∘ Y)) sequentially
```

The problem is that, to *prove* a sequence of random variables converge in distribution to another one, the function f is fixed in the proof (by `rpt STRIP_TAC`, e.g.) and all we know is that it's "bounded and continuous". Note that, e.g., we don't even know if the function is (any order) derivable/differentiable at any position. This lacking of information makes analytical proofs (based on  Lebesgue or Gauge calculus) impossible.

The "Portemanteau" (or "Portmanteau") theorem as above mentioned, provides a list of equivalent conditions which can be used as replacements when proving convergence in distribution. A notable one is the second condition which additionally asserts the that involved function is "Lipschitz continuous". Later, starting from it (by another lemma), one may further restrict the function class into functions which are arbitrary order derivable.

We only care about the first 6 conditions, with a proof path like this: `(2) ==> (4) <=> (5) ==> (6) ==> (3) ==> (1) ==> (2)`, so that `(1) <=> (2)` (the target theorem) can be obtained. The hard proofs are `(2) ==> (4)` and `(6) ==> (3)`, where I was totally stuck for 3 months...

I have added "abstract" definitions like this, to clearly represent each equivalent conditions and the shared antecedents:
```
[Portemanteau_antecedents_def]
⊢ ∀E X Y.
    Portemanteau_antecedents E X Y ⇔
    (∀n. subprobability_measure (space (B E),subsets (B E),X n)) ∧
    subprobability_measure (space (B E),subsets (B E),Y)
[Portemanteau_i_def]
⊢ ∀E X Y. Portemanteau_i E X Y ⇔ weak_converge_in_topology (mtop E) X Y
[Portemanteau_ii_def]
⊢ ∀E X Y.
    Portemanteau_ii E X Y ⇔
    ∀f. f ∈ BL E ⇒ weak_convergence_condition (mtop E) X Y f
[Portemanteau_iii_def]
⊢ ∀E X Y.
    Portemanteau_iii E X Y ⇔
    ∀f. bounded (IMAGE f 𝕌(:α)) ∧ f ∈ borel_measurable (B E) ∧
        Y (U (mtop E) f) = 0 ⇒
        weak_convergence_condition (mtop E) X Y f
[Portemanteau_iv_def]
⊢ ∀E X Y.
    Portemanteau_iv E X Y ⇔
    Y (mspace E) ≤ liminf (λn. X n (mspace E)) ∧
    ∀s. closed_in (mtop E) s ⇒ limsup (λn. X n s) ≤ Y s
[Portemanteau_v_def]
⊢ ∀E X Y.
    Portemanteau_v E X Y ⇔
    limsup (λn. X n (mspace E)) ≤ Y (mspace E) ∧
    ∀s. open_in (mtop E) s ⇒ Y s ≤ liminf (λn. X n s)
[Portemanteau_vi_def]
⊢ ∀E X Y.
    Portemanteau_vi E X Y ⇔
    ∀A. A ∈ subsets (B E) ∧ Y (mtop E frontier_of A) = 0 ⇒
        ((λn. X n A) ⟶ Y A) sequentially
```
and then proved "abstract" theorems like this:
```
[Portemanteau_i_eq_ii]
⊢ ∀E X Y.
    Portemanteau_antecedents_alt E X Y ⇒
    (Portemanteau_i E X Y ⇔ Portemanteau_ii E X Y)
[Portemanteau_i_imp_ii]
⊢ ∀E X Y. Portemanteau_i E X Y ⇒ Portemanteau_ii E X Y
[Portemanteau_ii_imp_iv]
⊢ ∀E X Y.
    Portemanteau_antecedents_alt E X Y ∧ Portemanteau_ii E X Y ⇒
    Portemanteau_iv E X Y
[Portemanteau_iii_imp_i]
⊢ ∀E X Y.
    Portemanteau_antecedents E X Y ∧ Portemanteau_iii E X Y ⇒
    Portemanteau_i E X Y
[Portemanteau_iv_eq_v]
⊢ ∀E X Y.
    Portemanteau_antecedents E X Y ⇒
    (Portemanteau_iv E X Y ⇔ Portemanteau_v E X Y)
[Portemanteau_v_imp_vi]
⊢ ∀E X Y.
    Portemanteau_antecedents E X Y ∧ Portemanteau_v E X Y ⇒
    Portemanteau_vi E X Y
[Portemanteau_vi_imp_iii]
⊢ ∀E X Y.
    Portemanteau_antecedents E X Y ∧ Portemanteau_vi E X Y ⇒
    Portemanteau_iii E X Y
```
And eventually, from `Portemanteau_i_eq_ii`, the following applied form is generated (by forwarding rewrites):

```
[weak_converge_in_topology_alt_Lipschitz]
⊢ ∀E X Y.
    (∀n. prob_space (space (B E),subsets (B E),X n)) ∧
    prob_space (space (B E),subsets (B E),Y) ⇒
    (weak_converge_in_topology (mtop E) X Y ⇔
     ∀f. bounded (IMAGE f 𝕌(:α)) ∧ Lipschitz_continuous_map (E,mr1) f ⇒
         ((λn. ∫ (space (B E),subsets (B E),X n) (Normal ∘ f)) ⟶
          ∫ (space (B E),subsets (B E),Y) (Normal ∘ f)) sequentially)
```

Here "B" is locally overloaded to the "general Borel" generator from open sets of any topology (and `E` is any metric space), in `sigma_algebraTheory` (perhaps this is the most general definition of "Borel" sigma-algebra):
```
   [general_borel_def]  Definition      
      ⊢ ∀top. general_borel top = sigma (topspace top) (open_in top)
```

Unfortunately, to connect to the original definition of "converge_in_dist", I have to show that the above Borel generator, when applying to the "Euclidean" topology for extreals (`extreal$ext_euclidean`), indeed give us the existing `Borel` sigma-algebra defined in `borelTheory`. This is still in progress.

Changes in other files are basically new lemmas needed by the above results. Note that, to balance the file sizes, I moved some more contents from `extreal` to `extreal_base`, and then sone extreal theorems from `martingale` to `extreal`, such that each file under "src/probability" does not go beyond 10,000 lines (which may not be very meaningful). But I think it's the right direction to move similar theorems into the same places whenever possible. And by shrinking a little the already large `martingale` theory, more theorems about martingales can be added in the future.

--Chun

[1] Klenke, A.: Probability Theory: A Comprehensive Course (3rd Edition). Springer Science & Business Media (2020).
[2] https://en.wikipedia.org/wiki/Convergence_of_measures#Weak_convergence_of_measures


mn200:
Awesome; thanks!

Issue #1615, opened 2025-08-18
Remove the use of `Feedback.wrap_exn`


ordinarymath:
In many theories, using the PolyML profiler reveals alot of time is spent in `String.concat`. Its quite often the `hottest function`. After getting PolyML to inline that it shows that the use is by `Feedback.format_err_rec`. Further inlining shows that 
`Feedback.wrap_exn` is to blame. Therefore by removing the use of `Feedback.wrap_exn` there should be noticeable speedups.

konrad-slind:
Note that `wrap_exn` exists because it performs a useful function, namely giving a useful back-trace on failure, so it shouldn't just be deleted from the system.  This issue is really about finding the places where `wrap_exn` is being used in code that isn't failing back up to the top-level.

ordinarymath:
I think the solution should more about having the HOL_ERR datatype support back trace paths. The slow bit was string concatenation.

mn200:
Term-printing in exception messages is also a waste of time; this should be avoided if possible.

ordinarymath:
This was done in https://github.com/HOL-Theorem-Prover/HOL/pull/1656 see https://github.com/HOL-Theorem-Prover/HOL/pull/1656#issuecomment-3369823465

PR #1636, opened 2025-09-03
Add a fast path to HO_PART_MATCH

### LINKED ISSUE #1614: Remove the use of `bound_vars` in HO_PART_MATCH
https://github.com/HOL-Theorem-Prover/HOL/blob/a185ee9ec7cea229878b6df1c913e13bf7c78668/src/1/Drule.sml#L1787C1-L1795C7
The speed of `bound_vars` in the standard kernel is very slow since dest_abs is very slow. This issue is about removing it and not calling it all during HO_PART_MATCH.

ordinarymath:
This works by noticing that if that there are no free variables in the function it is first order.
Partially fixes #1614. Also helps with #1615 as ho_match_term is a huge `wrap_exn` generator.

### DIFF
--- a/src/1/Drule.sml
+++ b/src/1/Drule.sml
@@ -1799,67 +1799,99 @@ in
          val sth = SPEC_ALL th
          val bod = concl sth
          val pbod = partfn bod
-         val possbetas =
-            mapfilter (fn v => (v, BETA_VAR v bod))
-                      (filter (can dom_rng o type_of) (free_vars bod))
-         fun finish_fn tyin ivs =
-            let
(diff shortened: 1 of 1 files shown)


mn200:
Do you have a measurable speedup from this change?

ordinarymath:
I've just did another test by rebasing and it does speedup CakeML's compiler/backend/proofs/word_to_stackProof by 1min
Theory "word_to_stackProof" took 10m22s to build
Theory "word_to_stackProof" took 9m23s to build

mn200:
Thanks for this!

PR #1637, opened 2025-09-03
faster Sub_and_cond.

ordinarymath:
Code was repeatedly wasting time computing stuff. PART_MATCH is eager in the work it does and this gets rid of the repeated computation of the eager bit.

### DIFF
--- a/src/num/arith/src/Sub_and_cond.sml
+++ b/src/num/arith/src/Sub_and_cond.sml
@@ -112,10 +112,20 @@ fun SUBEQC_CONV t =
     else NO_CONV
   end t
 
-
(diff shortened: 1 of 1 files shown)


ordinarymath:
Btw this PR seems to be only speeding up CakeML or modern files that use the modern simplifier tactics. miscTheory becomes 5 seconds faster with this change while bin/build appears to have a slight negligible slowdown and as far I'm clueless to what's causing it. From my testing it seems to only be built once by numheap so am hoping this slowdown is a constant cost by code generation.

mn200:
Thanks!

PR #1638, opened 2025-09-05
Strengthening Bayes rule by removing an unnecessary precondition.

Eric-C-Hall:
Strengthening Bayes rule by removing an unnecessary precondition. While it is true that allowing the probability of the event B to be zero will implicitly cause us to be dividing a value by zero, in HOL4, division by zero has some value, and since anything multiplied by zero is zero, the left-hand-side and the right-hand-side will both be zero in this case.

### DIFF
--- a/src/probability/probabilityScript.sml
+++ b/src/probability/probabilityScript.sml
@@ -9322,10 +9322,12 @@ QED
 
 Theorem BAYES_RULE :
     !p A B. prob_space p /\ A IN events p /\ B IN events p /\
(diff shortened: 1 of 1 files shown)


binghe (approved):
Good finding. Usually, for `cond_prob p A B` to be meaningful, `prob p B <> 0` is required, but in this theorem we have somthing like `x / 0 * 0 = unspecified * 0 = 0`, making the theorem conclusion hold. It’s good to have one less antecedent for this theorem when applying it.

mn200:
CI errors are unrelated.  Thanks!

PR #1639, opened 2025-09-05
speedup Solve_ineqs

ordinarymath:
PolyML doesn't support tail mod cons and has to
waste time and memory allocating a list for it to be only checked if it's null.


mn200:
Thanks!

PR #1645, opened 2025-09-07
Converge in dist alt lipschitz

binghe:
Hi,

Our current definition of "convergence in distribution" (in `probabilityTheory`) is wrong for random variable (r.v.) having infinite values, because of the use of the function `real` which converts extended reals back to reals:
```
   [converge_in_dist_def]  Theorem (probabilityTheory, old definition)
      ⊢ ∀X Y p.
          (X ⟶ Y) (in_distribution p) ⇔
          ∀f. bounded (IMAGE f 𝕌(:real)) ∧ f continuous_on 𝕌(:real) ⇒
              ((λn. expectation p (Normal ∘ f ∘ real ∘ X n)) ⟶
               expectation p (Normal ∘ f ∘ real ∘ Y)) sequentially
```
One can imagine that, if the random variables `X n` is a constant function returning `PosInf`, by the above definition one may wrongly derive that they converge to the constant r.v.'s of zero, because `real PosInf = real NegInf = 0`.  However, this definition is good for its use of `continuous_on` (from `real_topologyTheory`), and the whole definition was added when we didn't know how to define "continuity" for functions of type `:extreal -> real` (and `:extreal -> extreal`).

Now all these problems are resolved. In this PR, I "upgrade" the above definition by the following new one, which fully supports all extreal-valued r.v.'s:
```
   [converge_in_dist_def]  Theorem      
      ⊢ ∀X Y p.
          (X ⟶ Y) (in_distribution p) ⇔
          ∀f. f ∈ C_b ext_euclidean ⇒
              ((λn. expectation p (Normal ∘ f ∘ X n)) ⟶
               expectation p (Normal ∘ f ∘ Y)) sequentially
```
where `C_b` is overloading `bounded_continuous` taking any topology as parameter:
```
   [bounded_continuous_def]  Definition      
      ⊢ ∀top f.
          C_b top f ⇔
          continuous_map (top,euclidean) f ∧ bounded (IMAGE f 𝕌(:α))
```

The old definition is still available (in `distributionTheory`), but now becomes an equivalent theorem for finite-valued r.v.'s:
```
[converge_in_dist_alt_continuous_on] (distributionTheory)
⊢ ∀X Y p.
    prob_space p ∧ (∀n. real_random_variable (X n) p) ∧
    real_random_variable Y p ⇒
    ((X ⟶ Y) (in_distribution p) ⇔
     ∀f. bounded (IMAGE f 𝕌(:real)) ∧ f continuous_on 𝕌(:real) ⇒
[...]

PR #1652, opened 2025-09-13
More alternative definitions of "convergence in distribution"

binghe:
Hi,

Following #1645, this PR adds more alternative definitions of "convergence in distribution" (`converge_in_dist`) for the (future) proof of CLT (central limit theorem):

```
[converge_in_dist_alt_Lipschitz_real] (distributionTheory)
⊢ ∀X Y p.
    real_random_variable Y p ⇒
    ((X ⟶ Y) (in_distribution p) ⇔
     ∀f. f ∈ BL mr1 ⇒
          expectation p (Normal ∘ f ∘ real ∘ Y)) sequentially)
```

This theorem is the starting point of further restricting the involved function class (`f`) to from bounded Lipschitz to higher differentiable functions, because the type of `f` is back to `:real -> real`, aligned with calculus. All intermediate definitions and lemmas serve for this proof.

--Chun


mn200:
Thanks!

Issue #1643, opened 2025-09-06
Implement SPECL in the kernel


ordinarymath:
SPEC_ALL is extremely slow for large terms. Using Thm.Specialize which will use lazy beta conv only helps the standard kernel. Given that GENL is already in the kernel i don't see why SPECL can't be.

ordinarymath:
This is actually also an issue that `can't` be mitigated with `Thm.Specialize` where if it's called e.g. with a term like ``! A B. A /\ B``and if you call SPECL with 2 terms [A,B] where `A` is a really big term you get
SPEC ``B`` ``!B. A /\ B`` which has to traverse the whole of A anyway. One workaround is to generate variables and the specializing so that you get `(genvar1) /\ (genvar2)` and a resulting INST call does to do the actual work.

ordinarymath:
Another thing to mention thou is than in practice most of the times where automation is attempting call SPEC and facing performance problems, it's possible to avoid the calls at all and use INST instead.

xrchz:
Is SPEC fast when you provide the bound variable? If so maybe the implementation of SPECL could also use INST

mn200:
With any number of `SPEC`s, the issue is that de Bruijn indices have to be replaced by a real terms.  When iterated, this results in many traversals.  With `INST` only one traversal is needed.  Using `lazy_beta_conv` can only ameliorate this a little.

PR #1653, opened 2025-09-13
Speedup SPECL on stdknl using explicit substitutions

ordinarymath:
Note that SPEC will get rid of every internal closures so this code ensure resulting term is the same. 

I would ideally want to do https://github.com/HOL-Theorem-Prover/HOL/issues/1643 but I don't have enough time to think whether I'm doing it correctly.


mn200:
What's the speedup like?

ordinarymath:
I recall that my benchmark file of word_to_stackProof didn't speed up that much. But the speed of bin/build did speed up. I'll find some time to generate the numbers this weekend.

mn200:
Thanks!

PR #1646, opened 2025-09-08
Higher Order Derivatives and Differentiability

binghe:
Hi,

As we know, in HOL's `limTheory` (originally by John Harrison, I believe), it already has formalised the "derivative" of real functions and the concept of "differentiable" i.e. functions having derivatives at certain points (and in `derivativeTheory` there are other equivalent concepts such `has_derivative` and `differentiable_on`):
```
   [diffl]  Definition      
      ⊢ ∀f l x. (f diffl l) x ⇔ ((λh. (f (x + h) − f x) / h) -> l) 0

   [differentiable]  Definition      
      ⊢ ∀f x. f differentiable x ⇔ ∃l. (f diffl l) x
```
However, they are hard to use when users simply want to write expression involving functions and their derivatives such as `f(x) + f'(x)` as in textbooks. Furthermore, there's currently no concept of "higher derivatives" (e.g. `f''(x)`, `f'''(x)`) formalised in HOL.  This PR resolves this issue. This is mainly the work of our student Kai Phan as part of her (in progress) formalisation of the Central Limit Theorem (CLT, of Probability Theory). I moved her code into `limTheory` with necessary minor fixes and renames.

By using the choice operator, it's possible to define function-like higher-order derivatives (and then higher-order differentiable):

```
   [diffn_def]  Definition      
      ⊢ (∀f x. diffn 0 f x = f x) ∧
        ∀m f x. diffn (SUC m) f x = @y. (diffn m f diffl y) x
   
   [higher_differentiable_def]  Definition      
      ⊢ (∀f x. higher_differentiable 0 f x ⇔ T) ∧
        ∀m f x.
          higher_differentiable (SUC m) f x ⇔
          (∃y. (diffn m f diffl y) x) ∧ higher_differentiable m f x
```
Note that the constant is named "diffn" instead of "diff" to prevent potential naming conflicts with unknown user code. Besides, `diff` is already a constant in the near `polyTheory`.

With `diffn`, the usual additivity and chain formula, etc. for (first order, i.e. `diffn 1`) derivatives now has the following more elegant form:
```
   [diffn_add]  Theorem      
      ⊢ ∀f g.
          (∀t. higher_differentiable 1 f t) ∧
          (∀t. higher_differentiable 1 g t) ⇒
          (diffn 1 (λx. f x + g x) = (λx. diffn 1 f x + diffn 1 g x))

   [diffn_mul]  Theorem      
      ⊢ ∀f g.
          (∀t. higher_differentiable 1 f t) ∧
[...]

PR #1654, opened 2025-09-17
More theorems of Higher Order Derivatives and Differentiability

KaiPhan:
Hi,

This extends PR [#1646](https://github.com/HOL-Theorem-Prover/HOL/pull/1646). 
In `limTheory` I add the zero/constant higher-derivative facts and a small reflection lemma for `x |-> f (a - x)` . 

```
[higher_differentiable_neg_sub]  Theorem
∀n f a.
      (∀x. higher_differentiable n f x) ⇒
      ∀x. higher_differentiable n (λx. f (a − x)) x
```

In `transTheory` I complete the symmetric side of Taylor: 
- the case `x < a` (`TAYLOR_THEOREM_NEG`), 
- its **SIGMA form** (`TAYLOR_THEOREM_NEG'`), 
- and a general version that covers both sides (`TAYLOR_THEOREM_ALL_LT`). 

All Taylor results are derived from the existing `MCLAURIN_NEG` / `MCLAURIN_ALL_LT`.

--Kai

### DIFF
--- a/src/real/analysis/limScript.sml
+++ b/src/real/analysis/limScript.sml
@@ -2677,3 +2677,64 @@ Proof
  >> qexists ‘n'’ >> rw [Abbr ‘g’]
  >> MATCH_MP_TAC higher_differentiable_imp_n1 >> simp []
 QED
+
+Theorem higher_differentiable_0 :
+    ∀n x. higher_differentiable n (λx. 0) x
+Proof
+    Induct_on ‘n’ >- (gs [higher_differentiable_def])
+ >> rw [higher_differentiable_def, FORALL_AND_THM]
+ >> qexists ‘0’ >> rw []
+ >> Induct_on ‘n’ >- (gs [higher_differentiable_def, DIFF_CONST])
+ >> rw [GSYM diffn_SUC, diffn_const]
+ >> ‘∀x. higher_differentiable n (λx. 0) x’
+   by (rw [] >> MATCH_MP_TAC higher_differentiable_mono >> qexists ‘SUC n’ >> gs [])
+ >> gs [higher_differentiable_def]
+QED
+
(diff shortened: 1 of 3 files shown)


binghe:
Let's do some changes before users start to use these theorems:

1. We should NOT create any new theorem based on the old `sum (0,n)`. Instead, all new theorems should use the new `SIGMA`-based statements. This means, the theorem `TAYLOR_THEOREM` should be changed into a local lemma with names like `TAYLOR_THEOREM_lemma[local]`, and the theorem `TAYLOR_THEOREM'` should be renamed to `TAYLOR_THEOREM`.  Please do this for all your other theorems and in your CLT proof we should only deal with `SIGMA` instead of `sum (0,n)`.

2. Note that your antecedent `(∀m t. m < n ∧ a ≤ t ∧ t ≤ x ⇒ higher_differentiable (SUC m) f t)` is equivalent to the shorter form `(∀t. a ≤ t ∧ t ≤ x ⇒ higher_differentiable n f t)`, due to `higher_differentiable_mono`.  This is what I did when proving `TAYLOR_THEOREM'` from your `TAYLOR_THEOREM`.  Please make sure all variants of Taylor theorems added in this and the previous PRs have the shorter antecedents.

3. The word "THEOREM" in the theorem names `TAYLOR_THEOREM_ALL_LT`, etc. looks weird and redundant. I tend to follow the naming conventions of the existing `MCLAURIN` and only name it as `TAYLOR_ALL_LT`.  That is, the word "THEOREM" gets removed from all variants of Taylor theorems added recently.

KaiPhan:
Thanks a lot for the review! I’ve updated the PR.

mn200:
Thanks for your work on this!

PR #1657, opened 2025-09-21
Use Theory syntax in examples

dnezam:


mn200:
Wow!  Thanks for this.

PR #1659, opened 2025-09-22
Rename theory while to while_

dnezam:
In the long term, we may want to drop the "Theory" suffix from theories, which would cause issues with `whileTheory`, as `while` is a keyword in Standard ML.

Less importantly, the HOL + SML parser does not like the fact that there is a keyword in the list of ancestors, since already with the Theory syntax, the "Theory" suffix is not mentioned anymore.

Appending a keyword with "_" seems to be a common solution (see PEP 8). An additional upside is that if we go with snake case as the naming convention for theories (as is mostly the case in CakeML), `while_` would be valid (in contrast to `While`).
Also, `functor` was already renamed to `functor_`.

### DIFF
--- a/Manual/Description/theories.stex
+++ b/Manual/Description/theories.stex
@@ -5005,11 +5005,11 @@ It is a curious fact that higher order logic, although a logic of
 total functions, allows the definition of functions that don't
 seem total, at least from a computational perspective. An example
 is \holtxt{WHILE}-loops. The following equation is derived in theory
-\theoryimp{while}:
+\theoryimp{while\_}:
 %
 \begin{hol}
 \begin{alltt}
->>__ open whileTheory
+>>__ open while_Theory
 ##thm WHILE
 \end{alltt}
 \end{hol}
@@ -5104,7 +5104,7 @@ If one wants a specified result even when the predicate is everywhere false, the
 \end{alltt}
 \end{hol}
 
-More theorems for reasoning about \holtxt{LEAST}, \holtxt{OLEAST} and \holtxt{WHILE} may be found in theory \theoryimp{while}.
+More theorems for reasoning about \holtxt{LEAST}, \holtxt{OLEAST} and \holtxt{WHILE} may be found in theory \theoryimp{while\_}.
 
 
 %\section{Partial orders}
--- a/doc/next-release.md
+++ b/doc/next-release.md
@@ -42,6 +42,7 @@ Incompatibilities
 
 -   `Preterm.eq` has been replaced with `Preterm.veq` which returns `true` if and only if the two arguments are variables with the same names and types.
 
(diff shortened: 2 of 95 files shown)


mn200:
I am not keen on theory names ending with trailing underscore characters and would rather see `While` and `Functor`.

dnezam:
I see, thanks for the feedback. I'll reroll this PR into changing `while` into `While`, and `functor_` into `Functor`.

binghe:
I'm OK with `FunctorTheory` because it's under `examples/` but I hope we avoid capitalized theory names in core theories, to be consistent with (most) others. Since the `WHILE` in `whileTheory` is Hoare logic's ["while" rule](https://en.wikipedia.org/wiki/Hoare_logic#While_rule), I suggest we rename `whileTheory` to `hoareTheory` or `hoare_logicTheory`. In this way, more related contents from Hoare Logic can safely go into it in possible future.

dnezam:
> [.. ] but I hope we avoid capitalized theory names in core theories, to be consistent with (most) others.

Agreed. I would be happiest if we could follow snake case in theory names everywhere. Instead of `Functor`, I would suggest `cat_functor`; `cat` seems to already be part of some theory names.

@binghe suggestions sound good if we want to keep it general. In case we want to keep it more specific, I would suggest `while_loop` or `hoare_while`.

binghe:
`while_loop` looks good to me. If `cat` actually means "category" (not very sure), I suggest `category_functor` instead of `cat_functor`. It's meaningless to save those 5 characters...

dnezam:
Yeah, I don't mind having a longer name.
To summarize, the proposal is to apply the following changes to the theory names:

`while` -> `while_loop`
`functor_` -> `category_functor`

Personally, I am quite happy with these names. If @mn200 signs off on this, I can work on these changes.

mn200:
I don't mind `category_functor`, but I already have a change of my own to make the `while` to `While` fix (along with a change to the kernel to catch and raise as error use of reserved words for theory names).

The use of capital letters here is not unprecedented:
```
./1/ConseqConvScript.sml
./HolSmt/HolSmtScript.sml
./opentheory/compat/OpenTheoryFunctionScript.sml
./opentheory/compat/HOL4combinScript.sml
./opentheory/compat/OpenTheoryRelationScript.sml
./opentheory/compat/OpenTheoryBoolScript.sml
./opentheory/compat/HOL4boolScript.sml
./integer/OmegaScript.sml
./integer/DeepSyntaxScript.sml
./string/ASCIInumbersScript.sml
./Boolify/src/EncodeScript.sml
./Boolify/src/EncodeVarScript.sml
./Boolify/src/DecodeScript.sml
./Boolify/src/CoderScript.sml
```

mn200:
See fe3ce020a

mn200:
And 0a5b891

PR #1661, opened 2025-09-23
Lift out some calls to PURE_REWRITE_CONV

ordinarymath:
This reduces the number of SPEC_ALL calls originating from PURE_REWRITE_CONV.

### DIFF
--- a/src/list/src/listSimps.sml
+++ b/src/list/src/listSimps.sml
@@ -266,6 +266,8 @@ local
 
    val APPEND_LEFT_ASSOC_CONV =
       PURE_REWRITE_CONV [GSYM listTheory.APPEND_ASSOC]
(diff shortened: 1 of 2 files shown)


mn200:
I wonder if the failures are caused by timeouts on trying to do doc-files translation.  @binghe : could you look into updating the CI environment with GNU `parallel`, which may speed things up?

binghe:
> I wonder if the failures are caused by timeouts on trying to do doc-files translation. @binghe : could you look into updating the CI environment with GNU `parallel`, which may speed things up?

OK, I'm rebuilding docker CI images with the `parallel` package installed. I will send a PR with updated docker definitions once finished (needs several hours due to multi-architecture builds).

mn200:
Thanks!

PR #1662, opened 2025-09-24
normalize the result of case_rwlist and cache it.

ordinarymath:
A huge amount of time is spent calling SPEC_ALL when adding the results of case_rwlist to the rewrites or simpset. By caching this expensive operation is now only done whenever the TypeBase is updated.

### DIFF
--- a/src/basicProof/BasicProvers.sml
+++ b/src/basicProof/BasicProvers.sml
@@ -696,17 +696,45 @@ fun PURE_FULL_CASE_TAC (g as (asl,w)) =
  in Cases_on `^t` end g;
 
 local
-  fun tot f x = f x handle HOL_ERR _ => NONE
-in
+
+fun tot f x = f x handle HOL_ERR _ => NONE
+
 fun case_rws tyi =
     List.mapPartial I
        [Lib.total TypeBasePure.case_def_of tyi,
         tot TypeBasePure.distinct_of tyi,
         tot TypeBasePure.one_one_of tyi]
 
+fun decompose (th::res) =
+   let val th = SPEC_ALL th
+       val t = concl th
+   in
+     if is_conj t then (let val (th1,th2) = CONJ_PAIR th
+                        in decompose (th1::th2::res) end)
+     else (th::decompose res)
+   end
+  | decompose [] = []
+
+fun raw_case_rwlist tb =
+  decompose (itlist (fn tyi => fn rws => case_rws tyi @ rws)
+        (TypeBasePure.listItems tb) []);
+
+val cache = ref (NONE: (TypeBase.typeBase * thm list) option)
+
+in
+
 fun case_rwlist () =
- itlist (fn tyi => fn rws => case_rws tyi @ rws)
-        (TypeBase.elts()) [];
+    case (!cache) of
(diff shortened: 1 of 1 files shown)


ordinarymath:
This **significantly speedups** my benchmark of word_to_stackProof in CakeML and is pretty much the main calls of SPEC_ALL in that file.

tanyongkiam:
Could you be more precise about what "significantly speedup" means quantitatively?

mn200:
Seems very plausible; I'm not sure I realised that this is what EVERY_CASE_TAC was doing (because “**no-one should ever use that tactic**” :) ; maybe with this change it won't be so egregious.  (Exact numbers _would_ be good.)

src/basicProof/BasicProvers.sml
```diff
@@ -696,17 +696,45 @@ fun PURE_FULL_CASE_TAC (g as (asl,w)) =
...
+
+fun tot f x = f x handle HOL_ERR _ => NONE
+
 fun case_rws tyi =
     List.mapPartial I
        [Lib.total TypeBasePure.case_def_of tyi,
         tot TypeBasePure.distinct_of tyi,
         tot TypeBasePure.one_one_of tyi]
 
+fun decompose (th::res) =
```
mn200: This code is simple enough that it's not a huge deal, but I was wondering why not just call `Cond_rewr.mk_cond_rewrs` or similar?

src/basicProof/BasicProvers.sml
```diff
@@ -696,17 +696,45 @@ fun PURE_FULL_CASE_TAC (g as (asl,w)) =
...
+        (TypeBasePure.listItems tb) []);
+
+val cache = ref (NONE: (TypeBase.typeBase * thm list) option)
+
+in
+
 fun case_rwlist () =
- itlist (fn tyi => fn rws => case_rws tyi @ rws)
-        (TypeBase.elts()) [];
+    case (!cache) of
```
mn200: You should use the TypeBase's listener API  (`register_update_fn`) so that you can see the new types arrive and add them to your cache rather than having to rebuild it every time.  The stateful simpset does the same.   (The ability to actually change the `tyinfo` value before it gets incorporated is interesting; but I'm not sure it's used, and if we dropped it, we could make the TypeBase code just use `Listener`.)

ordinarymath: Is there a way to listen for when a type is deleted?

mn200: If we made this an actual listener, the listener would be on a change type that included addition and deletion of info values.  I don’t think there are many deletions in non-interactive uses, and having out-dated stuff in a cache isn’t likely to be much of a problem.

src/basicProof/BasicProvers.sml
ordinarymath: It's being called by the REWRITE_TAC and the simplifier which have different behavior and only the SPEC_ALL call is expensive.

ordinarymath: Wait I just realized BODY_CONJUNCTS does what I want will change to using it
edit: done

ordinarymath:
@mn200 another ping on this?

mn200:
Looks good!

PR #1665, opened 2025-09-26
Add UNCURRY_EQ to AllCaseEqs,

ordinarymath:
Also add a UNCURRY_PRED theorem and add it to AllCasePreds


mn200:
Seems like your storage of additional case-Eq theorems, and the functions for adding them, should be using ThmSetData.

PR #1668, opened 2025-09-27
Add longmulTheory

myreen:
Multiplication of two 64-bit words results in numbers that need 128-bit words to not truncate the result. This script shows how one can use only operations over 64-bit words and still compute the complete 128-bit result (as two 64-bit words).

One can see this script as providing a recipe for how CakeML's [LongMul](https://github.com/CakeML/cakeml/blob/9948c0a3282205d92e6637d9915a5f4b107b93e2/compiler/encoders/asm/asmScript.sml#L90C11-L90C18) instruction can be implemented on architectures where no such instruction exists.

### DIFF
--- a/examples/machine-code/multiword/longmulScript.sml
+++ b/examples/machine-code/multiword/longmulScript.sml
@@ -5,7 +5,7 @@
   the complete 128-bit result (as two 64-bit words).
 
   Main definition: longmul64_def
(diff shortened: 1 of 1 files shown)


mn200:
I'm not sure this will ever get checked in a context that doesn't fail given the fail on develop (as of 63e7cae504bc), so will merge this.

PR #1669, opened 2025-09-29
vim: fix issue with `>>tac`

wildptr:
Make vim correctly strip leading `>>` without following space.

See https://hol.zulipchat.com/#narrow/channel/509224-.F0.9F.93.9D-Text-Editors/topic/vim.3A.20stripping.20leading.20tacticals/with/540800755


mn200:
Thanks for the Zulip link; based on reactions there, I figure this is fine to merge.  I'll give @xrchz some time to weigh in.

PR #1671, opened 2025-09-30
[real_sigma] optimized proof of jensen_pos_convex_SIGMA, etc.

binghe:
Hi,

The proofs of the following two theorems (Jensen's Inequality) in `real_sigmaTheory` were very slow, even when they were still under `examples/miller`:

```
    [jensen_concave_SIGMA]  Theorem      
      ⊢ ∀s. FINITE s ⇒
            ∀f g g'.
              SIGMA g s = 1 ∧ (∀x. x ∈ s ⇒ 0 ≤ g x ∧ g x ≤ 1) ∧
              f ∈ concave_fn ⇒
              SIGMA (λx. g x * f (g' x)) s ≤ f (SIGMA (λx. g x * g' x) s)
   
   [jensen_convex_SIGMA]  Theorem      
      ⊢ ∀s. FINITE s ⇒
            ∀f g g'.
              SIGMA g s = 1 ∧ (∀x. x ∈ s ⇒ 0 ≤ g x ∧ g x ≤ 1) ∧
              f ∈ convex_fn ⇒
              f (SIGMA (λx. g x * g' x) s) ≤ SIGMA (λx. g x * f (g' x)) s
```

I finally took a look at their proof in details, and found that it's due to an inefficient use of the condition simplifier, where it repeatedly attempts to prove complicated (false) antecedents in assumptions. Instead of saying the simplifier is bad, I tend to accuse the ineffient way done by the proof and have fixed the proof to avoid trapping into loops insider the simplifier.

After reworking parts of these proofs, the building time of `real_sigmaTheory` has reduced from 48s to 19s, in Docker CI builds.

P.S. If anyone managed to build HOL with PolyML running in byte-code mode, the above mentioned bad proofs will cause the building time of `real_sigmaTheory` totally unacceptable -- after 7 hours it's still not finishing at this file.

--Chun

### DIFF
--- a/src/real/real_sigmaScript.sml
+++ b/src/real/real_sigmaScript.sml
@@ -3461,12 +3461,24 @@ Proof
     rw [REAL_SUM_IMAGE_sum, SUM_EQ']
 QED
 
-val REAL_SUM_IMAGE_IN_IF_ALT = store_thm
-  ("REAL_SUM_IMAGE_IN_IF_ALT",``!s f z:real.
-         FINITE s ==> (REAL_SUM_IMAGE f s = REAL_SUM_IMAGE (\x. if x IN s then f x else z) s)``,
-  RW_TAC std_ss []
-  >> MATCH_MP_TAC REAL_SUM_IMAGE_EQ
-  >> RW_TAC std_ss []);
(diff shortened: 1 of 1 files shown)


mn200:
Nice improvement!

PR #1675, opened 2025-10-01
Replace refs to whileTheory

tanyongkiam:
This gets rid of references to "whileTheory" and "whileScript" (grep shows none left).

In addition, it also replaces loads of "while" theories that I heuristically found within 5 lines of "Theory", "Ancestor" or "open".


mn200:
Thanks for these!

PR #1676, opened 2025-10-01
Use Definition syntax in `src` and `examples`

dnezam:
The bulk of the rewrites have been done using a script I wrote based on a HOL + SML parser that lives on a branch on my fork.

Essentially, it transforms something like
```
val foo = Define `bar = T`
```
```
Definition foo: bar = T
End
```
Note how we took the name of the local SML name (`foo`) instead of trying to parse the quote (which would give us `bar`), as that is much easier. In the majority of the cases, the local SML name and the one produced by `Define` match up; hopefully the cases where they don't have been caught by `-t3`.

I tried to keep the changes as minimal as possible. In particular, I did not attempt to do much "styling" beyond perhaps removing an empty line between the `Definition` line and the quote or replacing the start quote with a space instead of deleting it (which keeps alignment for single-line definitions).

Some files redefined `Define`, defined various derivatives, or set some flags. I tried to get rid of as many of these cases as possible, which mainly consisted of removing `Defn.def_suffix` and `computeLib.auto_import_definitions`. For the latter, I tried my best to use `Define` and `zDefine` (and add `nocompute` to existing `Definition` usages) to account for this.

There are still some usages of `Define` left; those were a bit too non-trivial, and this PR is already far too large anyway.


mn200:
Many thanks!

PR #1678, opened 2025-10-02
HolSmt: replace SPECL with INST where possible

someplaceguy:
The latter is much faster when dealing with big terms.

Thanks to @ordinarymath for finding the problem and suggesting the solution!


mn200:
Thanks for this!

PR #1680, opened 2025-10-03
Prove alternative set_byte and word_of_bytes thms

talsewell:
If we know the byte being set is zero, then set_byte is equivalent to just a bitwise-or of the byte (cast and shifted into place) and the original contents. In particular, word_of_bytes can be shown to be just a bitwise-or of all the bytes being inserted, assuming we're not in a funny overflow case.

This makes at least some downstream proofs about word_of_bytes much easier.

This code was copied across from a scratch context in an earlier version of HOL4, hopefully it just builds in this context, but I haven't even tested that yet.

### DIFF
--- a/src/n-bit/byteScript.sml
+++ b/src/n-bit/byteScript.sml
@@ -96,6 +96,19 @@ Proof
   rw[set_byte_def,byte_index_def]
 QED
 
+Theorem set_byte_eq_or:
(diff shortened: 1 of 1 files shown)


xrchz:
Looks like it builds!

xrchz:
I guess you might want to include "F" or "le" in the theorem name since it's specialised to little-endian. (Alternatively, generalise it to big-endian too.)

mn200:
Thanks!

PR #1682, opened 2025-10-04
Replace many `save_thm` instances with `Theorem` in `src`

dnezam:
Finally, a PR with a much more reasonable size :)

I've also manually removed some instances of `save_thm`, which were either behind a `map` or within a `local` or `let`. 
In general, I don't think much is gained from using `map save_thm` instead of using `Theorem`.
Regarding `local`, `Theorem`s right before the end need a `;`, otherwise `unquote` seems to get confused on where its end is, leading to syntax problems.
If `let` is used on the right-hand side of `Theorem`, it seems that it is necessary to parenthesize it.


dnezam:
Oops, forgot to push my local changes...

dnezam:
fwiw, `bin/build -t3` worked with b685546e128b1c84ada4c1e2458e3fbffd268fb9

mn200:
Will wait for CI to check that my merge has done the right thing.

dnezam:
Thanks for resolving the conflict!

PR #1683, opened 2025-10-06
Replace many instances of (Q.)new_definition with Definition (nocompute)

dnezam:
The commit 87b93cea503896614da80552e06803554d8a6b1c (Fix errors) probably deserves to be looked at in a bit more detail, as I had to remove the quantifiers from `SUBMAP_DEF`.


dnezam:
`bin/build -t3` works locally.

dnezam:
The latest force push just fixes a commit message.

mn200:
Thanks!  The need to drop the universal quantifiers is a side effect of the way `Hol_defn` & friends take the quotation apart for figuring out "stuff" before definitions actually get made.

PR #1684, opened 2025-10-07
Remove more `new_definition` and some `Term` in `src`

dnezam:
Previous rewrites did not account for terms of the form
```
Term `...`
```

In this PR, I removed `new_definition` from `src` where possible. In the cases where it wasn't (because we are too early for `TotalDefn`), I replaced `Term` with double quotes, which should be equivalent.


dnezam:
Passes `bin/build -t3` locally (based on 5181e7d077e8867fc9c3bf74fd7aa7dd82f17821)

mn200:
Thanks!

PR #1686, opened 2025-10-08
Add new BISIM_REL theorems and cleanup old syntax

FieryIceStickie:
- Update lemma2, lemma3, lemma3' to use Theorem syntax
- Add BISIM_REL_strong, BISIM_REL_sym, and BISIM_REL_sym_strong


mn200:
I'm not super keen on using `_def` suffix for things that are not definitions.  In this sort of situation, I'd prefer `_thm` as the suffix.

mn200:
Thanks!

PR #1690, opened 2025-10-11
Improve HolSmt's Z3 proof replay

someplaceguy:
This PR contains improvements in replaying Z3 proofs. It also adds a new function to `Parse`:

* Implement a `term_to_string_by_grammar` (similar to `print_term_by_grammar`)
* HolSmt: use custom grammar for printing terms

    Otherwise, converting a term to a string would cause a hang for
    certain huge terms, even when setting `max_print_depth` to a low
    value.

* HolSmt: handle more Z3 rewrite rules
* HolSmt: handle `BBLAST_PROVE` exception

    `BBLAST_PROVE` by default raises a `Match` exception when a counterexample is found.

    However, the counterexample may be spurious due to `BBLAST_PROVE` assigning some
    value to a Z3 variable, even though the Z3 variable may have a different (and
    yet-to-be-known) value.

    To fix this issue, we temporarily configure `BBLAST_PROVE` to not try generating
    counterexamples (there's no point in doing so when replaying a Z3 proof) and
    then catch the `HOL_ERR` exception as usual so we can continue trying other
    proving strategies.

* HolSmt: bump supported Z3 and cvc5 versions

    cvc5 1.3.0 required bumping the supported SMT-LIB version to prevent
    it from spamming the self-test output with warning messages.

    Unfortunately, this cvc5 version also causes a regression in one of
    the self-tests, thus it was disabled for cvc5.

### DIFF
--- a/Manual/Description/HolSmt.tex
+++ b/Manual/Description/HolSmt.tex
@@ -200,8 +200,8 @@ delete temporary files automatically, \eg, when \HOL{} exits).
 
 \subsection{Installing SMT solvers}
 
-\ml{HolSmtLib} has been tested with cvc5~1.1.2, Yices~1.0.40, and
-Z3~4.12.6. Later versions may or may not work.  (Yices~2 is not
+\ml{HolSmtLib} has been tested with cvc5~1.3.0, Yices~1.0.40, and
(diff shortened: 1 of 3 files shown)


mn200:
Great stuff!  Thanks.

PR #1691, opened 2025-10-13
Fix strip_binder_conv

ordinarymath:
There's no point in handling failures that occured while running the conversion


ordinarymath:
@mn200 I think behavior will only change if conversions that are not a conversion are not passed in. Do I document this change?

mn200:
Thanks for this.  The documentation says it should fail if `conv body` fails, so I think the code achieves that better now, and the docs can be left as is.

mn200:
Thanks again.

PR #1692, opened 2025-10-13
Replace many (Q.)prove instances with Theorem ..[local] in src

dnezam:
Commits that deserve an extra look: 0a795fedf42d7475abffe6ecd6f34f5537c5b7a1 and 97c6a970b38c86fe3d71372ba59666550ad43de6.

Their commit messages highlight some changes that I manually applied.

### DIFF
--- a/src/relation/bisimulationScript.sml
+++ b/src/relation/bisimulationScript.sml
@@ -82,6 +82,48 @@ Proof
       PROVE_TAC [BISIM_def] ]
 QED
 
+Theorem BISIM_REL_sym:
+  symmetric (BISIM_REL ts)
+Proof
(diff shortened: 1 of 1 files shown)


dnezam:
Might conflict with #1686, so perhaps it is better to merge that one first.

dnezam:
`bin/build -t3` works locally

src/pred_set/src/pred_setScript.sml
mn200, line 1: This code is a bit gross though it achieves a useful effect.  It should be reimplemented using a TheoryDelta listener.

mn200: Thanks for making the simple workaround!

PR #1696, opened 2025-10-15
Replace many Hol_datatype instances with Datatype

dnezam:
It turned out to be more work than expected since I ended up manually removing `of` and `=>` from quotations.


mn200:
Thanks!  This is a big one!

PR #1697, opened 2025-10-16
Refactor Term.compare to not allocate memory

ordinarymath:
PolyML does not optimize Term.compare into a multiple argument function when `(p as (..,..))` is present. This means that in the comb case, Term.compare has to allocate 2 tuples. The pattern `case (..,..) of` is a more common optimization which is indeed done by PolyML.


mn200:
Thanks for this!

PR #1698, opened 2025-10-17
More real random variable and expectation theorems

KaiPhan:
This commit adds several supporting theorems about real random variable and expectation, which will be used in later parts of the CLT formalisation.

### DIFF
--- a/src/probability/probabilityScript.sml
+++ b/src/probability/probabilityScript.sml
@@ -2356,7 +2356,7 @@ Proof
  >> MATCH_MP_TAC abs_pow_le_mono >> art []
 QED
 
-Theorem integrable_finite_expectation :
+Theorem integrable_absolute_moments_mono :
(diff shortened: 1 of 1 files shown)


src/probability/probabilityScript.sml
```diff
@@ -2270,6 +2356,25 @@ Proof
  >> MATCH_MP_TAC abs_pow_le_mono >> art []
 QED
 
+Theorem integrable_finite_expectation :
```
binghe: This theorem is about the inclusive of absolute moments of random variable (if order n integrable than all m < n orders integrable too) but the current theorem seems irrelevant. Please choose a new name.

KaiPhan: Thanks! I’ve renamed it to integrable_absolute_moments_mono

mn200:

PR #1701, opened 2025-10-21
Borel approximation of Lebesgue-measurable sets

binghe:
Hi,

This PR is an important step before we finally obtain the equivalence between Lebesgue and gauge integrations (for Borel-measurable functions intergrable in both of them). Let be briefly describe this work from the beginning.

In the old `examples/probability/lebesgue_measureTheory`, HVG people ever defined the "Lebesgue" measure space by gauge integration of indicator functions over "lines", i.e. closed intervals in form [-n,n].
```
   [lebesgue_def]  Definition      
      ⊢ lebesgue =
        (𝕌(:real),{A | ∀n. indicator A integrable_on line n},
         (λA. sup {Normal (integral (line n) (indicator A)) | n ∈ 𝕌(:num)}))
```
This definition uses only concepts from `integrationTheory`, however, it's indeed a valid measure space, and the measurable sets here, namedly "Lebesgue-measurable sets", (magically) *contains* all Borel sets:
```
   [measure_space_lebesgue]  Theorem
      ⊢ measure_space lebesgue

   [lborel_subset_lebesgue]  Theorem      
      ⊢ measurable_sets lborel ⊆ measurable_sets lebesgue
```

It's known that there are much more Lebesgue(-measurable) sets than Borel(-measurable) sets. In fact, the cardinality of all Borel sets is the same with `univ(:real)`, while the cardinality of Lebesgue sets is the same with `univ(:real set)`, a larger cardinal. (This formal proof is hard, not done yet in HOL4 and nobody needs it.)

Although there are much more Lebesgue sets, for any of them, it's a bit surprising that we can find a corresponding Borel set, whose measure is very close (by a given small positive number) to it. Below is the basic "approximation" theorem which first finds a countably sequence of closed intervals (Borel sets for sure) which covers the given Lebesgue set:
```
   [approximation_thm]  Theorem
      ⊢ ∀E e.
          E ∈ measurable_sets lebesgue ∧ 0 < e ⇒
          ∃J. (∀i. closed_interval (J i)) ∧
              (∀i j. i ≠ j ⇒ nonoverlapping (J i) (J j)) ∧
              E ⊆ BIGUNION (IMAGE J 𝕌(:num)) ∧
              lmeasure E ≤ suminf (lmeasure ∘ J) ∧
              suminf (lmeasure ∘ J) ≤ lmeasure E + Normal e
```
By doing a big union to all the asserted Borel sets, the next theorem find a single Borel set with the same property:
```
   [lebesgue_approximation]  Theorem      
      ⊢ ∀E e.
          E ∈ measurable_sets lebesgue ∧ 0 < e ⇒
          ∃s. s ∈ subsets borel ∧ E ⊆ s ∧ lmeasure E ≤ lambda s ∧
              lambda s ≤ lmeasure E + Normal e
```
Furthermore, for any Lebesgue "negligible" set (whose measure is zero), we can find another Borel null set which is super set of it:
```
   [negligible_approximation]  Theorem      
      ⊢ ∀E e.
          negligible E ∧ 0 < e ⇒
          ∃s. s ∈ subsets borel ∧ E ⊆ s ∧ lambda s ≤ Normal e
   
   [negligible_approximation_null_set]  Theorem      
      ⊢ ∀E. negligible E ⇒ ∃s. s ∈ null_set lborel ∧ E ⊆ s
```

These proofs are hard, in the sense that, proof steps involving both integration systems are used together in the same proof, with also deep real topology, countable sets results. `integerTheory` and `intrealTheory` are used. And I also depend on some lemmas from `lift_ieeeTheory` (I didn't move these lemmas to `realTheory` to repect the original author.)

With this big preparation, now the equivalence between Lebesgue and gauge integration should follow shortly.

--Chun

### DIFF
--- a/src/probability/lebesgue_measureScript.sml
+++ b/src/probability/lebesgue_measureScript.sml
@@ -3274,9 +3274,6 @@ Proof
  >> MATCH_MP_TAC REAL_LE_DIV >> simp [POW_POS]
 QED
 
-val _ = export_theory ();
-val _ = html_theory "lebesgue_measure";
-
 (* References:
 
   [1] Schilling, R.L.: Measures, Integrals and Martingales (Second Edition).


binghe:
Schermata 2025-10-21 alle 16 51 04

I have moved `lebesgue_measureTheory` from `examples/probability` to `src/probability`, thus into core library. This is because: 1) the `lebesgue_measureTheory` does not depend on `probabilityTheory` (and `martingaleTheory`), 2) it's built in parallel with `martingaleTheory`, and therefore has no effect in the total building time of HOL (when in parallel). 3) It's current size is over 3K lines, almost the same with `real_borelTheory`, and will keep grow.

mn200:
Thanks for this.  I've fixed the issue with otknl elsewhere.

PR #1703, opened 2025-10-24
Replace Triviality with Theorem with local tag

dnezam:
As democratically decided on Zulip using a poll :)


mn200:
Can you change the manuals (principally DESCRIPTION, I should think) too please?

mn200:

PR #1706, opened 2025-10-26
Replace new_recursive_definition with Definition

dnezam:


mn200:
I don't care about `FCONS` because it's only used internally to datatype construction, but have you checked the `EVAL` situation with the other two?  The definitions look as if they could be used fine, but if there's other code already making them `EVAL`-able then that should presumably be adjusted.

dnezam:
It slipped my mind that I need to add `[nocompute]` here, sorry. I'll add those tags tomorrow (or well, today after waking up)

ordinarymath:
https://github.com/HOL-Theorem-Prover/HOL/blob/875095ab2782f1a84d1d29eb07d7b7accd3a1644/src/algebra/base/combinatoricsScript.sml#L1881-L1888
for PROD it's compute is added later.

The stuff in list you can probably modify
https://github.com/HOL-Theorem-Prover/HOL/blob/875095ab2782f1a84d1d29eb07d7b7accd3a1644/src/list/src/listScript.sml#L5559-L5567

mn200:
Awesome; thanks!

PR #1708, opened 2025-10-26
Replace Datatype call with modern syntax

dnezam:


dnezam:
`bin/build -t3` passes locally

PR #1709, opened 2025-10-27
Add STRIP_ALL_THEN and CHOOSE_ALL_THEN

ordinarymath:
Add STRIP_ALL_THEN and CHOOSE_ALL_THEN

Only STRIP_ALL_THEN is exposed in the signature.

STRIP_ALL_THEN should behave like REPEAT_TCL STRIP_THM_THEN
and
CHOOSE_ALL_THEN should behave like REPEAT_TCL CHOOSE_THEN

This change is intended to speedup record construction time
where STRIP_ALL_THEN is being called via STRUCT_CASES_TAC

### DIFF
--- a/src/1/Thm_cont.sml
+++ b/src/1/Thm_cont.sml
@@ -292,9 +292,8 @@ local
    fun varyAcc v (V, l) = let val v' = gen_variant Parse.is_constname "" V v in (v'::V, v'::l) end
    (* There are actual cases where strip_exists differ from this function *)
    fun strip_exists1 tm =
-   let fun dest_exist_opt tm = SOME (dest_exists tm) handle HOL_ERR _ => NONE
-       fun strip A tm =
-           case dest_exist_opt tm of
+   let fun strip A tm =
+           case Lib.total dest_exists tm of
                NONE => (List.rev A, tm)
              | SOME (x,tm') => strip (x::A) tm'
    in


src/1/Thm_cont.sml
```diff
@@ -260,6 +260,61 @@ val CHOOSE_THEN: thm_tactical =
       end
       handle HOL_ERR _ => raise ERR "CHOOSE_THEN" ""
 
+(* same as REPEAT_TCL CHOOSE_THEN but faster *)
+local
+   fun varyAcc v (V, l) = let val v' = gen_variant Parse.is_constname "" V v in (v'::V, v'::l) end
+   (* There are actual cases where strip_exists differ from this function *)
+   fun strip_exists1 tm =
+   let fun dest_exist_opt tm = SOME (dest_exists tm) handle HOL_ERR _ => NONE
```
mn200: The `dest_exist_opt` function is `Lib.total dest_exists`.

mn200 (changes requested):
Please write a documentation file for `STRIP_ALL_THEN` in `help/Docfiles`.  Use the template or other examples for formatting.

ordinarymath:
I've figured out why this is way slower than using MP with a induction thm + GENL.
https://github.com/HOL-Theorem-Prover/HOL/blob/a902f2db73f6a761d57905ec18eb06291ad8a9b9/src/1/Thm_cont.sml#L104-L133
This bug fix means that the thm passed through the continuation gets bigger and bigger with more hyps and resulting call of CHOOSE has to check for var_occurs for all of the hyps.

src/1/Thm_cont.sml
ordinarymath: I don't like using Lib.total due to it also handling non `HOL_ERR _` errors but in this case it doesn't matter since there isn't going to be non `HOL_ERR _` errors so I can change it if you insist.

ordinarymath:
> I've figured out why this is way slower than using MP with a induction thm + GENL.
> 
> https://github.com/HOL-Theorem-Prover/HOL/blob/a902f2db73f6a761dEVERY_TCL (map X_CHOOSE_THEN vars)57905ec18eb06291ad8a9b9/src/1/Thm_cont.sml#L104-L133
> 
> This bug fix means that the thm passed through the continuation gets bigger and bigger with more hyps and resulting call of CHOOSE has to check for var_occurs for all of the hyps.

This is worked around using a new function called X_CHOOSE_THENL

src/1/Thm_cont.sml
mn200: Please change it; the fewer identifiers in a piece of code the better

mn200 (changes requested):

mn200:
Thanks!

PR #1712, opened 2025-10-27
Clean up "magic" at the top of pred_set

dnezam:
I also took the opportunity to add some documentation and an example.

I was pleasantly surprised about how easy this was, despite it being my first time doing something like this. It would have been even better if I could have hovered over definitions and seen documentation, but I guess that could be added in the long term.


mn200:
I'm glad this was such a pleasant experience!

PR #1713, opened 2025-10-28
Get rid of precondition on LENGTH_TL and make it a default rewrite

ordinarymath:
Also make some other theorems simp

### DIFF
--- a/src/list/src/listScript.sml
+++ b/src/list/src/listScript.sml
@@ -2876,7 +2876,8 @@ Proof
   \\ Q.X_GEN_TAC`z`
   \\ rw[PULL_EXISTS]
   \\ res_tac
-  \\ Cases_on`x` \\ fs[]
+  \\ Cases_on`x`
+  \\ full_simp_tac(arith_ss) []
 QED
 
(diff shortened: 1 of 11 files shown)


src/list/src/listScript.sml
```diff
@@ -1136,10 +1136,10 @@ Proof
 GEN_TAC THEN Induct THEN SRW_TAC[] [FOLDR, MAP]
 QED
 
-Theorem LENGTH_TL:
-   !l. 0 < LENGTH l ==> (LENGTH (TL l) = LENGTH l - 1)
+Theorem LENGTH_TL[simp]:
+   !l. (LENGTH (TL l) = LENGTH l - 1)
```
dnezam: I think it should be possible to remove the parentheses now that the implication is gone.

ordinarymath:
bin/build -t works. -t3 is definitely still broken thou

ordinarymath:
bin/build -t3 works locally

ordinarymath:
The last push was for removing the stuff that I added accidentally

mn200:
Thanks!

PR #1714, opened 2025-10-28
Apply eta reduction to left-hand side of LIST_REL_MAP2

dnezam:
@mn200 and @digama0 agree that the new version is strictly better than the original ([#❓ Support > Change LIST_REL_MAP2](https://hol.zulipchat.com/#narrow/channel/494306-.E2.9D.93-Support/topic/Change.20LIST_REL_MAP2/with/542254591)).

I've made the change, but I am curious how the eta reduction makes this "strictly better".

### DIFF
--- a/doc/next-release.md
+++ b/doc/next-release.md
@@ -14,7 +14,7 @@ Contents
 -   [Bugs fixed](#bugs-fixed)
 -   [New theories](#new-theories)
 -   [New tools](#new-tools)
--   [New Examples](#new-examples)
+-   [New examples](#new-examples)
 -   [Incompatibilities](#incompatibilities)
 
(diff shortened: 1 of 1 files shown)


konrad-slind:
Because matching variable `R` against a term always succeeds while matching
`\a b. R a b` against a term requires that the term itself be a lambda.
Hence the latter imposes an extra impediment against `LIST_REL_MAP2` firing
as a rewrite rule.

dnezam:
Ah, that makes sense. Thank you for the explanation!

dnezam:
`bin/build -t3` works as well!

mn200:
Thanks!

PR #1715, opened 2025-10-29
Make records with huge number of fields faster

ordinarymath:
STRUCT_CASES_TAC has blowup where every call of Thm.CHOOSE which can't be optimized away


mn200:
Thanks!

PR #1716, opened 2025-10-29
Use simp instead of export_rewrites at some places in src

dnezam:
Some files are pretty finicky when it comes to when exactly the `simps` are added (`integer`, `list`, `pred_set`, and `bitstring` (?)), so some `export_rewrites` remain.

More reasons for `export_rewrites` remaining are:
- I don't think `new_specification` supports tags in its name
- Some custom functions for generating and saving theorems (in particular in `algebra/construction`)
- Apparently `simp` is not a valid tag in early theories such as `num/theories/prim_recScript.sml`
- Adding things to the simpset that were automatically generated or proved somewhere else

### DIFF
--- a/src/num/theories/prim_recScript.sml
+++ b/src/num/theories/prim_recScript.sml
@@ -57,6 +57,7 @@
 Theory prim_rec[bare]
 Libs
   HolKernel boolLib Prim_rec Parse simpLib boolSimps
+  BasicProvers[qualified] (* enable [simp] tag *)
 
 
 type thm = Thm.thm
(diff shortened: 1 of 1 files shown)


dnezam:
`bin/build -t3` works

mn200:
`simp` is a valid tag whenever `BasicProvers` is imported.  In the case of `prim_rec`, if you delete the lines saying `BasicProvers.export_rewrites` and then attempt to use the tag, it won't work because the script will lose its dependency on  `BasicProvers`. You need to add `BasicProvers` to the `Libs` field.

dnezam:
Learned something new, thanks :)

PR #1719, opened 2025-10-31
Replace store_thm calls with Theorem in examples

dnezam:
It feels like I am committing a crime just opening a PR of this size, but here we are. Maybe it makes sense to ask ourselves which files we care about and restrict changes to those?

To be clear, I don't have a problem making these changes; it's just that I don't think anyone can claim to have reviewed this beyond glancing over the diffs.

### DIFF
--- a/examples/vector/matrixScript.sml
+++ b/examples/vector/matrixScript.sml
@@ -923,7 +923,6 @@ Proof
     ASM_SIMP_TAC bool_ss[GSYM MATRIX_COMPOSE, MATRIX_I]]
 QED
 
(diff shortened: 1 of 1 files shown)


dnezam:
`bin/build -t3` works (b91bd71f71137078ab67fbc1d33684805727a5a7, before I removed the comment)

binghe:
The only good thing here (for me, and perhaps for others) is that there's no much room left for such grammar changes, i.e. the peace is finally coming.

mn200:
Our long national nightmare is finally over? :)

dnezam:
I think we are still a few (tens of) thousands of lines away from that :melting_face:

PR #1721, opened 2025-11-02
Refactor congprocs to use the UNCHANGED exception for failure

ordinarymath:
Note that some other minor changes. congprocs that raise non HOL_ERR _ are no longer caught. Also EQ_CONGPROC raises UNCHANGED when enountering atoms e.g. variables or constants. This changes don't affect simplifier behavior only possibly affecting code that directly calls the traverser.

### DIFF
--- a/src/simp/src/simpLib.sml
+++ b/src/simp/src/simpLib.sml
@@ -392,7 +392,7 @@ fun getlimit (SS ss) = #limit ss
        refl x
      end
    in
-     CONGPROC mk_refl th
+     QCHANGED_CONV o CONGPROC mk_refl th
    end
  in
    TRAVRULES {relations = rels,
--- a/src/simp/src/Opening.sml
+++ b/src/simp/src/Opening.sml
@@ -252,7 +252,7 @@ in fn {relation,solver,depther,freevars} =>
     if allunch then
       (* note critical link between this exception and traversal code in
          Traverse.FIRSTCQC_CONV *)
-      raise mk_HOL_ERR "Opening" "CONGPROC" "Congruence gives no change"
+      raise UNCHANGED
     else (trace(3,PRODUCE(tm,"congruence rule",final_thm)); final_thm)
   end
 end;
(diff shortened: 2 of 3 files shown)


ordinarymath:
I've tested the files that uses congLib which afaik is the only thing that calls the traverser other the simplifier and they still work.

src/simp/src/simpLib.sml
mn200, line 1: I'm not sure I like the introduction of QCHANGED_CONV here; it will flip the UNCHANGED generated by your other changes back into a HOL_ERR for "weakened" relations.  This will then allow other congruence rules to have a go at a term that has already been successfully traversed once.

ordinarymath: I've not fully changed the simplifier to use UNCHANGED. The weakener uses HOL_ERR _ to signal UNCHANGED.

ordinarymath: https://github.com/HOL-Theorem-Prover/HOL/blob/a0d42fb732ef4446cc6f34444aafb75a90e2b22d/src/simp/src/Traverse.sml#L135-L149

https://github.com/HOL-Theorem-Prover/HOL/blob/a0d42fb732ef4446cc6f34444aafb75a90e2b22d/src/simp/src/Traverse.sml#L278-L286
This bits of code are relevant for understanding the traverser

mn200: Sure; when working at a particular level, UNCHANGED is not used.  But the result from a cong-proc is seen inside FIRSTCQC_CONV, and if you've changed the UNCHANGED back into a generic HOL_ERR, the FIRSTCQC code will cause another depther to be tried.  (As per the big comment above the implementation.)  EQ_CONGPROC gets to throw a HOL_ERR because it's tried last as the default (when the others have genuinely failed), but I'm not 100% sure weakened ones should be the same.

ordinarymath: Weakeners don't prevent other weakeners from firing it's using FIRST_CONV
https://github.com/HOL-Theorem-Prover/HOL/blob/a0d42fb732ef4446cc6f34444aafb75a90e2b22d/src/simp/src/Traverse.sml#L272

mn200: Ha!  Good point, thanks.

mn200:
Thanks!

PR #1722, opened 2025-11-05
Reduce allocations in user_split_ident

ordinarymath:
This doesn't actually matter in practice but the code doesn't become more complicated


mn200:
Thanks for this; the old code looks like it is retaining some vestige of a path no longer taken.

PR #1725, opened 2025-11-05
Use exn_to_string to display error in Vim plugin

tanyongkiam:
General.exnMessage seems to raise Match on the new HOL_ERR exceptions

Debugged with @dnezam and @ordinarymath


mn200:
Thanks!

PR #1736, opened 2025-11-20
Some more overload tidying for SINGL

ordinarymath:


ordinarymath:
Ok this doesn't work due to inferior not working for overloaded function like definitions.

PR #1738, opened 2025-12-02
reduce calls to srw_ss in lib files 

ordinarymath:
The first commit might be able to be separated out but the second change depends on the first. bin/build of higher levels are probably broken


ordinarymath:
bin/build -t3 works locally

mn200:
thanks!

PR #1740, opened 2025-12-04
Standalone Theorems to Canonise for Sigma Algebras

jdyeager:
These are the results from my large library of results that pertain to sigma algebras and I can argue stand on their own merits (and lemmas/definitions for the cause).

The motif of these results, is theorems that help in proving some property holds for sets in a sigma algebra; the conclusion of two of the theorems being `subsets (sigma sp sts) ⊆ P`. The first such result is a weaker version of `SIGMA_PROPERTY`, but with simpler preconditions that are nicer for working forward in the cases where it might be useful. The second such result is the pi-lambda theorem.

### DIFF
--- a/src/probability/sigma_algebraScript.sml
+++ b/src/probability/sigma_algebraScript.sml
@@ -6075,6 +6075,48 @@ Proof
  >> rw [IN_APP]
 QED
 
+(* ------------------------------------------------------------------------- *)
+(*  Pi-Lambda Theorem                                                        *)
+(* ------------------------------------------------------------------------- *)
+
+(*
+These are the results for algebras from my own accumulated library
+that I believe stand on their own as something useful for future users.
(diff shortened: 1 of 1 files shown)


binghe:
I've been using the following lemma (long existed) to do the same thing as your `SIGMA_PROPERTY_WEAK`:
```
   [SIGMA_SUBSET]  Theorem      
      ⊢ ∀a b.
          sigma_algebra b ∧ a ⊆ subsets b ⇒
          subsets (sigma (space b) a) ⊆ subsets b
```
But your version may be easier to apply sometimes, and its conclusion has the same "shape" with other σ-property theorems (a good thing). I have no objections to merge this PR, thank for your work (and looking forward to see more).

mn200:
Thanks Jared!

PR #1743, opened 2025-12-21
Add etqLib: convenience wrapper for goaltree mode

charles-cooper:
## Summary

`etqLib` provides a convenience wrapper for goaltree mode that compiles tactic strings at runtime, eliminating the need to duplicate tactic names:

```sml
(* Before: must write name twice *)
et ("strip_tac", strip_tac);

(* After: just the string *)
etq "strip_tac";
```

### API

- `etq : string -> proof` - compile and apply tactic (30s compilation timeout)
- `etq_tmo : real -> string -> proof` - same, with custom timeout

### Location

Placed in `src/AI/sml_inspection/` alongside `smlExecute` (which provides `tactic_of_sml`). This is required due to HOL's build order: `src/proofman` is built into `hol.state0` before `src/AI/sml_inspection`, so the dependency can only go this direction.

### Auto-loading

Added to `tools-poly/prelude2.ML` so `etq` is available immediately in interactive sessions. **This is optional** - happy to remove if maintainers prefer explicit loading. Users can alternatively add to `~/.hol-config.sml`:

```sml
load "etqLib"; open etqLib;
```

## Test plan

```sml
gt `p /\ q ==> q /\ p`;
etq "strip_tac";
etq "conj_tac";
etq "first_assum ACCEPT_TAC";
etq "first_assum ACCEPT_TAC";
p();
(* Output: strip_tac \\ conj_tac >- ( first_assum ACCEPT_TAC) >- ( first_assum ACCEPT_TAC) *)
```

### DIFF
--- a/src/AI/sml_inspection/etqLib.sml
+++ b/src/AI/sml_inspection/etqLib.sml
@@ -5,8 +5,23 @@ open HolKernel Abbrev
 
 val ERR = mk_HOL_ERR "etqLib"
 
+fun print_goal g =
+  HOLPP.prettyPrint(print, !Globals.linewidth) (goalStack.pp_goal g)
+
+fun print_goals [] = print "Goal proved.\n"
+  | print_goals [g] = print_goal g
+  | print_goals gs =
+      (print (Int.toString (length gs) ^ " subgoals:\n\n");
+       List.app (fn g => (print_goal g; print "\n")) (List.rev gs))
+
 fun etq_tmo timeout s =
-  proofManagerLib.et (s, smlExecute.tactic_of_sml timeout s)
+  let
+    val pf = proofManagerLib.et (s, smlExecute.tactic_of_sml timeout s)
+    val goals = proofManagerLib.top_goals ()
+  in
+    print_goals goals;
+    pf
+  end
   handle HOL_ERR _ => raise ERR "etq" ("Cannot compile tactic: " ^ s)
 
 val etq = etq_tmo 30.0


tools-poly/prelude2.ML
```diff
@@ -3,6 +3,9 @@ val () = PolyML.print_depth 0;
 
 open bossLib  (* Any others? *)
 
+val _ = load "etqLib";
```
xrchz, line 6: I think this should go into bossLib rather than the prelude if it's gonna be made "auto-loaded".

charles-cooper: claude:
```quote
  The issue is build order - bossLib is in src/boss which is built before bin/hol, while etqLib depends on smlExecute from src/AI/sml_inspection which is built after bin/hol.

  We can't add it to bossLib directly. Options:
  1. Keep in prelude2.ML (runs after heap loads, so deps exist)
  2. Remove auto-load entirely, let users add to ~/.hol-config.sml
```

xrchz: I would probably remove the auto-load of this then. But would like to hear from another reviewer.

charles-cooper: in hol-agents i'm just defining it directly there. it may be better that way, since we can iron out any usage issues before upstreaming

konrad-slind: There's a substantial amount of PolyML-specific stuff in `src/AI/sml_inspection`. At a glance it seems disruptive to pull out the bits used in `smlExecute` and put them in a different place just so `eqtLib` could be put under `bossLib`. On the other hand, `eqtLib` should not depend on code living in `src/AI`, and the actual amount of code   isn't much. So I'd probably copy the parts needed for `eqtLib` and put them in somewhere like `src/portableML/poly` and then implement `eqtLib` under `src/proofman`. This should make it available from `bossLib`, as @xrchz is recommending.

charles-cooper:
not planning to use this any time soon, closing

PR #1746, opened 2025-12-23
Measurable Function Results to Canonise

jdyeager:
A collection of theorem and lemmas therefor from my library of accumulated results.

These theorems fall into three main categories:

- Results about general measurable function that had previously been shown only for Borel measurable functions.
- Results for measurable sets based on `borel` (real-valued) measurable functions.
- Results for `borel` (real-valued) measurable functions.


mn200:
Thanks!

PR #1748, opened 2025-12-24
Speed up dining_cryptos.

ordinarymath:
The speedups comes from preventing automation
from traversing internally in strings.


ordinarymath:
There 2 more theorems like this in dining_cryptos which can in theory be speedup in a similar way but would require modifying the automation of LEAKAGE_COMPUTE_CONV.

PR #1757, opened 2026-01-02
Improve error messages for constructors in listSyntax

ordinarymath:
Most of this code could probably be simplified with the use of HolKernel.syntax_fn{n} but those are stronger and might break stuff which I don't want to fix now.


ordinarymath:
Could @konrad-slind or @mn200 take a look at this?

konrad-slind:
Looks good! Thank you.

PR #1759, opened 2026-01-03
Add a way to replay THEN sequentially more accurately for vim

ordinarymath:
This is bound to hE or whatever your local leader is configured to.


mn200:
Does this still need to be a draft?

mn200:
Please resuscitate if and when it's ready for merging.

PR #1764, opened 2026-01-06
Make computeLib API functional

xrchz:
Main functions now return updated compset, which no longer is an implicit ref. Now the_compset is explicitly a ref (and the functions affecting it are still stateful).

Here's a summary of the core changes made:

  clauses.sig - Updated signatures:
  - add_thms, add_extern, add_thmset, set_skip, scrub_const, scrub_thms now return compset
  - from_term now returns compset * db dterm

  clauses.sml - Implementation changes:
  - Removed outer ref from compset type
  - assoc_clause returns (compset, ref) tuple
  - All modification functions now return updated compsets
  - from_term threads the compset through its continuation-passing style

  computeLib.sig - Public API changes:
  - Modification functions (add_thms, add_conv, set_skip, etc.) now return compset
  - the_compset is now compset ref
  - add_datatype_info returns compset
  - Extenders changed to (compset -> compset) list
  - extend_compset returns compset

  computeLib.sml - Implementation changes:
  - the_compset is now a ref
  - Global functions (add_funs, del_consts, etc.) update the ref
  - EVAL_CONV dereferences the_compset
  - extend_compset, add_datatype_info, etc. return compsets

### DIFF
--- a/examples/l3-machine-code/arm8/asl-equiv/l3_equivalenceLib.sml
+++ b/examples/l3-machine-code/arm8/asl-equiv/l3_equivalenceLib.sml
@@ -124,14 +124,14 @@ val l3_run_tac =
 (***** custom compset *****)
 local
 
-  fun add_type ty cmp =
+  fun add_type (ty, cmp) =
     computeLib.add_datatype_info cmp (Option.valOf (TypeBase.fetch ty))
-  fun add_types l cmp = List.foldl (fn (ty, cmp) => add_type ty cmp) cmp l
+  fun add_types l cmp = List.foldl add_type cmp l
 
   val tyvars = [alpha, beta, gamma, delta, etyvar, ftyvar]
   fun mk_cmp_type (thy, (name, arity)) =
     mk_thy_type {Args = List.take (tyvars, arity), Thy = thy, Tyop = name}
-  fun add_all_types thy cmp =
+  fun add_all_types (thy, cmp) =
     add_types (map (mk_cmp_type o pair thy) (types thy)) cmp
 
   fun add_defs defs cmp = computeLib.add_thms defs cmp
@@ -140,7 +140,7 @@ local
   fun unpack_defn_thm (DefnBase.STDEQNS t) = t
     | unpack_defn_thm (DefnBase.OTHER   t) = t
 
-  fun add_all_defs thy cmp =
+  fun add_all_defs (thy, cmp) =
     add_defs (DefnBase.thy_userdefs {thyname = thy} |>
               map (unpack_defn_thm o #thm)) cmp
 
@@ -186,8 +186,8 @@ in
   val cmp = computeLib.add_conv
     (``$= : α word -> α word -> bool``, 2, QCHANGED_CONV blastLib.BBLAST_CONV) cmp
   val cmp = bitstringLib.add_bitstring_compset cmp (* has to come after BBLAST_CONV *)
-  val cmp = List.foldl (fn (thy, cmp) => add_all_types thy cmp) cmp thys
-  val cmp = List.foldl (fn (thy, cmp) => add_all_defs thy cmp) cmp thys
+  val cmp = List.foldl add_all_types cmp thys
+  val cmp = List.foldl add_all_defs cmp thys
   val cmp = computeLib.add_thms [
               armv86aTheory.ExecuteA64_def,
               armv86aTheory.DecodeA64_def,
--- a/examples/l3-machine-code/common/utilsLib.sml
+++ b/examples/l3-machine-code/common/utilsLib.sml
@@ -1036,11 +1036,11 @@ in
 end
 
 local
-   fun add_datatype ty cmp =
+   fun add_datatype (ty, cmp) =
      computeLib.add_datatype_info cmp (Option.valOf (TypeBase.fetch ty))
 in
    fun add_datatypes l cmp =
-     List.foldl (fn (ty, cmp) => add_datatype ty cmp) cmp l
+     List.foldl add_datatype cmp l
 end
 
 type inventory = {C: string list, N: int list, T: string list, Thy: string}


xrchz:
I was going to also replace all the `unit -> compset` functions with `compset` values. But I'm not so sure this is a good idea, since compsets are still mutable values. Thoughts?

More context from Claude Code:

Let me trace through what actually happens with the current code:
```
  val bool_compset = ... (* static value with refs for COND, /\, \/, etc. *)

  (* Scenario 1: Adding rules for a NEW constant *)
  val cs_a = add_thms [ADD_CLAUSES] bool_compset
  (* + doesn't exist in bool_compset, so assoc_clause creates a NEW ref *)
  (* bool_compset is UNCHANGED *)

  (* Scenario 2: Adding rules for an EXISTING constant *)
  val cs_b = add_thms [COND_EXPAND] bool_compset
  (* COND exists in bool_compset, so assoc_clause returns the SAME ref *)
  (* The ref is MUTATED - bool_compset now has COND_EXPAND too! *)
```
  So pollution only affects constants already in the base compset.

  Correctness implications:

  1. Computed results: Should still be correct - adding valid rewrite rules doesn't produce wrong answers, just potentially different reduction paths.
  2. Termination: Could be affected if added rules interact badly with existing ones (e.g., create loops). But this is a general compset hazard, not specific to sharing.
  3. Semantic violation: The abstraction that bool_compset is "just boolean rewrites" is broken. Code expecting only boolean rules might behave unexpectedly.
  4. Accumulation over time: In a long-running session, bool_compset could accumulate many rules for `COND`, `/\`, etc., slowing down evaluation.
  5. Non-determinism: Different execution orders could produce different compset states, making behavior harder to reason about.

  Realistic risk assessment:

  For bool_compset, the existing constants are: `COND`, `/\`, `\/`, `==>`, `<=>`, `~`, `LET`, `literal_case`, `T`, `F`. Adding more rules for these specific constants is relatively rare in practice - most uses add rules for new constants like arithmetic or data structures.

xrchz:
I think a good approach to solving that might be to introduce an operation to "seal" a compset, which prevents updating of any of its existing constants.

ordinarymath:
How are you planning to deal with `(=)`? Every new datatype will declare new rewrite rules for that constant. It could potentially just be special cased thou.

xrchz:
I'd be surprised if `(=)` isn't already treated quite specially. I don't think it's gonna be a problem.

I am liking the sealing idea, so gonna implement it and see how it goes.

  So the API would be:
```
  val seal : compset -> compset
  (* Marks compset as sealed. Adding to existing constants will error. *)

  val copy : compset -> compset
  (* Creates unsealed copy with fresh refs for all constants. *)
```
  Behavior:
  - `add_thms [new_const_thm] sealed_cs` → works, new constant gets fresh ref
  - `add_thms [existing_const_thm] sealed_cs` → raises exception
  - `add_thms [existing_const_thm] (copy sealed_cs)` → works, modifies the copy

```
  fun assoc_clause (Compset {sealed, dict}) cst =
    case Redblackmap.peek (dict, cst) of
      SOME rl =>
        if sealed then
          raise ERR "assoc_clause" "cannot add to existing constant in sealed compset"
        else
          (Compset {sealed=false, dict=dict}, rl)
    | NONE =>
        let val mt = ref (EndDb, NONE)
            val dict' = Redblackmap.insert (dict, cst, mt)
        in (Compset {sealed=false, dict=dict'}, mt)
        end

  fun seal (Compset {dict, ...}) = Compset {sealed=true, dict=dict}

  fun copy (Compset {dict, ...}) =
    Compset {sealed=false, dict=Redblackmap.map (fn r => ref (!r)) dict}
```

xrchz:
> How are you planning to deal with `(=)`? Every new datatype will declare new rewrite rules for that constant. It could potentially just be special cased thou.

What new rewrite rules are you thinking of actually?

ordinarymath:
The constructor rewrite rules e.g. for lists it would be `x :: xs = y :: ys <=> x = y /\ xs = ys` and `[] = []` +
`x :: xs = [] <=> F` `[] = x :: xs <=> F`

xrchz:
> The constructor rewrite rules e.g. for lists it would be `x :: xs = y :: ys <=> x = y /\ xs = ys` and `[] = []` + `x :: xs = [] <=> F` `[] = x :: xs <=> F`

So I think we want to special case (=) for this reason. Would it make sense to simply always allow additional rules for (=) or do we need to be more nuanced?

xrchz:
Alternatively, keep it simple without special cases and just use `copy` whenever we need to extend a compset with new equations for existing constants, (=) included, constructor distinctness equations included.

ordinarymath:
Equality can be distinguished by the type of the equality constant.  It's probably beneficial to do so anyway as a perf trick since otherwise you have a huge amount of matches for `(=)`. So for equality you use a type net to individual references. So the code when matching eqc, calls type_of on eqc and looks up the sub rewrite rules/conversions to use.

xrchz:
So I've decided to make this PR just about making `computeLib` have a functional API, but not removing the `unit -> compset` thunks. That's because the design of `seal` and `copy` that I chose is maybe more controversial, as well as the treatment of equality on constructors -- the approach taken on the new `compute_values` branch is simply to `copy` whenever needed, but we might also want to consider @ordinarymath's ideas.

src/compute/src/computeLib.sig
```diff
@@ -28,7 +28,7 @@ sig
   val CBVn_CONV       : int -> compset -> conv
   val WEAK_CBV_CONV   : compset -> conv
 
-  val the_compset     : compset
```
mn200, line 31: Ultimately, this explicit reference should go away and the AncestryData API should be used instead.  But I'm happy to break the work up.

mn200 (approved):

mn200:
Will merge this by Friday if @acjf3 hasn't weighed in beforehand.

mn200:
Thanks for your work on this!

PR #1766, opened 2026-01-10
Compset values

xrchz:
This is an addition to #1764 that makes more significant design decisions on how to handle the fact that compsets, even with a functional computeLib API, are still mutable. See discussion on that PR.


mn200:
In for a penny, in for a pound!  Thanks.

PR #1768, opened 2026-01-10
Replace free_vars with FVL in HO_PART_MATCH

### LINKED ISSUE #1621: Remove the use of `free_vars` in HO_PART_MATCH
https://github.com/HOL-Theorem-Prover/HOL/blob/24e40cd1c7f7e235a5c9f32f692c0d9ae04ddb2b/src/1/Drule.sml#L1802C1-L1804C71
HO_PART_MATCH calls free_vars. `Term.free_vars` is much slower than `Term.FVL`.   
This might just be replacing `free_vars bod` with `HOLset.listItems (FVL [bod])` if the order doesn't matter.

xrchz:
Fixes #1621

### DIFF
--- a/src/1/Drule.sml
+++ b/src/1/Drule.sml
@@ -1823,7 +1823,8 @@ in
          val lconsts =
              HOLset.intersection (FVL [pbod] empty_tmset, hyp_frees th)
          val ltyconsts = HOLset.listItems (hyp_tyvars th)
-         val flex_candidates_vlist = filter (can dom_rng o type_of) (free_vars bod)
+         val flex_candidates_vlist =
+            filter (can dom_rng o type_of) (HOLset.listItems (FVL [bod] empty_tmset))
       in
          if (null flex_candidates_vlist) then
             (fn tm =>


ordinarymath:
The code looks ok. Only way to find out if it's faster thou is to benchmark it.

xrchz:
> The code looks ok. Only way to find out if it's faster thou is to benchmark it.

Can you do that?

ordinarymath:
> > The code looks ok. Only way to find out if it's faster thou is to benchmark it.
> 
> Can you do that?

I'll run some benchmarks now

mn200:
> > > The code looks ok. Only way to find out if it's faster thou is to benchmark it.
> > 
> > 
> > Can you do that?
> 
> I'll run some benchmarks now

What's the verdict?

ordinarymath:
I've realized that this should not be hot code at all and would be changed by my fix to the simplifier.

ordinarymath:
This is slower. I suspect it might be due to HOLset.listItems taking longer maybe could be faster using HOLset.filter.

mn200:
If the PR code is slower, I guess it should be closed.

PR #1776, opened 2026-01-13
Make it so that RM get optimized to take in 3 args in PolyML

ordinarymath:
This slightly reduces tuple allocations.

### DIFF
--- a/src/0/Term.sml
+++ b/src/0/Term.sml
@@ -891,7 +891,7 @@ local
   val tymatch = Type.raw_match_type
   open KernelSig
 in
-fun RM [] theta = theta
+fun RM [] (tmS,tyS) = (tmS,tyS)
   | RM (((v as Fv(n,Ty)),tm,scoped)::rst) ((S1 as (tmS,Id)),tyS)
      = if bound_by_scope scoped tm
        then MERR "Attempt to capture bound variable"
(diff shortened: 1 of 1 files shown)


ordinarymath:
I'm wondering whether it would be clearer to just curry RM so it becomes
```
fun RM [] tmS tyS =
```

ordinarymath:
Btw I think there are a few other opportunities 
```
e.g. you could define a pairlist datatype
`datatype ('a,'b) pairlist = Pairnil | PairCons of 'a * 'b * ('a,'b) pairlist`
 saving a pointer indirection + tuple allocation
```
You could make it so that it's taking
```
fun RM pat obj res (tmS,tyS) =
and then in the end case do
case res of
   [] => (tmS,tyS)
 | ((pat,obj)::res) => RM pat obj res (tmS,tyS)
```
saves a cons allocation every Abs and Comb recursion.
But probably the extra complexity isn't really that worth it.
Also most cases of calls to RM are when the pattern is already known and you can implement a way faster version which allocates a single array when checking for a match if it becomes something someone really wants to optimize.

mn200:
It's a shame that this idiom is the one that Poly/ML handles better.  What impact does the change have on build times?

ordinarymath:
Hmm it doesn't seem to speed up or slow down the speed of match_term beyond showing a reduced number of allocations. Also PolyML doesn't seem to optimize it into a loop which might be causing slower tail recursive calls that makes argument passing slower.

mn200:
Without obvious advantage, I'd prefer not to touch the kernel.

PR #1777, opened 2026-01-13
Make pmatch use its own case keyword

### LINKED ISSUE #1673: pmatch should use `pmatch ... of ...` instead of using `case ... of ...`
@myreen suggested [on Zulip](https://hol.zulipchat.com/#narrow/channel/494306-.E2.9D.93-Support/topic/Why.20are.20cases.20being.20turned.20into.20PMATCH.3F/near/513137547) that pmatch should use  “pmatch … of … “ as its syntax, instead of “case … of …”.

I am not entirely sure what else this would include. In particular, I wonder whether we want to deprecate/remove `ENABLE_PMATCH_CASES()` and `dtcase`.

xrchz:
Fixes #1673

### DIFF
--- a/src/pattern_matches/patternMatchesSyntax.sml
+++ b/src/pattern_matches/patternMatchesSyntax.sml
@@ -876,15 +876,18 @@ val temp_enable_pmatch =
                 addup = K o temp_add_user_printer,
                 up = userprinter_info}
 
-(* Note: enable_pmatch uses temp_add_preterm_processor and temp_add_user_printer
-   because Parse lacks permanent versions with matching signatures. *)
+(* TODO: A permanent enable_pmatch would require:
+   1. A permanent add_preterm_processor, which doesn't currently exist
+   2. Parse.add_user_printer takes (string * term) but userprinter_info is
+      (string * term * userprinter), so the types don't match
+   See GitHub issue for future work.
 val enable_pmatch =
     add_pmatch {get = term_grammar,
                 arule = K o Parse.add_rule,
-                add_ptmproc =
-                  (fn s => fn pp => K (temp_add_preterm_processor s pp)),
-                addup = K o temp_add_user_printer,
-                up = userprinter_info}
+                add_ptmproc = ...,
+                addup = K o add_user_printer,
+                up = ...}
+*)
 
 val grammar_add_pmatch =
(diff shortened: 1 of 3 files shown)


src/pattern_matches/patternMatchesSyntax.sml
```diff
@@ -868,10 +868,19 @@ end
...
+val temp_enable_pmatch =
     add_pmatch {get = term_grammar,
                 arule = K o Parse.temp_add_rule,
-                rmtmtok = K o Parse.temp_remove_termtok,
+                add_ptmproc =
+                  (fn s => fn pp => K (temp_add_preterm_processor s pp)),
+                addup = K o temp_add_user_printer,
+                up = userprinter_info}
+
+(* Note: enable_pmatch uses temp_add_preterm_processor and temp_add_user_printer
```
mn200: I don't think this will work, precisely because you're not making the preterm-processor and user-printer changes permanent.  If you want to enable this make-pmatch-permanently-switched-on functionality (which certainly has its appeal), please test that you have implemented it.  (Use a `theory_tests` subdirectory where you can put script files `A` and `B` with `B` building on `A`, with `A` calling `enable_pmatch`.  Then the code in `B` can check to see if things have worked.  My claim is that your code won't achieve this.)

xrchz: But how do you make them permanent? I don't think there are permanent versions of those in Parse. (The idea here was to get things building with the temp functions that do exist and all the reviewer for advice; I didn't think it would work as-is either.)

mn200: There is a permanent `add_user_printer` (with a help file), but currently no permanent preterm-processor.  The facility for doing this would have to build a code library analogous to what is done for `userPP` and `absynPostProcessor` in `term_grammar`. 

Alternatively, give up on having a permanent version entirely (we didn't have it before).

xrchz: I don't want to give up on it but defer it to a future PR. I'll do that and make another issue.

mn200 (approved):

mn200:
Thanks for putting in the time on this!  The Moscow ML CI failure is unrelated so will happily merge.

Issue #1602, opened 2025-08-07
Rename buildheap executable


xrchz:
When looking at running processes while Holmake or hol is running there's something called `buildheap` that is actually the name of the hol process. This issue is to name it `hol` or something more descriptive/recognisable instead.

xrchz:
From discussion on Zulip, I am gathering some more specific desiderata:
- Combine the `buildheap` and `hol` executables into a single executable (`hol`) that uses subcommands (commandline arguments) to select what functionality is required
- Remove the shell wrapper, make the executable do everything it needs to without a wrapper

mn200:
The important current use-cases:

- `hol repl`  starts the repl for interactive proof
- `hol lsp` starts an LSP server
- `hol buildheap` builds a heap
- `hol runscript` runs a script file for building a theory

This seems fairly clean.   The `repl` sub-command could be made optional so that `hol` on its own just starts the standard REPL (if you want additional options to that invocation, you might insist that `repl` subcommand be present textually). There might be some value in keeping `hol.bare` as a shell-script because it would correspond to `hol repl --bare` or similar.

digama0:
I don't see the advantage of `hol.bare` over `hol --bare`

mn200:
Two fewer characters and it's just a single token, duh.   :-)

But yes, if you just scan all the arguments and don't find a sub-command, assuming `repl` to be the sub-command is simple to explain and implement.

xrchz:
I think this was closed by #1781

Issue #171, opened 2014-06-12
hol --help


myreen:
I can't seem to find any documentation about command-line arguments for `hol`. `Holmake`'s command-line arguments are well documented, both in the _Description_ and also `Holmake --help`. 

Are there command-line arguments for `hol`? If there are, then there should also be support for `hol --help` for listing the available command-line arguments.
##

xrchz:
`hol --help` does produce output... But maybe it's not the output you want? This seems related to #1602

xrchz:
After #1781 would you say this is done @myreen ?

myreen:
The latest `hol` binary has very nice `--help` output. Thanks!

PR #1781, opened 2026-01-14
Rename buildheap to hol with subcommand-based CLI

xrchz:
Replace the buildheap executable and shell wrapper scripts with a unified `hol` command that uses subcommands for different modes of operation.

## New CLI structure

```
  hol [poly-options] [subcommand] [options] [files...]

  Subcommands:
    repl       Start interactive REPL (default if no subcommand)
    lsp        Start LSP server
    buildheap  Build a heap from object files (requires -o)
    run        Run a script file for side effects and exit
    heapname   Print the heap path (from Holmakefile or default)

  Options:
    --holstate=FILE, -b FILE   Load base heap state
    --bare                     Use minimal heap (hol.state0)
    --poly                     Use bare Poly/ML (no HOL state)
    --exe=FNNAME               Create standalone executable (buildheap only)
    --quiet, -q                Reduce verbosity
    --mt=N                     Set thread count
```

### Poly/ML
- Rename tools-poly/buildheap.ML to tools-poly/hol.ML
- Add subcommand parsing with REPL as default
- Embed heapname logic directly (eliminates separate heapname binary)
- Add `--bare` flag (replaces hol.bare shell script)
- Add `--exe` option for creating standalone executables
- Load prelude files before user scripts in REPL mode
- Remove shell wrapper generation from configure.sml

### Moscow ML
- Consolidate `hol` and `hol.noquote` into single `hol` script
- Add `--bare` flag for minimal environment
- Add `--noquote` flag to disable the quotation filter
- Remove `emit_hol_unquote_script` function

### Build system
- Update Holmake to use `hol run` for script execution
- Update all Holmakefiles to use `hol buildheap`
- Update documentation (English and Italian translations)

## ML-specific options

| Option | Poly/ML | Moscow ML |
|--------|---------|-----------|
| `--bare` | ✓ | ✓ |
| `--poly` | ✓ | - |
| `--noquote` | - | ✓ |

This is a breaking change with no backward compatibility for the old `buildheap` command name.

Solves #1602 and #171.

### DIFF
--- a/Manual/Translations/IT/Description/misc.tex
+++ b/Manual/Translations/IT/Description/misc.tex
@@ -719,11 +719,11 @@ This can make developing big mechanisations considerably more pleasant.
 
 \subsection{Generating \HOL{} Heaps}
 
-The easiest way to generate a \HOL{} heap is to use the \texttt{buildheap} executable that is built as part of the standard build process for (Poly/ML)~\HOL.
-This program takes a list of object files to include in a heap, an optional heap to build upon (use the \texttt{-b} command-line switch; the default is to use the heap behind the core \texttt{hol} executable), and an optional name for the new heap (the default is the traditional Unix \texttt{a.out}).
+The easiest way to generate a \HOL{} heap is to use the \texttt{hol buildheap} subcommand that is part of the standard \texttt{hol} executable for (Poly/ML)~\HOL.
+This subcommand takes a list of object files to include in a heap, an optional heap to build upon (use the \texttt{-b} command-line switch; the default is to use the heap behind the core \texttt{hol} executable), and a required name for the new heap (the \texttt{-o} switch).
 Thus the command-line
 \begin{alltt}
-   buildheap -o realheap transcTheory polyTheory
+   hol buildheap -o realheap transcTheory polyTheory
 \end{alltt}
 would build a heap in the current directory called \texttt{realheap}, and would preload it with the standard theories of transcendental numbers and real-valued polynomials.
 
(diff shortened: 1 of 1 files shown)


mn200:
Looking forward to it!

xrchz:
I think I want to get rid of `hol.noquote` too as part of this - any objection?

mn200:
> I think I want to get rid of `hol.noquote` too as part of this - any objection?

None at all.

mn200:
- There are now conflicts because of the `tools/utils` change; 
- There are still corrupted accented characters in the Italian files.

xrchz:
I think this is good now? I didn't see any corrupted characters - do you still see them, in which file if so?

mn200:
Of course, you haven't introduced them (they were there before, and perhaps they are handled fine by LaTeX and we shouldn't be concerned), but in the Italian Tutorial's `intro.tex` : 

> Il passo successivo � eseguire il programma \texttt{build}. Questo dovrebbe

xrchz:
I don't think it should block this PR if they weren't introduced here.

mn200:
I note now that `hol --help` lists `lsp` as a sub-command, and `--lsp` as an option.  Which is right?

xrchz:
> I note now that `hol --help` lists `lsp` as a sub-command, and `--lsp` as an option. Which is right?

I think `--lsp` is dead code and should be removed. The `lsp` subcommand seems to be the correct one.

xrchz:
Made #1794 to rectify this

PR #1788, opened 2026-01-19
Make FRONT [] = []

### LINKED ISSUE #1648: Make FRONT [] = []
Similar to how its currently true that `TL [] = []` having `FRONT [] = []` would be allow for some preconditions on theorems to be removed.

xrchz:
Similar to how TL [] = [], define FRONT [] = [] to allow removing preconditions from several theorems.

Changes to listTheory:
- FRONT_DEF: Add FRONT [] = [] as base case
- FRONT_NIL[simp]: New theorem extracting FRONT [] = []
- LENGTH_FRONT[simp]: New unconditional theorem LENGTH (FRONT xs) = LENGTH xs - 1
- LENGTH_FRONT_CONS: Remove [simp] (superseded by LENGTH_FRONT)
- MAP_FRONT: Remove precondition (ls <> [] ==>) making it unconditional

Changes to rich_listTheory:
- BUTLASTN_1: Remove precondition (~(l = []) ==>) making it unconditional

Closes #1648

### DIFF
--- a/src/list/src/listScript.sml
+++ b/src/list/src/listScript.sml
@@ -2066,13 +2066,13 @@ Proof
   REWRITE_TAC [FRONT_DEF, NOT_CONS_NIL]
 QED
 
-Theorem LENGTH_FRONT_CONS:
+Theorem LENGTH_FRONT_CONS[simp]:
  !x xs. LENGTH (FRONT (x::xs)) = LENGTH xs
 Proof
 Induct_on ‘xs’ THEN ASM_SIMP_TAC bool_ss [FRONT_CONS, LENGTH]
 QED
 
-Theorem LENGTH_FRONT[simp]:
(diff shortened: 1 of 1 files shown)


mn200:
Looks good to me!

ordinarymath:
I don't get why LENGTH_FRONT is unsafe as a simp rule

ordinarymath:
The termination issues in PSL are probably by the proof adding rewrites like `LENGTH v = SUC(LENGTH(FRONT v))` to the goal. This is a bad rewrite rule in general as the rhs is more complicated.

xrchz:
I would suggest merging this and adding more simp rules if you want them later.

ordinarymath:
Sure I'm fine with that

ordinarymath (approved):

PR #1790, opened 2026-01-20
Document prelude files

xrchz:


tools-poly/README.md
mn200, line 1: It's not clear to me that we should have READMEs scattered around the various directories, or if we should (just) have the documentation that is in the developers manual.

tools/unquote-init.sml
mn200, line 1: The comment about quotation being handled by modifications to PolyML.Compiler hooks is incorrect.  The HOL REPL uses a modified copy of Poly/ML's REPL code to do its custom lexing.  This code is in `tools-poly/holrepl.ML` and is called from `hol.ML` in the same directory.

xrchz: Changed

tools-poly/README.md
xrchz: Fair - made them short and point to developers manual

mn200:
Looks great; thanks.  The CI problems are not yours.

PR #1791, opened 2026-01-22
Fix ty_sub to not recursively check theta = []

ordinarymath:
Branch predictors aren't good enough to make this not important


ordinarymath:
NB: kernel level change

PR #1792, opened 2026-01-22
Slightly speed up term_pmatch

ordinarymath:
The speed up is from when you have a long comb term where it's like `f a1 a2 ... an` and the code would be doing
repeat rator `f a1 a2 ... an`
repeat rator `f a1 a2 ... a(n - 1)` etc to check if the head is a variable It's not a major perf issue but it does affect performance


ordinarymath:
Note that this kinda mainly affects artificial examples since in practice the comb chains are fairly short.

mn200:
Thanks!

PR #1793, opened 2026-01-22
Refactor traverser

ordinarymath:
This speeds up benchmarks when code is just descending and no rewrites are applied. Also fixes major perf issue where COND_REWR_CONV is not partially applied.


ordinarymath:
```
val string_grd_tm = ``"cakeml"``
fun test n = if n = 0 then () else
   let val _ = ignore (SIMP_CONV (ss) [] string_grd_tm)
               handle UNCHANGED => ()
   in test (n - 1) end
```
This has been my main benchmark. 30.8 seconds instead of 40.9 seconds this is including the change to #1791 and #1792

mn200:
Very nice; thanks!

PR #1794, opened 2026-01-22
Remove dead --lsp option from hol command

xrchz:
The --lsp option was parsed and stored in the options record but never used. The LSP server is triggered by the "lsp" subcommand (e.g., "hol lsp"), not by the --lsp flag.

Removed the lsp field from option_record and all related code.


mn200:
Thanks!

PR #1795, opened 2026-01-23
Central Limit Theorem

KaiPhan:
This PR adds Central Limit Theorem proof, based on [1] .

This PR turns the theorem `converge_in_dist_alt_C3` into a definition of convergence in distribution, and therefore removes all remaining cheats in the CLT proof.

[1] Chung, K. L.: A Course in Probability Theory, Elsevier, 2000.


mn200:
Thanks for your work on this!

PR #1800, opened 2026-01-28
Some central_limitTheory cleanups

binghe:
Hi,

This PR slightly cleaned up the newly added `central_limitTheory`, where some lemmas are already available in upstream theories, some are moved to upstream. I also changed it to the modern theory grammar.

--Chun


mn200:
Thanks!

PR #1801, opened 2026-01-29
Pair of fixes for mcandidate

tanyongkiam:
The commit updates the hash for armv8.6-asl-snapshot

The bug in regexpLib was due to incorrect use of the new API for compsets. Found thanks to self-tests and with @Claude


mn200:
Thanks for these!

PR #1816, opened 2026-02-07
Start investigating a faster version of type_vars_in_term

ordinarymath:


ordinarymath:
Opening a PR to test all kernels

ordinarymath:
As far as I'm aware this change is only needing documentation. Other bonus points maybe could be to change the expknl implementation of the `type_vars_in_term` function but other than that nothing else should be needed.

ordinarymath:
I also modified the experimental kernel which could use a review

mn200:
Thanks; I will remove the unnecessary `let`s myself.

PR #1817, opened 2026-02-09
Updating the definition of a cycle

Eric-C-Hall:
Updating the definition of a cycle to exclude trivial cycles. A cycle in a simple finite graph ought to be at least 3 nodes long, because a cycle containing only two nodes is nothing more than traveling backwards and forwards on an edge, which is trivial. In particular, the definition of a tree as a connected graph with no cycles would make no sense if we allowed a cycle of length 2. The previous definition of a cycle required that the sequence of vertices was of at least length 3, however, this was a buggy implementation, because the first and last vertices are the same vertex, so a loop containing 3 vertices only contains 2 distinct vertices. So this should be updated to require a loop to contain at least 4 vertices, so that we have at least 3 distinct vertices, in order to exclude the trivial situation where we simply travel backwards and forwards on a single edge.

### DIFF
--- a/examples/generic_graphs/fsgraphScript.sml
+++ b/examples/generic_graphs/fsgraphScript.sml
@@ -679,7 +679,7 @@ Definition trail_def:
 End
 
 Definition cycle_def:
-  cycle g vs ⇔ walk g vs ∧ ALL_DISTINCT (TL vs) ∧ 3 ≤ LENGTH vs ∧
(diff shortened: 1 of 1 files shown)


mn200:
Thanks!

mn200:
NB: please use standard git "practice" for writing your commit messages: see [this page](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) for conventional advice.

Issue #1527, opened 2025-06-12
Holmake source-based hashing for dependencies


xrchz:
Recursively compute source-based hashes for theory-target dependencies, for #1522.

mn200:
A particular theory `.dat` file depends on its script file and the theory files it loads.  An existing `.dat` file already records the hashes of its ancestor `.dat` files (which do the same with *their* ancestors).  This much would seem to give us "recursive" hashing of the `.dat` file.  

It seems like all we really need to do then is to also include the hash of the corresponding `Script.sml` file in the `.dat` file that it generated.

xrchz:
Actually do we even need that? Can we not reuse build outputs if the .dat files are the same even if the script files changed?

I think the problem is more in the lookup direction: if all you have is script files, how do you know what hash to check the cache for to avoid building?

mn200:
Well, the script file hashes could be the keys that you looked up in the cache... 

It seems a rebuild has to happen if

1. the script file is newer than the .dat file, and 
2. the script's hash is not a key in the cache

Something is wrong if (probably requiring a rebuild or a re-pull from the cache) if:

1. the script file's hash is in the cache as a key, but 
2. the stored .dat file has hashes for ancestors that are not present

mn200:
Can we please please not reinvent the wheel here?  This is not so different from building/compiling in a normal PL environment so there is probably existing work to consult.

xrchz:
I don't like the idea of the .dat files depending on irrelevant (to the generated theory) script file changes.

But I'm happy for the script file hash to be included somewhere for the purposes of this cache.

Regarding following existing practice maybe @digama0 or @pylasnier already knows the design of another build system we ought to copy?

mn200:
Yeah, agreed.  We shouldn't do what I suggested earlier and have script-file hashes in `.dat` files. 

Given that the `.dat` files then contain all of the relevant hashes of their ancestors, and those ancestors do likewise, it seems like this particular sub-task is already done.

pylasnier:
The original Nix system is the work of [Eelco Dolstra as described in his PhD thesis](https://edolstra.github.io/pubs/phd-thesis.pdf). The original work is probably worth looking at for more detail, but I think I can abbreviate it for the purposes of HOL.

[...]

PR #1820, opened 2026-02-11
Add cachekey option to Holmake

xrchz:
Finish #1527

### DIFF
--- a/Manual/Description/misc.tex
+++ b/Manual/Description/misc.tex
@@ -311,6 +311,12 @@ In addition, there are three special targets that can be used:
 \noindent Finally, users can directly affect the workings of \holmake{}
 with the following command-line options/flags:
 \begin{description}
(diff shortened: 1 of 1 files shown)


xrchz:
@mn200 are these build failures actually related to these changes?

mn200:
> @mn200 are these build failures actually related to these changes?

No; they're problems with the Manual that I caused.

mn200:
When you say you sort the dependencies to get a canonical list, how do you know that the prefix of those paths (the directories leading down to where the sources are) won't vary?  

In particular, if I have a pair of projects with an `INCLUDES` link between them, then the user's choice as to where those projects are within their hosting file-system might cause all of `projectParent`’s theories to come before _or_ after `projectChild`’s.

xrchz:
We're not looking at path at all when sorting dependencies. They're sorted by filename then content (hash) for ties.

mn200:
Please document the option in the `Holmake` section of the DESCRIPTION manual.

PR #1824, opened 2026-02-15
Fix bug in suspend

### LINKED ISSUE #1822: Proofs containing suspend fail unexpectedly
Consider the following theory:
```
Theory suspend_test
Ancestors
  num

Theorem test_proof:
  ∀n. P 0 ∧ (∀n. P n ⇒ P (SUC n)) ⇒ P n
Proof
  ho_match_mp_tac INDUCTION >> rpt strip_tac
  >> simp[]
QED

Theorem test_cheat:
  ∀n. P 0 ∧ (∀n. P n ⇒ P (SUC n)) ⇒ P n
Proof
  ho_match_mp_tac INDUCTION >> rpt strip_tac
  >> cheat
QED

Theorem test_suspend:
  ∀n. P 0 ∧ (∀n. P n ⇒ P (SUC n)) ⇒ P n
Proof
  ho_match_mp_tac INDUCTION >> rpt strip_tac
  >> suspend "base"
QED

Resume test_suspend[base]:
  simp[]
QED

Finalise test_suspend
```

If you try to build this, you get:
```
Saved theorem _______ "test_proof"
Saved CHEAT _________ "test_cheat"

Exception raised at boolLib.store_thm_at:
at Tactical.TAC_PROOF:
  Failed to prove theorem "test_suspend":
Can't alpha convert
Proof of

∀n. P 0 ∧ (∀n. P n ⇒ P (SUC n)) ⇒ P n

failed.
```

Given that the theorem and cheat versions work as expected, I would have expected the suspend version to also work and produce something equivalent to the proved version.

xrchz:
Closes #1822

### DIFF
--- a/src/marker/markerLib.sml
+++ b/src/marker/markerLib.sml
@@ -841,7 +841,7 @@ fun RESCONJS th =
                   | SOME (th1, th2) =>
                     recurse (th2::th1::A) rest
     in
-      recurse [] [th]
+      rev (recurse [] [th])
     end
 
 fun RESCONJS_TAC g =
@@ -850,18 +850,7 @@ fun RESCONJS_TAC g =
 fun SUSPIMPS_TAC g =
     TRY (CONV_TAC elimsi_conv THEN DISCH_THEN ASSUME_TAC THEN SUSPIMPS_TAC) g
 
-(* Strip closure foralls that wrap suspimp structures. Only strips foralls
-   when suspimps are present underneath, so as not to disturb foralls that
-   are part of the original goal. *)
-fun has_suspimp_under_foralls t =
-    is_suspimp t orelse
-    (is_forall t andalso has_suspimp_under_foralls (snd (dest_forall t)))
-
-fun CLOSURE_GENTAC (asl, w) =
-    if has_suspimp_under_foralls w then REPEAT GEN_TAC (asl, w)
-    else ALL_TAC (asl, w)
-
-val RESUME_TAC = RESCONJS_TAC THEN CLOSURE_GENTAC THEN SUSPIMPS_TAC
+val RESUME_TAC = RESCONJS_TAC THEN SUSPIMPS_TAC
 
 
 (* th is the theorem corresponding to the suspended goal (which the user has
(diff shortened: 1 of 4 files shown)


mn200:
It looks as if you recreate the goal by stripping all external universally quantified variables. This is not guaranteed to recreate the goal exactly if it originally looked like
```
!p q. p /\ q = q /\ p
```
and you did one `strip_tac` before suspending. 

Your tweak to the `suspended_to_goal` function is also unnecessary; that function is never used and should just be deleted.

xrchz:
Do you have a suggested/preferred solution? We could store more info under the `suspendlabel` that tells us how many variables to strip?

mn200:
Yes, perhaps that's easiest: the argument could be a variable whose name is the string of the number.   This won't be a common case as I think it's only an issue when you suspend a goal that has remaining quantification and which has no assumptions.

src/marker/markerLib.sml
```diff
@@ -880,47 +902,50 @@ fun prove_suspended0 nm s_t th tgt =
...
+                SOME (nm,nc,t0) => if nm = label then SOME (nc,t0) else NONE
               | NONE => NONE
     in
       case List.mapPartial myhyp (hyp th) of
           [] => raise ERR "extract_suspended_goal"
                         ("No such label in theorem: " ^ label)
-        | [t] => t
-        | ts => list_mk_resconj ts
+        | [(nc,t)] => (nc, t)
+        | ncts as ((nc,_)::_) => (nc, list_mk_resconj (map #2 ncts))
```
mn200: Privileging the first hypothesis's count seems odd here (in the last, catch-all case); do you have a test case for this scenario?  It seems as if the code might attempt to strip `nc` foralls off the top of a "res"-conjunction which will in fact have no quantifiers at all.  Moreover, the counts for the subsidiary goals will be ignored and so there may be extra quantifiers in the extracted goals that should be, but won't be, stripped.

Perhaps in this (admittedly) uncommon case, you should keep the labels for the hypotheses underneath the res-conjunction, and then that will tell `RESUME_TAC` how many to strip (currently your code just looks for suspimp-assumptions, which is wrong).

xrchz: Should be fixed now

mn200:
Thanks!

PR #1825, opened 2026-02-16
Proof trace export: replayable kernel-level proof certificates

charles-cooper:
## Summary

Add opt-in proof trace export to HOL4. When enabled, every theory build
produces a `.pftrace.zst` file recording every kernel inference step
needed to derive its exported theorems. These traces are independently
verifiable by replaying through the kernel — eliminating DISK_THM trust.

```bash
HOL_TRACE_PROOFS=1 bin/build    # build with tracing
bin/hol < src/postkernel/verify_all_traces.sml   # verify all 180 traces
```

**Result:** 180/180 theories traced and verified with zero fallbacks.
277 MB total (zstd-compressed), 25.3M proof steps, 27K exports.

## Motivation

Theorems loaded from `.dat` files carry a `DISK_THM` oracle tag — there's
no independent way to verify they follow from kernel inference rules. For
high-assurance applications (e.g., formal verification of smart contracts),
this trust gap matters. Proof traces close it: if a trace replays
successfully through the kernel, the exported theorems are sound.

### stdknl hooks, not a kernel fork

The kernel (`std-thm.ML`) gets minimal additions (~200 LOC diff):
- `thm` datatype extended with an `int` trace ID
- `trace_step` sum type with 42 constructors (one per inference rule)
- `trace_hook : (trace_step -> unit) option ref` — called after each inference
- `trace_export_hook` — called at theory export time

The hooks are **provably inert**: each theorem is fully computed before
the hook runs, the hook's return value is discarded, and exceptions are
caught. When `trace_hook` is `NONE` (the default), the only overhead is
8 bytes per `thm` value and one ref deref per inference rule.

**Alternative considered:** A full kernel fork (`tracknl`). We prototyped
this first (~2200 LOC) and it worked, but maintaining a parallel kernel
is fragile. The hook approach keeps a single kernel with zero behavioral
change when tracing is disabled.

### Stream-to-disk, not proof-in-thm

Each kernel inference streams its step to a per-process temp file
immediately. At `export_theory()`, backward reachability from exported
roots determines which steps survive, and only those are written to the
final trace. This gives:

- **Zero proof memory overhead** during proof search (ML GC is unaffected)
- **Minimal final traces** (only export-reachable steps)
- **No external minimization tool** needed

**Alternative considered:** Storing proof DAGs in the `thm` datatype
(like `otknl`). This lets ML GC drop unreachable proofs, but keeps all
runtime-reachable proofs in memory (simpsets, caches, databases) — not
just export-reachable ones. The stream approach gives strictly smaller
traces with zero memory overhead.

### Two-phase export with raw-first safety

Export writes the trace twice:

1. **Phase 1 (raw):** Stream temp file through zstd compressor, writing
   all reachability-minimized steps. This always produces a valid trace.

2. **Phase 2 (optimized):** Allocate dedup + pruning structures, re-scan
   the temp file, write an optimized trace, atomically rename over Phase 1.

If Phase 2 fails (e.g., OOM on a very large theory), the raw trace from
Phase 1 is still valid. Phase 2 achieves significant size reduction:

- **51% step dedup** via FNV-1a hash-consing (27M duplicate steps merged)
- **63% term pruning** (22M dead terms removed)
- **25% type pruning**

The two-phase approach avoids adding memory pressure during Phase 1 —
the dedup hash tables and canonical maps (60-112 MB for large theories)
are only allocated in Phase 2.

### boolTheory: POLY_CLINE_OPTIONS injection

`boolTheory` builds before `hol.state0` exists (in `--poly` mode), so
`TraceRecord` isn't in the heap yet. We inject it via:

```makefile
# src/bool/Holmakefile
ifdef HOL_TRACE_PROOFS
POLY_CLINE_OPTIONS += $(dprot $(SIGOBJ)/TraceRecord.uo)
endif
```

This uses Holmake's existing `POLY_CLINE_OPTIONS` mechanism to load
`TraceRecord.uo` before `boolScript` runs.

### TRUST provenance

TRUST entries (references to ancestor theorems) include the source theory
name, extracted from `Dep.DEP_SAVED` tags. This enables efficient lookup
during chain verification without structural matching against all ancestors.

## Architecture

```
                    ┌─────────────┐
                    │ std-thm.ML  │  trace_hook ref (SOME when tracing)
                    │  (kernel)   │──────────────────────────────┐
                    └──────┬──────┘                              │
                           │ trace_export_hook                   │ trace_step
                           ▼                                     ▼
                    ┌─────────────┐                     ┌──────────────┐
                    │ Theory.sml  │                     │TraceRecord.sml│
                    │ (postkernel)│                     │ (postkernel)  │
                    └─────────────┘                     └──────┬───────┘
                                                               │ per-PID temp files
                                                               ▼
                                                       ┌──────────────┐
                                                       │TraceExport.sml│
                                                       │ (postkernel)  │
                                                       └──────┬───────┘
                                                               │ .pftrace.zst
                                                               ▼
                                                       ┌──────────────┐
                                                       │ReplayTrace.sml│
                                                       │  (verifier)   │
                                                       └──────────────┘
```

**Trust boundary:** Only `std-thm.ML` is trusted. `TraceRecord`, `TraceExport`,
and `ReplayTrace` are all outside the trust boundary — if they have bugs,
replay catches them.

## File changes

| File | LOC | Purpose |
|------|-----|---------|
| `src/thm/std-thm.ML` | +200 | trace ID, trace_step datatype, hooks, tr calls |
| `src/thm/std-thmsig.ML` | +51 | expose trace_step, hooks in signature |
| `src/thm/otknl-thm.ML` | +2 | no-op trace_export (signature compat) |
| `src/postkernel/TraceRecord.sml` | 406 | recording: intern tables, temp files, hooks |
| `src/postkernel/TraceExport.sml` | 652 | export: reachability, dedup, pruning, compression |
| `src/postkernel/ReplayTrace.sml` | 977 | verifier: parse, replay, chain verification |
| `src/postkernel/Theory.sml` | +10 | call trace_export_hook at theory export |
| `src/postkernel/FORMAT.md` | 134 | formal trace format specification |
| `src/bool/Holmakefile` | +9 | POLY_CLINE_OPTIONS for boolTheory tracing |
| `src/proofman/Holmakefile` | +5 | load TraceRecord into hol.state0 |
| Benchmarks/verification | ~330 | bench_traces.sml, bench_build.sh, verify_all |

## Trace format

Text-based, one line per entry. Sections: header, types, terms, steps, exports.
See `src/postkernel/FORMAT.md` for the full grammar.

```
HOL4_PROOF_TRACE 1
THEORY combin
PARENTS bool
ANCESTOR bool sha256:abc123...
COUNTS 178 1852 3209

Y 0 V 'a
Y 1 O min bool 0
M 0 V x 0
M 1 A 0 0
P 0 REFL 0
P 1 ASSUME 1
P 2 MP | 0 1
E I_THM 3208
```

## Performance

Build overhead with tracing enabled:

| | Time |
|---|---|
| Normal build | 540s |
| Traced build | 1563s |
| Overhead | +189% |

Export phase breakdown (aggregated across 180 theories):

| Phase | Time | % |
|---|---|---|
| Reachability (mark_live + renumber) | 250s | 17% |
| Raw write (Phase 1 + zstd) | 430s | 29% |
| Dedup (FNV hash scan) | 372s | 25% |
| Prune (type/term liveness) | 34s | 2% |
| Optimized write (Phase 2 + zstd) | 320s | 21% |

zstd compression itself is ~5s per pass — the ML string processing
(TextIO.inputLine, String.tokens, formatting) dominates.

Replay benchmarks:

| | |
|---|---|
| Total traces | 180 files, 277 MB zstd |
| Total steps | 25.3M (after dedup) |
| Parse throughput | 3.7 MB/s |
| Replay (per-theory avg) | 2.2s |
| Chain verification (probability, 75 theories) | 303s |

## Reproducing benchmarks

```bash
# Full build comparison + phase breakdown:
bash src/postkernel/bench_build.sh

# Replay-only benchmarks (after traced build):
bin/hol < src/postkernel/bench_traces.sml

# Verify all traces:
bin/hol < src/postkernel/verify_all_traces.sml
```

## Commit history

There are 34 commits on this branch including iterative development and
bug fixes. Happy to squash/rewrite into logical units if that's preferred.

## Future work

- **Binary format** (varint encoding) for smaller traces
- **Streaming parse** in replayer (currently reads all lines into memory)
- **COMPUTE verification in standalone chain replay**
  (in-session verification works for all 1282 COMPUTE steps via
  `cv_computeLib.cv_compute`; standalone `verify_chain` treats them
  as trusted because it doesn't load theories)
- **Export optimization:** single-pass dedup, or skip Phase 2 for
  faster builds (raw traces are ~2x larger but fully valid)

### DIFF
--- a/src/postkernel/prooftrace/MergeTrace.sml
+++ b/src/postkernel/prooftrace/MergeTrace.sml
@@ -660,7 +660,8 @@ fun mark_live (data : file_data) (prev : liveness)
   : liveness * int list (* unresolved parent trace_ids *)
     * bool (* true if any COMPUTE entry was newly marked live *)
     * (string * string) list (* unresolved type defs: (thy, tyop) *)
-    * (string * string) list (* unresolved const defs: (thy, name) *) =
+    * (string * string) list (* unresolved const defs: (thy, name) *)
+    * int list (* newly live P trace_ids *) =
   let
     val live_y = #live_y prev
     val live_t = #live_t prev
@@ -669,6 +670,7 @@ fun mark_live (data : file_data) (prev : liveness)
     val live_p_count = #live_p_count prev
     val unresolved = ref ([] : int list)
     val found_compute = ref false
+    val newly_live_p = ref ([] : int list)
 
     val dp = get_deps data
 
@@ -693,6 +695,7 @@ fun mark_live (data : file_data) (prev : liveness)
             val idx = id - p_min
             val _ = Array.update(live_p, idx, true)
             val _ = live_p_count := !live_p_count + 1
+            val _ = newly_live_p := id :: !newly_live_p
             val _ = if PIntMap.mem id (#compute_ids dp)
                     then found_compute := true else ()
         in List.app mark_thm parents;
@@ -751,7 +754,8 @@ fun mark_live (data : file_data) (prev : liveness)
        !unresolved,
        !found_compute,
        unresolved_tyops,
-       unresolved_consts)
+       unresolved_consts,
+       !newly_live_p)
     end
   end
 
@@ -1055,15 +1059,16 @@ fun merge {trace_paths : (string * string) list,
                 else ()
         val _ = pass1_progress ()
         val prev_lv = get_liveness path
+        val prev_live_count = lv_p_size prev_lv
         (* Skip if all requested IDs are already live *)
         val dominated = List.all
           (fn id => lv_p_mem prev_lv id) needed_thm_ids
       in if dominated then () else
       let
         val (lv, unresolved, found_compute,
-             unresolved_tyops, unresolved_consts) =
+             unresolved_tyops, unresolved_consts,
+             newly_live) =
           mark_live data prev_lv needed_thm_ids
-        val prev_live_count = lv_p_size prev_lv
         val new_live_count = lv_p_size lv
       in
         pass1_live_thms :=
@@ -1100,14 +1105,29 @@ fun merge {trace_paths : (string * string) list,
           | NONE => ())
           unresolved_consts;
 
-        (* Only enqueue NAME/LOAD for NEWLY live entries
-           (not in prev_lv but in lv) to avoid redundant re-processing *)
-        let val _ = ()
+        (* Enqueue NAME/LOAD for newly-live entries *)
+        let
+          val newly_live_set =
+            let val s = Array.array(
+                  if #p_min_id data < 0 then 0
+                  else #p_max_id data - #p_min_id data + 1, false)
+                val base = #p_min_id data
+            in List.app (fn id =>
+                 let val idx = id - base
+                 in if idx >= 0 andalso idx < Array.length s
+                    then Array.update(s, idx, true) else ()
+                 end) newly_live;
+               s
+            end
+          fun is_newly_live id =
+            let val idx = id - #p_min_id data
+            in idx >= 0 andalso idx < Array.length newly_live_set
+               andalso Array.sub(newly_live_set, idx)
+            end
         in
         (* Enqueue NAME ancestor exports *)
         List.app (fn (id, anc_thy, anc_name) =>
-          if lv_p_mem lv id andalso
-             not (lv_p_mem prev_lv id) then
+          if is_newly_live id then
             case Redblackmap.peek(thy_path_map, anc_thy) of
               SOME anc_path =>
                 let val anc_data = load_file anc_path
@@ -1124,8 +1144,7 @@ fun merge {trace_paths : (string * string) list,
 
         (* Enqueue LOAD ancestor theorems by trace_id *)
         List.app (fn (id, anc_thy, anc_trace_id) =>
-          if lv_p_mem lv id andalso
-             not (lv_p_mem prev_lv id) then
+          if is_newly_live id then
             case Redblackmap.peek(thy_path_map, anc_thy) of
               SOME anc_path =>
                 enqueue (anc_path, [anc_trace_id])


xrchz:
Can be this on expk as well as stdknl?

xrchz:
Can the additions to postkernel be at least in a subdirectory grouped together? (Possibly shouldn't be in postkernel at all, but it's fine, let's see what Michael thinks of that)

src/postkernel/TraceExport.sml
```diff
@@ -0,0 +1,652 @@
...
+    val n_types = length types
+    val n_terms = length terms
+    val objdir = ".hol/objs"
+    val _ = (OS.FileSys.mkDir ".hol" handle OS.SysErr _ => ())
+    val _ = (OS.FileSys.mkDir objdir handle OS.SysErr _ => ())
+    val outname = objdir ^ "/" ^ thyname ^ "Theory.pftrace" ^ compress_ext
+
+    fun open_compress path =
+      let val p : (TextIO.instream, TextIO.outstream) Unix.proc =
+            Unix.execute("/bin/sh", ["-c", compress_cmd (shell_quote path)])
```
xrchz: this needs to be made portable -- compression should be an option too

src/thm/std-thmsig.ML
```diff
@@ -1,4 +1,55 @@
...
+    | TR_INST        of thm * thm * {redex: term, residue: term} list
+    | TR_Beta        of thm * thm
+    | TR_Mk_comb     of thm * thm * thm * thm
+    | TR_Mk_abs      of thm * thm * thm * term
+    | TR_Specialize  of thm * thm * term
+    | TR_ORACLE      of thm * string
+    | TR_AXIOM       of thm * term
+    | TR_DEF_TYOP    of thm * thm * string * string
+    | TR_DEF_SPEC    of thm * thm
+    | TR_DISK_THM    of thm * string
```
xrchz: why is this treated differently from oracle?

src/postkernel/prooftrace/ReplayTrace.sml
```diff
@@ -0,0 +1,977 @@
...
+                         else tm (opd 0)
+            in Thm.TRANS (parent 0) (Thm.ABS bvar (parent 1)) end
+        | "Specialize" => Thm.SPEC (tm (opd 0)) (parent 0)
+
+        (* --- Opaque steps: trust with recorded statement --- *)
+        | "DISK_THM" => mk_trust_thm operands
+        | "ORACLE" => mk_trust_thm (tl operands)  (* skip tag *)
+        | "AXIOM" =>
+            Thm.mk_oracle_thm "REPLAY_TRUST" ([], tm (opd 0))
+        | "DEF_SPEC" =>
```
xrchz: this seems wrong - definitions should be made properly, not turned into oracle thms

src/postkernel/TraceExport.sml
charles-cooper: Fixed in 450a3c8. Compression is now optional: at load time we probe for zstd, then gzip, falling back to writing uncompressed `.pftrace` files if neither is available. The `LargeInt` / `Time.toMilliseconds` portability issues for Moscow ML are also fixed in the same commit.

*co-authored by claude opus 4.6*

src/thm/std-thmsig.ML
charles-cooper: `TR_DISK_THM` records when a theorem is loaded from a saved theory file (via `disk_thm`). It is distinct from `TR_ORACLE` because it represents a different trust boundary: `DISK_THM` means "this theorem was proven in a previously-built theory and saved to disk," while `ORACLE` means "this theorem was asserted by an external oracle (SAT solver, etc.)."

During replay, both are currently resolved via the ancestor context (TRUST entries). When a full chain is replayed from scratch, DISK_THM entries resolve against the replayed ancestor's exports and produce clean tags (no oracle). The distinction matters for diagnostics: if a replayed theorem still has oracle tags, knowing whether the source was DISK_THM vs ORACLE tells you whether the gap is from a missing ancestor trace vs a genuinely external oracle.

*co-authored by claude opus 4.6*

src/postkernel/prooftrace/ReplayTrace.sml
charles-cooper: Fixed in 6c31516. The replayer now calls the actual definition primitives (`prim_type_definition`, `gen_prim_specification`, `prim_specification`) with the witness theorem from the trace. The trace format for `DEF_SPEC` was extended to include the theory name and constant names needed for replay.

Result: `boolTheory` now replays with 191/191 clean exports (0 oracle tags). A 30-theory chain verification (`verify_chain "." "hol"`) shows all theories OK.

*co-authored by claude opus 4.6*

charles-cooper:
Re expk: yes, the trace hooks are mechanical to add to other kernels — the `trace_step` datatype and `trace_hook` ref are ~50 lines that slot in the same way. Will do this as a follow-up.

Re subdirectory: agreed, will group TraceRecord/TraceExport/ReplayTrace into a subdirectory. Need to check Holmakefile implications but should be straightforward.

The latest two commits address the other review items:
- **450a3c8**: compression is now optional (probes zstd → gzip → uncompressed fallback); Moscow ML portability fixes (`LargeInt.int` → `int`, portable timing)
- **6c31516**: definitions replay properly via actual primitives (`prim_type_definition`, `gen_prim_specification`); axioms use `mk_axiom_thm`; bool theory replays with 191/191 clean exports (0 oracle tags)

*co-authored by claude opus 4.6*

src/thm/std-thmsig.ML
xrchz: DISK_THM is implemented as an oracle in HOL though. We can see the tags on the resulting theorem either way. I don't think the trace format needs to distinguish them. It should use oracle either way, and diagnostics can report differently based on _which_ oracles are in the theorem tag if desired.

charles-cooper:
Update on the review items:

**expk support (F4):** On closer inspection, expk already shares `std-thm.ML` / `std-thmsig.ML` with stdknl (the Holmakefile only branches for otknl). So trace hooks are already present for expk builds — no additional work needed. otknl has its own recording via OpenTheory logging, so the no-op stub there is appropriate.

**Subdirectory (F5):** Done in 2eaabfd — moved TraceRecord, TraceExport, ReplayTrace and helper scripts to `src/postkernel/prooftrace/`. Added to build sequence. All references use sigobj paths so no other changes needed.

Full build + 180 theory traces + 30-theory chain verification all pass after the move.

*co-authored by claude opus 4.6*

src/thm/std-thmsig.ML
```diff
@@ -1,4 +1,55 @@
...
+    | TR_Beta        of thm * thm
+    | TR_Mk_comb     of thm * thm * thm * thm
+    | TR_Mk_abs      of thm * thm * thm * term
+    | TR_Specialize  of thm * thm * term
+    | TR_ORACLE      of thm * string
+    | TR_AXIOM       of thm * term
+    | TR_DEF_TYOP    of thm * thm * string * string
+    | TR_DEF_SPEC    of thm * thm * string * string list
+    | TR_DISK_THM    of thm * string
+    | TR_COMPUTE     of thm * thm list * term * term
```
xrchz: I don't think we should include the result term in the trace. The result can be cleanly specified as a function of the inputs. This is no different from the other inference rules where we have a specification of what the rule does and therefore don't need to include the expected output explicitly in the trace.

src/thm/Thm.sml
```diff
@@ -0,0 +1,1488 @@
...
+fun dest_thm(t as THM(_,asl,w,_)) = (hyplist t,w) (* Compatible with old code.*)
+fun sdest_thm(THM(_,asl,w,_)) = (asl,w)
+fun make_thm R (tg,asl,c) =                      (* internal only *)
+  let val id = !trace_counter
+  in trace_counter := id + 1; Count.inc_count R; THM(tg,asl,c,id) end;
+
+fun trace_id (THM(_,_,_,id)) = id
+
+(* Proof trace step: typed representation of each inference rule application.
+   Convention: first thm = result, remaining thms = parents. *)
```
xrchz: OK I see you are actually including the result in _all_ the trace steps. Well I think it shouldn't be included, but happy to hear more about why you want it.

xrchz: Maybe I am misunderstanding how it works. Seems like these result thms are only included in memory but not actually exported in the on-disk trace format?

charles-cooper:
sorry, forgot to push 2eaabfd, it's pushed now

src/postkernel/prooftrace/bench_build.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
+# Benchmark build time with and without proof tracing.
```
xrchz: I think this should probably also be in the prooftrace dir

xrchz: some of the benchmarking stuff could potentially be better suited to developers (in the root dir)?

src/postkernel/prooftrace/FORMAT.md
```diff
@@ -0,0 +1,134 @@
+# HOL4 Proof Trace Format (`.pftrace`)
```
xrchz: should be in prooftrace dir

src/postkernel/FORMAT.md
```diff
@@ -0,0 +1,134 @@
...
+
+| Rule | Operands | Parents | Meaning |
+|------|----------|---------|---------|
+| `AXIOM` | `concl_id` | — | Axiom introduction |
+| `ORACLE` | `tag` `concl_id` `nhyps` `hyp_ids*` | — | Oracle theorem |
+| `DISK_THM` | `concl_id` `nhyps` `hyp_ids*` | — | Loaded from disk |
+| `TRUST` | `global_id` [`thy_name`] `concl_id` `nhyps` `hyp_ids*` | — | Ancestor theorem |
+| `DEF_TYOP` | `thy` `name` `concl_id` | 1 | Type definition |
+| `DEF_SPEC` | `concl_id` | 1 | Constant specification |
+| `COMPUTE` | `input_id` `result_id` | — | Computation (trusted) |
```
xrchz: COMPUTE should not be trusted. It's just as much a valid rule in the kernel as any other.

src/thm/Thm.sml
```diff
@@ -0,0 +1,1488 @@
+(* ===================================================================== *)
```
xrchz: I'm confused about why this file is appearing here new. Where is it copied from, and why do we need a copy?

charles-cooper:
### Demo: chain verification output (F7)

`verify_chain` replays 30 theories in dependency order, passing each theory's replayed exports as context for its descendants (eliminating `DISK_THM` oracle tags):

```
Replay chain: 30 theories in dependency order
  bool ... FAIL: 5/191 exports have oracle tags (fallback=5 placeholder=0)
  marker ... OK (29 exports, 8 trusted)
  num ... OK (10 exports, 3 trusted)
  normalForms ... OK (4 exports, 3 trusted)
  sat ... OK (24 exports, 23 trusted)
  combin ... OK (68 exports, 40 trusted)
  relation ... OK (262 exports, 190 trusted)
  prim_rec ... OK (47 exports, 40 trusted)
  quotient ... OK (100 exports, 66 trusted)
  pair ... FAIL: 90/115 exports have oracle tags (fallback=90 placeholder=0)
  arithmetic ... OK (524 exports, 500 trusted)
  numeral ... OK (42 exports, 37 trusted)
  sum ... FAIL: 57/63 exports have oracle tags (fallback=57 placeholder=0)
  one ... OK (12 exports, 7 trusted)
  option ... FAIL: 79/93 exports have oracle tags (fallback=79 placeholder=0)
  basicSize ... OK (9 exports, 5 trusted)
  reduce ... OK (3 exports, 3 trusted)
  cv ... OK (63 exports, 57 trusted)
  While ... OK (33 exports, 28 trusted)
  numpair ... OK (40 exports, 36 trusted, 31 compute-oracle)
  divides ... OK (89 exports, 87 trusted, 41 compute-oracle)
  pred_set ... FAIL: 2/838 exports have oracle tags (fallback=2 placeholder=0)
  ind_type ... FAIL: 7/36 exports have oracle tags (fallback=7 placeholder=0)
  list ... FAIL: 606/646 exports have oracle tags (fallback=606 placeholder=0)
  rich_list ... FAIL: 5/788 exports have oracle tags (fallback=5 placeholder=0)
  indexedLists ... OK (52 exports, 52 trusted, 15 compute-oracle)
  ConseqConv ... OK (41 exports, 19 trusted)
  quantHeuristics ... OK (121 exports, 112 trusted, 16 compute-oracle)
  patternMatches ... OK (99 exports, 84 trusted, 14 compute-oracle)
  hol ... OK (0 exports)

=== Chain summary: 22 OK, 8 FAILED ===
```

**22/30 theories verify cleanly** through the kernel with zero oracle-tagged exports ("trusted" = ancestor theorems resolved from the chain rather than from the session's `DISK_THM`). "compute-oracle" counts are COMPUTE steps that would need `cv_compute` to verify (the replayer supports a pluggable compute verifier, but it's not enabled in chain mode by default).

**8 failures** are all from a known limitation: replaying type definitions (`prim_type_definition`) in a session where those types already exist. The kernel doesn't error — it silently creates a corrupted type — so the replayer uses `mk_defn_thm` as a fallback, but the resulting theorem lacks the existential structure that subsequent `CHOOSE` steps need. This cascades: bool's 5 `itself_*` fallbacks propagate through pair → sum → option → list.

A standalone replay tool that runs in a bare kernel session (no preloaded theories) would get 30/30, since `prim_type_definition` succeeds on first definition. This is a follow-up item.

*co-authored by claude opus 4.6*

charles-cooper:
### Trace merging (F8)

Per-theory traces don't need merging during the build — each theory writes its own `.pftrace` file, parallel Holmake workers use PID-based temp files, no coordination needed.

However, a **post-build merge tool** could be a useful follow-up. The idea:

1. Read all `.pftrace` files in dependency order
2. Remap type/term/step IDs into a single global namespace (offset by cumulative counts from previous theories)
3. Replace TRUST entries (which embed full ancestor theorem statements) with direct step references to the ancestor theory's export step IDs
4. Write a single merged `.pftrace`

Benefits:
- **Simpler verification**: single-file replay, no chain resolution needed
- **Size reduction**: eliminates redundant TRUST entries (e.g. `arithmetic` alone has 500 ancestor references with full conclusions)
- **Solves loaded-session limitation**: a bare-kernel verifier processing the merged file in order sees every `prim_type_definition` as fresh → should get 30/30 clean

The TRUST→step-reference resolution is essentially what `verify_chain` already does with its `ancestor_ctx` map, so the core logic exists.

Probably better as a follow-up rather than expanding this PR's scope.

*co-authored by claude opus 4.6*

src/thm/std-thmsig.ML
charles-cooper: Agreed — will unify DISK_THM and ORACLE in the trace format. They're both opaque steps from the replayer's perspective, and the tag on the theorem already distinguishes them. Will update in a follow-up commit.

*co-authored by claude opus 4.6*

src/thm/std-thmsig.ML
charles-cooper: The result thm is only in the in-memory `trace_step` datatype — it's **not** written to the on-disk trace format. The on-disk format only records inputs (rule name + operand IDs + parent step IDs). The replayer reconstructs the result by calling the actual kernel rule.

The result is in the datatype because `record_hook` needs to call `Thm.trace_id` on the result to assign the step's ID, and it needs `Thm.dest_thm` for opaque steps (ORACLE, AXIOM, etc.) where the conclusion must be recorded. Having it in the constructor avoids a separate lookup.

Happy to restructure the datatype if you'd prefer the result separated out (e.g. a wrapper `trace_id * trace_step` instead of embedding it in each constructor).

*co-authored by claude opus 4.6*

src/thm/Thm.sml
charles-cooper: Correct — the result thms are only in the in-memory datatype, not exported to the on-disk trace. See reply on Thread 4 for why they're in the datatype.

*co-authored by claude opus 4.6*

src/postkernel/prooftrace/bench_build.sh
charles-cooper: Agreed — will move `bench_build.sh`, `bench_traces.sml`, and `verify_all_traces.sml` into `src/postkernel/prooftrace/`. Could also put benchmarking scripts in the root if that's preferred — let me know.

*co-authored by claude opus 4.6*

src/postkernel/prooftrace/FORMAT.md
charles-cooper: Agreed — will move `FORMAT.md` into `src/postkernel/prooftrace/`.

*co-authored by claude opus 4.6*

src/postkernel/FORMAT.md
charles-cooper: Agreed — COMPUTE is a kernel rule and should be replayed properly, not trusted. The replayer already supports this via a pluggable `compute_verifier` hook (uses `cv_computeLib.cv_compute` to re-evaluate and check the result matches). Currently it's not enabled by default in chain mode because the compute libraries aren't loaded, but it should be.

Will update the FORMAT.md description to remove "trusted" from COMPUTE and document the verification approach.

*co-authored by claude opus 4.6*

src/thm/Thm.sml
charles-cooper: This is a mistake — `src/thm/Thm.sml` is a build artifact (copy of `std-thm.ML`, created by the Holmakefile) and is in `.gitignore`. It was accidentally `git add -f`'d in commit 6c31516. Will remove it.

*co-authored by claude opus 4.6*

src/thm/std-thmsig.ML
mn200: There's a code-smell from having (big) datatype definitions repeated in `.sig` and `.sml` files.  Please extract the definition into something like `thm_ptrace_dtype.sml`:
```
  structure thm_ptrace_dtype =
  struct
     datatype ..
  end
```
and then write `datatype trace_step = thm_ptrace_dtype.trace_step` in the signature file; in the `.sml` file, just `open thm_ptrace_dtype`.

mn200:
Have you timed the effect of having a "is the trace hook `NONE`?" check on every inference rule when running in “normal mode”?

mn200:
I wonder if the `Count` tracing couldn't be merged with the tracing machinery.

charles-cooper:
> Have you timed the effect of having a "is the trace hook `NONE`?" check on every inference rule when running in “normal mode”?

| Variant | Wall-clock |
|---------|-----------|
| Hooks present, `trace_hook = NONE` | 9m06s |
| Hooks present, `trace_hook = NONE` | 9m26s |
| Hooks present, `trace_hook = NONE` | 9m13s |
| `fun tr _ = ()` (no-op, variant still constructed) | 9m06s |

All within ~20s variance

*co-authored by claude opus 4.6*

charles-cooper:
> I wonder if the `Count` tracing couldn't be merged with the tracing machinery.

can maybe fold counting into `tr` so there's one dispatch instead of two, but Count also works with tracing disabled (gated on `Count.counting`), so it can't fully depend on the trace hook.

 *co-authored by claude opus 4.6*

src/thm/std-thmsig.ML
charles-cooper: Makes sense. The wrinkle is that every `trace_step` constructor references `thm`, which is defined inside `Thm` itself (stdknl's `thm` has an extra `int` trace ID vs `KernelTypes.thm`). So `thm_ptrace_dtype` compiled before `Thm` can't see `thm`, and compiled after creates a cycle since `Thm` uses `trace_step` in every inference rule.

The existing `*_dtype` patterns (`type_tokens_dtype`, `quotation_dtype`) avoid this because they don't reference the enclosing module's types — they use type parameters instead. We could parameterize (`('thm, 'term, 'ty) trace_step`) but with 42 constructors that's pretty noisy.

Is there a pattern you had in mind that gets around the `thm` dependency, or would a Holmakefile-level textual include (cat the datatype into both files from a shared source) be acceptable?

*co-authored by claude opus 4.6*

mn200: You only need to parameterise on the `thm` position, so instances would be `thm trace_step`, and the declaration would be `datatype 'a trace_step = ....'a trace_step ...`

mn200:
> > Have you timed the effect of having a "is the trace hook `NONE`?" check on every inference rule when running in “normal mode”?
> 
> Variant	Wall-clock
> Hooks present, `trace_hook = NONE`	9m06s
> Hooks present, `trace_hook = NONE`	9m26s
> Hooks present, `trace_hook = NONE`	9m13s
> `fun tr _ = ()` (no-op, variant still constructed)	9m06s
> All within ~20s variance
> 
> _co-authored by claude opus 4.6_

I don't understand what these figures are.  I meant that I wanted to see timings for the proposed `Thm` *vs* what we currently have.

src/thm/std-thmsig.ML
xrchz: Done in [2255ee6](https://github.com/HOL-Theorem-Prover/HOL/pull/1825/commits/2255ee67885d6464723ad4e1201dbb99bbc19a4a)

charles-cooper:
> > > Have you timed the effect of having a "is the trace hook `NONE`?" check on every inference rule when running in “normal mode”?
> > 
> > 
> > Variant	Wall-clock
> > Hooks present, `trace_hook = NONE`	9m06s
> > Hooks present, `trace_hook = NONE`	9m26s
> > Hooks present, `trace_hook = NONE`	9m13s
> > `fun tr _ = ()` (no-op, variant still constructed)	9m06s
> > All within ~20s variance
> > _co-authored by claude opus 4.6_
> 
> I don't understand what these figures are. I meant that I wanted to see timings for the proposed `Thm` _vs_ what we currently have.

that last run was a run where claude edited `tr` directly instead of setting `trace_hook = NONE`, i guess its point was to measure trace_hook = NONE vs not calling the hook at all

vs develop:

 | Branch | Run 1 | Run 2 |                                                     
 |--------|-------|-------|                                                     
 | develop (8fa843af7) | 9m02s | 9m02s |                          
 | proof-traces (d0d90e36c, `trace_hook = NONE`) | 9m09s | 8m51s |

xrchz:
### Design revision: merged traces for from-scratch replay

After reviewing the design more carefully, I think the approach should
be revised around a simpler, more useful core use case.

#### Use case

The primary use case for proof traces is **independent verification**:
replay a complete proof from scratch to produce oracle-free theorems,
either in a fresh HOL4 session or in an independent proof checker.

The current per-theory chain verification (`verify_chain`) doesn't
serve this well. It replays inside a running HOL session where types
and constants already exist. This can be worked around (e.g., by
substituting existing constants instead of redefining), but it
overcomplicates the replayer with `already_defined` checks and
fallback paths. From-scratch replay avoids all of this.

Partial replay based on pre-built HOL4 ancestors doesn't add much
value over the existing `.dat` file approach.

#### New design

**Per-theory traces** become build intermediates. During
`HOL_TRACE_PROOFS=1` builds, each theory still produces a `.pftrace`
file recording its kernel steps. The format simplifies:

- Remove `PARENTS` and `ANCESTOR` header lines (not needed)
- `ORACLE DISK_THM` entries record only the source theory and
  theorem name (e.g., `ORACLE DISK_THM bool TRUTH`), not the full
  theorem statement. This requires threading the theorem name through
  `disk_thm` from `SharingTables` where it's already available.
  This avoids building/interning terms for ancestor theorems in the
  intermediate trace.
- Everything else stays the same

**A merge tool** combines per-theory traces into a single
self-contained trace file:

1. Read per-theory traces in dependency order (inferred from the
   filesystem / theory dependency graph)
2. Remap type/term/step IDs into a single global namespace
3. Replace `ORACLE DISK_THM <theory> <name>` entries with direct
   references to the ancestor theory's export step (resolved by name)
4. Write one merged `.pftrace` file

**The replayer** becomes simple: parse one file, replay every step
through the kernel from a bare session. All definitions succeed
(fresh kernel, no pre-existing types/constants). Only genuine
axioms and oracles remain as trust boundaries. No chain resolution,
no `already_defined` workarounds, no `compute_verifier` hooks.

#### What this removes

- `PARENTS`, `ANCESTOR` header lines
- The entire `verify_chain` apparatus (`toposort`, `get_parents`,
  `check_ancestors`, `digest_map`, `ancestor_ctx`, `ctx_lookup`,
  `ctx_add_exports`, `replay_file_ctx`)
- `already_defined` workarounds in DEF_TYOP/DEF_SPEC replay
- `ext_thm_cache` complexity in TraceRecord/TraceExport (no longer
  need to intern ancestor theorem statements)
- `compute_verifier` hook (COMPUTE should be replayed properly —
  see below)

#### COMPUTE replay

Currently COMPUTE records the result term and treats replay as opaque.
This is wrong: `Thm.compute` is a conservative kernel operation and
should be replayed through the same entry point. This requires
recording the initialization arguments (`cval_terms`, `cval_type`,
`num_type`) and code equation theorems in the trace. Working on the
right design for this.

#### Other recent cleanups (already committed)

- Extracted `trace_step` datatype into parameterized shared file
  (`trace_step_dtype.sml`) per mn200's review feedback
- Unified TRUST and ORACLE in the trace format
- Fixed FORMAT.md: removed phantom rules (DEDUCT_ANTISYM, ADD_ASSUM),
  added missing rules (Beta, EQ_IMP_RULE1/2, Specialize), fixed
  parent counts and operand descriptions
- Replay Mk_comb/Mk_abs through actual kernel entry points instead
  of decomposing into TRANS + MK_COMB / TRANS + ABS

#### Memory considerations

Type/term deduplication during recording uses `Redblackmap` keyed by
structural comparison. The merge tool will need to re-intern across
theories into a global namespace. For very large builds, memory
pressure from the global intern tables may need attention — will
benchmark and address as needed.

#### Build timing overhead

The baseline timing comparison (current `develop` kernel vs proposed
kernel with hooks disabled) will be done after the redesign settles.

mn200:
It seems like we should hold off a little on the merge then?

xrchz:
> It seems like we should hold off a little on the merge then?

Yes indeed! I'll convert this to draft until I'm ready to merge.

xrchz:
OK I ran these benchmarks, will post results next:
```
#!/usr/bin/env fish
git checkout c38cd4913
git clean -xdf
poly --script tools/smart-configure.sml
/usr/bin/time -v bin/build --nograph &| tee /tmp/bench-a.log
git checkout cc2167173
git clean -xdf
poly --script tools/smart-configure.sml
/usr/bin/time -v bin/build --nograph &| tee /tmp/bench-b.log
git clean -xdf
poly --script tools/smart-configure.sml
/usr/bin/time -v bin/build --nograph --trace &| tee /tmp/bench-c.log
cd examples/formal-languages/regular
/usr/bin/time -v ../../../bin/Holmake --trace &| tee /tmp/bench-c-regexp.log
cd -
find . -name "*.pft" -o -name "*.pft.zst" | \
  xargs du -ch | tee /tmp/bench-c-size.log
/usr/bin/time -v bin/prooftrace merge -o bench.pft \
  arithmetic.DIVISION \
  regexp_compiler.regexp_matcher_correct \
  &| tee /tmp/bench-d.log
zstd bench.pft
ls -lh bench.pft bench.pft.zst | tee -a /tmp/bench-d.log
/usr/bin/time -v bin/prooftrace replay bench.pft \
  &| tee /tmp/bench-e.log
```

More rationale for the commands is in https://github.com/charles-cooper/HOL/blob/proof-traces/src/postkernel/prooftrace/BENCHMARKS.md

The machine info is 
```
11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
32GB RAM (2x Corsair DRAM DDR4 16GiB)
Non-Volatile memory controller: Intel Corporation SSD 670p Series [Keystone Harbor] (rev 03)
```

xrchz:
```
(develop)
	Command being timed: "bin/build --nograph"
	User time (seconds): 1264.85
	System time (seconds): 83.80
	Percent of CPU this job got: 223%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 10:04.00
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 1642600
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 5
	Minor (reclaiming a frame) page faults: 71336699
	Voluntary context switches: 439348
	Involuntary context switches: 10749
	Swaps: 0
	File system inputs: 0
	File system outputs: 255848
	Page size (bytes): 4096
	Exit status: 0

(proof-traces)
	Command being timed: "bin/build --nograph"
	User time (seconds): 1265.07
	System time (seconds): 83.75
	Percent of CPU this job got: 221%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 10:07.94
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 1450936
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 25
	Minor (reclaiming a frame) page faults: 71091299
	Voluntary context switches: 497756
	Involuntary context switches: 12065
	Swaps: 0
	File system inputs: 0
	File system outputs: 459328
	Page size (bytes): 4096
	Exit status: 0

	Command being timed: "bin/build --nograph --trace"
	User time (seconds): 1639.80
	System time (seconds): 95.02
	Percent of CPU this job got: 216%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 13:20.12
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 2102080
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 84
	Minor (reclaiming a frame) page faults: 75985255
	Voluntary context switches: 643018
	Involuntary context switches: 13431
	Swaps: 0
	File system inputs: 0
	File system outputs: 11999256
	Page size (bytes): 4096
	Exit status: 0

	Command being timed: "../../../bin/Holmake --trace"
	User time (seconds): 354.23
	System time (seconds): 13.08
	Percent of CPU this job got: 172%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 3:33.06
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 3356040
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 10723302
	Voluntary context switches: 208557
	Involuntary context switches: 2347
	Swaps: 0
	File system inputs: 0
	File system outputs: 4885072
	Page size (bytes): 4096
	Exit status: 0

size of compressed pft files:
1.4G	total

Found 195 trace files
Pass 1: determining live entries...
Pass 2: writing merged trace...
Merged trace: 140272 types, 2137594 terms, 8293283 thms -> bench.pft
	Command being timed: "bin/prooftrace merge -o bench.pft arithmetic.DIVISION regexp_compiler.regexp_matcher_correct"
	User time (seconds): 326.86
	System time (seconds): 21.24
	Percent of CPU this job got: 129%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 4:28.89
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 17439096
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 0
	Minor (reclaiming a frame) page faults: 13644476
	Voluntary context switches: 756077
	Involuntary context switches: 3350
	Swaps: 0
	File system inputs: 16
	File system outputs: 597800
	Page size (bytes): 4096
	Exit status: 0

-rw-r--r-- 1 ramana ramana 292M Feb 26 16:00 bench.pft
-rw-r--r-- 1 ramana ramana  69M Feb 26 16:00 bench.pft.zst

Replaying: /home/ramana/HOL1/bench.pft
<<HOL warning: Type.mk_vartype: non-standard syntax>>
<<HOL warning: Type.mk_vartype: non-standard syntax>>
<<HOL warning: Type.mk_vartype: non-standard syntax>>
OK DIVISION: bool$!
  (λn.
       prim_rec$< num$0 n ==>
       bool$!
         (λk.
              bool$/\
                (k =
                 arithmetic$+ (arithmetic$* (arithmetic$DIV k n) n)
                   (arithmetic$MOD k n)) (prim_rec$< (arithmetic$MOD k n) n)))
OK regexp_matcher_correct: regexp_compiler$dom_Brz_alt balanced_map$empty
  (balanced_map$singleton (regexp$normalize r) one$one) ==>
regexp_compiler$regexp_matcher r s = bool$IN s (regexp$regexp_lang r)

2/2 exports verified clean
	Command being timed: "bin/prooftrace replay bench.pft"
	User time (seconds): 86.36
	System time (seconds): 3.90
	Percent of CPU this job got: 270%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 0:33.42
	Average shared text size (kbytes): 0
	Average unshared data size (kbytes): 0
	Average stack size (kbytes): 0
	Average total size (kbytes): 0
	Maximum resident set size (kbytes): 7626996
	Average resident set size (kbytes): 0
	Major (requiring I/O) page faults: 9
	Minor (reclaiming a frame) page faults: 3029447
	Voluntary context switches: 68213
	Involuntary context switches: 531
	Swaps: 0
	File system inputs: 73936
	File system outputs: 0
	Page size (bytes): 4096
	Exit status: 0
```

xrchz:
Let me just copy the main results table for easy viewing:
## Results

| Run | Description | Wall time | Peak RSS | Notes |
|-----|-------------|-----------|----------|-------|
| (a) | Baseline build | 10m04s | 1.64G   | |
| (b) | PR build, tracing off | 10m08s | 1.45G | |
| (c) | PR build, tracing on | 13m20s | 2.1G | total .pft.zst size: 1.4G |
| (c') | regexp_compiler (tracing on) | 3m33s | 3.36G | |
| (d) | Merge | 4m29s | 17.44G | merged: raw (292M) / zstd (69M) (ratio 4.23) |
| (e) | Replay | 33s | 7.63G | |

xrchz:
Redid the benchmarks on latest commits:
| Run | Description | Wall time | Peak RSS | Notes |
|-----|-------------|-----------|----------|-------|
| (a) | Baseline build | 10m04s | 1.63G   | |
| (b) | PR build, tracing off | 10m08s | 1.41G | |
| (c) | PR build, tracing on | 12m43s | 1.8G | total .pft.zst size: 1.4G |
| (c') | regexp_compiler (tracing on) | 2m46s | 1.48G | |
| (d) | Merge | 4m37s | 15.63G | merged: raw (304M) / zstd (68M) (ratio 4.47) |
| (e) | Replay | 32s | 7.55G | |

xrchz:
Some possible bugs observed on `oven1`:
1. Holmake silently accepts prefixes of the flag "--trace" (this is wrong not just for "--trace"). (N.B. we want options after targets to still work)
2. Prooftrace files contain only `N` and `E` lines: the proof tracing in the kernel isn't getting loaded for some reason (despite the `--trace` flag).
3. Proof tracing doesn't give any indication whether it's on or not in the command output: 1 line indicating tracing is on would be helpful feedback.

xrchz:
I think this branch is currently not in a great state and is becoming too complicated. I will create fresh PRs in cleaner chunks, and clean up the design and implementation too.

PR #1829, opened 2026-02-21
HolSmt: add cvc5 Alethe proof replay

charles-cooper:
Add `CVC_TAC`/`CVC_PROVE` that invoke cvc5 with Alethe proof output and replay the proof in HOL4, reducing oracle usage.

## New files

- **`Alethe_Proof.sml`** (35 LOC): Proof datatypes (`ASSUME`, `STEP`, `ANCHOR`)
- **`Alethe_ProofParser.sml`** (492 LOC): Parse cvc5 Alethe proof output — handles `:named` annotations, rational/negative literals, grouped output, `define-fun`/`declare-fun`/`declare-sort`, proof deduplication
- **`Alethe_ProofReplay.sml`** (1117 LOC): Replay ~50 Alethe rules via `unit_resolution`, `MK_COMB`, `TAUT_PROVE`, `ARITH`, `METIS` fallbacks; post-processing removes extra hypotheses

## Modified files

- **`CVC.sml`**: `CVC_SMT_Prover` with `--produce-proofs`, proof parsing, replay, `--tlimit=10000` to prevent cvc5 hanging on hard proof generation
- **`HolSmtLib.sig/sml`**: Export `CVC_TAC`/`CVC_PROVE`
- **`selftest.sml`**: Add `thm_CVCp`/`sat_CVCp` test coverage; disable `CVCp` for tests where cvc5 cannot produce Alethe proofs (`pow` with `num` exponent, `realinv`, certain quantifier goals, word concatenation)

## Design notes

This code was entirely AI-generated. Design, architecture, and style feedback is very welcome.

The replay engine follows the same pattern as `Z3_ProofReplay.sml`:
- Each Alethe rule gets a dedicated replay function
- `unit_resolution` (adapted from Z3) handles clause resolution directly via kernel operations, avoiding expensive METIS search
- Oracle fallbacks (`Alethe_hole`, `Alethe_sko`, `Alethe_unknown`, `Alethe_fallback`) are used when replay fails, tagged for traceability
- METIS has a tight limit (50ms / 50 inferences) to prevent non-termination

Key known limitations:
- Many `hole` steps (cvc5 proof gaps) fall back to oracle
- Skolemization (`sko_ex`/`sko_forall`) uses oracle
- `forall_inst` sometimes fails on complex instantiations
- Some cvc5 proofs are unsupported (e.g., `div_by_zero` Skolem)
- Natural number arithmetic proofs can be very large (~2000 steps for simple subtraction goals)

Tested with cvc5 1.3.2.

Co-authored-by: Claude Opus 4.6 (Anthropic)

### DIFF
--- a/src/HolSmt/Alethe_ProofReplay.sml
+++ b/src/HolSmt/Alethe_ProofReplay.sml
@@ -89,6 +89,16 @@ local
   (* auxiliary proof functions                                               *)
   (***************************************************************************)
 
+  (* A static simpset for real arithmetic. We avoid srw_ss() because it is
+     dynamic (its contents change as theories are loaded), which is
+     undesirable in a decision procedure. real_ss is static and covers
+     most real-number simplification; we add REAL_ADD_LINV/RINV for
+     goals like -(1*x) + x = 0 that arise from Alethe proofs. *)
+  val alethe_ss =
+    simpLib.++ (realSimps.real_ss,
+    simpLib.rewrites [realTheory.REAL_ADD_LINV, realTheory.REAL_ADD_RINV,
+                      realTheory.POW_ONE])
+
   (* prove |- t by trying various automation *)
   fun auto_prove name t =
     let val (l, r) = boolSyntax.dest_eq t
@@ -102,7 +112,7 @@ local
     handle Feedback.HOL_ERR _ =>
     simpLib.SIMP_PROVE bossLib.std_ss [] t
     handle Feedback.HOL_ERR _ =>
-    simpLib.SIMP_PROVE (bossLib.srw_ss()) [] t
+    simpLib.SIMP_PROVE alethe_ss [] t
     handle Feedback.HOL_ERR _ =>
     raise ERR name ("failed to prove: " ^ Hol_pp.term_to_string t)
 
@@ -949,7 +959,7 @@ local
         metis_prove [] target
         handle Feedback.HOL_ERR _ =>
         (* 6. Simplifier (handles e.g. -(1*x) + x = 0) *)
-        simpLib.SIMP_PROVE (bossLib.srw_ss()) [] target
+        simpLib.SIMP_PROVE alethe_ss [] target
         handle Feedback.HOL_ERR _ =>
         (* 7. Evaluation *)
         Drule.EQT_ELIM (bossLib.EVAL target)


someplaceguy:
> * **`CVC.sml`**: `CVC_SMT_Prover` with `--produce-proofs`, proof parsing, replay, `--tlimit=10000` to prevent cvc5 hanging on hard proof generation

I highly recommend not using `--tlimit` because it makes proofs unreliable: whether they succeed or not will depend on how fast your computer is.

Given your choice of timeout, this becomes especially clear on proofs that take close to 10,000 ms on modern hardware: you will find that they will succeed on some machines but not others, or they may even succeed or fail on the same machine depending on how much background activity there is going on. If some user is running cvc5 under some kind of emulator, slow hardware or heavily-loaded machine then this problem becomes even worse.

If you really want to hardcode a timeout in the code, I suggest using cvc5's `--rlimit` or `--rlimit-per` options together with a fixed `--seed / --random-seed`, which make proofs deterministic. That said, I'm not sure that it's such a good idea because the resource usage will likely change depending on the exact cvc5 version, and therefore whether proofs succeed or not, will depend on the exact cvc5 version that the user has installed.

With Z3 I decided not to set any kind of timeout or resource limit at all, by default: this means that during the self-tests and during a `Holmake` build, Z3 will keep trying to find a proof indefinitely (i.e. until it succeeds). I think this is a reasonable choice because in these situations, Z3 is only being used for proofs in places where it's already known that it can find a proof.

For interactive development, where it's not clear whether Z3 can find a proof or not, I expect users to write a simple wrapper script that sets a reasonable timeout and tell HOL4 to use that (rather than the Z3 binary directly).

src/HolSmt/Alethe_ProofReplay.sml
```diff
@@ -0,0 +1,1117 @@
...
+  fun arith_prove t =
+    intLib.ARITH_PROVE t
+    handle Feedback.HOL_ERR _ =>
+    RealField.REAL_ARITH t
+    handle Feedback.HOL_ERR _ =>
+    raise ERR "arith_prove" ("failed: " ^ Hol_pp.term_to_string t)
+
+  (* try to prove a tautology by METIS *)
+  (* METIS with a time/inference limit to avoid non-termination *)
+  val metis_limit : mlibMeter.limit = {time = SOME 0.05, infs = SOME 50}
```
someplaceguy: I highly recommend not putting a time limit here for the exact same reason I recommended not using `--tlimit`. An inference limit is fine, though.

charles-cooper:
@someplaceguy 13e1beea0

charles-cooper:
fwiw i think for agent-directed workflows, builtin timeouts (that can maybe be increased but never removed entirely) are good because agents have no sense of time -- where a human would kill after a few seconds or whenever they get impatient, agents are happy as a clam to run requests that last forever and does your system lol. but this can be handled at a higher level i guess (e.g. in https://github.com/charles-cooper/hol4-mcp everything is guarded with timeouts)

someplaceguy:
> fwiw i think for agent-directed workflows, builtin timeouts (that can maybe be increased but never removed entirely) are good because agents have no sense of time -- where a human would kill after a few seconds or whenever they get impatient, agents are happy as a clam to run requests that last forever and does your system lol. but this can be handled at a higher level i guess (e.g. in https://github.com/charles-cooper/hol4-mcp everything is guarded with timeouts)

Indeed, I think that should be done at a higher level, because timeouts only make sense in some contexts (interactive development, agents) but not others (self-tests, Holmake builds).

For what it's worth, I'm simply using the following script for interactive development:
```bash
#!/usr/bin/env bash

set -euo pipefail

cvc5 --tlimit=60000 "$@"
```
And then I set `HOL4_CVC_EXECUTABLE=/path/to/script.sh`.
And I'm doing something similar for Z3.

someplaceguy:
> ## Design notes
> This code was entirely AI-generated. Design, architecture, and style feedback is very welcome.

So, I'm still trying to figure out how impressed I should be with Claude's abilities in this PR  :laughing: .

This PR revealed an important gap in the HolSmt self-tests: the test suite is not checking that the implementation is not cheating by using oracle tags.

And indeed Claude seems to have taken advantage of this situation. Once I added this oracle check (my first vibecoded feature, although quite small and easy to review), many self-tests are failing to do a full proof replay:

```
Running functional tests...
.................................<<HOL warning: Selftest.expect_thm: Test of solver 'cvc5 (proofs)' failed: proof replay produced oracle tag(s): Alethe_hole>>
...................................................................<<HOL warning: Selftest.expect_thm: Test of solver 'cvc5 (proofs)' failed: proof replay produced oracle tag(s): Alethe_hole>>
.....<<HOL warning: Selftest.expect_thm: Test of solver 'cvc5 (proofs)' failed: proof replay produced oracle tag(s): Alethe_hole>>
..................................<<HOL warning: Selftest.expect_thm: Test of solver 'cvc5 (proofs)' failed: proof replay produced oracle tag(s): Alethe_hole>>
.....<<HOL warning: Selftest.expect_thm: Test of solver 'cvc5 (proofs)' failed: proof replay produced oracle tag(s): Alethe_hole>>
............................<<HOL warning: Selftest.expect_thm: Test of solver 'cvc5 (proofs)' failed: proof replay produced oracle tag(s): Alethe_fallback, Alethe_hole, Alethe_subproof>>
.....<<HOL warning: Selftest.expect_thm: Test of solver 'cvc5 (proofs)' failed: proof replay produced oracle tag(s): Alethe_fallback, Alethe_hole>>
.....<<HOL warning: Selftest.expect_thm: Test of solver 'cvc5 (proofs)' failed: proof replay produced oracle tag(s): Alethe_fallback, Alethe_hole>>
.........<<HOL warning: Selftest.expect_thm: Test of solver 'cvc5 (proofs)' failed: proof replay produced oracle tag(s): Alethe_fallback, Alethe_hole, Alethe_subproof>>
.....<<HOL warning: Selftest.expect_thm: Test of solver 'cvc5 (proofs)' failed: proof replay produced oracle tag(s): Alethe_fallback, Alethe_hole, Alethe_subproof>>
```
(... etc).

Note that Z3's proof replay does not use any oracle tags, all of them are still succeeding despite this oracle check that I just added to my local HOL checkout.

Strictly speaking, it's not exactly cheating because the PR description does mention that one of the limitations is that oracles are being used during proof replay:

> Key known limitations:
> 
> * Many `hole` steps (cvc5 proof gaps) fall back to oracle

However, the problem is that if most proofs are using oracles to summon theorems out of nothing, then proof replay is not being very useful.

I'm now trying to disable those tests that are using such oracles and trying to see how much of a problem this is, i.e. how many tests of cvc5 proof replay both 1) succeed and 2) can be solved by cvc5 but not with HOL4's built-in automated procedures (i.e. HolSmt's `thm_AUTO`). Because, if the working steps of cvc5's proof replay are mostly relying on HOL4's built-in procedures such as metis and arith_tac, then the proof replay is not really adding anything useful yet...

src/HolSmt/Alethe_ProofReplay.sml
```diff
@@ -0,0 +1,1117 @@
...
+                handle Feedback.HOL_ERR _ =>
+                  raise ERR "replay_equiv" "premise not an equality"
+            val eq_imp = Thm.MP (Thm.SPEC r (Thm.SPEC l boolTheory.EQ_IMP_THM)) p
+            val imp1 = Thm.CONJUNCT1 eq_imp
+            val imp2 = Thm.CONJUNCT2 eq_imp
+          in
+            (* try building the target clause from imp1/imp2 *)
+            metis_prove [imp1, imp2] target
+          end
+      | _ => raise ERR "replay_equiv" "expected 1 premise"
```
someplaceguy, line 636: There seems to be a bug here due to precedence:

```sml
src/HolSmt/Alethe_ProofReplay.sml:599: warning: Matches are not exhaustive.
  Found near
    case prems of
       [p] =>
       metis_prove [...] target
       handle
          Feedback.HOL_ERR _ => let val ...; ... in ... ... end |
          _ => raise ...
```

That is, the branch `| _ => (...)` actually seems to actually be pattern-matching on exception handling (i.e. `handle`) rather than the `case` statement.

src/HolSmt/Alethe_ProofReplay.sml
```diff
@@ -0,0 +1,1117 @@
...
+            prove_tautology clause
+            handle Feedback.HOL_ERR _ =>
+            Thm.mk_oracle_thm "Alethe_subproof" ([], target)
+        in (cache_step s id thm, thm) end
+    | "symm"               =>
+        let val thm = case prems of
+              [p] => Thm.SYM p
+                handle Feedback.HOL_ERR _ =>
+                metis_prove [p] (clause_to_disj clause)
+            | _ => prove_tautology clause
```
someplaceguy, line 1152: Same precedence bug here?
```sml
src/HolSmt/Alethe_ProofReplay.sml:959: warning: Matches are not exhaustive.
  Found near
    case prems of
       [p] =>
       Thm.SYM p handle Feedback.HOL_ERR _ => ... ... (... ...) | _ => ... ...
```

charles-cooper:
yea i will try to address. btw is it possible to disable oracle tags completely in hol builds (both bin/build and holmake)..? seems like not good to have those leak into proofs by accident

someplaceguy:
> yea i will try to address. btw is it possible to disable oracle tags completely in hol builds (both bin/build and holmake)..? seems like not good to have those leak into proofs by accident

You can do `Globals.show_tags := true;` which will print a theorem's tags when the theorem is printed. I'm not sure if they can be disabled, though, perhaps @mn200 can chime in.

In either case, in a few minutes I will submit a PR to check for oracle tags in the HolSmt self-tests, perhaps you can just use that for now.

someplaceguy:
> yea i will try to address. btw is it possible to disable oracle tags completely in hol builds (both bin/build and holmake)..? seems like not good to have those leak into proofs by accident

Regarding those gaps in the proof steps, which are currently being handled with oracle tags, I'm wondering if Claude can be coerced into transforming them into unproven assumptions which are added into the final proved theorem.

This would avoid the use of oracle tags and although it would not help in making more tests succeed, it might still be useful, because those assumptions could be proved by a user afterwards when doing interactive development.

In fact, I even started adding such a feature myself to the Z3 proof replay implementation, although unfortunately I haven't been able to complete it yet due to other priorities.

charles-cooper:
> > yea i will try to address. btw is it possible to disable oracle tags completely in hol builds (both bin/build and holmake)..? seems like not good to have those leak into proofs by accident
> 
> Regarding those gaps in the proof steps, which are currently being handled with oracle tags, I'm wondering if Claude can be coerced into transforming them into unproven assumptions which are added into the final proved theorem.
> 
> This would avoid the use of oracle tags and although it would not help in making more tests succeed, it might still be useful, because those assumptions could be proved by a user afterwards when doing interactive development.
> 
> In fact, I even started adding such a feature myself to the Z3 proof replay implementation, although unfortunately I haven't been able to complete it yet due to other priorities.

hmm so you mean remove oracle tags completely?

someplaceguy:
> hmm so you mean remove oracle tags completely?

Yes. And instead, for those proof steps that the current implementation can't replay or prove, add them as unproven assumptions to the final theorem returned by cvc5's proof replay procedure.

Although, this is just an idea, not sure if it can be feasibly implemented.

charles-cooper:
i'm all for it, altho seems like it should be in a follow up PR otherwise this will get too much scope creep

charles-cooper:
unless you think it will be better to restructure the code starting in this PR..?

someplaceguy:
> unless you think it will be better to restructure the code starting in this PR..?

The Z3 proof replay code (Z3_TAC / Z3_PROVE) always tries to replay proofs without using any oracles. If it can't, it fails. But if the user is fine with using Z3 as an oracle, then Z3_ORACLE_TAC / Z3_ORACLE_PROVE is available, which does not even try to replay any steps.

This PR implements a hybrid of those: it tries to replay the cvc5 proof, but for the proof steps that fail, it behaves as an oracle. I think this can be confusing, because the user may think cvc5 has successfully replayed a proof, when in fact the theorem may not have been proven at all.

On the other hand, a procedure that replays some proof steps may be better than one that doesn't replay any, as it can increase confidence that the proof is correct (but obviously, a partial proof may contain fatal unchecked mistakes, so it's not a replacement for a full proof).

I've submitted PR #1830 which forbids proof replay implementations from using oracle tags and axioms (enforced only in self-tests and in Z3's existing implementation, but you can add a similar check for cvc5 in this PR, if/when appropriate).

Personally, I think there are at least these 2 options:
1. For now, keep the current cvc5 implementation, but rename `CVC_{PROVE,TAC}` to (perhaps) `CVC_PARTIAL_{PROVE,TAC}`, to indicate that these procedures can succeed even though they only did a partial proof replay, not the full replay. A clear explanation of this, and their behavior as oracles, should also be added to the HolSmt documentation. Perhaps then we can merge this PR, and later improve the implementation on a subsequent PR, if desired.
2. Or, change the current cvc5 proof replay implementation to completely remove the usage of oracles (perhaps adding unproven replay steps as assumptions, as I mentioned earlier, or just making proof replay fail, like Z3 is doing now). And then merge this PR.

Personally, I'm fine with either of those options. But I don't think we should merge the implementation as it is now, because calling this implementation "proof replay" when in fact it's behaving (at least partially) as an oracle is a bit unexpected... :confused:

charles-cooper:
yea, going to work on option 2) and leave the partial replayers to a future PR

charles-cooper:
Pushed commit 044d59ed1 implementing option 2: all oracle usage removed from Alethe proof replay.

**What changed:**
- `mk_oracle_thm` is no longer called anywhere in `Alethe_ProofReplay.sml` — proof replay now fails on unsupported steps rather than silently producing oracle-tagged theorems
- Fixed several bugs that were masked by the oracle fallback (e.g. `auto_prove` was returning wrong theorems for non-reflexive equalities via `REFL`)
- Properly implemented `replay_bind` (FORALL_EQ/EXISTS_EQ from cached sub-steps), `replay_forall_inst` (SPEC + excluded middle), `replay_cong` (recursive dest_comb for n-ary→binary ops), and subproof discharge (DISCH + IMP→DISJ)
- Parser fix: `@p_N` named terms now persist across subproof scope boundaries
In follow-up commits (`5cae7539e`, `4eb79d723`):
- Improved `replay_hole` with targeted tactic dispatch: bitvector → `WORD_DECIDE`, quantified propositional → strip quantifiers + `TAUT_PROVE` + rewrap, first-order → `METIS_PROVE`, ground terms → `EVAL`
- Added Abbrev marker unfolding in `not_implies1` rule

**Test status:**
- 305 proof replay tests enabled (all pass locally, zero oracle tags)
- Remaining exclusions: integer div/mod, `1^y` EXP, num×variable multiplication, MIN/MAX transitivity, `realinv` on fractions — each with explanatory comment
`CVC_PARTIAL_PROVE`/`CVC_PARTIAL_TAC` (oracle fallback for user convenience) can be added in a separate PR later if desired.

*Co-authored with Claude Opus 4.6*

charles-cooper:
removing the proof size limit btw, that seems silly

mn200:
It looks as if the regression failures are your problem.

charles-cooper:
Recent pushes:

- `996e00870`: fix CI — add `all_simplify` rule handler + upgrade CI Dockerfile to cvc5 1.3.2 (CI had cvc5 1.1.2 which emits `all_simplify` instead of `refl`)
- `43257599d`: add `srw_ss()` simplifier fallback in `auto_prove` and `replay_hole` dispatch — handles goals like `-(1 * x) + x = 0` where `std_ss` lacks necessary rewrites. Re-enables test `0 < (1:real) pow y`.
- `e20d7b9cf`: merge with develop (picks up oracle tag checking from #1830)
- `f48ac2a98`: fix `thm_CVCp` to pass `check_oracles=true` after merge

**Current state**: 306 active proof replay tests, 0 oracle tags, 21 excluded tests. The excluded tests are all due to upstream cvc5 limitations (returns `unknown` for nonlinear EXP/pow, very slow proof generation for `42*x` multiplication, empty proof for `realinv` fractions) or missing parser support (`@bit_of`). None are replay bugs.

The stdknl/expk CI failures on earlier commits were due to the develop merge conflict on `expect_thm` — should be fixed now with `f48ac2a98`.

Co-authored-by: Claude (Anthropic, claude-opus-4-6)

mn200:
- Using `srw_ss()` in a decision procedure is not necessarily a great idea.  Can you find an existing, static, simpset that deals with all the troublesome cases you've encountered to date?  (E.g., something from `realSimps`.)
- @binghe do we have `cvc5` installed on the Docker image?  
    - Should we distribute our own snapshots of it as part of HOL (as we do with `minisat`)?

charles-cooper:
Good point about `srw_ss()`. I've replaced it with a static simpset:

```sml
val alethe_ss =
  simpLib.++ (realSimps.real_ss,
  simpLib.rewrites [realTheory.REAL_ADD_LINV, realTheory.REAL_ADD_RINV,
                    realTheory.POW_ONE])
```

`real_ss` covers most real-number simplification. The three extras handle goals like `-(1*x) + x = 0` and `1 pow y = 1` that arise from Alethe proofs. All 306 functional tests pass with this change (commit aa5dcf2e8).

_co-authored by claude opus 4.6_

binghe:
> * Using `srw_ss()` in a decision procedure is not necessarily a great idea.  Can you find an existing, static, simpset that deals with all the troublesome cases you've encountered to date?  (E.g., something from `realSimps`.)
> 
>     * @binghe do we have `cvc5` installed on the Docker image?
>       
>       * Should we distribute our own snapshots of it as part of HOL (as we do with `minisat`)?

We do have `cvc5` (version 1.1.2, recent) installed in the Docker image (`amd64` only), shown in the `Dockerfile` definition (note that the entry command in `PATH` is `cvc5-Linux` for historical reasons):
```
ARG CVC5_VERSION="1.1.2"

RUN if [ "linux/amd64" = "$TARGETPLATFORM" ]; then \
    wget -q -O cvc5-Linux-static.zip \
         https://github.com/cvc5/cvc5/releases/download/cvc5-${CVC5_VERSION}/cvc5-Linux-static.zip; \
    unzip cvc5-Linux-static.zip; \
    rm cvc5-Linux-static.zip; \
    ln -s cvc5-Linux-static/bin/cvc5 cvc5-Linux; \
fi
```
I don't think we need to distribute it as part of HOL. We do that for minisat because minisat 1.x (with proof logging) is subsumed by 2.x (no proof logging) and is not available in most OS system packages.

charles-cooper:
fyi this PR updates the cvc binary in the dockerfile to 1.3.2

mn200:
Thanks for your work on this!

src/HolSmt/Alethe_ProofReplay.sml
```diff
@@ -0,0 +1,1302 @@
...
+        in (cache_step s id thm, thm) end
+    | "not_symm"           =>
+        (* ⊢ ¬(a = b) from premise ⊢ ¬(b = a), or vice versa *)
+        let val target = clause_to_disj clause
+            val thm = case prems of
+              [p] =>
```
someplaceguy, line 1158: The newest changes seem to have another precedence issue here as well:

```sml
src/HOL/src/HolSmt/Alethe_ProofReplay.sml:1158: warning: Matches are not exhaustive.
Found near
  case prems of
     [p] =>
     let val ... = ...; val ...; ... in ... ... p end
     handle Feedback.HOL_ERR _ => ... ... target | _ => ... ...
```

someplaceguy:
I'm trying to test this PR locally but at least one of the self-tests seems to either be stuck on an infinite loop or performing excruciatingly slowly.

I'm using cvc5 1.3.2, though, so it might be emitting a slightly different proof certificate than cvc5 1.3.5... I'll try to debug this as soon as I have a chance.

someplaceguy:
> I'm trying to test this PR locally but at least one of the self-tests seems to either be stuck on an infinite loop or performing excruciatingly slowly.
> 
> I'm using cvc5 1.3.2, though, so it might be emitting a slightly different proof certificate than cvc5 1.3.5... I'll try to debug this as soon as I have a chance.

It looks like the self-tests aren't hanging, they're just really slow. Also, it looks like this is not due to cvc5 1.3.2, I think the same is happening with cvc5 1.3.5.

In particular, in PR #1830 which I submitted recently, the self-tests complete in Github's CI in ~220 seconds:

2231.4 holsmt-selftest.log                                     src/HolSmt(202s)     OK
2231.4 Finished $(HOLDIR)/src/HolSmt [#theories: 1]                         (222.090s) 

But in this PR, looking at the CI logs it looks like the HolSmt self-tests are taking more than 50 minutes:

5180.0 holsmt-selftest.log                                     src/HolSmt (50m)     OK
5180.0 Finished $(HOLDIR)/src/HolSmt [#theories: 1]                        (3028.160s)

src/HolSmt/selftest.sml
```diff
@@ -55,8 +55,14 @@ fun expect_thm check_oracles name smt_tac t =
           top_structure_of holerr ^ "." ^ top_function_of holerr ^
           " " ^ locn.toString (top_location_of holerr) ^
           ", message: " ^ message_of holerr ^ ")")
+    val (oracles, _) = Tag.dest_tag (Thm.tag thm)
+    val oracle_tags = List.filter (String.isPrefix "Alethe_") oracles
   in
-    if null (Thm.hyp thm) andalso Thm.concl thm ~~ t then ()
+    if not (null oracle_tags) then
```
someplaceguy, line 61: After PR #1830, this check (as well as the changes above) are not necessary anymore. A complete oracle and axioms check is already being done a few lines below.

PR #1830, opened 2026-02-21
HolSmt: check that theorems produced by proof replay don't contain unexpected oracle tags 

someplaceguy:
This PR has been motivated by discussion in #1829.

Note: I've also added an unrelated commit, for Z3's proof replay not to rely on the type parser, as it can cause problems depending on what the grammar looks like at any given time.

**Edit:** Added a commit that reuses Sanity's already-existing implementation of an oracle check. This existing implementation also checks for unexpected axioms, which makes this even more fool-proof.

**Note:** As a result, I've made Sanity's `check_tags` function available to outside code by adding it to `Sanity.sig`.

cc @charles-cooper

### DIFF
--- a/src/HolSmt/selftest.sml
+++ b/src/HolSmt/selftest.sml
@@ -15,7 +15,7 @@ val _ = Globals.show_assums := true
 val _ = Globals.show_types := true
 val _ = wordsLib.add_word_cast_printer ()
 *)
-
+val _ = Feedback.set_trace "PP.avoid_unicode" 1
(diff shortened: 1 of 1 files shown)


charles-cooper:
seems good to me, is there some way to ban oracles at the build level?

mn200:
> seems good to me, is there some way to ban oracles at the build level?

You can get a hook into `export_theory` to check everything being exported, and the code that loads this change could be preloaded the same way fastbuild and the like are done.

mn200:
Thanks for your work on this!

PR #1837, opened 2026-02-26
Cachekey: resolve sigobj symlinks to find .dat files

### LINKED ISSUE #1828: cachekey looks for `.dat` files in HOL/sigobj when a dependency is a “core” file
After a build, in `src/floating-point`
```
Holmake --cachekey binary_ieeeTheory
```
responds with 
```
Holmake: --cachekey: dependency arithmeticTheory.dat (/home/mn200/regression-hols/stdknl/sigobj/arithmeticTheory.dat) does not exist
```
because its analysis of the dependencies sees a dependency on `sigobj/arithmeticTheory.uo` and maps that (with the `dep_to_hashable` function) to a dependency on `sigobj/arithmeticTheory.dat` (which doesn't exist).

I think the right approach is probably to special-case dependencies on `sigobj`, and to follow the symbolic link from `sigobj/arithmeticTheory.uo` to find out where the `.dat` file is.  Alternatively, the “link to sigobj” code could also link the `.dat` files...

This will be an issue for all dependencies on theory files built under `HOL/src`.

xrchz:
When --cachekey converts a Theory .uo/.ui dependency to the corresponding .dat file, dependencies from sigobj/ would fail because only .uo/.ui files are symlinked there, not .dat files.

Fix by falling back to resolving the .uo symlink via realPath to discover the actual directory where the .dat file resides.

Closes #1828.

### DIFF
--- a/tools/Holmake/core/Holmake.sml
+++ b/tools/Holmake/core/Holmake.sml
@@ -1384,8 +1384,8 @@ fun do_cachekey thyname =
                                     val real_uo = OS.FileSys.realPath uo_fspath
                                     val real_dir = OS.Path.dir real_uo
                                     val dat_name = fromFile (DAT s)
-                                    val dat_fspath = toFSpath
-                                        (OS.Path.concat(real_dir, dat_name))
+                                    val dat_fspath =
+                                        OS.Path.concat(real_dir, dat_name)
                                   in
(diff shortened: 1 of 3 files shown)


mn200:
Can you add a regression test please?  The example from the issue is fine.

PR #1842, opened 2026-03-02
Vim/HOL: fix invisible syntax

talsewell:
Change HOLIdents to render as "Normal" class rather than "Ignore". The "Ignore" syntax class doesn't seem to be used much, isn't set by all colour schemes, and is invisible in the default "dark" theme (printing as black on black).

This affects the majority of HOL text (as opposed to ML text).

I will declare that I would also prefer to use Operator rather than Underlined for HOL operators, but I'll leave that as a matter of preference.


mn200:
I guess @xrchz’s heart emoji means he approves but will let him weigh in further if he likes.

xrchz:
I don't have much to add, lgtm, I'm using kakoune more than vim currently.

mn200:
Thanks!

PR #1844, opened 2026-03-04
metis: on classify BUG, print the HOL_ERR

talsewell:
The "metisTools.classify: shouldn't fail" message appeared recently. The real culprit of the error was problematic use of delete_const elsewhere. However it would have helped if the BUG message didn't hide the contents of the HOL_ERR that was generated. Modify things so that the HOL_ERR is printed out, which might help a user diagnose a similar problem.

### DIFF
--- a/src/metis/metisTools.sml
+++ b/src/metis/metisTools.sml
@@ -414,7 +414,11 @@ local
     end;
 in
   fun classify fms = order (First,empty) fms
-    handle HOL_ERR _ => raise BUG "metisTools.classify" "shouldn't fail";
+    handle HOL_ERR e => (
+        print "metisTools.classify: error in classifying:\n";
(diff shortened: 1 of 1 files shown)


konrad-slind:
Actually, the integer argument to `prettyPrint` is the line length, so "1" is a bold choice.

talsewell:
Argh, good reading.

It has actually run, and seemed to print normally, so maybe it ignores really daft line length suggestions?

While producing that incantation, it did occur to me that it seemed unnecessarily complicated to print out the contents of a HOL_ERR in the default way, and I wondered if that was one reason it hadn't been done.

talsewell:
Is the better incantation `print (format_ERR e);`? No idea how I missed `format_ERR` the first time.

konrad-slind:
Yes, that would be a bit higher level.

PR #1845, opened 2026-03-04
Implement frontend for HOLSource

xrchz:
With this PR, we implement a proper frontend for HOLSource.  We use the term
HOLSource to mean any code that goes through `unquote`, which is (to my
understanding, at least) basically everything from libraries to theories once
Holmake is built.

## AST

`HOLSourceAST` defines the AST, and includes HOL-exclusives such as
`Definition`, `Theorem`, `Theory` and the entirety of Standard ML with some
relaxations. For example, I think it is possible to represent:

``` val it = 4 + _; ```

in the AST. The advantage of this is that we can add useful features, as long as
it is possible to desugar into Standard ML (done by `HOLSourceExpand`).

There are two more points worth calling out about the AST: for one, it keeps
track of the locations of all tokens. Thus, one can use it to do very
fine-grained replacements. For those that were around, this is the feature I
used to do those big rewrites to "modern" syntax (and is also the main reason I
wanted a HOLSource parser). Besides that, the location information can most
likely be used in more fancy code completion features.  Another point worth
calling out is that the AST can represent "bad" Standard ML.  For example, it
can represent something like:

``` val _ = (4 + 3 val _ = 4 ```

Note how the first declaration is missing the closing parenthesis.

## Parser

`HOLSourceCore` implements a recursive descent parser.  What is worth calling
out here is that we do not have the intermediate step of a list of
tokens. Instead, tokens are produced and consumed as parsing goes on.

## Desugaring

`HOLSourceExpand` is responsible for desugaring HOLSource into Standard ML.  To
be able to use of some non-Standard ML features, `expandDec` takes some extra
arguments such as `canBindStr`.

`HOLSourcePrinter` is used to "print" the desugared AST.

Note that these two steps do not necessarily preserve location information.
However, `HOLSourceExpand` does place results into a `DecExpansion` where
applicable, so the location information is not lost.

## Incompatibilities

Almost all HOL code should keep working as is, but there are a few places to
watch out for:

1. No space after `:` in `Definition`

The following is not allowed:

```
Definition:foo x = x
End
```

instead, it must be written as

```
Definition: foo x = x
End
```

2. No mismatched quotes allowed

This means that one cannot open with a unicode quote and close with an ASCII
quote or any kind of other strange combination such as opening with the closing
unicode quote.

3. Implicit insertion of semicolon

In `HOLSource.ToSML.mkPushTranslator`, a semicolon is added after each
declaration.  My understanding is that a semicolon tells the compiler to finish
the current "block", which in practice means that type variables might not get
inferred properly, requiring some extra annotations.

## Miscellaneous

We now have a `DString` module, which implements dynamic strings based on
Word8Array.array.

## Contributors

Most of the work was done by Mario Carneiro (@digama0), with Daniel Nezamabadi
(@dnezam - the guy who wrote this comment) going around filling holes and fixing bugs ~~and ranting
about MoscowML~~.  The initial draft for the AST we sketched out together at
Chalmers and was based on the excellent `smlfmt` by @shwestrick
([repo](https://github.com/shwestrick/smlfmt)).

Co-Authored-By: Mario Carneiro <di.gama@gmail.com>
Co-Authored-By: Daniel Nezamabadi <55559979+dnezam@users.noreply.github.com>

### DIFF
--- a/tools/parsing/HOLSourceAST.sig
+++ b/tools/parsing/HOLSourceAST.sig
@@ -5,6 +5,11 @@ type ident = int * string
 
 type fileline = {file: string, line: int, col: int}
 
+(* A list of 'exp separated by tokens, e.g "e1 , e2 , e3" stored as
+  {args = [e1, e2, e3], seps = [SOME ",", SOME ","], ...}.
+  |seps| is either |args| - 1 for a normal list, |args| for a list with trailing
+  separator (e.g. "[a, b,]", an SML extension), or 0 for fully synthetic lists.
+  A NONE is for a missing separator which was inserted during parse error repair. *)
 type 'exp separated = {args: 'exp list, seps: int option list, stop: int}
 
 datatype 'a seq =
--- a/tools/parsing/HOLSourceParser.sig
+++ b/tools/parsing/HOLSourceParser.sig
@@ -1,7 +1,14 @@
 signature HOLSourceParser = sig
 
+(* A scope represents the local state of the parser, containing the list of infixes.
+  A key-value pair tok->(n, false) means that tok is a left associative infix
+  operator of precedence n, and tok->(n, true) is a right associative infix.
+  The scope is updated by declarations, both locally and at the top level. *)
 type scope = (string, int * bool) Binarymap.dict
 
+(* The initial scope, used by default at the beginning of a HOL file.
+  This includes all of the infixes from the SML basis library, as well as
+  built-in infixes from the tactic library such as >> and >- . *)
 val initialScope: scope
 
 type result = {
--- a/tools/parsing/HOLSourceAST.sml
+++ b/tools/parsing/HOLSourceAST.sml
@@ -5,7 +5,7 @@ structure HOLSourceAST :> HOLSourceAST = struct
 
    The AST is capable of representing ill-formed files. For example, Lists
    contain a record that holds `right: int option` and `stop: int`. If the
-   closing bracket is missing, `right` will be `None`, and `stop` will indicate
+   closing bracket is missing, `right` will be `NONE`, and `stop` will indicate
    where the parser guesses the list should have ended.
 
    The basic shape of the AST is based on Sam Westrick's excellent smlfmt, which

### CODE REVIEW

**Comment by dnezam** [comment]
Various bits of things that have been discussed on Zulip and during the CakeML meeting:

- [x] Move canBindStr to Systeml and do not pass it as an argument
- [x] Uncomment the quotefilter test, and replace expected with the current state instead
- [x] Rename HOLSourceCore to HOLSourceParser

**Comment by dnezam** [comment]
Updated comment to include some Elisp to jump to the correct position based on byte offset for users of the greatest editor this world has to (and will ever) offer.

**Comment by mn200** [comment]
Why is `HOLParserOld` still required?

File: developers/releasing-hol
Comment (mn200) [inline developers/releasing-hol:1]: I don't object to futzing around with `rm` *vs* `/bin/rm`, but why is it in this PR?

File: doc/hol-mode/Makefile
Comment (mn200) [inline doc/hol-mode/Makefile:1]: What's this?

File: Manual/Description/definitions.stex
Comment (mn200) [inline Manual/Description/definitions.stex:1]: Why is a PR about HOLSource messing around with the layout of code-blocks in the DESCRIPTION manual?

**Comment by digama0** [comment]
> Why is `HOLParserOld` still required?

I want to provide some space for migration. In particular, vscode uses the old parser and I want to fix it in a separate PR

File: Manual/Description/definitions.stex
Comment (digama0) [reply Manual/Description/definitions.stex:1]: I believe most of the things here are due to `Datatype` meaning something special in col 0, while it is apparently also the name of a function in these examples. Probably the function should be renamed to not be a keyword...

File: developers/releasing-hol
Comment (dnezam) [reply developers/releasing-hol:1]: At the time I was working with GNU Guix, and couldn't run any of the tests because of the hard coded locations.

File: doc/hol-mode/Makefile
Comment (digama0) [reply doc/hol-mode/Makefile:1]: seems like something odd is happening in the diff, the file history here dates back 27 years and I think the only change in this PR is the /bin/rm -> rm thing

File: tools/Holmake/tests/quote-filter/input [hunk 1]
```diff
@@ -90,8 +90,8 @@ Quote bar_x:
 End
 
 val x = #(LINE) + 10;
-val #(LINE=11) y = #(LINE)
-val s = #(FILE) #(FILE=foo.sml)
+#(LINE=11) val y = #(LINE)
```
mn200, line 93: Why this change in the test?

digama0: `#(LINE=N)` is now parsed as a `dec`, rather than being allowed anywhere like a kind of comment. (The first version had it parsed as an exp which expands to `()` but I think a dec makes more sense.) Otherwise the AST ends up completely nuts.

developers/releasing-hol
mn200: Fixing these (and the use of `#!/bin/bash`) should be a separate PR.

digama0: *Assigned to @dnezam*

Manual/Description/definitions.stex
mn200: Agreed with the latter idea.  In the meantime, I think the documentation should just be adjusted to use the keyword form.

doc/hol-mode/Makefile
mn200: It's certainly bizarre, but as it stands, this PR will resurrect an ancient and unwanted directory and `Makefile` within that directory...

digama0: oh, in that case it's probably the result of a modify/delete merge. In any case, this should be handled in the /bin/rm PR

tools/parsing/HOLSourceAST.sig
```diff
@@ -0,0 +1,362 @@
+signature HOLSourceAST = sig
+
+(* (start, content) *)
+type ident = int * string
+
+type fileline = {file: string, line: int, col: int}
+
+type 'exp delimited = {args: 'exp list, delims: int option list, stop: int}
```
mn200: It would be helpful to have comments about what some of these fields mean.  (E.g., what is the `delims : int option list`?)

digama0: The delimiters between the elements. `NONE` if the delimiter is missing and inserted

tools/parsing/HOLSourceParser.sig
```diff
@@ -0,0 +1,22 @@
+signature HOLSourceParser = sig
+
+type scope = (string, int * bool) Binarymap.dict
```
mn200, line 7: Please document what this (`scope`) is encoding and how it is doing so.

tools/parsing/HOLSourceAST.sig
mn200: Like commas in a list?  I can cope with this (though “separators” would be my preferred term; somehow I think of delimiters as things that appear to the left and right of objects like quotes, brackets, parentheses). But please comment (or document in `help/Docfiles`) either way.

developers/releasing-hol
dnezam: Hopefully done.

doc/hol-mode/Makefile
digama0: @dnezam this file should not exist in the PR and is likely the result of a merge conflict

dnezam: Pushed a commit where that gets deleted.

tools/parsing/HOLSourceAST.sig
```diff
@@ -0,0 +1,367 @@
+signature HOLSourceAST = sig
+
+(* (start, content) *)
+type ident = int * string
+
+type fileline = {file: string, line: int, col: int}
+
+(* A list of 'exp separated by tokens, e.g "e1 , e2 , e3" stored as
```
mn200, line 8: I guess the `seps` (thanks for nomenclature adjustment!) are actually token positions (being `int`s) rather than the strings that appear in the comment, but I think that equivalence is probably clear enough throughout that this is OK.

PR #1852, opened 2026-03-04
metis: BUG-print: more conventional err print

talsewell:
Follow-up to previous PR.

Use format_ERR rather than a probably-incorrect HOLPP incantation. Thanks to Konrad Slind for spotting the issue.

### DIFF
--- a/src/metis/metisTools.sml
+++ b/src/metis/metisTools.sml
@@ -416,7 +416,7 @@ in
   fun classify fms = order (First,empty) fms
     handle HOL_ERR e => (
         print "metisTools.classify: error in classifying:\n";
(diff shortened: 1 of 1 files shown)


mn200:
If this code that "shouldn't fail" reliably fails given out-of-date constants, can't it be changed so that it just raises an exception (not printing anything) with message "Out-of-date constant detected"

talsewell:
Michael: yes, that would be a better fix, if you want to write a bunch of new code for detecting that scenario.

This is just a 3-line change that exposes some diagnostic information that (in my opinion) should never have been discarded, which also happens to clarify what the issue is in the case you are proposing to detect.

mn200:
OK

Issue #1839, opened 2026-03-01
Soundness bug in cv_if


xrchz:
Similar to #1838 we have another bug in compute

xrchz:
```
Theory cvIfBug
Ancestors cv arithmetic
Libs cv_computeLib

(* f x = cv_if (cv_pair (cv_num 0) (cv_num 0)) (cv_num 1) (cv_num 0) *)
val g_def = new_definition("g_def",
  “g (x:cv) = cv_if (cv$Pair (cv$Num 0) (cv$Num 0)) (cv$Num 1) (cv$Num 0)”);

(* Interpreter: condition is Pair(Num 0, Num 0) ≠ Num 0, takes "then" branch *)
val bad_thm = cv_compute [g_def] “g (cv$Num 0 : cv)”;
(* bad_thm: |- g (Num 0) = Num 1    (WRONG!) *)

(* Correct result via cv_if3: cv_if (Pair r s) p q = q *)
val cv_if3 = cj 3 cv_if_def |> Q.GENL[‘r’,‘s’,‘p’,‘q’];
val correct = TRANS (Q.SPEC‘cv$Num 0’ g_def)
                    (SPECL [“cv$Num 0”, “cv$Num 0”, “cv$Num 1”, “cv$Num 0”] cv_if3);
(* correct: |- g (Num 0) = Num 0 *)

(* Combine: |- Num 0 = Num 1 *)
val false_eq = TRANS (SYM correct) bad_thm;

(* Derive F from |- Num 0 = Num 1 using injectivity of Num and SUC≠0 *)
val zo = EQ_MP (Q.SPECL[‘0’,‘1’] $ Q.GENL[‘m’,‘n’]Num_11) false_eq
val zs = TRANS zo ONE
val ff = MP (Q.SPEC‘0’SUC_NOT_ZERO) (SYM zs)

val true = concl ff ~~ boolSyntax.F
val [] = hyp ff
val (["DISK_THM"],[]) = Tag.dest_tag $ Thm.tag ff
```

xrchz:
The fix is something like this in Compute.sml: Change line 120 from:

 ```sml
   | If (x, y, z) =>
       exec funs env (if exec funs env x = cv_zero then z else y)
 ```

 to:

 ```sml
   | If (x, y, z) =>
       exec funs env (case exec funs env x of
                        Num n => if n = Arbnum.zero then z else y
                      | Pair _ => z)
 ```

[...]

PR #1853, opened 2026-03-05
Fix for cv_if issues

myreen:


xrchz:
Just to document what's happening here: the plan is to revert to HOL4's cv_if equations from before #1839 was solved, but then change the Candle equations to match HOL4's, since this is likely to involve fewer knock-on effects and still results in a state where HOL4's implementation matches the verified implementation.

mn200:
Hoping that Candle soon confirms our leap into the dark here :)

myreen:
> Hoping that Candle soon confirms our leap into the dark here :)

The Candle proofs have now been updated: https://github.com/CakeML/cakeml/pull/1351

PR #1854, opened 2026-03-05
Add --min option to hol

xrchz:
This makes it possible to start a hol session with access to the kernel but before boolTheory is loaded, which is useful for replaying proofs "from min" avoiding any DISK_THM oracles.


mn200 (approved):

mn200:
Thanks!

PR #1857, opened 2026-03-06
Add more arguments to compute_prf

xrchz:
I think this is necessary for replay to work when using #1309. Possibly we also should add the initialisation arguments on the same proof constructor.


xrchz:
Closing in favour of #1858

PR #1860, opened 2026-03-06
Proof trace converter

digama0:
* [ ] Depends on #1309 

Adds a parser for the tracing kernel outputs.


xrchz:
I think we shouldn't do this (and favour #1876), unless you also want the ProofTraceConverter module for other reasons.

PR #1843, opened 2026-03-02
HolSmt: nonlinear arithmetic proof replay for Z3 and cvc5

charles-cooper:
_co-authored by claude opus 4.6_

Adds nonlinear arithmetic (NIA/NRA) proof replay support to HolSmt's Z3 and cvc5 Alethe proof replayers, using the NLArith/SOSLib prover tower from #1836.

## Changes

When Z3 emits `th_lemma[arith]` or cvc5 emits Alethe `hole` steps involving nonlinear terms (e.g. `x*y >= 0`, `x*x >= 0`, sign lemmas), linear arithmetic replay fails. This PR adds a fallback to NLArith/SOSLib:

- **Library.sml**: `is_nonlinear` syntactic detection (walks term for variable–variable products in real/int types) and `nla_prove` type-dispatch (real→REAL_NLA→REAL_SOS, int→INT_SOS, num→NUM_SOS_RULE)
- **Z3_ProofReplay.sml**: NLA attempted *before* the linear tactic chain (see design note below)
- **Alethe_ProofReplay.sml**: NLA fallback in `arith_prove` and `replay_hole`
- **selftest.sml**: 5 NIA/NRA test cases (3 int, 2 real) with Z3, Z3 proof replay, and cvc5 proof replay

## Design notes

**NLA before linear tactics in Z3 replay**: In `Z3_ProofReplay.arith_prove`, the NLA fallback runs *before* the linear tactic chain (`arith_prove_linear`), not after. The linear chain's `RW_TAC` can mangle nonlinear goals (rewriting products into sums) so that NLA/SOS can no longer recognise the polynomial structure. The `is_nonlinear` guard ensures this early dispatch only fires on goals containing variable–variable products.

**REAL_SOS fallback after REAL_NLA**: `nla_prove` tries `REAL_NLA` first, then falls back to `REAL_SOS`. This is needed because `REAL_NLA` cannot handle strict inequalities with equality preconditions (e.g. goals arising from `ARITH_MULT_SIGN` proof steps), where `REAL_SOS` succeeds.

[...]

PR #1862, opened 2026-03-07
SOS: fix difficult test case

someplaceguy:
Fix the test case as discussed in the Zulip chat.

Also added it to the selftest.

_Authored-by: GPT-5.4_

cc @charles-cooper

### DIFF
--- a/src/real/RealArith.sig
+++ b/src/real/RealArith.sig
@@ -35,7 +35,7 @@ sig
                              thm list * thm list * thm list -> 'a
 
  (* for REAL_LINEAR_PROVER, 0: nothing, 1: minimal, 2+: details *)
-  val verbose_level        : int ref (* default: 1 *)
+  val verbose_level        : int ref (* default: 0 *)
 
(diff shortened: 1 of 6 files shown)


mn200:
Thanks!

mn200:
Grr.  This causes regression failures when `csdp` is not installed.  Please test your code in an environment without `csdp` in addition to one that does have it.

mn200:
I don't want to mask out all the tests that are now failing with “CSDP not found” messages because these cases used to work.  

Please sort this out.

someplaceguy:
> Grr. This causes regression failures when `csdp` is not installed. Please test your code in an environment without `csdp` in addition to one that does have it.

Sorry, I did not think of testing without csdp installed.

Perhaps it would be easier to just revert this PR? As far as I understand, similar changes have been done in #1843.

mn200:
I've just removed the SOS selftest from the regression checks.  Nothing else depends on this code behaving one way or another, and while you're still developing I think it's OK to have it in a half-broken/half-working state.  Once things finalise a bit better, the tests can be turned on.

PR #1868, opened 2026-03-13
Add mlb file for building the kernel modules with MLton 

xrchz:
Make it possible to build executables that include the HOL kernel (not the `Theory` layer currently) with MLton.

### DIFF
--- a/src/thm/mlton/README.md
+++ b/src/thm/mlton/README.md
@@ -1,0 +1,13 @@
+# Building HOL kernel modules with MLton
+
+MLton must be invoked with `-default-type intinf` so that the default
+`int` type is `IntInf.int`, matching the assumption made throughout
+the HOL codebase.
+
+For example:
+
+    mlton -default-type intinf thm-stdknl.mlb
+
+or for the experimental kernel:
+
+    mlton -default-type intinf thm-expk.mlb


src/portableML/mlton/Exn.sml
mn200, line 1: Why is this different from the Moscow ML version, which has a signature ascription and provides `get_res` and `get_exn` entry-points?

src/portableML/mosml/PrettyImpl.sml
mn200, line 1: Moscow ML doesn't seem to provide `FixedInt`, so this code looks as if it breaks the Moscow ML implementation.

src/thm/mlton/thm-stdknl.mlb
mn200, line 1: Groups of related files here should be bundled into `.mlb` files of their own.

src/portableML/poly/Arbnumcore.sml
mn200, line 1: Most of these changes seem unnecessary.

xrchz: The issue was trying to use this code on mlton (admittedly before I learned the options to enable IntInf as int on mlton...). I will see if the unchanged code works with that option on.

src/portableML/mlton/Exn.sml
xrchz: Oh I think I didn't see I could reuse that! Thanks.

src/thm/mlton/thm-stdknl.mlb
xrchz: Done

src/thm/mlton/README.md
```diff
@@ -0,0 +1,13 @@
+# Building HOL kernel modules with MLton
+
+MLton must be invoked with `-default-type intinf` so that the default
+`int` type is `IntInf.int`, matching the assumption made throughout
+the HOL codebase.
```
mn200, line 5: I think that the HOL codebase is actually int-agnostic (because Poly/ML actually changed its default).

xrchz:
```
mlton selftest-stdknl.mlb
Error: ../../portableML/poly/Arbnumcore.sml 8.25-8.34.
  Variable in structure disagrees with signature (type): zero.
    structure: val zero: [int]
    defn at: ../../portableML/poly/Arbnumcore.sml 15.5-15.8
    signature: val zero: [LargeInt.int]
    spec at: ../../portableML/poly/Arbnumcore.sig 6.7-6.10
```

xrchz: Is there a request here - like you want the comment to be reworded somehow? Or for `Arbnumcore` to be changed?

mn200: Yes, for all that you are not getting `mlton` to compile things successfully, HOL does *not* require IntInf.  Indeed, when I run `poly < tools/smart-configure.sml` I get 
```
Loading system specific functions
Note: Int.maxInt = SOME 4611686018427387903
Compiling system specific functions (sml)
Beginning configuration.
```
as part of the output.  If `Int` has a maximum, it's certainly not `IntInf`, so the comment is misleading.

mn200: I'll fix it.

PR #1815, opened 2026-02-06
Radon-Nikodym Derivative Results

jdyeager:
This pull request contains a few results about Radon-Nikodym derivatives (and supporting lemmas). Specifically, about changing the base of integration, multiplying RN derivatives, and getting the inverse of an RN derivative.
```
   [RN_deriv_integral]  Theorem
      ⊢ ∀m v f.
          f ∈ Borel_measurable (measurable_space m) ∧
          (∀x. x ∈ m_space m ⇒ 0 ≤ f x) ∧ sigma_finite_measure_space m ∧
          measure_space (m_space m,measurable_sets m,v) ∧ v ≪ m ⇒
          ∫ (m_space m,measurable_sets m,v) f = ∫ m (λx. (v / m) x * f x)
```
```
   [RN_deriv_mul]  Theorem
      ⊢ ∀m u v.
          sigma_finite_measure_space m ∧
          sigma_finite_measure_space (m_space m,measurable_sets m,u) ∧
          sigma_finite_measure_space (m_space m,measurable_sets m,v) ∧
          u ≪ m ∧ v ≪ (m_space m,measurable_sets m,u) ⇒
          AE x::m.
            (u / m) x * (v / (m_space m,measurable_sets m,u)) x = (v / m) x
```
```
   [RN_deriv_inv]  Theorem
      ⊢ ∀m v.
          sigma_finite_measure_space m ∧
          sigma_finite_measure_space (m_space m,measurable_sets m,v) ∧
          v ≪ m ∧
          measure_absolutely_continuous' m (m_space m,measurable_sets m,v) ⇒
          AE x::m.
            RN_deriv' (m_space m,measurable_sets m,v) m x = ((v / m) x)⁻¹
```

A note about the internals for record-keeping purposes. This work uses a definition (from my original implementation) that has been converted to a local `Overload` (so that it can be obscured). The definition is effectively just the property that the usual definition selects from.
```
Overload RN_deriv_property[local] = “\sa nu mu. {f | f ∈ Borel_measurable sa ∧
    (∀x. x ∈ space sa ⇒ 0 ≤ f x) ∧
    (∀s. s ∈ subsets sa ⇒ (f * (space sa,subsets sa,mu)) s = nu s)}”
```

This could be made into a definition and results about it or connecting it to `RN_deriv` could be made non-local if ever that would be useful in the future.

PR #1872, opened 2026-03-16
Promote RN_deriv_property from overload to definition

binghe:
Hi,

The recently added overload `RN_deriv_property` by #1815 is not a good use of `overload`, as it's unstable under default rewriting rules of `simp` (the gspec gets destroyed immediately when doing `simp` on `f IN RN_deriv_property sa nu mu`).
```
Overload RN_deriv_property[local] = “\sa nu mu. {f | f ∈ Borel_measurable sa ∧
    (∀x. x ∈ space sa ⇒ 0 ≤ f x) ∧
    (∀s. s ∈ subsets sa ⇒ (f * (space sa,subsets sa,mu)) s = nu s)}”
```

I think this overloading can be promoted to the following definition without changing its type.  If we understand it as a "property" of a function, then it should be naturally defined as a predicate, instead of the set of all such functions, even in HOL their types are the same:
```
   [RN_deriv_property_def]  Definition      
      ⊢ ∀sa nu mu f.
          RN_deriv_property sa nu mu f ⇔
          f ∈ Borel_measurable sa ∧ (∀x. x ∈ space sa ⇒ 0 ≤ f x) ∧
          ∀s. s ∈ subsets sa ⇒ (f * (space sa,subsets sa,mu)) s = nu s
```
In comparision with the existing `RN_deriv` (/), as indicated by the following theorems, this property is almost everythere equivalent to the canonnical `RN_deriv` and all functions satisfying the property are "almost the same" (i.e. no practical usefulness on using different ones of them):
```
   [RN_deriv_property_almost_RN_deriv]  Theorem
      ⊢ ∀sa mu nu f g.
          sigma_finite_measure_space (space sa,subsets sa,mu) ∧
          sigma_finite_measure_space (space sa,subsets sa,nu) ∧
          nu ≪ (space sa,subsets sa,mu) ∧ f ∈ RN_deriv_property sa nu mu ⇒
          AE x::(space sa,subsets sa,mu).
            f x = (nu / (space sa,subsets sa,mu)) x
   
   [RN_deriv_property_almost_unique]  Theorem      
      ⊢ ∀sa mu nu f g.
          measure_space (space sa,subsets sa,mu) ∧
          sigma_finite_measure_space (space sa,subsets sa,nu) ∧
          f ∈ RN_deriv_property sa nu mu ∧ g ∈ RN_deriv_property sa nu mu ⇒
          AE x::(space sa,subsets sa,mu). f x = g x
```

Note that I didn't change any statement of the existing, recently added theorems. (Minor formating aiming to truncate long lines, and optional ASCII conversions were done in the code changes.)

--Chun

### DIFF
--- a/src/probability/lebesgueScript.sml
+++ b/src/probability/lebesgueScript.sml
@@ -10114,21 +10114,21 @@ QED
 (* Helper simps for later results *)
 
 Theorem m_space_density[simp]:
-    ∀m f. m_space (density m f) = m_space m
+    !m f. m_space (density m f) = m_space m
 Proof
     simp[density_def]
 QED
 
 Theorem measurable_sets_density[simp]:
-    ∀m f. measurable_sets (density m f) = measurable_sets m
+    !m f. measurable_sets (density m f) = measurable_sets m
 Proof
     simp[density_def]
 QED
 
-Theorem sig_alg_density[simp]:
-    ∀m f. measurable_space (density m f) = measurable_space m
+Theorem measurable_space_density[simp] :
+    measurable_space (density m f) = measurable_space m
 Proof
-    simp[density_def]
+    simp [density_def]
 QED
 
 (* References:
--- a/src/probability/martingaleScript.sml
+++ b/src/probability/martingaleScript.sml
@@ -9433,12 +9433,6 @@ Proof
  >> MATCH_MP_TAC pos_fn_integral_density_reduce >> art []
 QED
 
-Theorem measurable_space_density[simp] :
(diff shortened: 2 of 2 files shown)


mn200:
Thanks!  I will wait for the CIs.  

In general, I'd prefer not to have formatting wars over ASCII *vs* Unicode (or `\\` *vs* `>>` *etc.*).  
Unless we:
- determine a canonical One True Way, and 
- that we should enforce it,

I think the original author’s choices can stand.

binghe:
Thanks for your suggestion.

PR #1874, opened 2026-03-17
Use exportSmallToFD for tracing

digama0:
Adaptation to use `exportSmallToFD` from https://github.com/digama0/polyml/compare/exportSmall , which reduces memory usage during the export.


mn200:
It would be nice to have tests of `--trknl` in the CI machinery.

PR #1737, opened 2025-11-28
Parameter-Dependent Integrals

binghe:
Hi,

This PR adds two major lemmas for "Parameter-Dependent Integrals" (Chapter 12 of [1]). The first one "continuty lemma" says that if a (parameterized) function `u(x,t)` is continuous (w.r.t. `t`), so is its integration (w.r.t. `x`). The second "differentiable lemma" says that its integration is differentiable if itself is, i.e. the derivative operation and integration is exchangeable:
```
   [continuity_lemma]  Theorem (martingaleTheory)
      ⊢ ∀s m u.
          measure_space m ∧ open s ∧
          (∀t. t ∈ s ⇒ integrable m (Normal ∘ u t)) ∧
          (∀x. x ∈ m_space m ⇒ (λt. u t x) continuous_on s) ∧
          (∃w. integrable m w ∧ (∀x. x ∈ m_space m ⇒ 0 ≤ w x ∧ w x ≠ +∞) ∧
               ∀t x. t ∈ s ∧ x ∈ m_space m ⇒ Normal (abs (u t x)) ≤ w x) ⇒
          (λt. real (∫ m (Normal ∘ u t))) continuous_on s

   [differentiable_lemma]  Theorem (martingaleTheory)
      ⊢ ∀s m u.
          measure_space m ∧ open s ∧ connected s ∧
          (∀t. t ∈ s ⇒ integrable m (Normal ∘ u t)) ∧
          (∀x. x ∈ m_space m ⇒ (λt. u t x) differentiable_on s) ∧
          (∃w. integrable m w ∧ (∀x. x ∈ m_space m ⇒ 0 ≤ w x ∧ w x ≠ +∞) ∧
               ∀t x.
                 t ∈ s ∧ x ∈ m_space m ⇒
                 Normal (abs (diff1 (λt. u t x) t)) ≤ w x) ⇒
          ∀t. t ∈ s ⇒
              integrable m (λx. Normal (diff1 (λt. u t x) t)) ∧
              ((λt. real (∫ m (Normal ∘ u t))) has_vector_derivative
               real (∫ m (λx. Normal (diff1 (λt. u t x) t))))
                (at t within s)
```
The above theorems are based on Lebesgue integration over a general measure space. Specializing it to `lborel` with the equivalence between Lebesgue and Guage integration used, we have the following two practical theorems (in `examples/probability/distributionTheory`):
```
[gauge_continuity_lemma] (distributionTheory)
[...]

PR #1875, opened 2026-03-18
Add differentiable_lemma_revisited, etc.

binghe:
Hi,

Recall in PR #1737, I added some lemmas about "Parameter-Dependent Integrals". For a two-parameter function, say `u(t,x)`, the goal is to pass the derivative operation on `t` into the integration on `x`. The conclusion of the following main results (`diff1 (λt. real (∫ m (Normal ∘ u t))) t = real (∫ m (λx. Normal (diff1 (λt. u t x) t)))`) quite illustrates the idea:
```
   [differentiable_lemma]  Theorem (martingaleTheory)
      ⊢ ∀s m u.
          measure_space m ∧ open s ∧ connected s ∧
          (∀t. t ∈ s ⇒ integrable m (Normal ∘ u t)) ∧
          (∀x. x ∈ m_space m ⇒ (λt. u t x) differentiable_on s) ∧
          (∃w. integrable m w ∧ (∀x. x ∈ m_space m ⇒ 0 ≤ w x ∧ w x ≠ +∞) ∧
               ∀t x.
                 t ∈ s ∧ x ∈ m_space m ⇒
                 Normal (abs (diff1 (λt. u t x) t)) ≤ w x) ⇒
          ∀t. t ∈ s ⇒
              integrable m (λx. Normal (diff1 (λt. u t x) t)) ∧
              ((λt. real (∫ m (Normal ∘ u t))) has_vector_derivative
               real (∫ m (λx. Normal (diff1 (λt. u t x) t))))
                (at t within s)
```
The above proof is not easy but is also not hard (because it can be done fully inside Lebesgue integration using some convergence theorems there). The textbook version of the above lemma was learnt from Chapter 12 of Shilling's book [1]:
Schermata 2026-03-18 alle 13 33 18

I was hoping to the use this lemma to as a key step in the equivalent definitions of "convergence in distribution" (part of the CLT project), but, unfortunately, in that case there's no way to satisfy the boundedness antecedents in the above lemma (the bound function doesn't exist). I was stuck there.

Occasionally, in the same book, but at the ending "Problems" part of another Chapter (14), I found the following "revisited" version of the same lemma with very different antecedents (in particular, a bound function is no more needed):
Schermata 2026-03-18 alle 13 31 52

The formal proof of this "revisited" differentiability lemma is the main contribution of this PR. Below is the formal statements:
```
[differentiable_lemma_revisited] (distributionTheory)
⊢ ∀m u.
    sigma_finite_measure_space m ∧ (∀t. integrable m (Normal ∘ u t)) ∧
    (∀x. x ∈ m_space m ⇒ (λt. u t x) differentiable_on 𝕌(:real)) ∧
    (∀x. x ∈ m_space m ⇒ integrable lborel (Normal ∘ diff1 (λt. u t x))) ∧
    (∀t. integrable m (λx. Normal (diff1 (λt. u t x) t))) ∧
    (λt. real (∫ m (λx. Normal (diff1 (λt. u t x) t)))) continuous_on 𝕌(:real) ∧
    (∀k. compact k ⇒
         ∫ lborel (λt. ∫ m (λx. Normal (abs (diff1 (λt. u t x) t))) * 𝟙 k t) ≠
         +∞) ∧
    (λ(x,t). Normal (diff1 (λt. u t x) t)) ∈
    Borel_measurable (measurable_space m × borel) ⇒
    ∀t. ((λt. real (∫ m (Normal ∘ u t))) has_vector_derivative
         real (∫ m (λx. Normal (diff1 (λt. u t x) t)))) (at t)
```
This is one of the hardest proofs I ever worked with. It's not very long, but the proof involves the Fundamental Theorem of Calculus (of Gauge integration), Fubini theorem (of Lebesgue integration) and the equivalence between the two integration system, a mixing style.  Due to the dependencies it can only live under `examples/probability`. (I spent two months for building the determination and one week to actually do it...)

Now it seems possible to fill the missing gaps in the CLT project.

--Chun

[1] Schilling, R.L.: Measures, Integrals and Martingales (2nd Edition).
      Cambridge University Press (2017).

### DIFF
--- a/examples/probability/central_limitScript.sml
+++ b/examples/probability/central_limitScript.sml
@@ -7,7 +7,7 @@ Ancestors
   pair combin option prim_rec arithmetic pred_set list rich_list real iterate
   seq transc real_sigma lim topology real_topology extreal sigma_algebra
   measure lebesgue_measure real_borel borel lebesgue martingale extreal_base
-  probability derivative distribution stochastic_process
+  probability derivative integration distribution stochastic_process
 Libs
   pred_setLib hurdUtils realLib Diff
 
@@ -1551,7 +1551,7 @@ Theorem has_integral_x_cubic_std_normal_density :
 Proof
   rpt STRIP_TAC
   >> ONCE_REWRITE_TAC [GSYM REAL_SUB_NEG2]
-  >> HO_MATCH_MP_TAC integrationTheory.FUNDAMENTAL_THEOREM_OF_CALCULUS
+  >> HO_MATCH_MP_TAC FUNDAMENTAL_THEOREM_OF_CALCULUS
   >> rw [IN_INTERVAL]
   >> MATCH_MP_TAC HAS_VECTOR_DERIVATIVE_AT_WITHIN
   >> REWRITE_TAC [has_vector_derivative_x_cubic_normal_density]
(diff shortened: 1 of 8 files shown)


mn200:
Very nice; thanks!

PR #1882, opened 2026-03-25
Add converge_in_dist_alt_C3 (assuming integrable_std_normal_quartic)

binghe:
Hi,

Recall that the "convergence (of random variable sequence `X(n)` to another random variable `Y`) in distribution", denoted (nicely) by `(X ⟶ Y) (in_distribution p)`, is defined in `probabilityTheory`:
```
   [converge_in_dist_def]  Theorem
      ⊢ ∀X Y p.
          (X ⟶ Y) (in_distribution p) ⇔
          ∀f. f ∈ C_b ext_euclidean ⇒
              ((λn. expectation p (Normal ∘ f ∘ X n)) ⟶
               expectation p (Normal ∘ f ∘ Y)) sequentially
```
But this definition is currently NOT used in the statement of our Central Limit Theorem (submitted in PR #1795), because an imporatant equivalent definition of the above `converge_in_dist` wasn't available. Now with this huge PR, the missing gap is (almost) done:
```
[converge_in_dist_alt_C3] (distributionTheory)
⊢ ∀p X Y N.
    prob_space p ∧ ext_normal_rv N p 0 1 ∧ integrable_std_normal_quartic ∧
    (∀n. real_random_variable (X n) p) ∧ real_random_variable Y p ⇒
    ((X ⟶ Y) (in_distribution p) ⇔
     ∀f. f ∈ CnR 3 ⇒
         ((λn. expectation p (Normal ∘ f ∘ real ∘ X n)) ⟶
          expectation p (Normal ∘ f ∘ real ∘ Y)) sequentially)
```
With this theorem, the new statement of CLT is updated to the following:
```
[central_limit_theorem] (central_limitTheory)
⊢ ∀p X N.
    prob_space p ∧ ext_normal_rv N p 0 1 ∧ integrable_std_normal_quartic ∧
    (∀i. real_random_variable (X i) p) ∧
    (∀n. indep_vars p X (λi. Borel) (count n)) ∧
    (∀i. expectation p (X i) = 0) ∧
    (∀i. expectation p (λx. (abs (X i x))³) < +∞) ∧
    (∀i. variance p (X i) < +∞) ∧ (∀i. variance p (X i) ≠ 0) ∧
    ((λn.
          absolute_third_moments p X (SUC n) /
          (sqrt (second_moments p X (SUC n)))³) ⟶ 0) sequentially ⇒
    CLT p X N
```
where the constant `CLT` is defined in this way (for reuse purposes of other CLT versions in the future):
```
[...]

PR #1885, opened 2026-03-27
Added integrable_std_normal_quartic

KaiPhan:
This PR extends [PR #1882](https://github.com/HOL-Theorem-Prover/HOL/pull/1882), completes the proof of integrable_std_normal_quartic, removes the remaining assumption in converge_in_dist_alt_C3, and includes minor changes to theorem names in the Central Limit Theorem.

### DIFF
--- a/src/probability/real_borelScript.sml
+++ b/src/probability/real_borelScript.sml
@@ -4316,17 +4316,6 @@ Proof
  >> REWRITE_TAC [CONTINUOUS_ON_EXP]
 QED
 
-Theorem in_measurable_borel_borel_aniv :
-  numeric_negate ∈ borel_measurable borel
-Proof
-  Know ‘$real_neg = \x. -1 * x’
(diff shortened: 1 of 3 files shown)


src/probability/real_borelScript.sml
```diff
@@ -4316,6 +4316,17 @@ Proof
  >> REWRITE_TAC [CONTINUOUS_ON_EXP]
 QED
 
+Theorem in_measurable_borel_borel_aniv :
```
binghe: The last part of this theorem name should be "ainv", meaning **a**rithmetic **inv**ersion, i.e. numeric negation (`-`). And previously I have already moved this theorem into `real_borelTheory`. Please remove it and use the one with correct name instead.

binghe:
The antecedent `integrable_std_normal_quartic /\ ..` should be also removed from the theorem `central_limit_theorem` (in `central_limitTheory`), otherwise it cannot build.

mn200:
Thanks!

PR #1888, opened 2026-03-30
Prove integrable_std_normal_density directly; add CLT_Lyapunov

binghe:
Hi,

This PR slightly improves some lemmas in `distributionTheory` including the lemma added in previous #1885.

Note that currently there's no way to calculate `∫ lborel (λx. Normal_density mu sig x) = 1` in HOL4, because there's no close-form antiderivative for `exp(-(x pow 2)/2)` and the standard calculation requires "change of variable" formula for (2-dimensional) Lebesgue integration, which is not yet formalised. Therefore the following lemma has to assume the existence of normal r.v., with a proof essentially ``P ==> P``:
```
[integral_normal_density]
⊢ ∀p X mu sig.
    prob_space p ∧ normal_rv X p mu sig ⇒
    integrable lborel (λx. Normal_density mu sig x) ∧
    ∫ lborel (λx. Normal_density mu sig x) = 1
```

But if the goal is just to show `integrable lborel (λx. Normal_density mu sig x)`, this is indeed possible without any extra assumption (newly added in this PR):
```
[integrable_std_normal_density] (distributionTheory)
⊢ integrable lborel (λx. Normal_density 0 1 x)
```
This is because, the involved function is bounded in the domain `[-1,1]` (abs x <= 1), while in other values its absolute value is smaller than `abs (x * std_normal_density)`, which has antiderivative and is integrable. Based on this result, many lemmas in `distributionTheory` now have less antecedents and the caller proof is slightly smaller.

I also added `CLT_Lyapunov` (based on `central_limit_theorem`) into `central_limitTheory` with less antecedents:
```
Theorem CLT_Lyapunov :
  !p X N. prob_space p /\ ext_normal_rv N p 0 1 /\
         (!n. real_random_variable (X n) p) /\
         (!n. indep_vars p X (\i. Borel) (count n)) /\
         (!n. integrable p (\x. (abs (X n x)) pow 3)) /\
         (!n. expectation p (X n) = 0) /\
         (!n. variance p (X n) <> 0) /\
         ((\n. absolute_third_moments p X (SUC n) /
               sqrt (second_moments p X (SUC n)) pow 3) --> 0) sequentially
         ==> CLT p X N
```
Note that `(!n. integrable p (\x. (abs (X n x)) pow 3))` implies the r.v.'s has finite variance and finite (absolute) third moments.  The name `CLT_Lyapunov` indicates it's just one version of CLT, and there may be other versions in the future. (I suggest using this version in the paper.)

--Chun

### DIFF
--- a/examples/probability/central_limitScript.sml
+++ b/examples/probability/central_limitScript.sml
@@ -6361,8 +6361,6 @@ Theorem central_limit_theorem :
       CLT p X N
 Proof
     RW_TAC std_ss [CLT_def]
- >> MP_TAC (Q.SPECL [‘p’, ‘real o N’] integrable_std_normal_quartic)
- >> DISCH_TAC >> fs []
  >> Q.ABBREV_TAC ‘s = λn. sqrt (second_moments p X n)’ >> fs []
  >> Q.ABBREV_TAC ‘b = λn. absolute_third_moments p X n’ >> fs []
  >> Q.ABBREV_TAC ‘R = λn x. ∑ (λi. X i x) (count (SUC n)) / s (SUC n)’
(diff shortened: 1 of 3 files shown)


mn200:
Thanks!

PR #1890, opened 2026-03-30
Add infrastructure for building from replayed theorems

xrchz:
This adds the possibility of calling `build` (or `Holmake` or `hol`) with the option `--thmsrc=tr` (the default being `--thmsrc=dat`) to specify that when loading a theory instead of reading the `Theory.dat` file, the system should expect to (also) find a `Theory.tr.gz` file there and should read that (and process it with `ProofTraceReplay`) instead, to get theorems with no `DISK_THM` oracles proved by the kernel at load time.

(The intention is to in future also support `--thmsrc=pft`, using the PFT replay module being developed on the `replay` branch.)


tools-poly/hol.ML
```diff
@@ -435,6 +439,21 @@ fun main() = let
   val _ = diag ("holpathdb extensions: "^
                 String.concatWith ", " (map print_ext exts))
   val _ = List.app holpathdb.extend_db exts
+  (* Initialize theorem source if --thmsrc was specified (via flag or env var) *)
```
mn200, line 442: I'm super un-keen on environment variables.  Do you think this is a route that users will want to use?

tools/Holmake/poly/BuildCommand.sml
```diff
@@ -474,9 +475,12 @@ fun make_build_command (buildinfo : HM_Cline.t buildinfo_t) = let
 
 
   fun system s =
-    Systeml.system_ps
-      (if relocbuild then Systeml.build_after_reloc_envvar ^ "=1 " ^ s
-       else s)
+    let val pfx = (if relocbuild then Systeml.build_after_reloc_envvar ^ "=1 "
```
mn200, line 478: OK: is this where `HOL_THMSRC` is needed?

src/postkernel/prooftrace/ThmSrc_tr.sml
```diff
@@ -0,0 +1,114 @@
...
+         gen_prim_specification, etc.) as side effects of replaying
+         the definition proofs.
+         We chDir to the directory containing the .dat file so that
+         ProofTraceReplay can find <thyname>Theory.tr.gz there.
+         The path may be a HOL-level path (e.g., src/bool/boolTheory.dat)
+         that maps to .hol/objs/ via HFS_NameMunge. *)
+      val fspath = case HFS_NameMunge.HOLtoFS path of
+                       SOME {dir, ...} => dir
+                     | NONE => OS.Path.dir path
+      val olddir = OS.FileSys.getDir ()
```
mn200: Having to change directory before calling the function seems bad.

tools-poly/hol.ML
xrchz: This is just for internal plumbing imo -- users should use the flag. But I don't know a way to implement the flag propagation properly that doesn't use an environment variable. do you?

src/postkernel/prooftrace/ThmSrc_tr.sml
xrchz: How big of a deal is that? Like ProofTraceReplay could be enhanced to take a path instead of just a theory name, and that's probably the "right" fix. But I thought it wasn't necessary in this PR. Can change if you prefer.

mn200: Having the "right" API (one that takes a path) from the start seems like it would be a good idea.

tools-poly/hol.ML
mn200: Got it.

mn200:
Can you get it to try again now that the top of `develop` is actually building?

src/postkernel/prooftrace/ThmSrc_tr.sml
xrchz: OK I have improved the ProofTraceReplay API

xrchz:
> Can you get it to try again now that the top of `develop` is actually building?

Yes. So... I have tried but I don't know why mosml isn't getting `Subst.uo`. Can you help?

dnezam:
My local mosml build fails with this message instead:
```
Holmake: Compiling ProofTraceParser.sig
Holmake: Compiling ProofTraceParser.sml
File "ProofTraceParser.sml", line 16, characters 14-25:
! type 'a ptr = Word64.word
!               ^^^^^^^^^^^
! Cannot find file Word64.ui
```
Searching for "Word64" on the GitHub repo of mosml doesn't result in any hits, so maybe it's that (?)

xrchz:
@mn200 would it make sense to somehow exclude this option from mosml? (do you know how to do that?)

dnezam:
`Systeml.sml` does provide `haveWord64`, but I am not sure what an elegant way to handle this is. I am guessing that one needs to put the `Word64` parts at least into a different file and then, depending on `haveWord64` include (or exclude) that file from the build process.
I could find one such example in [hmcore.ML#L50](https://github.com/HOL-Theorem-Prover/HOL/blob/94828f2b2a2de9eeb83c71d5455c4b184ebc7492/tools/Holmake/hmcore.ML#L50).

By the way, it appears that it is not just MoscowML, but also PolyML on 32-bit systems that are lacking the `Word64` module (going by [tools-poly/configure.sml#L264](https://github.com/HOL-Theorem-Prover/HOL/blob/94828f2b2a2de9eeb83c71d5455c4b184ebc7492/tools-poly/configure.sml#L264)). Interestingly, it appears that SML/NJ ships `Word64` on all platforms. What is target dependent is Int{31,63} and Word{31,63} (see [smlnj/legacy/base/system/Basis/basis.cm](https://github.com/smlnj/legacy/blob/39914627460733493dee3e034821d3417c9af2f7/base/system/Basis/basis.cm#L150)).

mn200:
Perhaps you can do different things at the `Holmakefile` level, as is done in `src/portableML/poly/Holmakefile` ?

dnezam:
I saw the `src/portableML/poly/Holmakefile`, but I didn't want to go through the trouble of spelling out everything. Do you think that would be better?

EDIT: Actually, I'm not sure what that would look like. `src/portableML/poly/Holmakefile` seems to switch implementations, whereas with my change it won't get build at all (which was my understanding of the request)

PR #1894, opened 2026-04-03
holide: switch from HolParserOld to new HOLSource parser

xrchz:
Replace the old HolParserOld-based implementation of HOL_IDE with one based on the new HOLSource* parser infrastructure (HOLSourceParser, HOLSourceExpand, HOLSourcePrinter, HOLSourceAST). This brings tools-poly/holide.ML in line with the hol4-vscode extension's holide.sml.

Key changes:

- The initialize function now uses a per-declaration compile loop: HOLSourceParser parses each declaration, HOLSourceExpand expands HOL syntax to SML, HOLSourcePrinter prints it, and PolyML.compiler compiles it. Source positions are tracked via a span stack to provide accurate byte offsets back to the original source.

- Add annotation infrastructure for correlating PolyML parse trees with the HOL AST: the built datatype, props datatype, and annotateExp/ annotateDec/annotateSigexp/annotateStrexp functions. These are exported in the signature but not yet wired into the initialize callbacks (for future use by the LSP server).

- Add prelude/postPrelude for IDE-specific setup (oracle-based fast proving). prelude() uses PolyML.compiler to defer resolution of HOL kernel symbols to runtime.

- Add compile ref and noCompile for controlling whether PolyML actually compiles (vs just parsing), with a commented-out autoconf hack for CPNoCompile which is not available in Poly/ML 5.9.2.

- The holParseTree callback type changes from HolParserOld.Simple.decl to HOLSourceAST.dec. This is transparent to lsp-server.ML which passes (fn _ => ()) for this callback.

The signature is backward-compatible with lsp-server.ML, which continues to work without changes.

### DIFF
--- a/tools-poly/holide.ML
+++ b/tools-poly/holide.ML
@@ -154,7 +154,9 @@ fun prettyToString pp = let
 
 fun trProps [] = []
   | trProps (PolyML.PTprint p :: ls) = PPrint (prettyToString (p 5)) :: trProps ls
-  | trProps (PolyML.PTdeclaredAt p :: ls) = PDeclaredAt p :: trProps ls
+  | trProps (PolyML.PTdeclaredAt {file, startLine, endLine, startPosition, endPosition} :: ls) =
+    PDeclaredAt {file = file, startLine = FixedInt.toInt startLine, endLine = FixedInt.toInt endLine,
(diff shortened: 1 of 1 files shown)


mn200:
Happy to merge if @digama0 is.

digama0 (approved):
Both the old and new code are unused, so this is low risk and we can do more fixes later if needed.

mn200:
Keep chugging along!

PR #1895, opened 2026-04-04
Fix non-mlton build

dnezam:
Ideally we first merge a PR that adds CI for checking whether Holmake builds without mlton.


mn200:
This has nothing to do with mlton; it's bcb3357.

mn200:
Moreover, `heapname.ML` is now dead-code and doesn't need fixing (rather it should be removed).

tools-poly/heapname.ML
mn200, line 1: Don't edit this file.

mn200 (changes requested):

tools-poly/heapname.ML
dnezam: Deleted it as mentioned in the previous comment.

mn200:
Thanks!

PR #1896, opened 2026-04-04
CI: Check that pre-Holmake works without mlton

dnezam:
At the moment it looks like nothing is explicitly checking that building the pre-Holmake tools without mlton works.
This change aims to fix this by first running smart-configure.sml with MLTON explicitly set to NONE first, followed by cleaning up and then building as usual.


dnezam:
Fails as expected. The fix is (probably) in https://github.com/HOL-Theorem-Prover/HOL/pull/1895

mn200:
Can you rerun and check for it passing now please?

mn200:
Thanks!

PR #1900, opened 2026-04-11
Improved proofs of permutator_thm, selector_thm, etc.

binghe:
Hi,

This PR reworked several proofs in `examples/lambda/barendregt`. These are early proofs which can be proved easiler by later more general lemmas (some of them were left as TODOs), but no statements are changed.

--Chun


mn200:
Nice improvements!

Issue #1899, opened 2026-04-09
cv_auto_trans: from_to proof scales badly with large datatypes


xrchz:
## Summary

`cv_auto_trans` takes over 12 minutes on a trivial identity function when the argument type is a recursive datatype with ~50 constructors, some of which have many arguments. This should be near-instant.

## Minimal reproducer

The following self-contained 80-line script takes **~12 minutes** to build, where essentially all the time is spent generating the `from_to_kont_thm` during `cv_auto_trans idk_def`:

```sml
Theory cvMin
Ancestors
  arithmetic list string cv cv_std
Libs
  cv_transLib

Datatype:
  leaf = LA | LB num | LC string | LD (num list)
End

Datatype:
  tree = Leaf num | Node tree tree | Big (tree list)
End

Datatype:
  kont
  = K0 kont
  | K1 num kont
  | K2 string kont
  | K3 (num option # string) kont
  | K4 tree kont
  | K5 (tree list) kont
  | K6 leaf kont
  | K7 num num tree kont
  | K8 leaf (num list) kont
  | K9 string num kont
  | K10 tree num kont
  | K11 (num list) (num list) kont
  | K12 num (num list) (num list) kont
  | K13 kont
  | K14 string num num (num list) kont
  | K15 num num (num list) (num list) kont
  | K16 kont
  | K17 (num list) kont
  | K18 num kont
  | K19 tree kont
  | K20 num kont
  | K21 (leaf list) kont
  | K22 num kont
  | K23 kont
  | K24 kont
  | K25 (tree list) kont
  | K26 num kont
  | K27 string kont
  | K28 tree kont
  | K29 (leaf # num list) kont
  | K30 tree tree kont
  | K31 (string list) kont
  | K32 num tree kont
  | K33 num num kont
  | K34 string kont
  | K35 num num kont
  | K36 kont
  | K37 num num kont
  | K38 kont
  | K39 bool string (num list) num (tree option) kont
  | K40 num num kont
  | K41 kont
  | K42 kont
  | K43 kont
  | K44 num num bool kont
  | Wide1 num (num option # string) ((string # num) list) (tree list) num (num list) bool num kont
  | Wide2 num (num option # string) ((string # num) list) (num list) num (num list) bool num kont
  | Wide3 (num list) num bool bool kont
  | K48 (tree list) kont
  | K49 num kont
  | Done
End

Definition idk_def:
  idk (x: kont) = x
End

[...]

Issue #1889, opened 2026-03-30
Disk thms don't remember their axioms


xrchz:
It's not entirely clear to me what `#2(Tag.dest_tag(Thm.tag th))` is supposed to say about `th`, but if it's supposed to say something like "the axioms that this theorem's proof depended on", then exporting to `dat` and reloading with `DISK_THM` oracle destroys this information.

I'm opening this issue to get clarity on whether it's a bug that the axioms aren't retained, and to record them for disk thms if so.

xrchz:
It's possible this is deliberate if the only point of storing the axioms in the tag is to ensure they are uptodate on export.

mn200:
It's certainly deliberate, though I'm not now 100% about the motivation.  Certainly, for the bulk of HOL's theorems, I think we'd be needlessly storing all 5 axioms as part of their tag.

konrad-slind:
For most theorems, the tracked axioms would only be interesting when there
are additional axioms beyond the ones given for HOL.
In that case, the "5 axioms" could be taken as "always there so not
tracked". But then the extras should be tracked.

There are also the logical puzzles like "Can X be proved without Choice" in
which case tracking the base axioms would be interesting.

PR #1901, opened 2026-04-11
Improve speed of some cv auto proofs

xrchz:
Work on #1899


xrchz:
This reduces the runtime of the example that motivated this (in #1889) from ~12 minutes to ~9 seconds.

myreen (approved):

Issue #1531, opened 2025-06-14
Add download-from-cache for theory build outputs to Holmake


xrchz:
Add a function to Holmake that serves as an alternative to the compile/build `fooScript.sml` approach to producing theory build outputs and instead downloads `fooTheory{.sig,.sml,.dat}` from a given URL. The other outputs (.ui, .uo, logs?) are either not to be included or generated as normal by Holmake if they are necessary for its operation.

PR #1911, opened 2026-04-19
Add support for Holmake building from a cache

mrichards30:
This feature addresses https://github.com/HOL-Theorem-Prover/HOL/issues/1531 by adding support for using pre-built theory files for Holmake.

This adds `Holmake --write-cache <location>` and `Holmake --use-cache <location>` to create and use the cache respectively.
image

- `Holmake --use-cache <location>` runs a regular Holmake build but causes Holmake to check the cache before any theory build
- `Holmake --write-cache <location>` runs recursively on all theories in the current folder or subfolders
- `Holmake --write-cache <location> fooTheory` can write a specific theory to the cache
- The cachekey feature has also been refactored a little bit. Previously its logic was tied into the `--cachekey` CLI, but I needed a string-producing version. Its tests still pass. 
- These features are only added for PolyML and just warn the user if they try to use them with mosml.


charles-cooper:
maybe there should be a default cache. like `--use-cache` and `--write-cache` are always on

mn200 (changes requested):
Looks pretty good, but if you're going to pass an output function around (you've called it `talk` and pass it in via `try_cache`), please instead use the `output_functions` type (see `core/Holmake_tools`).  One standard value of this type is generated as the `Holmake` process starts up.

mrichards30:
> Looks pretty good, but if you're going to pass an output function around (you've called it `talk` and pass it in via `try_cache`), please instead use the `output_functions` type (see `core/Holmake_tools`). One standard value of this type is generated as the `Holmake` process starts up.

done 🙂

mrichards30:
@charles-cooper i definitely agree. is it ok if i look into it for a follow up pr after this one gets merged? dont have heaps of time atm

mn200 (changes requested):
Looks neat; please add test-cases for behaviours under `tools/Holmake/tests` and then ensure they're exercised as part of a `bin/build`.  If the default becomes use of a cache (not unreasonable), the tests can be for behaviours where the cache is bogus/unavailable/etc.

mrichards30:
ok i've added tests and integrated with the new work from #1913
lmk if there's anything else

mn200:
Thanks!

Issue #1909, opened 2026-04-17
Resume fails with `--fast`


myreen:
I've noticed that `Holmake --fast` fails in case `suspend`/`Resume` is used. This is quite inconvenient.

Here's a simple reproduction of the error. Have a file `fooScript.sml` with the content:
```
Theory foo
Ancestors
  list

Theorem lemma:
  ∀n. n < n + 1:num
Proof
  Cases
  >- fs []
  >- suspend "bar"
QED

Resume lemma[bar]:
  simp []
QED

Finalise lemma
```

With `Holmake fooTheory` this builds fine.

With `Holmake --fast fooTheory` this file fails to build. The error is "No suspension with name `lemma`".

xrchz:
The proposal in #1883  would likely solve this too

xrchz:
Solved as part of #1912

Issue #1883, opened 2026-03-25
Feature request: cross-file suspend/resume


charles-cooper:
_co-authored by claude opus 4.6_

## Motivation

The current suspend/resume mechanism is a great tool for structuring proofs within a single script file. However, it's limited to intra-file use: the suspension DB is an in-memory `Sref`, the `ExportTheory` hook enforces that all suspensions are finalised before the script ends, and nothing is persisted to the theory file on disk.

There are cases where it would be useful to suspend a subgoal in one theory and resume it in a downstream theory — for example:

- **Layered developments**: A foundational theory establishes the main structure of a proof but a subgoal requires machinery (definitions, lemmas) that naturally belongs in a later theory. Currently, you must either anticipate and manually state the subgoal as a standalone lemma, or restructure the theory hierarchy.

- **Large proof efforts**: Splitting a proof across files can help manage complexity. The suspend/resume pattern already captures the exact goal context (assumptions, free variable bindings), which is more convenient than manually reconstructing the subgoal statement as an independent lemma.

## Current behaviour

- `suspend` stores the incomplete theorem in `boolLib.suspensionDBref` (in-memory `Symtab`)
- `ExportTheory` hook raises an error if any suspensions remain unfinalised
- `Finalise` must appear in the same script file

## Possible approach

One way to support this:

1. Allow suspended theorems to be serialised into the theory file (e.g., as a new kind of stored theorem with `suspendlabel` hypotheses intact)
2. Provide a way to import/access them in downstream theories (e.g., `load_suspension "TheoryName" "thm_name"`)
3. `Finalise` in the downstream theory would discharge the suspension hypotheses and save the completed theorem (possibly back-patching the original theory, or storing it in the finalising theory)
4. The `ExportTheory` check could be relaxed to allow explicitly marked "deferred" suspensions

## Open questions

- Should the finalised theorem live in the original theory or the one that finalises it?
- How to handle theory rebuilds — if the upstream theory changes, suspended goals may no longer match
- Should there be a mechanism to declare *which* suspensions are intentionally deferred vs accidentally left open?

mn200:
I agree this is an appealing feature.  Another issue is merging resumptions that are handled in distinct child (or grandchildren, ...) theories.  If the exported suspension is
```
  [S1, S2] |- Phi
```
[...]

Issue #1881, opened 2026-03-24
Allow Resume to be rerun without error


xrchz:
If you redo the proof of a suspension, things don't work. You should be able to rerun the proof of a suspension even if you ran it already, without having to rerun the original suspension producer.

mn200:
I think this will preclude reuse of suspension names.  Currently, it's possible to write
```
Theorem foo: Phi 
Proof
  ... >> suspend "blah"
QED
```
and to then call `suspend "blah"` inside the resumption of `blah`.  If you want the desired behaviour, the system will have to forbid this.  

I don't at all think this is a problem btw.

xrchz:
Yeah I'm happy to make that trade (rerun allowed in exchange for label reuse disallowed)

xrchz:
I think this is solved by #1912

PR #1912, opened 2026-04-19
Functional suspend/Resume

xrchz:
Should solve #1883, #1881, #1909


src/1/boolLib.sml
```diff
@@ -263,43 +223,36 @@ in
...
+                       "; apply them at Finalise time instead")
         else ();
-        update_suspensionDB true (n,th);
-        th
+        (* Save the suspended theorem as an ordinary exported theorem so
+           that it persists in the theory file and can be resumed /
+           finalised in the current script or in downstream theories.
+           Rebinding is allowed so that re-running the Theorem block in
+           an interactive session is not disruptive. *)
+        trace("Theory.allow_rebinds", 1)
```
mn200: Rebinds are always allowed interactively.  The `trace` call is unnecessary.

mn200 (changes requested):
I don't like the way the state is encoded in names within `DB`.  I'd prefer to see both exported unfinalised theorems and their resumption results stored in an ancestry-data structure (a dictionary) so that they can be readily found without scanning all ancestors.  That way you also avoid the horrible code that avoids having to give them SML compatible names. When a normally suspended theorem has been fast-proved it *will* be in the normal DB and the code for resuming them can lookup that precise name as a fallback.

src/1/boolLib.sml
xrchz: removed

xrchz:
Agreed: here's a new version using `AncestryData` instead of `DB`.

src/marker/markerLib.sml
```diff
@@ -947,48 +947,324 @@ fun prim_resume (th,label,tac) =
...
+
+      suspension.theorems
+        name (string)  ->  the suspended parent theorem
+          (its conclusion is the original statement, and each still-open
+           subgoal is recorded as a suspendlabel hypothesis)
+
+      suspension.resumptions
+        (parent_thy, parent_name, label)  ->  list of proved
+          |- suspendlabel "lab" G  theorems, one per matching
+          hypothesis of the parent (ordered to match the hyp order).
```
mn200: This comment is dodgy; the “hyp order” is not a well-defined notion: hypotheses are stored as a set, so when rendered as a list, you’ll get them out as sorted by `Term.compare`.

src/marker/markerLib.sml
```diff
@@ -947,48 +947,324 @@ fun prim_resume (th,label,tac) =
...
+  open ThyDataSexp
+in
+
+(* --- suspension.theorems store --- *)
+
+(* Key: (thy, name).  Theory qualification allows distinct theorems
+   with the same name in different theories. *)
+type susp_key = string * string
+
+fun susp_key_compare ((t1,n1), (t2,n2)) =
```
mn200: Should use `Portable.pair_compare`

src/marker/markerLib.sml
```diff
@@ -947,48 +947,324 @@ fun prim_resume (th,label,tac) =
...
+     update_susp_global (apply_susp_delta d))
+
+(* --- suspension.resumptions store --- *)
+
+(* Key: (parent_thy, parent_name, label).  Theory qualification lets
+   sibling theories resume unrelated parents that happen to share a
+   name. *)
+type res_key = string * string * string
+
+fun res_key_compare ((t1,n1,l1), (t2,n2,l2)) =
```
mn200: Should use `Portable.pair_compare`

src/marker/markerLib.sml
```diff
@@ -947,48 +947,324 @@ fun prim_resume (th,label,tac) =
...
+          [] => NONE
+        | [(thy, th)] => SOME (thy, th)
+        | _ =>
+            (* Multiple matches - ambiguous. For now, just pick the first.
+               A future enhancement could raise an error and require
+               qualified syntax. *)
+            SOME (hd matches)
+    end
+
+(* Main entry point: supports both "foo" and "thyname$foo" forms *)
```
mn200: We tend to use `.` as the separator when referring to theorems.  `$` is used for constant-naming within the logic, but `.` for theorems. For example, theorem references in the LaTeX mungers.

src/marker/markerLib.sml
```diff
@@ -947,48 +947,324 @@ fun prim_resume (th,label,tac) =
...
+val _ = boolLib.suspended_theorem_recorder :=
+    (fn (n, th) => record_suspension_delta
+                     (AddSuspended ((Theory.current_theory(), n), th)))
+
+(* ----------------------------------------------------------------------
+    Finding the parent theorem for a Resume / Finalise.
+
+    Strategy:
+     1. Look up in the suspension.theorems store (current theory merged
+        with ancestry) \u2014 this is the normal case.
```
mn200: What is this \u2014 thing?

src/marker/markerLib.sml
```diff
@@ -947,48 +947,324 @@ fun prim_resume (th,label,tac) =
...
+     2. Fall back to DB.fetch: when Holmake --fast is in effect, the
+        tactic body of Theorem...suspend...QED is not run, so the
+        theorem is saved to DB normally (without any suspendlabel
+        hypothesis) and does not appear in the suspension store.
+        Resume and Finalise both detect this by finding a parent in
+        DB whose hyp set has no relevant suspendlabel \u2014 see callers.
```
mn200: And again

src/marker/markerLib.sml
```diff
@@ -947,48 +947,324 @@ fun prim_resume (th,label,tac) =
...
+      suspension.resumptions
+        (parent_thy, parent_name, label)  ->  list of proved
+          |- suspendlabel "lab" G  theorems, one per matching
+          hypothesis of the parent (ordered to match the hyp order).
+
+    Finalise consults both stores (merged across the current theory
+    and its ancestors) to assemble the clean theorem.
+   ---------------------------------------------------------------------- *)
+
+local
```
mn200: `local` is annoying when debugging interactively; I can't imagine that `ThyDataSexp`'s bindings will cause problems elsewhere; please remove.

mn200 (changes requested):

src/marker/markerLib.sml
dnezam: Turns out that that's — (em dash).

src/marker/markerLib.sml
xrchz: removed, thanks

src/marker/markerLib.sml
xrchz: yes, thanks. done.

src/marker/markerLib.sml
xrchz: good idea, thanks. done.

src/marker/markerLib.sml
xrchz: fixed (using "--").

src/marker/markerLib.sml
xrchz: done, thanks.

mn200:
Thanks for this; please go ahead the close the connected GitHub issues.

Issue #381, opened 2016-11-30
Holmake to rebuild based on theory hashes rather than timestamps


xrchz:
This would for example avoid cascading rebuilds if you just add a comment to a script file (and it produces exactly the same theory).

xrchz:
In my vision for this, it would also obviate a rebuild in the following situation. `barTheory` depends on `fooTheory`. You add a theorem to `fooTheory` which is not used in `barTheory`. Then `barTheory` does not need to be rebuilt even after `fooTheory` is rebuilt to include the new theorem.

mn200:
This seems unlikely.  Presumably foo's hash has just changed.  In that situation, bar has to rebuild.  However, *its* hash won't change (I suppose).  If baz only depends on bar, it won't need to be rebuilt. On the other hand, suppose the change to foo has redefined an important constant that bar uses, but the theorems that bar states have the same statements so that when bar is rebuilt it doesn't change its external nature.  (Say the tactics proving bar's theorems are robust in the face of whatever change was made.)

Now theory baz is only explicitly a descendent of bar (so the graph records a link form bar to baz, but not from foo to baz), but it does have a theorem in it that depends on foo.  Maybe that theorem is now false... So, we actually have to rebuild everything in baz too (though there'd be lots of similar seeming situations when we'd prefer not to).

marioxcc:
If this is implemented, it would be better to _move_ the old theory and then compare the new and the old file byte for byte, file instead of overwriting it with the new theory file and then comparing hashes. Although probably theory hashes will never collide in practice, this is not provable, and thus if one were to model the build process, including this optimization, one could not prove that it is correct, not even probabilistically (in cryptography it is common to assume that the hash is a function chosen at random; this approach is used under the name of “pseudorandom function model” but note that this assumption does not match reality).

xrchz:
I started writing this issue again :woman_facepalming: ...

This issue is a proposal to replace the timestamps (based on modification time of the script file that built the theory) currently used to identify theories (these are recorded in the `Theory.dat` files) with (recursive) content hashes. The content to hash is the `Theory.dat` file content (or the logically relevant subset of it in case anything can be omitted).

[...]

PR #1913, opened 2026-04-19
Avoid rebuilds based on mtime, using cachekey instead (#381)

xrchz:
When Holmake reruns a script file that happens to produce a Theory.dat whose content is identical to the existing one (for example because the edit to fooScript.sml was comment-only, or was made and then reverted), we would previously still overwrite fooTheory.dat, fooTheory.sml and fooTheory.sig, advancing their mtimes and causing Holmake's timestamp-based dep-check to cascade rebuilds to every descendant theory.

Change export_theory so that each of the three generated files is first written to a sibling tempfile (<path>.new), and only renamed over the destination when content actually differs (SHA1 compared). If the content matches, the tempfile is removed and the existing destination file (and its mtime) is left alone, so descendants are not invalidated.

Key points:

* We work in real filesystem paths here rather than HOL paths, because HFS_NameMunge.HOLtoFS routes generated theory files into .hol/objs/ based on their extension, and a ".new" suffix is not recognised.  The tempfile therefore has to be constructed as <real-dst>.new, and we explicitly ensure the munged output directory exists before opening the tempfile.

* The SHA1 of the new .dat, which is already computed for embedding into fooTheory.sml, is reused when deciding whether to replace the old .dat (no extra hash pass).  For .sig/.sml the helper hashes on demand.

* The rename-or-discard step uses OS.FileSys.rename / remove on real paths directly, bypassing HOLFileSys's munging (which would re-munge the .new-suffixed tempfile path incorrectly).

* No Holmake change: Holmake continues to use mtime-based rebuild decisions.  The win is purely from export_theory no longer bumping mtimes unnecessarily.

Adds a selftest (tools/Holmake/tests/idempotent_export) that verifies:

  - a comment-only edit to baseScript.sml does not advance baseTheory.dat's mtime, and therefore does not rebuild childTheory;
  - a real content change (added theorem) to baseScript.sml does advance both baseTheory.dat and childTheory.dat mtimes;
  - a mere `touch` of baseScript.sml (reruns the script but produces byte-identical output) does not advance baseTheory.dat's mtime.

This is a first, Holmake-free step towards #381; later steps can teach Holmake itself to consult content hashes rather than timestamps, but this change on its own is enough to eliminate the most common source of unnecessary cascading rebuilds.


xrchz:
**Updated PR description** (the original description is now outdated)

This PR addresses issue #381 by teaching Holmake to skip rerunning theory scripts when the recursive content hash of their inputs hasn't changed since the last successful build.

### Background

The original approach (first commit in this branch) attempted to make `export_theory` idempotent by comparing SHA1 hashes before overwriting generated files. However, this doesn't work because Holmake's `BuildCommand.sml` preemptively deletes the expected output files (`.dat`, `.sml`, `.sig`) before re-running the script, so `export_theory` never sees an existing destination to compare against. That approach was therefore reverted.

### Current Approach

Instead of working around the deletion in `export_theory`, the rebuild-avoidance logic is moved up to Holmake itself using cachekey-based stamps:

1. **New `HM_Cachekey` module** (`tools/Holmake/core/`): Exposes a `compute_for_node` / `compute_for_deps` API that computes the recursive content hash (cachekey) for a theory target's dependencies. Also provides stamp-file helpers (`stamp_path_for_datfile`, `read_stamp`, `write_stamp`, `remove_stamp`).

2. **New `--rebuild=<strategy>` option**: Accepts `cachekey` (new default) or `mtime`. Under `--rebuild=cachekey`, Holmake checks whether:
   - the target file exists,
   - all dependencies are built, and
   - the stamp file records a cachekey equal to the current one.
   
   If all conditions hold, the script is skipped and the node is marked `Succeeded` immediately. Otherwise, the existing mtime-based `needs_building` logic applies.

3. **Stamp file workflow**: On successful script completion, `BuildCommand` writes the cachekey to `<thy>Theory.cachekey` (next to the `.dat` file). The stamp is removed at the start of every script run, so a failed or interrupted build cannot leave a stale claim of being up-to-date.

4. **`--rebuild=mtime` escape hatch**: Restores the traditional timestamp-only behaviour for users who prefer it.

### What This Means in Practice

- **Touch without content change**: `touch fooScript.sml` followed by `Holmake` does *not* rebuild `fooTheory.dat` (the cachekey is unchanged).
- **Edit-and-revert**: Editing a script and then reverting the edit (without an intervening build) does *not* rebuild the theory.
- **Real content change**: Adding a theorem to `fooScript.sml` *does* rebuild `fooTheory.dat` and cascades to dependent theories.
- **Missing target**: Deleting `fooTheory.dat` forces a rebuild even if the stamp is valid (safety check).

mn200:
How does this relate to #1911 ?

xrchz:
> How does this relate to #1911 ?

I think they would both just work complementarily. I'm not sure. @mrichards30 ?

mrichards30:
> > How does this relate to #1911 ?
> 
> I think they would both just work complementarily. I'm not sure. @mrichards30 ?

looks like this branch works well with #1911 after pulling these changes in, ~but my holmake is running noticably slower? idk if its just my laptop going through something today though. have you noticed anything @xrchz~ ignore that, i think my laptop is just dying

ok my laptop is back to normal. retested and it looks great- i can use the cache and it doesnt trigger a bunch of rebuilds!

mn200:
Full steam ahead!

PR #1915, opened 2026-04-21
Add a patch to asl-equiv download script

tanyongkiam:
Prevents it from generating sig docs which, in turn, cause an Overflow in the parser (DArray / DString) when it tries to load the entire comment at once.


mn200:
Thanks!

PR #1918, opened 2026-04-22
perf: avoid unnecessary uptodate checks

digama0:
This came up while benchmarking some cakeml changes that will generate a lot of simple definitions. When there are a lot of registered constants, every definition becomes more expensive due to these outofdate checks, which are not needed most of the time. An easy way to detect whether it is needed is to use an epoch counter which is incremented every time you retire an id. If the counter hasn't changed since last time you looked, then nothing new could have been retired and therefore the outofdate check is redundant. This yields a 5% improvement in HOL src compile time on my machine, and about an 18% improvement on the definition-heavy code I was testing.

Also drive-by fixed a few warnings in touched files.

### DIFF
--- a/src/parse/AncestryData.sml
+++ b/src/parse/AncestryData.sml
@@ -36,7 +36,7 @@ type ('d,'v) info =
       apply_delta : 'd -> 'v -> 'v, tag : string,
       parents : {thyname : string} -> string list}
 
-fun fpluck f [] = NONE
+fun fpluck _ [] = NONE
   | fpluck f (h::t) = case f h of
                           NONE => Option.map (fn (v,r) => (v,h::r)) (fpluck f t)
                         | SOME v => SOME ((h,v),t)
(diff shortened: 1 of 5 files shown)


src/parse/AncestryData.sml
```diff
@@ -191,7 +191,21 @@ fun make {info : ('delta,'value)adata_info, get_deltas, delta_side_effects} =
...
+fun gen_other_tds {tag,dec,enc,P} = let
+  val last_scan_epoch = ref ~1
+  (* Memoize only retire-sensitive events.  The scan's predicate P
+     (via ThmSetData.export_with_ancestry) composes retire checks with
+     DB.lookup, so it can change on *any* event that mutates the binding
+     table — NewBinding, UpdBinding, DelBinding.
+     When retire_epoch is unchanged and the event is one that only fires
+     after a retire (NewConstant / NewTypeOp / DelConstant / DelTypeOp),
+     we know no entry's P result can have changed since the last scan. *)
+  fun memoable (TheoryDelta.NewConstant _) = true
```
mn200, line 203: This is the same as the `retire_memoable` from before; can we share these in a common ancestor?

src/prekernel/KernelSig.sml
```diff
@@ -14,13 +14,23 @@ struct
...
 
-  fun name_of_id r = case !r of (n,utd) => n
-  fun uptodate_id r = case !r of (n,utd) => utd
+  (* Monotone counter bumped on every retire_id.  Lets downstream listeners
+     (ThmSetData, AncestryData, GrammarDeltas) skip their O(n) "scan all
+     stored deltas for staleness" work when no constant/type has been
+     retired since the last scan.  A constant's uptodate flag only flips
+     via retire_id, so the counter is a sound summary of "has any stored
+     delta potentially become stale?". *)
+  val retire_counter = ref 0
```
mn200, line 23: Please use the `Sref` API for this; in a distant future where things become more concurrent; it feels like this reference really should be locked when updated (as happens with `Sref.update`)

mn200 (changes requested):

Issue #1922, opened 2026-04-24
Suspend inside a resume now fails


mrichards30:
We're seeing a regression on develop after #1912 that means suspends inside "resumptions" don't work anymore. 

I think the issue is that functional suspend/resume means a theorem's suspension label list doesnt grow as its "resumptions" add new suspensions. So when you write `Resume some_thm[some_label]` we're only looking in the original `some_thm` for `some_label` and not finding it if it was added by a different `Resume`.

Example theory that works pre-refactor and fails now
```
Theory suspNested

Theorem nested:
  p ∧ (p ⇒ r) ⇒ p ∧ r
Proof
  strip_tac >> suspend "goal"
QED

Resume nested[goal]:
  conj_tac
  >- suspend "p"
  >> simp[]
QED

Resume nested[p]:
  simp[]
QED
        
Finalise nested
```

xrchz:
It should be possible to fix this by (recording and) searching additionally in the labels of the resumption theorems? Like, don't make it imperative, just additionally store data associated with the resumption theorem that is available for lookup by later resumptions.

mrichards30:
fixed by #1923

PR #1923, opened 2026-04-24
Fix nested suspense/resume proofs

mrichards30:
This addresses #1922 by fixing nested suspense/resume proofs and adds a test.

The main things this PR does to fix this is:
- Changed `extract_suspended_goal` to take a list of theorems, so it can search the parent and any of its resume proofs for a label 
- When recording a resumption proof, also store it under the key of each nested suspendlabel hypothesis it carries, which lets us use `lookup_resumption` on a label during `resume` to find its respective suspend
- `assemble_finalised` now loops until no suspendlabel hypotheses are left, since `PROVE_HYP` can introduce new ones if nested suspends are present


mn200:
Thanks for this!

PR #1927, opened 2026-04-26
Add PFT replay module

xrchz:
This commit introduces the PFT format for HOL proof traces and a standalone tool for replaying proofs in that format in HOL4.

In the future we intend to also integrate such replay in the system using the --thmsrc option of Holmake/build.


mn200:
Go wild!

PR #1930, opened 2026-04-27
Fix Holmake parallel cross-directory dependencies

charles-cooper:
_co-authored by OpenAI ChatGPT_

## Summary

This fixes parallel Holmake graph construction for clean cross-directory builds.

When a needed target is in an `INCLUDES` directory but has not been built yet, Holmake previously inserted a `Failed/NoCmd` placeholder for the external target. After `mkneeded`, that node remained unbuildable, so parallel builds could schedule dependent work before the external `.ui`/`.uo` products existed.

The fix expands missing external targets from scanned include directories in place:

- existing external targets remain `Succeeded/NoCmd` placeholders
- missing external targets in scanned directories are rebuilt as real graph nodes in their own directory
- missing external targets outside the recursive scan, such as standard `sigobj` entries, retain the old placeholder behavior

This keeps one global scheduler and avoids spawning nested Holmake processes.

## Validation

- compiled `bin/Holmake` from the patched sources
- checked `tools/Holmake/tests/depchain1/dir3` with `Holmake -n --jobs=2`
- confirmed the generated graph includes cross-directory dependency nodes and does not expand standard `sigobj` targets into a self-dependency loop
- added and ran `tools/Holmake/tests/parallel_deps`, which cleans and rebuilds the existing `depchain1/dir3` cross-directory dependency chain with `--jobs=2`

### DIFF
--- a/tools/Holmake/core/Holmake.sml
+++ b/tools/Holmake/core/Holmake.sml
@@ -407,6 +407,7 @@ type incmap = (hmdir.t, {incs:dirset,pres:dirset}) Binarymap.dict
 type dirinfo = {incdirmap : incmap, visited : hmdir.t Binaryset.set,
                 ancestors : hmdir.t list (* most recent hd of list *)}
 type 'a hmfold =
+     incmap ->
      {includes : string list, preincludes : string list} ->
      (string -> unit) ->
      hmdir.t ->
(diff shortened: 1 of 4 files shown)


mn200:
> avoids spawning nested Holmake processes

indicates that Claude does not know how `Holmake` works; there is only ever one `Holmake` process and it does not spawn nested instances of itself.

charles-cooper:
> > avoids spawning nested Holmake processes
> 
> indicates that Claude does not know how `Holmake` works; there is only ever one `Holmake` process and it does not spawn nested instances of itself.

that's fine but there are definitely bugs here, they are just hard to reproduce

PR #1934, opened 2026-04-29
quiet bare top level opens

### LINKED ISSUE #1933: Old syntax is very awkward in interactive sessions in emacs
Old syntax is borderline unusable in emacs now. It used to be the case that one can just send the entire text to the HOL buffer in emacs and things would run smoothly. 
```
open HolKernel Parse boolLib bossLib;
open arithmeticTheory;

val _ = new_theory "test";


val _ = export_theory();
```
Now the bindings from the theories are printed, which takes a long time. 

This is not a critical bug in the sense that everyone should switch to the new Theory syntax. I am writing this issue to point  out that backwards compatibility is hurt.

digama0:
fixes #1933


mn200:
Thanks!

PR #1936, opened 2026-05-01
Claude generated docs

mn200:
Documentation changes mostly generated by Claude.  Requested @digama0  and @dnezam as reviewers for the two "Overview" commits that attempt to document (in SML comments) the goalFrag/TacticParse "suite", and `tools/parsing`. 

I'm more than ready to be told Claude has been inaccurate, or even flat out wrong. 

I've tested locally at `build -t3` so I'm fairly confident nothing has broken.

Its summary of its commits:

## Summary

Two related strands of documentation work, organised into 12 commits
on top of `develop`:

1. **New `Overview` blocks** at the bottom of nine `.sig` files
   covering the HOL source-preprocessing pipeline, the tactic
   parser/linearizer, the interactive proof-fragment manager, and
   the per-theory persistent-value abstraction.
2. **Stale-comment cleanup** — four refreshes of comments that had
   drifted out of date, and six deletions of clearly-superseded code
   blocks preserved as comments.

Plus one related change: the long-standing "Test Cases" block at the
bottom of `src/finite_maps/enumTacs.sml` is promoted into
`src/finite_maps/selftest.sml` as proper `convtest` / `require_msg`
assertions, so regressions in those conversions now surface as build
failures.

Net diff: +1374 / −791. Verified with `bin/build -F -t` (full
sequence including selftests) at HEAD.

## New `Overview` documentation

Each block sits at the bottom of its `.sig`, banner-style, and
narrates the data model, design rationale, and groupings of API
entries — modelled on the existing exemplar in
`src/proofman/goalFrag.sig` (which is also brought into this branch
because it lived as an uncommitted local edit).

| File | Lines | Topic |
|---|---|---|
| `src/proofman/goalFrag.sig` | 87 | Interactive proof-fragment manager |
| `src/parse/TacticParse.sig` | 122 | Tactic-source parser, printer, and `tac_frag` linearised form |
| `tools/parsing/AttributeSyntax.sig` | 57 | `Theorem foo[attrs]:` bracketed-attribute parser |
| `tools/parsing/HOLSource.sig` | 67 | Top-level driver of the parse/expand/print pipeline |
| `tools/parsing/HOLSourceAST.sig` | 126 | AST datatypes, span discipline, `fileline` event-log machinery |
| `tools/parsing/HOLSourceExpand.sig` | 112 | Macro-expansion of HOL keywords (`Theorem`, `Definition`, `Theory`, …) |
| `tools/parsing/HOLSourceParser.sig` | 81 | Recursive-descent parser, scope/infix table, error reporters |
| `tools/parsing/HOLSourcePrinter.sig` | 81 | Span-aware token-stream emitter |
| `tools/parsing/HolParserOld.sig` | 112 | Legacy lex-driven translator (kept as a fallback / reference) |
| `src/parse/AncestryData.sig` | 169 | Per-theory persistent-value abstraction underpinning ThmSetData and friends |

## Stale-comment refreshes

| Commit | File | What was wrong |
|---|---|---|
| `778e160c4` | `src/HolSat/satConfig.sml` | Header pointed at `SatSolver.sml`; actual file is `SatSolvers.sml` |
| `778e160c4` | `src/HolSat/dimacsTools.sml` | Header described a global `sat_var_map` ref of type `(int * (term * int) set) ref`; current code threads a `(svm, sva)` Redblackmap pair |
| `778e160c4` | `src/parse/ParseDatatype.sml` | Single BNF described only the `id "of" <under_constr>` form; rewritten as two grammars (Hol_datatype `parse_phrase` style and Datatype-block `parse_hphrase` style) selected by the caller |
| `778e160c4` | `src/parse/Overload.sml` | Print-side described as a record-keyed map returning a single identifier; current type is an `LVTermNet` keyed by term shape with timestamp-based tie-breaking |
| `e1c103dfa` | `src/parse/Overload.sig` | `oi_strip_comb` comment claimed `SOME(f, args, pattern)`; signature shows `(term * term list) option` |
| `e1c103dfa` | `src/postkernel/Theory.sml` | "CREATE THE INITIAL THEORY SEGMENT" banner body referenced `timestamp` / `con_wrt_disk` fields that are no longer on the `segment` record (timestamps now live in a separate `metadata` ref) |
| `b7278dfa3` | `src/num/reduce/conv-old/Arithconv.sml` | Header implied outright supersedence; the file is in fact still selected (over the rewriting-based replacement in `../conv/`) by the OpenTheory logging kernel — clarified the kernel-conditional retention via `tools/sequences/core-theories` |
| `f9d24ab36` | `src/HolSmt/SmtLib_Parser.sml` | FIXME enumerated 5 supported commands "as of 2011-05-20"; the parser actually handles 10 (adds `set-info`, `declare-const`, `check-sat`, `get-proof`, `exit`) and explicitly rejects everything else |

## Deletions of superseded / dead code

| Commit | File | Lines | Why |
|---|---|---|---|
| `d30644e70` | `src/quotient/src/quotient.sml` | −275 | Block opening "It is now superceeded by the more general code for a partial equivalance"; body was the old normal-equivalence construction kept verbatim. `define_quotient_type` already routes through the partial-equivalence path |
| `d30644e70` | `src/HolSat/satTools.sml` | −5 | "isSuccess expected to be primitive in next release of Moscow ML" — Moscow ML caveat decades obsolete |
| `d30644e70` | `src/quantHeuristics/quantHeuristicsLibBase.sml` | −29 | 28-line block of commented-out REPL fragments including a tap into a non-existent `xxx` debug ref |
| `cdceb6641` | `src/datatype/EnumType.sml` | −143 | "Examples" block referencing `simpls` (no longer in `define_enum_type`'s result record) and `prove_initiality_thm`/`define_case`/`prove_cases_thm`/`case_cong_thm` (private to the structure). Couldn't be revived as-is; equivalent enum-shape coverage already exists via `Hol_datatype` tests in `src/datatype/selftest.sml` |
| `e2dda83f4` | `src/1/TypeBase.sml` | −90 | Two consecutive commented-out `pp_strip_case` implementations both flagged broken by their own comments; the active path uses `strip_case` directly via `term_pp.init_casesplit_munger` |
| `ca8e396a3` | `examples/miller/formalize/numContext.sml` | −23 | "old---hopefully obsolete" alternative `mod_mult_rewr`; live replacement directly above |

## Test promotion

`6b5b8ca02` ports the 161-line "Test Cases" comment block from
`src/finite_maps/enumTacs.sml` into the existing
`src/finite_maps/selftest.sml`. Each conversion now has a real
expected-output assertion (32 `convtest`, 6 `thmCheck`-via-`require_msg`
for the non-equational `OWL_*` / `set_TO_OWL` etc.) with outputs
captured from a one-off run in HOL. The original block also exercised
private helpers (`REVERS_CONV`, `smerge_*`, `incr_*`, `sinter_CONV`,
`sdiff_CONV`, `bt_to_ol_*`); those tests stay behind the structure
boundary and are dropped — the rest exercise the public `enumTacs` API.

## Decisions not acted on

A few candidates surfaced during the scan but were intentionally left:

- **`src/integer/CooperShell.sml:446`** — self-flagged "god-awful HACK"
  in `find_triangle_eliminable`; needs a real algorithm change rather
  than a comment fix, deferred.
- **`tools/mlyacc/*` and `tools/mllex/mllex.sml`** — preserved SML/NJ
  CVS revision-history blocks at file tops. Probably intentionally
  kept from the upstream import; no policy change here.
- **`src/metis/folTools.sml`** — `recent_fol_problems` ref initially
  flagged as "dead infrastructure", but `save_fol_problem` is in fact
  called from the same file (the ref defaults to `NONE` for a no-op
  but can be flipped at the REPL for ad-hoc debugging). Not stale,
  not changed.

## Test plan

- [x] `bin/build -t --seq=tools/sequences/upto-parallel` (core build) — green at HEAD
- [x] `bin/build -F -t` (full build + selftests) — green at HEAD; `finite-maps-selftest.log (3s) OK`, all 38 new assertions pass

## Notes for the reviewer

- The branch was started against `8bfa052cb` and `develop` has since
  advanced; a rebase or merge will be needed before merging.
- All commits include a `Co-Authored-By: Claude Opus 4.7 …` trailer.


tools/parsing/HOLSourceAST.sig
```diff
@@ -365,3 +365,129 @@ val updateCursor: cursor * int -> cursor
...
+
+      - 'exp separated = {args, seps, stop}
+        A comma/semicolon-separated list.  |seps| is normally
+        |args| - 1; |args| if a trailing separator was permitted (SML
+        extension, e.g. "[a, b,]"); 0 for synthetic lists.  A
+        separator entry of NONE marks an inserted separator from
+        parse-error repair.
+      - 'a seq = Empty | One of 'a | Many
+        Used wherever the surface allows zero, one, or
+        parenthesised-many items (e.g. tyvarseq, datatype tycons).
```
dnezam, line 402: I feel like these would be better at the actual declaration

tools/parsing/HOLSourceAST.sig
```diff
@@ -365,3 +365,129 @@ val updateCursor: cursor * int -> cursor
...
+        LetInEnd, App, AndAlso, OrElse, Handle, Raise, IfThenElse,
+        While, Case, Fn) plus the HOL quotation forms (HOLFullQuote,
+        HOLQuote) and #(LINE)/#(FILE) pragmas, and the synthetic
+        ExpExpansion / ExpEmpty / ExpBad markers.
+      - row : record-row entries (DotDotDot, LabEq, LabAs,
+        LabExpansion).
+      - dec : top-level declarations.  Standard SML
+        (DecVal/Fun/Type/Datatype/Abstype/Exception/Local/Open/Infix
+        /Infixr/Nonfix/Structure/Signature/Include/Sharing/Functor
+        /Exp), plus the HOL-extension declarations:
```
dnezam: I don't think there is much point in writing out the constructors in a comment, so this part can probably go

tools/parsing/HOLSourceAST.sig
```diff
@@ -365,3 +365,129 @@ val updateCursor: cursor * int -> cursor
...
+          HOLDefinition                       - Definition ... End
+          HOLDatatype                         - Datatype: ... End
+          HOLQuoteDecl                        - Quote id [= e]: ... End
+          HOLInductiveDecl                    - [Co]Inductive id: ...
+          HOLType                             - Type / Overload
+          HOLSimpleThm                        - Theorem id[attrs] = exp
+          HOLTheoremDecl                      - Theorem ... QED
+          HOLResume / HOLFinalise             - suspended-proof support
+          HOLLinePragmaWith / HOLFilePragmaWith
+                                              - #(LINE=...) / #(FILE=)
```
dnezam: In `HOLSourceAST.sml` and `HOLSourceAST.sig` at least there are examples for each of these (some of which I'd say are better), so these can go too

tools/parsing/HOLSourceAST.sig
```diff
@@ -365,3 +365,129 @@ val updateCursor: cursor * int -> cursor
...
+
+        plus DecMosmlPrimVal for Moscow ML compatibility, DecBad for
+        unparseable input, and DecExpansion as an "original ~> result"
+        wrapper produced by HOLSourceExpand.
+      - funarg, sigexp, strexp, qdecl : the structure / signature /
```
dnezam: The DecMosmlPrimVal can go, `DecBad` and `DecExpansion` are good but should probably go into the `sig`/`sml` file instead

tools/parsing/HOLSourceAST.sig
```diff
@@ -365,3 +365,129 @@ val updateCursor: cursor * int -> cursor
...
+        per-declaration attributes carried on Theorem / Definition /
+        Type / Resume / Finalise.  HOLAncestors and HOLLibs are the
+        two header forms.
+      - defn_label_id : labels inside DefinitionLabel
+        (HOLConjLabel for `[~name]`, HOLLabel for `[id]`).
+      - maybe_quoted : an identifier that may have been written either
+        bare (UnquotedId) or in double quotes (QuotedId).
+      - constraint : `:` vs `:>` for opaque vs transparent ascription.
+      - type_kw : selects between `type`, `eqtype`, and the Moscow ML
+        `prim_type` / `prim_eqtype` / `prim_EQtype` forms.
```
dnezam, line 450: Seems kind of pointless; I would remove this

tools/parsing/HOLSourceAST.sig
```diff
@@ -365,3 +365,129 @@ val updateCursor: cursor * int -> cursor
...
+
+    Construction helpers:
+
+      - mkIdent / mkString / mkInt / mkList / mkTuple / mkApp /
+        mkLabEq / mkRecord build synthetic nodes using a single
+        position parameter; the parser uses these when emitting
+        expansion results.
+      - decodeStr / encodeStr move between SML string literals and
```
dnezam, line 458: I would have hoped that the names are obvious enough; I would delete this too

tools/parsing/HOLSourceAST.sig
```diff
@@ -365,3 +365,129 @@ val updateCursor: cursor * int -> cursor
...
+        their raw character contents, handling \NNN / \uXXXX /
+        line-continuation escapes.
```
dnezam, line 460: Probably better at the declaration site

tools/parsing/HOLSourceAST.sig
```diff
@@ -365,3 +365,129 @@ val updateCursor: cursor * int -> cursor
...
+
+    Span accessors.  For each AST sort there is a *Stop (and where
+    useful *Start, *Span) function: idStop / idSpan, tyStart / tyStop
+    / tySpan, tybindStop, conbindStop, datbindStop, expStart /
+    expStop / expSpan, exbindStop, valbindStop, mosmlPrimvalbindStop,
+    sigexpStart / sigexpStop / sigexpSpan, strexpSpan, sigbindStop,
+    structbindStop, headerStop, decStart / decStop / decSpan,
+    functorbindStop.  These walk only as far as needed to find the
+    extent.
```
dnezam: Not particularly useful, would probably remove

tools/parsing/HOLSourceAST.sig
```diff
@@ -365,3 +365,129 @@ val updateCursor: cursor * int -> cursor
...
+    pointer that walks the event sequence; newCursor / updateCursor
+    advance it.  mkFileline body evts memoises a cursor and returns a
+    function (int -> fileline) that callers can use to translate an
+    arbitrary offset back to {file, line, col} -- amortised cheap when
+    queries arrive in roughly increasing offset order.
+
+    isOnlyComments tests whether a substring of the source consists
+    entirely of whitespace and (possibly nested) (* ... *) comments;
+    used by the printer and expander to decide where lossless
+    line-pragma insertions are safe.
```
dnezam: Maybe my search is broken, but it is only used in `HOLSourceExpand`, and there it is used in the context of inductive definitions that make use of labels. I would remove this

tools/parsing/HOLSourceAST.sig
```diff
@@ -365,3 +365,129 @@ val updateCursor: cursor * int -> cursor
...
+    (FileEvent, LineEvent, LineColEvent) capturing changes in the
+    file/line/col cursor as the input is consumed.  Each event carries
+    the character offset at which it takes effect; eventPos extracts
+    it.  Events are stored in an `events` record alongside the initial
+    filename, in a DArray for binary search.  A `cursor` is a stateful
+    pointer that walks the event sequence; newCursor / updateCursor
+    advance it.  mkFileline body evts memoises a cursor and returns a
+    function (int -> fileline) that callers can use to translate an
+    arbitrary offset back to {file, line, col} -- amortised cheap when
+    queries arrive in roughly increasing offset order.
```
dnezam, line 477: This is the part I don't know well, so I cannot say much about its correctness. At least it is not stating something obvious like "mkApp makes an App"

tools/parsing/HOLSourceExpand.sig
```diff
@@ -9,3 +9,115 @@ val expandDec:
...
+(* ----------------------------------------------------------------------
+    Overview
+   ----------------------------------------------------------------------
+
+    HOLSourceExpand is the macro-expansion stage of the source
+    pipeline: it takes a parsed HOLSourceAST.dec carrying the HOL-
+    specific surface forms (Theorem, Definition, Datatype, Inductive /
+    CoInductive, Type, Overload, Quote, Theory, Resume, Finalise, the
+    #(LINE=)/(FILE=) pragma variants, etc.) and rewrites it into pure
+    SML by emitting calls to the HOL kernel and library functions that
```
dnezam: Do we really call the HOL kernel? Maybe it is better to say that everything is desugared into function calls

tools/parsing/HOLSourceExpand.sig
```diff
@@ -9,3 +9,115 @@ val expandDec:
...
+             a sequence of `open`s grouped by qualified-vs-unqualified,
+             with structure aliases for each [alias=...] entry, and
+             optional HOL_Interactive open-bracketing
+      EndTheory  ~>  Theory.export_theory () (with optional
+                     Feedback.set_trace "TheoryPP.include_docs" 0
+                     before, when [no_sig_docs] was set on Theory)
+      Resume label[smlname=x, attrs]: tac QED
+          ~> val x = markerLib.resume {label_name, suspension_name} tac
+      Finalise id[attrs]
+          ~> val id = markerLib.finalise_suspended_thm "id[attrs]"
```
dnezam, line 86: Perhaps this is a matter of taste, but maybe one should just refer to unquote instead of running the risk of having outdated docs at some point

tools/parsing/HOLSourceExpand.sig
```diff
@@ -9,3 +9,115 @@ val expandDec:
...
+    keeping line directives anchored to the original source positions.
+
+    Inputs to expandDec:
+
+      - fileline : int -> fileline
+        Resolves a character offset to {file, line, col}.  Used to
+        choose between location-tagged versions of the runtime
+        helpers, e.g. boolLib.save_thm vs save_thm_at, Q.store_thm vs
+        store_thm_at, TotalDefn.qDefine vs located_qDefine, depending
+        on whether the source originates in a known file.
```
dnezam, line 39: Not sure about this part. Either way, does it really matter what it is used for?

tools/parsing/HOLSourceExpand.sig
```diff
@@ -9,3 +9,115 @@ val expandDec:
...
+      - parseError : int * int -> string -> unit
+        Reports any unrecognised or malformed attributes encountered
+        during expansion (e.g. unknown keys on [attrs], unsupported
+        argument shapes for [smlname], [induction], etc.).
```
dnezam: I feel like it missed the important part, namely that this is a function that takes a span and a message, which allows the user to decide what to do on failure. In particular, the use could raise an exception or delete your hard drive on parse errors

Example:
```
parseError (p, p + 2) "unicode escapes are not supported"
```

tools/parsing/HOLSourceExpand.sig
```diff
@@ -9,3 +9,115 @@ val expandDec:
...
+          ~> val id = markerLib.finalise_suspended_thm "id[attrs]"
+
+    Pure-SML decls are walked recursively rather than rewritten:
+    DecVal expands its expressions (treating top-level patterns
+    specially -- a top-level bare expression DecExp e becomes
+    `val it = e`, a non-top one becomes `val _ = e`), DecLocal /
+    DecStructure / DecSignature / DecFunctor / DecInclude descend
+    into their bodies, etc.  DecMosmlPrimVal becomes a vacuous
+    val-binding (Moscow ML's prim_val syntax loses its body).
+    DecBad turns into `val _ = raise Fail "malformed"`.
```
dnezam, line 95: People can read the code, this looks useless to me

tools/parsing/HOLSourceExpand.sig
```diff
@@ -9,3 +9,115 @@ val expandDec:
...
+
+    Expressions go through expandExp.  The notable transformation is
+    that SuccessorML "or-patterns" (Or {...}) are detected in
+    binding/pattern positions and either expanded into multiple match
+    arms (case / fn / handle), or rejected via parseError when they
+    appear somewhere multiple arms cannot be synthesised (val
+    bindings raise HasOrPat, which is caught and replaced with
+    ExpBad).
```
dnezam: It is indeed notable, and after looking at the code, I am not actually sure how Or patterns are handled.

tools/parsing/HOLSourceExpand.sig
```diff
@@ -9,3 +9,115 @@ val expandDec:
...
+      - magicBind, valPat, valWild, mkNameAttrs, mkLocString /
+        mkLocString' : internal builders for the common output shapes,
+        each parametrised by the position of the originating surface
+        token so generated decls inherit a sensible source span.
+
+    The result of expandDec preserves the AST shape and never rewrites
+    structure : the orig-bearing DecExpansion wrapper is the only
+    visible difference, and it carries enough information that
+    consumers can choose to display either the original surface text
+    or the elaborated SML.
```
dnezam, line 126: Does any of this matter?

tools/parsing/HOLSourceParser.sig
```diff
@@ -27,3 +27,84 @@ val parseSML: string -> (int -> string) ->
...
+    pre-populated with both the SML-basis infixes and the HOL tactic-
+    library operators -- THEN, THEN1, THENL, THEN_LT, THENC, ORELSE,
+    ORELSE_LT, ORELSEC, THEN_TCL, ORELSE_TCL, >>, >-, >|, >>>, >>-,
+    >~, >>~, >>~-, by, suffices_by, via, using, |->, |>, |>>, ?>, $,
+    --> and friends -- so that downstream tools see the same
+    associativity HOL programmers expect.  The parser updates the
+    scope as it encounters infix / infixr / nonfix declarations,
+    threading both inside and across local / structure / let blocks.
+
+    parseSML : string -> (int -> string) -> errSink -> scope -> result
```
dnezam, line 58: What is errSink? I don't think it should just randomly invent type abbreviations without saying so

tools/parsing/HOLSourceParser.sig
```diff
@@ -27,3 +27,84 @@ val parseSML: string -> (int -> string) ->
...
+      - filelineParseError print (body, events) : resolves spans to
+        file:line:col and emits compiler-style messages
+        ("foo.sml:12:34-12:40: parse error: ...").  This is what a
+        Holmake-driven build feeds in.
+
+    The parser is intentionally a single recursive-descent module,
+    not a parser-generator product; this is what makes it possible
+    to (a) feed input lazily and parse a declaration at a time, and
+    (b) recover from local syntax errors without abandoning the
+    rest of the file.
```
dnezam: Is (a) true though? Wasn't the previous lexer/parser/?? a parse-generator product but also lazy?

tools/parsing/HOLSourcePrinter.sig
dnezam, line 1: I haven't really worked on the printer, so I cannot say much on it. However, I get the impression that it mostly talks about obvious things, or is slightly inaccurate. For example, it says:
> This is what lets the printer
    reproduce the original line layout: as long as mkLine reports
    truthful line numbers (typically `#line o fileline` from the AST's
    event log), the printed text re-aligns to the source line breaks.

But suppose in the future there is a one-line construct that expands to 1000 lines; for some of the following constructs, the line numbers are going to wrong. This is just "best-effort".

dnezam (commented):
My impression is that most of it is unnecessary fluff or feels slightly inaccurate.

mn200:
Thanks for taking the time over this.  In one or two places, you mention that you think the comment is misplaced: that is my fault in that I deliberately asked for a big end-of-file overview comment.  I'll look into tweaking the other stuff you've identified.

tools/parsing/HOLSourceAST.sig
mn200: Deliberate per my top-level comment — keeping at EOF.

[Comment from Claude]

tools/parsing/HOLSourceAST.sig
mn200: Trimmed in 2c700585e — the verbose per-HOL-constructor surface-form list is gone; kept the dividing line between standard SML and HOL-extension decls plus the mention of `DecBad` / `DecExpansion`, which aren't surface forms.

[Comment from Claude]

tools/parsing/HOLSourceAST.sig
mn200: Yes, you're right — each HOL constructor already has the `(** Definition foo[attrs]: … *)` style annotation inline, so the duplicate one-liners at EOF were redundant. Dropped in 2c700585e; the Overview now points at the inline annotations instead.

[Comment from Claude]

tools/parsing/HOLSourceAST.sig
mn200: Partial — the explicit "Moscow ML compatibility" gloss for `DecMosmlPrimVal` is gone (the constructor is now just listed alongside the other SML ones). For `DecBad` / `DecExpansion`, see my top-level comment — keeping at EOF since they're the non-surface variants the dec-section reader most wants to know about. 2c700585e.

[Comment from Claude]

tools/parsing/HOLSourceAST.sig
mn200: Kept — `prim_type` / `prim_eqtype` / `prim_EQtype` aren't standard SML, and the note about Moscow ML lineage seemed worth keeping for readers who don't know that history.

[Comment from Claude]

tools/parsing/HOLSourceAST.sig
mn200: Kept — the names cover the broad strokes, but the `\NNN` / `\uXXXX` / line-continuation handling isn't obvious from `decodeStr` / `encodeStr` alone; that's the bit I wanted to surface.

[Comment from Claude]

tools/parsing/HOLSourceAST.sig
mn200: Deliberate per my top-level comment — keeping at EOF.

[Comment from Claude]

tools/parsing/HOLSourceAST.sig
mn200: Trimmed in 2c700585e — the exhaustive `*Stop` / `*Start` / `*Span` enumeration is gone; description now lists representative examples (`tyStart`/`tyStop`/`tySpan`, `expStart`/`expStop`/`expSpan`, `decStart`/`decStop`/`decSpan`) and the "walk only as far as needed" rationale.

[Comment from Claude]

tools/parsing/HOLSourceAST.sig
mn200: Noted, thanks — leaving the source-position-tracking section as-is. If you do spot inaccuracies on a closer pass, happy to revisit.

[Comment from Claude]

tools/parsing/HOLSourceAST.sig
mn200: Yes, my description was wrong on both counts — `isOnlyComments` has nothing to do with the printer or "line-pragma insertions". Its sole call site is `HOLSourceExpand`, filtering comment-only quote-literal fragments when splitting `Inductive` / `CoInductive` quotes at `DefinitionLabel` boundaries. Fixed in 2c700585e.

[Comment from Claude]

tools/parsing/HOLSourceExpand.sig
mn200: Yes, that overstated things — the expansion emits calls to library helpers (`Q.store_thm`, `TotalDefn.qDefine`, `Theory.new_theory`, `Parse.set_grammar_ancestry`, `IndDefLib.xHol_reln`, `markerLib.*`, etc.), not to the kernel proper. Reworded in 2c700585e.

[Comment from Claude]

tools/parsing/HOLSourceExpand.sig
mn200: Kept — the located vs non-located helper choice is the visible reason the `fileline` callback is parameterised in, so explaining that seemed worth keeping for someone reading the inputs section.

[Comment from Claude]

tools/parsing/HOLSourceExpand.sig
mn200: Yes, the previous wording only said "reports" rather than explaining the contract. Rewritten in 2c700585e: `parseError` is now described as a caller-supplied callback over `(span, message)`, with the caller picking the failure semantics (raise / log / accumulate) and the expander itself not aborting.

[Comment from Claude]

tools/parsing/HOLSourceExpand.sig
mn200: Kept — the surface→SML mapping is the part readers most often want to consult when reasoning about why a HOL declaration produced the SML it did, and a "see the source" pointer doesn't substitute well for inline expansion sketches. Happy to update specific entries if you spot drift.

[Comment from Claude]

tools/parsing/HOLSourceExpand.sig
mn200: Kept — the paragraph documents semantics that aren't obvious from the datatype declarations: top-level `DecExp e` ⇒ `val it = e` (non-top ⇒ `val _ = e`); `DecMosmlPrimVal` ⇒ vacuous binding (Moscow ML's `prim_val` loses its body); `DecBad` ⇒ `raise Fail "malformed"`. Reading `expandDec` reveals all of these, but having them stated saves the trip.

[Comment from Claude]

tools/parsing/HOLSourceExpand.sig
mn200: Tightened in 2c700585e — now reads "in `case`/`fn`/`handle` positions each alternative is duplicated across multiple match arms; in `val` bindings, where no such duplication is sound, the expander raises `HasOrPat`, catches it, reports through `parseError`, and emits `ExpBad`". Let me know if any of that is still inaccurate.

[Comment from Claude]

tools/parsing/HOLSourceExpand.sig
mn200: Kept — the paragraph captures the printer-relevant invariant ("expansion preserves AST shape; `orig` is preserved alongside `result`"), which is what lets consumers downstream choose to display the original surface text or the elaborated SML.

[Comment from Claude]

tools/parsing/HOLSourceParser.sig
mn200: Yes, the previous Overview wrote `errSink` in the type signature as if it were a defined SML abbreviation, but it's just prose shorthand. Fixed in 2c700585e: `errSink` is now explicitly called out as a prose-only abbreviation for `(DString.dstring * HOLSourceAST.events) -> int * int -> string -> unit`, and the full curried type is shown once.

[Comment from Claude]

tools/parsing/HOLSourceParser.sig
mn200: Fair — laziness / pull-style isn't actually a recursive-descent-vs-parser-generator distinction (it's an API property). Dropped that claim in 2c700585e; the section now keeps just the genuine point about hand-written recursive descent making local syntax-error recovery via `ExpBad` / `DecBad` / `right=NONE` closers straightforward.

[Comment from Claude]

tools/parsing/HOLSourcePrinter.sig
mn200: Yes, the previous wording overstated the guarantee — the printer can only emit forward newlines, so when a single surface line expands to many output lines (typical for a HOL declaration rewritten to a sequence of SML decls), the alignment overshoots and subsequent "jump to line N" requests become no-ops. Reframed in 2c700585e as best-effort, with the overshoot mode called out explicitly. For the rest of the Overview — happy to revise specific bits if you can flag where it's stating the obvious or being inaccurate.

[Comment from Claude]

mn200:
> Thanks for taking the time over this. In one or two places, you mention that you think the comment is misplaced: that is my fault in that I deliberately asked for a big end-of-file overview comment. I'll look into tweaking the other stuff you've identified.

And when I say “I’ll look into”, I mean, “I’ll get Claude to do it.”

PR #1937, opened 2026-05-01
Holmake cache UX refactor

mrichards30:
[On Zulip](https://hol.zulipchat.com/#narrow/channel/486798/topic/Holmake.20cache/near/591700635), some UX improvements for the Holmake cache were discussed.

This PR addresses these concerns by adding the following:
- to start, the existing `--use-cache` and `--write-cache` flags have been removed
- a default cache folder is now determined by the first of either `$XDG_CACHE_HOME` or `$HOME/.cache/HOL`
- bin/build and Holmake both now use the cache by default
  - both support`--no-cache` and `--cache-dir <directory>` flags to disable the cache and choose a specific folder to use a cache respectively (though the latter mostly only exists for tests)
  - bin/build will print its resolved cache folder at the start of the build
  - image
  - when on, it checks for an existing cache hit before it builds. if it doesn't find anything it locally builds and caches these artefacts for next time
- the github actions files have --no-cache added
- the cache is disabled if any side products (coproducts?) are found in the Holmakefile since it's probably not safe to assume these can be cached
- when there's a cheat AND the theory is cached, the cheat status shows instead when built 

I ran into an issue with the cache (mostly visible in tests with multiple identical fooScript.sml files in different folders) where since these would generate identical cache keys, only one version would be cached. Since cache keys are path independent, they would be given the same Theory.sml files; this is bad because Theory.sml files mention their path relative to $HOLDIR. 
To fix this I have changed TheoryPP.sml to write relative paths into Theory.sml, and now the working directory is moved to each theory during the build so that relative paths resolve correctly.

image

This PR also has a number of fixes to some apparent bugs that showed up after making caching the default - Claude helped with a lot of these.

- [This change](https://github.com/HOL-Theorem-Prover/HOL/pull/1937/changes#diff-17b4e7e67fc0262103b7952bdb3d69b84df625d79ca223533643c2b074f1f449R313) changes TheoryPP to use the  `#(FILE)` macro instead of writing absolute paths into built Theory.sml files.
- Not so much a bug, but [these tests](https://github.com/HOL-Theorem-Prover/HOL/pull/1937/changes#diff-6c0e98a42d97353b404c442e12ce2092280197ca5d0426f8710355c0711ef9eaR1) look for the generated HTML file which I've skipped for caching, so I've set them to run with the no-cache flag.
- [Here](https://github.com/HOL-Theorem-Prover/HOL/pull/1937/changes#diff-fd1a473aaef5d5da0c452b3f7873fdb497c78967c63fd03685fc8403556c0c82R807) is Claude's fix to allow side products to be loaded from subfolders (needed to migrate emitScript.sml to use side products).
- [Here](https://github.com/HOL-Theorem-Prover/HOL/pull/1937/changes#diff-b92b10d7b0a056e8455aa1726b6381dce173b8cb3a468a9ddc9fde51bd34cf05R36) is a fix for a confusing bug I ran into where for some reason mosml expanded #(FILE) to a relative path? This change just makes it match PolyML's handling.

### DIFF
--- a/tools/Holmake/poly/HM_CacheFetch.sml
+++ b/tools/Holmake/poly/HM_CacheFetch.sml
@@ -115,18 +115,6 @@ fun fetch base_url cachekey (ofns : Holmake_tools.output_functions) =
                 val _ = if ok
                         then info "Cache hit! local theory building can be skipped."
                         else warn "Only managed a partial cache hit; theory will be built locally."
-                val _ = if ok then
-                  List.app (fn {name,...} =>
-                    if String.isSuffix "Theory.dat" name then
-                      let val s = TextIO.openIn (to_dest_dir name)
-                          val c = TextIO.inputAll s
-                      in TextIO.closeIn s;
-                         if String.isSubstring "\"cheat\"" c then
-                           info "(used CHEAT)"
-                         else ()
-                      end handle _ => ()
-                    else ()) files
-                else ()
             in
                 ok
             end
--- a/tools/Holmake/poly/HM_CacheFetch.sml
+++ b/tools/Holmake/poly/HM_CacheFetch.sml
@@ -20,7 +20,7 @@ fun recMkDir dir =
 fun already_cached base key =
     OS.FileSys.access(OS.Path.concat(OS.Path.concat(base, "key"), key), [])
 
-fun upload base_url cachekey dir thyname (ofns : Holmake_tools.output_functions) =
+fun upload base_url cachekey dir filenames (ofns : Holmake_tools.output_functions) =
     case cachekey of
         HM_Cachekey.Missing _ => false
       | HM_Cachekey.Key key =>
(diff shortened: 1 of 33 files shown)


mn200:
Looking forward to it!  I don’t think we should be snaffling the .cache name though: perhaps .hol/cache, also allowing people to put their interactive config files under .hol.

mrichards30:
> Looking forward to it! I don’t think we should be snaffling the .cache name though: perhaps .hol/cache, also allowing people to put their interactive config files under .hol.

oops, that was just my typo. it's in $HOME/.cache/HOL - $HOME/.cache seems to be the standard cache location (on mac at least)
image

mn200:
Hmm.  I'm not mad keen on the relative paths.  What about making the key include the path back to the top .holpath?  Then "identical" script files are not identical anymore.

mrichards30:
> Hmm. I'm not mad keen on the relative paths. What about making the key include the path back to the top .holpath? Then "identical" script files are not identical anymore.

i think this is probably a better idea given how difficult it has been to get relative paths to work (and i still couldnt get them to work with mosml). @xrchz are you happy if i change the cache key code?

xrchz:
No don't change cacheKey. We should change the Theory.sml files to not use a literal path but rather use the #(FILE) macro to find its own filename and refer to the dat file that way (which should be a sibling file, i.e., the same file with a different extension).

tools/Holmake/poly/HM_CacheFetch.sml
```diff
@@ -96,14 +105,28 @@ fun fetch base_url cachekey (ofns : Holmake_tools.output_functions) =
...
                                  fetch_to_file (base_url ^ url) (to_dest_dir name))
                              files
                 val _ = if ok
                         then info "Cache hit! local theory building can be skipped."
                         else warn "Only managed a partial cache hit; theory will be built locally."
+                val _ = if ok then
+                  List.app (fn {name,...} =>
+                    if String.isSuffix "Theory.dat" name then
+                      let val s = TextIO.openIn (to_dest_dir name)
+                          val c = TextIO.inputAll s
```
mn200: Reading a whole .dat file just to see if there's something cheated in it is gross.  Perhaps drop this for now; we can put a "contains-a-cheat" tag into the .dat format later.

mn200 (changes requested):

tools/Holmake/poly/HM_CacheFetch.sml
mn200: Such a tag could be guaranteed to exist within the first few bytes/lines of the format so that a mindless check, not requiring any s-expression parsing, would be easy and fast.

tools/Holmake/poly/ProcessMultiplexor.sml
```diff
@@ -188,7 +188,7 @@ struct
             val _ = OS.FileSys.chDir dir
           in
             if try_cache () then
-              OS.Process.exit OS.Process.success
+              OS.Process.terminate OS.Process.success
```
mn200: Why the change from `exit` to `terminate`?  The former shuts the process down more cleanly.

mrichards30:
thanks for the review @mn200 ive fixed pr comments

mn200:
Let's doooooo it!

Issue #1595, opened 2025-08-04
Consistent SML structures for Sets and Finite Maps


konrad-slind:
Currently there's a mix of usage of structures`Binaryset` and `Binarymap`, defined in `<holdir>/tools-poly/poly, and also `Redblackset`,  `Redblackmap`, `HOLdict`, and `HOLset` from `<holdir>/portableML`.  The latter two structures are the middle two structures augmented with some pretty printing support. It seems there is a lot of redundancy here. Beyond the redundancy, more recent implementations, such as the SML/NJ Library's [ordered sets](https://www.smlnj.org/doc/smlnj-lib/Util/sig-ORD_SET.html#:~:text=The%20ORD_SET%20signature%20defines%20an,specific%20types%20are%20also%20provided.) structure, provide a richer API.

Task: install the sets and finite map functors from the SML/NJ library and adapt them as needed to rebuild HOL. 

Bonus points: define structures for sets and maps for HOL type variables, types, term variables, and terms. Put these somewhere in the vicinity of HolKernel (maybe in it?).

konrad-slind:
Nice to see clanker-tech applied to this unappealing chore!

mn200:
Shame it hasn't quite figured out all the nuances yet!  (It's still whirring away on my machine atm; Moscow ML's handling of functors doesn't make life easy.)

PR #1942, opened 2026-05-03
Phase 1 of Table.sml-based map migration (issue #1595)

mn200:
## Summary

Phase 1 of the consolidation requested in #1595: replace the
overlapping `Binarymap`/`Redblackmap`/`HOLdict` map libraries with
structures derived from `src/portableML/Table.sml` (the Isabelle-style
2-3 tree functor already vendored in tree).  Existing instantiations
are reused where possible -- `Symtab` (string), `Inttab` (int),
`Termtab` (term), `KNametab` (kname), `Typetab` (hol_type, promoted to
top-level as part of the SharingTables work here), `Symreltab` -- with
local Tab structures for the handful of maps whose key types don't
already have a top-level instantiation.

This PR is the dictionary half of the consolidation only; sets and
the polymorphic comparator-passing abstractions are deferred (see
"What's not in this PR" below).

## Why Table over SML/NJ ORD_MAP

The issue proposes SML/NJ Library's `ORD_SET`/`ORD_MAP` functors as
the destination.  This PR uses `Table` instead because it is already
in tree (no new dependency), the API is richer (`update_new`,
`default`, `map_default`, `cons_list`, `insert_list`, `join`,
`merge`, ...), and the "bonus points" task in #1595 -- pre-instantiated
maps for term/type/etc. -- is partly already done in tree (`Termtab`,
`KNametab`).

Happy to switch direction if there's a strong reason to prefer the
SML/NJ functors instead.

## Commit structure

Six commits, each independently buildable with `bin/build -F`:

1. **Migrate src/ map users to Table-derived structures** -- the bulk
   of the work: a per-directory walk through src/ (~33 directories,
   originally landed one commit each before being squashed for review).
2. **Switch SharingTables to Symtab/Idtab/Typetab/Termtab** -- a
   public-API surface that needed coordinated changes across consumers.
3. **Switch src/real/SOSLib to Termtab/MonoTab/IntPairTab/IntTripleTab**
   -- ~100 sites in one large internal file, no signature leak.
4. **Remove HOLdict and switch FullUnify Env to Symtab/Termtab** --
   delete a redundant Redblackmap wrapper after mass-renaming its
   28 consumers; rebuild FullUnify.Env on Symtab/Termtab.
5. **Migrate post-HOLdict consumers to specific Tabs** -- per-consumer
   migration of those 28 sites to the right Tab for their key type,
   plus follow-up exception-interface cleanup in TypeNet/LVTermNet/
   LVTermNetFunctor (each now has a local `exception NotFound`).
6. **Switch ThmSetData.simple_dictionary to Symtab** -- the last
   public-API dict surface in src/.

## What's *not* in this PR

Explicitly deferred from phase 1:

- **Sets** (Binaryset/Redblackset).  Table lacks union/intersection/
  difference/subset; adding them is its own change.
- **Polymorphic comparator-passing abstractions**: `Conv.memoize`
  (gates one site in `wordsLib`), `aiLib`, `mlib*` (Multiset, Subst,
  Termnet, Termorder, Thm), `HolSat` RBM/SVM.  These need
  restructuring rather than a straight migration since `Table` is a
  functor over a fixed key type.
- **Cross-directory** `OpenTheoryMap` (consumed by relation,
  coretypes, num/theories, datatype, many *Script.sml).
- **Internal-heavy** `src/HolSmt` (~12 files, dozens of sites).
- **`tools/`** -- blocked on having Table.sml available early enough
  in the build sequence (would need `tools/sequences/` futzing).
- **`examples/`** -- next phase, not started.

`Lib.dict` and `Lib.dict_topsort` intentionally stay polymorphic.

## Test plan

- [x] `bin/build -F` passes after each commit (verified during the
      original per-directory work; full build green on the regrouped
      branch as well)
- [x] No content drift vs. the pre-regroup commit history (empty diff
      between the regrouped tip and the original 44-commit tip)

### DIFF
--- a/src/1/DefnBaseCore.sml
+++ b/src/1/DefnBaseCore.sml
@@ -7,19 +7,25 @@ struct
 open HolKernel boolSyntax Drule
 
 type kname = KernelSig.kernelname
-type defnstore = (string * kname, string * thm) Binarymap.dict
+
+structure DStab = Table(struct
+  type key = string * kname
+  val ord = pair_compare(String.compare, KernelSig.name_compare)
+  fun pp (s,k) = HOLPP.add_string (s ^ "," ^ KernelSig.name_toString k)
+end)
+
 datatype defn_thm = STDEQNS of thm | OTHER of thm
 fun thm_of (STDEQNS th) = th | thm_of (OTHER th) = th
 type defn_presentation = {const: term, thmname: kname, thm : defn_thm}
+type defnstore = (kname * defn_thm) DStab.table
 
 
-val empty_dstore =
-    Binarymap.mkDict(pair_compare(String.compare, KernelSig.name_compare))
+val empty_dstore : defnstore = DStab.empty
 
 fun list_insert m k v =
-    case Binarymap.peek(m,k) of
-        NONE => Binarymap.insert(m,k,[v])
-      | SOME vs => Binarymap.insert(m,k,v::vs)
+    case KNametab.lookup m k of
+        NONE => KNametab.update (k,[v]) m
+      | SOME vs => KNametab.update (k,v::vs) m
 
 fun to_kid {Thy,Name,Ty} = {Thy = Thy, Name = Name}
 val prim_dest_const = to_kid o dest_thy_const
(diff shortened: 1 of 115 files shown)


konrad-slind:
This looks like a good start. `Table.sml` needs to have some documentation, which should explain its signature entry points and should also enumerate the common HOL instances (Symtab, TermTab, ...)

konrad-slind:
I unthinkingly merged to `develop` after making my comment. Hope this doesn't clobber anything ...

mn200:
Think it will be fine.  Agree about the need for extra documentation.

PR #1943, opened 2026-05-03
Fix SAT_ORACLE bogus theorems from dropped F conjuncts in to_cnf

### LINKED ISSUE #1170: SAT_ORACLE gives wrong answers
This is correct behaviour:
```
> minisatProve.SAT_PROVE ``b ==> T``;
val it = ⊢ b ⇒ T: thm
```
This is incorrect behaviour:
```
> minisatProve.SAT_ORACLE ``b ==> T``;
Exception- SAT_cex ⊢ b ⇒ ¬(b ⇒ T) raised
```
In other words, the oracle version of the interface to SAT returns a bogus theorem for this simple input.

The following shows that one can prove false from the bogus theorem.
```
> (minisatProve.SAT_ORACLE ``b ==> T``; fail()) handle minisatProve.SAT_cex th => th;
val it = ⊢ b ⇒ ¬(b ⇒ T): thm
> it |> GEN_ALL |> SPEC T |> REWRITE_RULE [];
val it = ⊢ F: thm
> Tag.dest_tag (tag it);
val it = (["DISK_THM", "HolSatLib"], []): string list * string list
```
The last line shows that the proof of false depends on a trusted oracle call made by `HolSatLib`. That is to say: this isn't a soundness bug in HOL but rather a bug report about `minisatProve.SAT_ORACLE`.

mn200:
## Summary

`SAT_ORACLE  “b ==> T”` previously produced

```
Exception- SAT_cex [oracles: HolSatLib] [] ⊢ b ⇒ ¬(b ⇒ T) raised
```

— a bogus oracle theorem from which `|- F` is derivable.

The root cause was in `to_cnf` (`src/HolSat/def_cnf.sml`). For `tm = b ⇒ T` the negated goal `~tm = b ∧ ¬T` is split by `GCONJUNCTS` into `[b, ¬T]`; `presimp_conv` collapses `¬T` to `F`. The dispatch loop then read

```sml
if is_T t orelse is_F t then (m,tops,defs,lfn)   (* skip *)
else …build clause…
```

Skipping a `T` conjunct is sound (`p ∧ T = p`); skipping an `F` conjunct is not (`p ∧ F = F`, not `p`). With the `F` clause gone, the CNF passed to MiniSAT was just `[b]`, MiniSAT correctly reported satisfiability, and `SAT_ORACLE` (which by design skips the post-hoc `satCheck` that `SAT_PROVE` runs) fabricated the bogus implication and raised `SAT_cex`.

This PR fixes the root cause: when `to_cnf` detects an `F` conjunct it raises a new `to_cnf_unsat` exception carrying `|- ¬tm`; `initialise` translates that into the existing `initexp` channel via `NOT_NOT_CONV` and `GEN_SAT` returns `|- tm`. Both `SAT_PROVE` and `SAT_ORACLE` benefit, and MiniSAT is no longer invoked for trivially-unsat preprocessing cases.

A regression test under `src/HolSat/selftest.sml` asserts `SAT_ORACLE \`b ==> T\`` returns `|- b ==> T`. The test runs without needing a working MiniSAT (the short-circuit is hit before any solver invocation).

### Follow-up not addressed here

Issue #1170's discussion also flags a deeper soundness concern: `SAT_ORACLE`'s "SAT verdict" branch fabricates an oracle theorem from the model without verifying it, so a buggy verdict from MiniSAT on any *other* input could still leak a bogus theorem. binghe (in the issue thread) eventually concluded that "perhaps the whole idea of SAT_ORACLE is wrong — without calling satCheck the intended oracle theorem cannot be determined." Hardening that path (e.g. always running `satCheck`, even on the oracle path) is left as a separate follow-up; this PR closes the specific reported reproducer.

Closes #1170.

## Test plan
- [x] Core build (`bin/build -t --seq=tools/sequences/upto-parallel`) green.
- [x] New regression test passes: `SAT_ORACLE proves `b ==> T` (#1170) … OK`.
- [x] Existing HolSat behaviour for genuinely-SAT inputs is unaffected — only the `is_F` branch in `to_cnf`'s dispatch loop changes; the `is_T` skip and the clause-building branch are untouched.

### DIFF
--- a/src/HolSat/Holmakefile
+++ b/src/HolSat/Holmakefile
@@ -1,6 +1,18 @@
 .PHONY: all
-all: $(DEFAULT_TARGETS)
+all: $(DEFAULT_TARGETS) selftest.exe
 
 ifeq ($(KERNELID),otknl)
 all: $(patsubst %Script.sml,%.ot.art,$(wildcard *Script.sml))
 endif
+
+selftest.exe: selftest.uo
+	$(HOLMOSMLC) -o $@ $<
+
+ifdef HOLSELFTESTLEVEL
+all: holsat-selftest.log
+
+holsat-selftest.log: selftest.exe
+	$(tee ./selftest.exe 2>&1,$@)
+endif
+
+EXTRA_CLEANS = selftest.exe holsat-selftest.log
--- a/src/HolSat/def_cnf.sig
+++ b/src/HolSat/def_cnf.sig
@@ -3,6 +3,7 @@ sig
 
   include Abbrev
   val presimp_conv : conv
+  exception to_cnf_unsat of thm
   val to_cnf : bool -> term ->
                (string option * int *
                 (term,term)satCommonTools.RBM.dict *
--- a/src/HolSat/def_cnf.sml
+++ b/src/HolSat/def_cnf.sml
@@ -17,6 +17,12 @@ val presimp_conv =
             [NOT_CLAUSES, AND_CLAUSES, OR_CLAUSES, IMP_CLAUSES, EQ_CLAUSES,
              COND_EXPAND, COND_CLAUSES])
 
+(* Raised by to_cnf when boolean simplification of one of the top-level
+   conjuncts of the input term reduces to F: the input is unsatisfiable
+   without help from the SAT solver. The carried theorem is |- ~tm where
+   tm is the term passed to to_cnf. *)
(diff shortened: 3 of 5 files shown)


binghe (approved):
It's been too long for me to remember this issue... but I think your code changes make sense: now it prevents dropping `F`-conjuncts, together with `T`-conjuncts.

PR #1947, opened 2026-05-04
Initial work on Böhm's (beta-)separation/separability thm

binghe:
Hi,

Previously Böhm's λη-completeness was proved by the η-separation/separability theorem:
```
[separability_thm] (lameta_completeTheory)
⊢ ∀X M N r.
    FINITE X ∧ FV M ∪ FV N ⊆ X ∪ RANK r ∧ 0 < r ∧ has_benf M ∧ has_benf N ∧
    M =/== N ⇒
    ∀P Q. ∃pi. Boehm_transform pi ∧ apply pi M === P ∧ apply pi N === Q
```
If the `apply pi M === P ∧ apply pi N === Q` in the conclusion of the above theorem were replaced with `apply pi M == P ∧ apply pi N == Q`, i.e. using `lameq` instead of `lameta`, the resulting so-called Böhm's (β-)separability theorem, is a more difficult result, whose proof requires a new device called virtual subterm (vsubterm), with the following long definition:

```
[vsubterm_def] (separabilityTheory, new)
⊢ (∀X M r. vsubterm X M [] r = SOME (M,r)) ∧
  ∀X M h p r.
    vsubterm X M (h::p) r =
    if solvable M then
      (let
         M0 = principal_hnf M;
         n = LAMl_size M0;
         vs = RNEWS r n X;
         M1 = principal_hnf (M0 ·· MAP VAR vs);
         Ms = hnf_children M1;
         m = LENGTH Ms;
         j = h − m;
         zs = RNEWS r (n + SUC j) X;
         z = LAST zs;
         M2 = if h < m then EL h Ms else VAR z
       in
         vsubterm X M2 p (SUC r))
    else NONE
```
In comparision with the existing `subterm` which returns `NONE` when the number of hnf children is not enough:
```
[subterm_def] (boehmTheory)
⊢ (∀X M r. subterm X M [] r = SOME (M,r)) ∧
  ∀X M h p r.
    subterm X M (h::p) r =
    if solvable M then
      (let
         M0 = principal_hnf M;
         n = LAMl_size M0;
         vs = RNEWS r n X;
         M1 = principal_hnf (M0 ·· MAP VAR vs);
         Ms = hnf_children M1;
         m = LENGTH Ms
       in
         if h < m then subterm X (EL h Ms) p (SUC r) else NONE)
    else NONE
```
The new `vsubterm` only returns `NONE` on unsolvable (sub)terms, otherwise it's like doing an inline η-expansion and returns a carefully chosen fresh `VAR z` as the subterm.

An imporatant way to evaluate the quallity of a definition is to see if it has nice properties. The existing `subterm` is a good definition because it has the following corresponce to Böhm trees (a subterm is NOT none iff the path is in the Böhm tree):
```
[BT_ltree_paths_thm] (boehmTheory, existing)
⊢ ∀X M p r.
    FINITE X ∧ FV M ⊆ X ∪ RANK r ⇒
    (p ∈ ltree_paths (BT' X M r) ⇔ subterm X M p r ≠ NONE)
```
And if two terms are β-equivalent (`M == N`), then the congruence property holds for their `subterm`s over the same (arbitrary) path:
```
[lameq_subterm_cong]
⊢ ∀X M N p r.
    FINITE X ∧ FV M ⊆ X ∪ RANK r ∧ FV N ⊆ X ∪ RANK r ∧ M == N ⇒
    (subterm X M p r = NONE ⇔ subterm X N p r = NONE) ∧
    (subterm X M p r ≠ NONE ⇒ subterm' X M p r == subterm' X N p r)
```

The new `vsubterm` is also good, and it seems that it has support of η-equivalence built in: another congruence property (of the same shape) holds for any two subterms which are βη-equivalent (`M === N`):
```
[lameta_vsubterm_cong]
⊢ ∀X M N p r.
    FINITE X ∧ FV M ⊆ X ∪ RANK r ∧ FV N ⊆ X ∪ RANK r ∧ M === N ⇒
    (vsubterm X M p r = NONE ⇔ vsubterm X N p r = NONE) ∧
    (vsubterm X M p r ≠ NONE ⇒ vsubterm' X M p r === vsubterm' X N p r)
```

Then, as an important step in proving the "Böhm out" technique (for single term), the following lemma shows the existence of a Βöhm transformation for passing certain closed substitution down to virtual subterms, maintaining the potential Böhm tree structure: (Note that the universal quantifier `p` has no restriction at all, i.e. arbitrary path works, even it's not in the Böhm tree, which is impossible for `subterm`)
```
[Boehm_transform_exists_lemma']
⊢ ∀X M p r.
    FINITE X ∧ FV M ⊆ X ∪ RANK r ⇒
    ∃pi.
      Boehm_transform pi ∧ is_ready (apply pi M) ∧
      FV (apply pi M) ⊆ X ∪ RANK (SUC r) ∧
      ∃v P.
        closed P ∧
        ∀q. q ≼ p ∧ q ≠ [] ∧ vsubterm X M q r ≠ NONE ⇒
            vsubterm X (apply pi M) q r ≠ NONE ∧
            vsubterm' X (apply pi M) q r = [P/v] (vsubterm' X M q r)
```
The construction of the needed Böhm transformation `pi` in the above lemma is very similar with the case of `subterm` but requires another new device called `vsubterm_width`, based on the existing `subterm_width`:
```
[vsubterm_width_def] (separabilityTheory)
⊢ ∀M p. vsubterm_width M p = MAX (subterm_width M p) (SUC (MAX_LIST p))
```

The successful proof of the above lemma shows that the "Böhm out" technique for multiple terms can also be proved for virtual subterms, but the proof will be slightly longer due to more cases which must handled in the proof (the existing lemma for `subterm` has already over 4K lines). By following basically the same proof path, replacing all `subterm` lemmas with corresponding `vsubterm` lemmas, I believe Böhm's (original) β-separability theorem can be formally proved.

--Chun

### DIFF
--- a/examples/lambda/barendregt/boehmScript.sml
+++ b/examples/lambda/barendregt/boehmScript.sml
@@ -526,6 +526,18 @@ Proof
  >> qexistsl_tac [‘X’, ‘M’, ‘r’, ‘n’, ‘vs’, ‘M1’] >> simp []
 QED
 
+Theorem FV_subterm_upperbound :
+    !X M p r. FINITE X /\ FV M SUBSET X UNION RANK r /\
+              subterm X M p r <> NONE ==>
+              FV (subterm' X M p r) SUBSET X UNION RANK (r + LENGTH p)
+Proof
+    rpt STRIP_TAC
+ >> fs [GSYM IS_SOME_EQ_NOT_NONE, IS_SOME_EXISTS]
+ >> Cases_on ‘x’
+ >> MP_TAC (Q.SPECL [‘p’, ‘X’, ‘M’, ‘q’, ‘r’, ‘r'’] subterm_rank_lemma)
+ >> NTAC 2 (rw [])
+QED
+
 (* NOTE: This theorem provides a better estimates for “FV (EL h Ms)”, than the
    above subterm_induction_lemma.
  *)
@@ -1112,12 +1124,14 @@ Proof
  >> qexistsl_tac [‘M’, ‘M0’, ‘n’, ‘m’, ‘vs’, ‘M1’] >> simp []
 QED
 
-(* NOTE: In the above theorem, when the antecedents hold, i.e.
+(* NOTE: Due to BT_ltree_paths_thm (below):
 
-         p IN ltree_paths (BT X M) /\ subterm X M p = NONE
+         p IN ltree_paths (BT' X M r) <=> subterm X M p r <> NONE
 
-   Then ‘subterm' X M (FRONT p)’ must be an unsolvable term. This result can be
-   even improved to an iff, as the present theorem shows.
+   The conclusion actually reduces to
+
+            ~(p <> [] /\ subterm X M (FRONT p) r <> NONE /\
+              unsolvable (subterm' X M (FRONT p) r))
  *)
 Theorem subterm_is_none_iff_parent_unsolvable :
     !p X M r. FINITE X /\ FV M SUBSET X UNION RANK r /\
@@ -1185,16 +1199,12 @@ QED
         ‘p NOTIN ltree_paths (BT X M)’, the conclusion (rhs) always holds.
  *)
 Theorem subterm_is_none_inclusive :
-    !X M p r. subterm X M p r = NONE <=>
-              !q. p <<= q ==> subterm X M q r = NONE
+    !X M p r. subterm X M p r = NONE <=> !q. p <<= q ==> subterm X M q r = NONE
 Proof
     rpt GEN_TAC
  >> reverse EQ_TAC
  >- (DISCH_THEN (MP_TAC o (Q.SPEC ‘p’)) >> rw [])
- >> Q.ID_SPEC_TAC ‘r’
- >> Q.ID_SPEC_TAC ‘M’
- >> Q.ID_SPEC_TAC ‘X’
- >> Q.ID_SPEC_TAC ‘p’
+ >> qid_specl_tac [‘p’, ‘X’, ‘M’, ‘r’]
  >> Induct_on ‘p’ >- rw [subterm_NIL]
  >> rw [subterm_def]
  >> Cases_on ‘q’ >> fs [subterm_def]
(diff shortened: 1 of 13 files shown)


binghe:
In `examples/lambda/basics/nomdatatype.sml`, previously it missed a right parenthesis. But surprisingly the (new?) parser just reported a warning (and luckly the code still works), which I also didn't see.    Is this behavior of the parser (only warning on mismatched parenthesis) desired, by design?

mn200:
Do you think the otknl failure is your fault?  (It seems unlikely.)

binghe:
> Do you think the otknl failure is your fault? (It seems unlikely.)

Actually it is caused by a new proof in `rich_listTheory`:
```
Theorem LAST_TAKE_EL :
    !l n. 0 < n /\ n <= LENGTH l ==> LAST (TAKE n l) = EL (PRE n) l
Proof
    simp [PRE_SUB1]
 >> Induct_on ‘l’ >> rw []
 >> simp [LAST_DEF]
 >> Cases_on ‘l = []’ >> fs []
 >- (‘n = 1’ by simp [] >> simp [])
 >> Cases_on ‘n <= 1’
 >- (‘n = 1’ by simp [] >> simp [])
 >> simp [EL_CONS]
 >> ‘PRE (n - 1) = n - 2’ by simp []
 >> simp []
QED
```

I think the use of `n - 2` in the proof makes `rich_listTheory` depend on `iSUB` of `numeralTheory`. I noticed it only because I saw a warning when building `src/boss/hol4-base-unint.art`:
```
WARNING: 2 different constants named HOL4.numeral.iSUB
```

But in `src/boss/hol4-base-unint.thy`, currently `rich-list` didn't import `numeral` (it didn't need it before). So I'm putting `numeral` into it now and I'm waiting for my local build to finish (the building process has passed the previous breaking point shown in CI log):
```
rich-list {
  import: bool
  import: arithmetic
  import: numeral
  import: marker
  import: combin
  import: pair
  import: list
  import: pred-set
  import: relation
  import: divides
  article: "../list/src/rich_list.ot.art"
}
```

P.S. I wouldn't call it "fault" because it's nothing wrong to add a new theorem into `rich_listTheory`, just the whole OpenTheory design is a bit vulnerable and sensitive to seemingly irrelevant changes...

mn200:
Thanks for this!

PR #1948, opened 2026-05-04
Use relative paths in Theory.sml files

mrichards30:
I'm working on improving the Holmake cache UX and have been stuck on changing the Theory.sml files to use relative paths (see #1937). This was getting a bit messy so I'm pulling it into its own PR.

I'm not sure if this approach is too hacky but after many attempts, this PR has PolyML printing relative paths and MosML still printing absolute paths (previously PolyML would print a path relative to $HOLDIR and MosML would print a full absolute path with no path variable).


mn200:
Not keen; but see comment on original PR.

xrchz:
see https://github.com/HOL-Theorem-Prover/HOL/pull/1937#issuecomment-4378370788 instead

PR #1950, opened 2026-05-05
Per-theory HTML docs + theorytool-driven theory graph

mn200:
Follow-up to [1ea272ca](https://github.com/HOL-Theorem-Prover/HOL/commit/1ea272ca02ea53bb4cc30229b5f8951325176b38) and the discussion under it about losing theorem statements from the htmlsigs/HOL Reference Page links and the role of doc emission.

## Summary

- `Theory.export_theory` writes a styled HTML doc per theory at `<src>/.hol/docs/<thy>Theory.html` (replacing the comment-style docs that the original commit moved into `<thy>Theory.txt`).  Theorem names are accent-coloured links wrapped around an Octicons `file-code` glyph that take you to the source location.
- Each `<thy>Script.sml` is mirrored to `<src>/.hol/docs/<thy>Script.html` with a per-line `<span id="Lnnn">`, so the source links land on the right line in any browser without depending on `:~:text` or editor URI schemes.
- Parent links in the HTML doc resolve across `.hol/docs/` directories using each parent's recorded metadata path.
- The theory dependency graph is now built via `src/portableML/rawtheory/theorytool` (the older `help/src-sml/DOT` script is removed): `theorytool --thygraph --url-base=DIR --paths-from=FILE` reads an explicit list of `<thy>Theory.dat` paths.  In `write_theory_graph`, that list is materialised by walking sigobj symlinks, so only production theories appear in the graph and there is no clash from theories that exist in more than one place under the tree.
- The `no_sig_docs` theory attribute keeps working with the same effect (now: skip the HTML doc and the script mirror); the underlying trace is renamed `TheoryPP.include_html_docs` for clarity.
- HOL syntax rendering of theorem statements is preserved — these docs go through the kernel's `pp_thm`, not raw-theory rendering.

## Test plan
- [x] `bin/build cleanAll && bin/build -F` succeeds end-to-end.
- [x] Sample doc page (e.g. `src/pred_set/src/.hol/docs/pred_setTheory.html`) shows the parent pills resolving across source dirs (e.g. to `src/num/extra_theories/.hol/docs/dividesTheory.html`) and theorem names link to `pred_setScript.html#Lnnn` with the expected line content.
- [x] `help/theorygraph/theories.svg` covers the production library; clicking any node opens the corresponding `.hol/docs/<thy>Theory.html`.
- [x] `Theory foo[no_sig_docs]` skips both `<thy>Theory.html` and `<thy>Script.html` for that theory, with the existing test under `tools/Holmake/tests/no_sig_docs/` still passing.


binghe (commented):
Hi,

It seems that you only changed what's after the theory graph, because when I clicked into the link of a theory in the HOL Reference Page, it still shows that same blank theory HTML as before: (e.g. `/help/src-sml/htmlsigs/numTheory.html`). This is not surprising, because I think all links on the page are generated by the same code from all SML signature files.

What's changed by this PR are those HTML files after clicking into the theory graph, they look fancy now but contains less information than before: previously there's a table of new types and constants defined in the theory. Now this table is gone.

I also see you have generated a HTML version of the original scripts, which can be reached by clicking theorem names in the _fancy_ (I don't like it, because it's space-wasting) theory HTML page, e.g. `src/num/theories/.hol/docs/numeralScript.html#L952`. I think this design is unnessary. At least I don't want to read the original SML proof script (even without no grammar highlight) in my browser. This design is also not disk space friendly, for large projects like CakeML it means the entire code base will be doubled, with a HTML version hidden inside those `.hol` directories.

For what I really need (as before): a compact **Table of Contents** (TOC) of definition and theory names at the beginning of each theory page, so that by clicking it I can jump to the definition body or theorem statements in later part of the page, this PR has done **nothing**.   This TOC is very important for reading large theories such as `real_topology`, `integration`, etc.

I think generating per-theory documents in `.hol` directory and use it for HOL Reference Page is not a good design.
Previously, by copying just the "help" directory to a Web server's root directory (e.g. a local server), at least it can provide a basic online reference page for HOL core theories. Now, the entire HOL working directory must be copied to do the same thing.

If user (like me) needs per-theory documents, they should explicitly do it by calling functions like `html_theory` inside their scripts, and the output should stay with the original script or can be specificed to any other location. Implicitly (and forcely) generating them into `.hol` directory is not useful, I think.

With all due respect, I suggest none of code changes in this PR being merged, not a single bit of them... as each of them make things even worse than before.

--Chun

binghe:
I really _appreciate_ if you can first revert [1ea272ca](https://github.com/HOL-Theorem-Prover/HOL/commit/1ea272ca02ea53bb4cc30229b5f8951325176b38), and (if still wanted) redo it as PR with more community people consulted before merging.

And if it's the performance of (new) parser which cannot handle large SML signatures with large comments, I blame the parser (and its coder) instead.

mn200:
I appreciate the desire for a ToC and will act on that.   We can't revert the removal of data from `.sig` files; that is clearly bad design and a weird conflation of user-facing and machine-facing data.

mn200:
As for wasting disk-space: I did
```
find . -name '*Script.sml' | xargs wc 
```
in my recent CakeML checkout and got 25MB of sources (`736678 3081344 25961143 total`).

I then did `du --si` and learnt that the whole directory structure (partially built with whatever it is I'm working on, but not completely by any stretch of the imagination) takes up 890MB. 

So, having an extra 25MB consumed with duplicate copies of the sources is a drop in the bucket.

mn200:
The setup has now changed a bit more; feedback from all definitely welcome!  (I appreciate it's a bit of work to checkout a PR just to see what it all looks like.)

mn200:
Incidentally, I also don't much like the idea of replicating the `src` structure if we were to extract these files for publishing elsewhere.  

But, that is what we have already been doing: see the suspicious URL behind "Source File" when you visit, e.g., https://hol-theorem-prover.org/trindemossen-2-helpdocs/help/src-sml/htmlsigs/Abbrev.html

That's broken, but those links worked for previous releases, and it looks as if I failed to properly move across the source file when I set up the pages for T-2.

mn200:
There's thus a separate task here: build infrastructure to make the web-page creation more reliable.  (Perhaps I just failed to run the script at HOL-Theorem-Prover/hol-webpages: https://github.com/HOL-Theorem-Prover/hol-webpages/blob/master/tools/mkhelpdocs

binghe:
The current state is basically acceptable (to me), **thanks**, but I think there are still some room to improve:

- In the per-theory HTML, `(none)` could be avoided by conditioning in the related code, to avoid outputing any section which is empty. For instance, in all core theories (except for `bool` and `coalgAxioms`),  the **Axioms** section is "(none)".  Note that some user theory may contain only definitions, or only theorems, thus you cannot assume any of these sections is non-empty.

- The "Parents" section (even the name is plural) _somehow_ always contain just one parent theory. This is not a regression caused by recent commit but a long confusion since I saw these pages. I think this information (a single parent theory somehow randomly picked from all parent theories) is meaningless (and can be removed from the page), unless you find a way to show all parent theories.

- Calling `html_theory` still generates the old HTML theory. This means that currently there are two piece of code doing the same job.

mn200:
- I explicitly asked for the (none); though I agree that it's perhaps a bit silly for axioms and the vast majority of theories that don't have any. 
- I am not sure what you mean: when I look at generated `cardinal` HTML for example, I see two parents : “permutes” and “wellorder”.  I'll rebuild everything I have locally to check also.
- You're right that `html_theory` should probably go.

mn200:
I also see, for example:
Screenshot 2026-05-06 at 16 23 11

binghe:
Sorry, my observation (on "Parent Theories") was wrong. It can be multiple, and it seems directly corresponding to the parent arcs of theory nodes in the Theory Graph.

If I understand it correctly, this seems to be the _direct_ parent theories: other parent theories who are (direct or indirectly) parent of these direct theories, will not show up.  But I may be still wrong, because the only (direct) parent theory of `sigma_algebra`, (somehow) is `nets`, which is quite unexpected, because it's even not explicitly opened:
```
Theory sigma_algebra
Ancestors
  arithmetic option pair combin pred_set topology iterate
  prim_rec metric real
Libs
  pred_setLib numLib hurdUtils jrhUtils res_quanTools
```

In this unexpected sense, I said it's meaningless.

mn200:
The weirdness arises because all theories in `src/probability` use the `real-heap` heap, and so they are automatically building in a context where all of *its* theories are already loaded.

mn200:
Latest commit addresses issue where axioms are unnecessarily reported as none.

PR #1954, opened 2026-05-08
fix vim interrupt kills hol process

nikos-alexandris:
Currently, vim interrupt (h-c) kills the whole terminal process instead of interrupting the executing thread. Now, instead of the interrupt broadcasting to all threads, we check if the executing thread is alive and interrupt it directly, else we do nothing.

Minimal reproduction: `echo Interrupt >>"$HOLDIR/tools/editor-modes/vim/fifo"` while a hol process is running kills the terminal.

### DIFF
--- a/tools/editor-modes/vim/vimhol.src
+++ b/tools/editor-modes/vim/vimhol.src
@@ -81,7 +81,12 @@ local
       NONE => (warn "Vimhol's tail gave eof\n"; restartTail())
     | SOME line =>
       case String.tokens Char.isSpace line of
-          ["Interrupt"] => (warn "Vim interrupt\n"; Thread.broadcastInterrupt ())
+          ["Interrupt"] => (
+            warn "Vim interrupt\n";
+            if Thread.isActive (!rpid) then
+              Thread.interrupt (!rpid)
+            else
+              warn "No proof thread to interrupt\n")
         | ["ReadFile",tmp] => let open Thread Mutex ConditionVar in
             warn ("Vim input "^tmp^"\n");
             lock m;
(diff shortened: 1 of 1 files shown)


tools/editor-modes/vim/vimhol.src
```diff
@@ -95,7 +100,6 @@ local
           end
         | [] => ()
         | _ => warn ("Got this rubbish from Vim: "^line);
-    Thread.testInterrupt () handle Interrupt => (removeQueue(); Thread.exit());
```
xrchz, line 98: Why do we need to remove this?

xrchz (approved):
Main change seems reasonable. Commented on the other

tools/editor-modes/vim/vimhol.src
nikos-alexandris: I assumed this was there because the previous code was broadcasting the interrupt. Since now we only interrupt the running thread this shouldn't do anything right?

xrchz: Idk maybe an interrupt could come from elsewhere though? If it shouldn't do anything maybe it is safe to leave as-is? But I don't mind actually, this should just be driven by what actually works

nikos-alexandris: Yes you are right i tested with this line left and it works fine. For some reason i thought this was handling only our custom interrupt. Ok i will revert this line change nice catch

nikos-alexandris: Done!

tanyongkiam (commented):
Tested and works for me.

PR #1958, opened 2026-05-09
Theory: write export outputs atomically via temp+rename

mn200:
## Summary
- Killing a Holmake build mid-`export_theory` previously left zero-byte `fooTheory.{sig,sml}` files; subsequent Holmakes mistook them as up-to-date and the build kept failing.
- Each output is now written to `<path>.<pid>.tmp` and renamed into place only after every output has finished. On a hard kill the final files keep their previous mtime (or stay missing on a first build), so Holmake correctly rebuilds.
- The pid suffix keeps the temp filename unique across concurrent Holmakes that share a theory dependency.

[Bug report on Zulip](https://hol.zulipchat.com/#narrow/channel/486798/topic/Bug.20report.20for.20Holmake.2FTheoryPP.20build).

## Test plan
- [x] `bin/build -t --seq=tools/sequences/upto-parallel` clean.
- [x] `tools/Holmake/tests/dirlock` selftest passes 5/5 (concurrent-Holmakes case is the regression test for the pid-suffix piece).

### DIFF
--- a/help/Docfiles/Portable.unique_tmp_suffix.md
+++ b/help/Docfiles/Portable.unique_tmp_suffix.md
@@ -1,0 +1,21 @@
+## `unique_tmp_suffix`
+
+``` hol4
+Portable.unique_tmp_suffix : unit -> string
+```
+
+------------------------------------------------------------------------
+
+A short string suitable for embedding in temp filenames so that two
+processes writing to the same logical path pick disjoint temp paths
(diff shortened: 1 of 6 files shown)


mn200:
Would be happy to have your feedback @talsewell .

talsewell:
I've just got around to looking at it, looks good to me!

PR #1962, opened 2026-05-11
markerLib: reject cyclic Resume proofs at recording time

### LINKED ISSUE #1960: Finalise loops when a resumption uses the original theorem
Running the following script file goes into an infinite loop (confusingly, if you don't know what to look for):
```
Theory bug

Theorem useful:
  (x ==> T) /\ (F ==> T)
Proof
  conj_tac >- suspend "x" >> suspend "F"
QED

Resume useful[x]:
  strip_tac >> ACCEPT_TAC TRUTH
QED

Resume useful[F]:
  MATCH_ACCEPT_TAC (cj 1 useful)
QED

Finalise useful
```

I don't know whether the correct behaviour is to somehow reject the theorem usage during resumption, or to just catch the cycle in Finalise, but the status quo is no good.

mn200:
## Summary
- Detect resumption proofs whose hypotheses include any suspendlabel of the parent theorem (the exact pattern in #1960: `MATCH_ACCEPT_TAC (cj 1 useful)`) and reject them in `markerLib.resume` before they reach the resumption store.
- Add a defence-in-depth progress check in `assemble_finalised`: bail out if any slab targeted in an iteration reappears in the result.

## Test plan
- [x] New regression test `src/basicProof/theory_tests/suspCycleScript.sml` reproduces the issue and asserts the cyclic resume call now raises `HOL_ERR` with `"cycle"` in the message.
- [x] `bin/build -t --seq=tools/sequences/upto-parallel` passes cleanly; existing `suspTest`, `suspNested`, `suspFast`, `suspRerun`, `suspCross{A,B}`, `suspSib{A,B,C,D}` are unaffected.

Closes #1960

### DIFF
--- a/src/basicProof/theory_tests/suspCycleScript.sml
+++ b/src/basicProof/theory_tests/suspCycleScript.sml
@@ -1,0 +1,51 @@
+Theory suspCycle[bare]
+
+(* Regression test for GitHub issue #1960.
+
+   If a Resume proof depends on the parent theorem's own
+   suspendlabel hypotheses (e.g. via `cj 1 useful` where `useful`
+   is the still-suspended parent), the resulting resumption
+   theorem re-introduces those hypotheses every time PROVE_HYP
+   discharges them, so Finalise's assembly loop never converges.
+
+   Reproduced from the issue:
+
+     Theorem useful:
+       (x ==> T) /\ (F ==> T)
+     Proof
+       conj_tac >- suspend "x" >> suspend "F"
+     QED
+     Resume useful[x]:
+       strip_tac >> ACCEPT_TAC TRUTH
+     QED
+     Resume useful[F]:
+       MATCH_ACCEPT_TAC (cj 1 useful)
+     QED
+     Finalise useful
+
+   Used to spin forever inside finalise_suspended_thm; now
+   `resume` rejects the cyclic proof up-front. *)
+
+Libs HolKernel Parse boolLib markerLib testutils
+
+Theorem useful:
+  (x ==> T) /\ (F ==> T)
+Proof
+  conj_tac >- suspend "x" >> suspend "F"
+QED
+
+Resume useful[x]:
+  strip_tac >> ACCEPT_TAC TRUTH
+QED
+
+val _ = shouldfail {
+  testfn = (fn () =>
+              resume {suspension_name = "useful", label_name = "F"}
+                     (MATCH_ACCEPT_TAC (cj 1 useful))),
+  printarg = (fn _ =>
+              "resume useful[F] with MATCH_ACCEPT_TAC (cj 1 useful)"),
+  printresult = (fn _ => "<unexpected success>"),
+  checkexn =
+    check_HOL_ERRexn (fn (_, fnm, msg) =>
+        fnm = "resume" andalso String.isSubstring "cycle" msg)
+} ()
--- a/src/marker/markerLib.sml
+++ b/src/marker/markerLib.sml
@@ -1252,6 +1252,23 @@ fun resume {suspension_name, label_name} tac =
                let
                  val goal = resumption_to_goal ncts
                  val sub_th = prove_goal (goal, tac)
+                 (* Reject resumption proofs that depend on the parent
+                    theorem's own suspendlabel hypotheses: those hyps
+                    will be re-introduced every time PROVE_HYP discharges
+                    them at Finalise, so assembly never converges. *)
+                 val parent_slabs = HOLset.filter (can dest_slab) (hypset th)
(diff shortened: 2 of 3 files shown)


src/basicProof/theory_tests/suspCycleScript.sml
```diff
@@ -0,0 +1,51 @@
...
+     QED
+     Resume useful[F]:
+       MATCH_ACCEPT_TAC (cj 1 useful)
+     QED
+     Finalise useful
+
+   Used to spin forever inside finalise_suspended_thm; now
+   `resume` rejects the cyclic proof up-front. *)
+
+Libs HolKernel Parse boolLib markerLib testutils
```
xrchz, line 29: wut? no Libs

xrchz: oh I see because this is [bare]. Sorry!

src/marker/markerLib.sml
```diff
@@ -1252,6 +1252,23 @@ fun resume {suspension_name, label_name} tac =
                let
                  val goal = resumption_to_goal ncts
                  val sub_th = prove_goal (goal, tac)
+                 (* Reject resumption proofs that depend on the parent
+                    theorem's own suspendlabel hypotheses: those hyps
+                    will be re-introduced every time PROVE_HYP discharges
+                    them at Finalise, so assembly never converges. *)
+                 val parent_slabs = HOLset.filter (can dest_slab) (hypset th)
```
xrchz, line 1259: I don't see how this is restricting to only avoiding dependencies on the current suspended theorem. Doesn't this prevent dependencies on _any_ suspended theorem?

xrchz: Maybe this is intended, but feels a bit more restrictive than necessary...

xrchz: If we want to be more permissive, we should also add a regression test that exercises it:
```
Theorem th1:
  ... suspend "l1" ...
QED

Theorem th2:
   ... suspend "l2" ...
QED

Resume th1[l1]:
   ... use th2 ...
QED

Resume th2[l2]:
  ... (don't use th1 or th2) ...
QED

Finalise th2

Finalise th1
```

xrchz (changes requested):
Please add a theory test that exercises using not-yet-finalised other theorems before finalisation. (Or we can discuss and decide we don't want to support that.)

mn200:
Claude wrote back:
> The check *is* scoped to just the current parent. `parent_slabs = HOLset.filter (can dest_slab) (hypset th)` operates on `th`, which is bound at `SOME (_, parent_thy, th) =>` — the parent theorem we're resuming. `HOLset.member (parent_slabs, h)` is alpha-equivalence against just that set, so a `slab` whose closed goal differs (i.e. one carried by some other suspended theorem) won't match.

> Concretely, your sketch passes my check at resume time: `Resume th1[l1]` proves a sub-goal using `th2`, the resulting `sub_th` carries `slab_th2_l2`, but `slab_th2_l2 ∉ parent_slabs` (different closed goal), so we don't raise.

> What does break, though, is `Finalise th1` later — independently of this PR. `find_res_for keys` lookups on `(parent_thy, parent_nm, label)` taken from the Finalise call, so when `slab_th2_l2` lands in the working theorem during assembly, the lookup is `(th1_thy, "th1", "l2")` rather than `(th2_thy, "th2", "l2")` where `rth_l2` actually lives. So it errors with `"No resumption proof found for label l2 of th1$th1"`. Slabs don't carry the owning theorem, so there's no way today to redirect the lookup. Supporting cross-theorem `Resume` dependencies would need either embedding the owner in the slab or having `find_res_for walk` all known suspension stores by label+goal — a real change, separate from the #1960 loop fix.

> Happy to land a test here that pins down the current behaviour: `Resume th1[l1]` using th2 succeeds at resume time, then `Finalise th1` errors (not loops) with the message above. That demonstrates the per-parent scoping of the new check while making the wider limitation explicit. If you'd like the actual feature, I'd suggest a separate issue — happy to file it.

mn200:
I have not thought about this code, nor Claude's changes to it, nor your comments at all.  Let me know if you think a human (other than you!) is required.

xrchz:
Yes please add the test illustrating the current behaviour and limitation and file a new issue to remove the limitation. I think the current change is fine for #1960.

mn200:
@xrchz if you're around, are you happy to approve the PR now that there's a fresh issue raised (#1963) with the appropriate test-case attached?  I will assume your last comment is implicit permission to merge if I don't hear from you otherwise.

xrchz (approved):

PR #1965, opened 2026-05-12
htmlsigs: index identifiers nested in inline sub-structures

### LINKED ISSUE #238: Link to Definition.new_type_definition is missing from HTML index of symbolic identifiers

mn200:
## Summary

- `Parsspec.processSpec` now recurses into the body of an inline `structure X : sig ... end`, so its val/typ/exc/con/datatype specs are recorded in `help/HOL.Help` and appear in `idIndex.html`.
- `Htmlsigs.processSig` line-scans the .sig file for those sub-structure blocks and passes the innermost structure name to `locate_docfile`, so identifiers inside one resolve to `<InnerStruct>.<id>.md`.
- `DefinitionDoc.sig` (the workaround sig for this issue) and its `aliasStrName` entry are retired.

Closes #238.

## Test plan

- [x] `makebase.exe` rebuilds cleanly.
- [x] `idIndex.html` contains entries for `new_type_definition`, `new_definition`, `new_specification`, `gen_new_specification`, each linking to `Theory.html`.
- [x] `Theory.html` hyperlinks each of those vals to its `help/Docfiles/HTML/Definition.<name>.html`, and all four files exist on disk.
- [x] Other nested-structure cases also work, e.g. `Parse.Unicode.unicode_version` is now indexed.
- [x] Top-level entries (`ASM_REWRITE_TAC`, `new_axiom`, …) unaffected; `DefinitionDoc` no longer appears in `idIndex.html`.
- [x] `bin/build -t --seq=tools/sequences/upto-parallel` and a subsequent full `bin/build` complete; `makebase` runs in well under a second.


xrchz (approved):
looks good assuming it works

PR #1967, opened 2026-05-12
add NoTrace and export_proof facilities

digama0:
## Summary
- Add a `TraceMode` structure (in `src/prekernel`) governing whether the trace kernel records proofs
- Add `Tracing.export_proof` which writes a theorem's proof tree to a file and rewrites the in-memory proof ref to `Exported_prf (file, thm_id)`, so subsequent trace serialization records the file reference instead of the full subtree
- Add `Thm.mark` and the corresponding `Mark_prf` proof constructor for inserting labeled markers in the proof trace (identity on the theorem, visible in the trace)
- Update `ProofTraceParser` (`sh_proof` datatype and `shProof` decoder), `ProofTraceReplay`, and `ProofTraceWalk` to handle the two new constructors

## Test plan
- [ ] Build with `--trknl` and confirm the kernel compiles
- [ ] Confirm `Thm.mark` produces a theorem whose proof ref is `Mark_prf`
- [ ] Confirm `export_proof` writes a separate file and rewrites the proof ref to `Exported_prf`
- [ ] Replay a trace containing `Mark_prf` and `Exported_prf` proofs via `ProofTraceReplay`

### DIFF
--- a/src/postkernel/Theory.sig
+++ b/src/postkernel/Theory.sig
@@ -55,6 +55,14 @@ sig
   val add_anonymous_thm      : thm -> int (* for ThyDataSexp only *)
   val export_theory          : unit -> unit
 
+(* Externalize a theorem's proof tree to a side .tr.gz file, then flip
+   the proof ref to Exported_prf. The in-memory proof tree becomes
+   garbage-collectable. Named form uses the given name; anonymous form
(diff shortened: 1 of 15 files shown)


mn200:
Thanks!

PR #1968, opened 2026-05-13
markerLib: support cross-theorem Resume dependencies

### LINKED ISSUE #1963: Cross-theorem Resume dependencies: Finalise can't resolve foreign-parent slabs
A `Resume` proof for theorem `th1` may legitimately depend on another suspended theorem `th2`. The resume-time check added in #1962 correctly allows this (the cycle check is scoped to the *current* parent's suspendlabel hypotheses, so `slab_th2_*` is not flagged). However, `Finalise th1` then fails: in `assemble_finalised`, `find_res_for` keys lookups on `(parent_thy, parent_nm, label)` taken from the Finalise call, so when `slab_th2_l2` lands in the working theorem during assembly the lookup is `(thy, "th1", "l2")`, while the actual resumption was stored under `(thy, "th2", "l2")`. The result is `No resumption proof found for label l2 of thy$th1` instead of successful assembly.

Slabs only carry the label name and the closed goal — not the owning theorem — so `find_res_for` has no way to redirect the lookup today.

## Minimal repro

```
Theorem th1: p ==> T Proof suspend "l1" QED
Theorem th2: q ==> T Proof suspend "l2" QED

Resume th1[l1]:  MATCH_ACCEPT_TAC th2  QED   (* legitimately depends on th2 *)
Resume th2[l2]:  strip_tac >> ACCEPT_TAC TRUTH  QED

Finalise th2  (* OK *)
Finalise th1  (* fails: "No resumption proof found for label l2 of thy$th1" *)
```

## Resolution options

1. Embed the owning theorem name in each slab (e.g. extend `suspendlabel` to carry `(thy, name, label)` rather than just `label`), so `find_res_for` can dispatch to the right resumption store.
2. Have `find_res_for` walk all known suspension stores keyed by label+closed-goal rather than by name. Brittle to label-name collisions.

Option (1) is the more principled fix.

## Notes

A `shouldfail` test in `src/basicProof/theory_tests/` pins down the current error mode (a clear `HOL_ERR`, not an infinite loop) so removing this limitation can be a focused change.

mn200:
## Summary
- A Resume proof that cites another still-suspended theorem inherits its `suspendlabel` hypotheses. `assemble_finalised` previously keyed lookups on the current parent's name, so those foreign slabs raised `"No resumption proof found ..."` at Finalise time.
- `save_thm_attrs` now stamps each fresh slab on a suspended theorem with its owning `(thy, name)` by encoding them into the label variable's name (delimited by SOH). `find_res_for` decodes the owner and routes the lookup accordingly, with a DB-fallback that reconstructs an rth from the clean theorem when the owner has already been Finalised.

## Test plan
- [x] New regression test `src/basicProof/theory_tests/suspCrossThmScript.sml` exercises the end-to-end case: `Resume th1[l1]` cites suspended `th2`, `Finalise th2` first, `Finalise th1` second.
- [x] `bin/build -t --seq=tools/sequences/upto-parallel` passes; existing `suspTest`, `suspCycle`, `suspNested`, `suspFast`, `suspRerun`, `suspCross{A,B}`, `suspSib{A,B,C,D}` all still pass.

Closes #1963

### DIFF
--- a/src/1/boolLib.sml
+++ b/src/1/boolLib.sml
@@ -110,33 +110,57 @@ end (* local open *)
 
 val def_suffix = ref "_def"
 
-fun has_suspended_subgoals th =
-    let
-      open HolKernel
-      fun issusp_hyp t =
-          let
-            val (fx,_) = dest_comb t
-            val (sl_c, lnm) = dest_comb fx
-            val {Thy,Name,...} = dest_thy_const sl_c
-          in
-            Thy = "marker" andalso Name = "suspendlabel" andalso
-            is_var lnm
-          end handle HOL_ERR _ => false
+(* If t = `suspendlabel <lab> g` for a variable <lab>, return SOME <lab>. *)
+local open HolKernel in
+fun slab_lab_var t =
+    let val (fx, _) = dest_comb t
+        val (sl_c, lnm) = dest_comb fx
+        val {Thy, Name, ...} = dest_thy_const sl_c
     in
-      isSome (HOLset.find issusp_hyp (hypset th))
-    end
+      if Thy = "marker" andalso Name = "suspendlabel" andalso is_var lnm
+      then SOME lnm else NONE
+    end handle HOL_ERR _ => NONE
+end
 
-(* The label variable's name encodes a closure-variable count in the
-   format "N name". Decode to extract the user-visible name and count. *)
+fun has_suspended_subgoals th =
+    isSome (HOLset.find (isSome o slab_lab_var) (Thm.hypset th))
+
+(* The label variable's name encodes the closure-variable count and
+   the user-visible label, and -- for slabs that have been stamped at
+   save-time -- the owning theorem (thy, name).  Two formats:
+
+   Legacy: "N name"
+        N is the closure-variable count, then a single space, then
+        the user's label.
+
+   Owner-stamped: "Nthythmnamename"
+        Same closure count and label, but with the owning (thy, name)
+        embedded between the count and the label, delimited by SOH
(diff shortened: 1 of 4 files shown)


src/1/boolLib.sml
```diff
@@ -110,33 +110,57 @@ end (* local open *)
...
+end
+
+fun has_suspended_subgoals th =
+    isSome (HOLset.find (isSome o slab_lab_var) (Thm.hypset th))
+
+(* The label variable's name encodes the closure-variable count and
+   the user-visible label, and -- for slabs that have been stamped at
+   save-time -- the owning theorem (thy, name).  Two formats:
+
+   Legacy: "N name"
```
xrchz, line 132: are these ever used? Why do we need to support them? We can assume users will rebuild all their theories with a new version of HOL.

xrchz:
I worry that the design (and implementation) of this feature might be too complicated to be worthwhile, for minimal benefit. Could I see a clear summary of the intended design? And an assessment of whether this is worthwhile vs just restricting the use of unfinalised theorems in general. I guess I'm also confused/concerned about what happens when unfinalised theorems are used in ordinary proofs (that don't involve further suspend/Resume) and then finalised much later (do the hypotheses retroactively disappear somehow?). In any case if this change is going to go ahead I think it should also include some documentation describing the design at a high level regarding how unfinished theorems can be used safely.

mn200:
Right.  I suspect that any use of a suspended theorem should be forbidden with a failure as a theory exits.

xrchz:
Can we change this PR to enforce (and document) that instead?

mn200:
Closing this as not the right way forward.  See #1963.

PR #1973, opened 2026-05-15
Disable recoding in tracing kernel

dnezam:
The fallback for recode32_path seems off, as it makes no use of Systeml.HOLDIR. On my system, this results in bool getting stuck and complaining about a file not being found. If I make the path use Systeml.HOLDIR, bool seems to get stuck as well (albeit without a "file not found" error).

Disabling recoding allows the tracer to make some progress.


dnezam:
Fixed on the recode32 branch, so this is unnecessary.

PR #1981, opened 2026-05-17
markerLib: forbid theorems consuming still-suspended theorems

### LINKED ISSUE #1963: Cross-theorem Resume dependencies: Finalise can't resolve foreign-parent slabs

mn200:
## Summary

- A theorem may now only consume already-finalised theorems. The only legal consumer of a still-suspended `X` is `Resume X[label]` for one of its unsolved subgoals.
- Cross-theorem citation (#1963) and self-citation of the parent in a Resume body (#1960) are rejected at the offending line — `boolLib.save_thm_attrs` catches the former, `markerLib.resume` catches both.
- An `ExportTheory` hook in markerLib acts as a safety net for any save path that bypasses `save_thm_attrs`.

This is the alternative design discussed in #1968 (replaces the find_res_for / owner-stamping routing that PR attempted).

## Test plan

- [x] `bin/build cleanAll && bin/build --no-cache --seq=tools/sequences/upto-parallel` — clean rebuild succeeds.
- [x] `src/marker/selftest.exe` — 20/20 OK.
- [x] `src/basicProof/theory_tests` — all 12 suspended-theorem regressions pass (suspCrossThm, suspCycle, suspCross{A,B}, suspSib{A,B,C,D}, suspNested, suspFast, suspRerun, suspTest).
- [x] `src/boss/theory_tests/gh1878Theory` — OK.

Closes #1963

### DIFF
--- a/src/1/boolLib.sml
+++ b/src/1/boolLib.sml
@@ -158,11 +158,11 @@ fun get_suspended_names th =
       HOLset.foldl foldthis [] (Thm.hypset th)
     end
 
-(* Shim for markerLib to install a recorder that routes suspended
-   theorems into the suspension.theorems AncestryData store.  This
-   module (boolLib) sits below markerLib in the dependency graph, so
-   we cannot call into it directly; markerLib populates the ref when
-   it loads.  Same pattern as term_pp.casesplit_munger. *)
+(* Shims for markerLib to install hooks that route suspended theorems
+   into, and consult, the suspension.theorems AncestryData store.
+   This module (boolLib) sits below markerLib in the dependency
+   graph, so we cannot call into it directly; markerLib populates the
+   refs when it loads.  Same pattern as term_pp.casesplit_munger. *)
 val suspended_theorem_recorder : (string * thm -> unit) ref =
     ref (fn _ => ())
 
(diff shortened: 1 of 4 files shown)


mn200:
Force push was to bring get the code merge-able.

xrchz (approved):
design sounds reasonable to me, and implementation looks fine (I haven't checked every last line)

Issue #1980, opened 2026-05-17
Holmake product cache silently returns stale Theory.dat when transitive parent hashes aren't in the cachekey


mn200:
Holmake's `--rebuild=cachekey` strategy can produce **silently unsound
cache hits**: a theory script that opens a library SML structure
(rather than its parent theory directly) computes a cachekey that
doesn't include the parent's `.dat` hash. After the parent has been
rebuilt with different content (e.g. as part of a kernel rebuild),
the cache returns the stale `.dat` and any downstream script that
opens the cached theory fails `link_parents` against the now-current
parent.

## Reproducer

A standalone regression test that fires this 100% of the time:

- Branch: [`claude/holmake-cachekey-transitive`](https://github.com/HOL-Theorem-Prover/HOL/tree/claude/holmake-cachekey-transitive)
- Commit: [`c10e6a3b5`](https://github.com/HOL-Theorem-Prover/HOL/commit/c10e6a3b5) — *Add regression test for Holmake cachekey transitive parents*
- Path: `tools/Holmake/tests/cachekey_transitive_parent/`

```
$ bin/Holmake selftest.exe && ./selftest.exe
cachekey_transitive_parent reproducer ...                                    FAILED!
Bug reproduced.
  Stale child.dat fetched from cache; consume's link_parents check
  failed against current parent.  This IS the bug under test
  (expected until fixed).
  ...
  <<HOL message: link_parents: the following parents of "child"
    should already be in the theory graph (but aren't):
    ("parent",5e72c13c6a786a747586b9ce1b109030ea59d63d)>>
```

The test is **not yet wired into**
`tools/Holmake/tests/parallel_tests/Holmakefile`'s `DIRNAMES` — adding
it would break `bin/build -t` until the underlying bug is fixed.

## How the bug arises

Setup:

```
inner/parentScript.sml         (* defines parentTheory             *)
inner/parentLib.sml/.sig       (* SML structure -- opens parentTheory *)
outer/childScript.sml          (* open parentLib (NOT parentTheory) *)
```

`childScript`'s direct deps are `[HolKernel.uo, Parse.uo, boolLib.uo,
parentLib.uo]`. None of those map to `UO(Theory _)` via the
type-driven discrimination in `HM_Cachekey.dep_to_hashable`
(`tools/Holmake/core/HM_Cachekey.sml:27-38`), and Holmake's
transitive-closure walk in `collect_all_dependencies`
[...]

PR #1982, opened 2026-05-17
Holmake: validate Theory.dat parent hashes at cache fetch time

mn200:
Proposed solution for #1980.

After fetching a cached `Theory.dat`, parse its recorded parent
hashes and verify each against the current on-disk parent.  On
mismatch, treat as a cache miss and rebuild locally; on match,
behave as before.  Implements the "Layer 1 — fetch-time parent-hash
validation" sketch from the issue.

The cachekey-computation gaps described in the issue (Layer 2)
remain in place; they're now an efficiency concern (false-miss
cache lookups) rather than a soundness concern, so this PR doesn't
chase them.

Current-parent lookup uses the dep graph's known directories first
(so downstream projects with their own theory hierarchies outside
core HOL's sigobj are covered), falling back to sigobj's symlinked
location.

Wires the `cachekey_transitive_parent` regression test (added in
the first commit) into `parallel_tests/Holmakefile` and updates its
README.  Existing cache selftests (`cachefetch`, `cachekey`,
`rebuild_cachekey`) continue to pass; real cache hits still
short-circuit the build.

May be the right resolution for #1980; opening for review.

### DIFF
--- a/tools/Holmake/poly/BuildCommand.sml
+++ b/tools/Holmake/poly/BuildCommand.sml
@@ -372,11 +372,37 @@ fun make_build_command (buildinfo : HM_Cline.t buildinfo_t) = let
                BuiltInCmd (BIC_BuildScript script_part, empty_incinfo))
               (* incinfos not consulted for comparison so empty value ok here *)
         end
+        (* Directories where parent Theory.dat files might live --
+           every directory that has appeared as a target or dep in the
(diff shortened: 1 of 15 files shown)


mn200:
@mrichards30 : happy to hear from you on this too (I'd make you a "reviewer" but you've not yet accepted invite to be a HOL contributor, which seems to be a prerequisite for that).

PR #1985, opened 2026-05-18
[lambda] The "cbpv-reasonable" work by Zhuo Zoey Chen et al.

binghe:
Code taken (with minor changes for `expk` build, modern Theory syntax, etc.) from:
https://github.com/ZhuoZoeyChen/cbpv-reasonable-HOL

Related Paper:
Zhuo Zoey Chen, Johannes Åman Pohjola, and Christine Rizkallah. A Verified Cost Model for Call-By-Push-Value. In 16th International Conference on Interactive Theorem Proving (ITP 2025). Leibniz International Proceedings in Informatics (LIPIcs), Volume 352, pp. 7:1-7:19, Schloss Dagstuhl – Leibniz-Zentrum für Informatik (2025) https://doi.org/10.4230/LIPIcs.ITP.2025.7


mn200:
Looks good.  I see from the referenced PR that you made against Zoey's repo that you were originally suggesting she should be the one making the PR here.  @ZhuoZoeyChen would you like do that?  (Or have you already discussed this with @binghe ?)

binghe:
I've sent that PR almost 8 months ago and didn't get any response. (There's absolutely no other private communications.) What I ever "suggested" was my personal wish and has no legal basis (I didn't see anyone such value my suggestions before.)

Note that the involved code were related under BSD 3-Clause License. Redistribution of these code in HOL, even without explicit permission from the original author, are still permitted by that license (all mentioned conditions are met.)

P. S. I do this only becase I need these code buildable with other lambda stuff inside HOL, (especially) under `expk` kernel, for understanding the related theory, possibly making changes, improvements, additions in the future. The fact that these code may now live with HOL code base, therefore always buildable and available to others who may be interested in, seems just side effects that the original author may not care about.

mn200:
I agree that this is totally "permitted".  I thought I'd just check also about the author's opinion. But yeah, if the repo is an orphan, it would be nice to give it a new home.

PR #1970, opened 2026-05-15
Multi-term "Boehm out" technique revisited (vsubterm_equivalent_lemma)

binghe:
Hi,

This PR follows #1947, adding an important lemma for multi-term "Boehm out" technique, which now works for λ-terms of arbitrary Böhm trees: (Recall that the key for β-separation (my idea) is to construct Böhm constructions which can manipulate (eventually, separate) multiple terms whose sets of Böhm tree paths are not identical, i.e. arbitrary.)
```
[vsubterm_equivalent_lemma] (separabilityTheory, new)
⊢ ∀X Ms p r pi.
    FINITE X ∧ p ≠ [] ∧ 0 < r ∧ BIGUNION (IMAGE FV (set Ms)) ⊆ X ∪ RANK r ∧
    pi = Boehm_construction' X Ms p ∧ EVERY solvable Ms ⇒
    (∀M. MEM M Ms ⇒ is_ready' (apply pi M)) ∧
    (∀q M.
       MEM M Ms ∧ q ≼ p ∧ vsubterm X M q r ≠ NONE ⇒
       vsubterm X (apply pi M) q r ≠ NONE ∧
       (solvable (vsubterm' X M q r) ⇔ solvable (vsubterm' X (apply pi M) q r))) ∧
    ∀q M N.
      MEM M Ms ∧ MEM N Ms ∧ q ≼ p ∧ vsubterm X M q r ≠ NONE ∧
      vsubterm X N q r ≠ NONE ⇒
      (equivalent (vsubterm' X M q r) (vsubterm' X N q r) ⇔
       equivalent (vsubterm' X (apply pi M) q r)
         (vsubterm' X (apply pi N) q r))
```
Note that, in the above lemma, which now involves virtual subterms (`vsubterm`) instead of `subterm`, the **only** requirement for the Βöhm tree path `p`, is `p ≠ []` (non-root). This is a huge improvements over the previous `subtree_equiv_lemma` which requires `p` must be valid path of all Βöhm trees of involved terms (`EVERY (λM. p ∈ ltree_paths (BT' X M r)) Ms`):
```
[subtree_equiv_lemma] (lameta_completeTheory, existing)
⊢ ∀X Ms p r.
    FINITE X ∧ p ≠ [] ∧ 0 < r ∧ Ms ≠ [] ∧
    BIGUNION (IMAGE FV (set Ms)) ⊆ X ∪ RANK r ∧
    EVERY (λM. p ∈ ltree_paths (BT' X M r)) Ms ⇒
[...]

PR #1986, opened 2026-05-18
[lambda] The new "agree upto" theorem (vsubterm_agree_upto_thm)

binghe:
Hi,

This PR follows #1970 and further proved a new "agree upto" theorem with no antecedents on the involved Böhm tree path:
```
[vsubterm_agree_upto_thm]
⊢ ∀X Ms p r.
    FINITE X ∧ 0 < r ∧ (∀M. MEM M Ms ⇒ FV M ⊆ X ∪ RANK r) ∧
    vsubterm_agree_upto X Ms p r ⇒
    ∃pi. Boehm_transform pi ∧ vsubterm_faithful p X Ms pi r
```
It will proved (soon) that the Böhm transformation asserted here can separate arbitrary two λ-terms who has bnf and not βη-equivalent.

--Chun

### DIFF
--- a/examples/lambda/barendregt/boehmScript.sml
+++ b/examples/lambda/barendregt/boehmScript.sml
@@ -2803,6 +2803,37 @@ Proof
  >> Q.EXISTS_TAC ‘q’ >> rw []
 QED
 
+Theorem subterm_width_of_unsolvables :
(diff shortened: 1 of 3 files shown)


mn200:
Thanks!

PR #1987, opened 2026-05-18
GitHub actions version upgrade (to Node.js 24)

binghe:
Hi,

You may have already seen those "Node.js 20 actions are deprecated." warnings when doing CI builds:

- "Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, docker/build-push-action@v5, docker/setup-buildx-action@v3. Actions will be forced to run with Node.js 24 by default starting June 2nd, 2026. Node.js 20 will be removed from the runner on September 16th, 2026. Please check if updated versions of these actions are available that support Node.js 24. To opt into Node.js 24 now, set the FORCE_JAVASCRIPT_ACTIONS_TO_NODE24=true environment variable on the runner or in your workflow file. Once Node.js 24 becomes the default, you can temporarily opt out by setting ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/"

The solution is to boost all GitHub actions to their next version (i.e. increase the version number by exact one), which supports Node.js 24. That's what this PR does. By greate backward compatibility (of joint efforts between GitHub and Docker), there's no need to change any parameter when calling those actions.

Please squash all commits (because it's only possible to edit the related workflow files one by one using GitHub's Web-based online editor.)

--Chun

### DIFF
--- a/.github/workflows/docker-ci-for-pr.yml
+++ b/.github/workflows/docker-ci-for-pr.yml
@@ -11,13 +11,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@v4
 
       - name: Build and push
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v6
         with:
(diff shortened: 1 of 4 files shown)


mn200:
Thanks for all your work on keeping this technology working!

PR #1990, opened 2026-05-19
[lambda] Böhm's Theorem (beta-separability)

binghe:
Hi,

Today is the day I finally formalised Böhm's β-separability theorem: Two terms `M` and `N` having βη-nf, if they are not βη-equivalent, then there's a Böhm transformation _separating_ them in the sense of only β-equivalence: (our previous ITP paper work has to do with βη-equivalence, i.e. a weaker result)
```
[beta_separability_final]
⊢ ∀M N.
    has_benf M ∧ has_benf N ∧ M =/== N ⇒
    ∀P Q. ∃pi. Boehm_transform pi ∧ apply pi M == P ∧ apply pi N == Q
```
If, additionally, `M` and `N` are closed term, then the asserted Böhm transformation reduces to just a list of terms appending to `M` and `N`:
```
[closed_beta_separability]
⊢ ∀M N.
    has_benf M ∧ has_benf N ∧ M =/== N ∧ closed M ∧ closed N ⇒
    ∀P Q. ∃L. M ·· L == P ∧ N ·· L == Q
```
If we understood `M` and `N` as two programs (or computable functions), the above theorem roughly says that if they are _syntatically different_, then they are also _semantically different_: there exists always a special input leading to different outputs from the two functions. This property of (untyped) λ-calculus as computation model, seems not shared by other models such as Turning machine.

A further corollary is the following results: two βη-"distinct" terms are also incompatible, i.e. they are inconsistent in the vanilla λ-theory (having only β-rule):
```
distinct_benf_imp_incompatible
⊢ ∀M N. has_benf M ∧ has_benf N ∧ M =/== N ⇒ M # N
```
It's easy to see that the previous proved λη-completeness result is an easy corollary, because inconsistence in *λ* implies inconsistence in *λη* (or any other λ-theories with arbitrary extra equations).

What's remaining here, is the optional work to prove the previous huge `[subtree_equiv_lemma]` as a corollary of the new `[vsubterm_equivalent_lemma]`, removing ~4k lines of proof scripts from the code base. Besides, I feel the `==` (β-equivalence) in all above theorems can be sharpen to `-b->*` (β-reduction) with some minor proof changes.

--Chun

### DIFF
--- a/examples/lambda/barendregt/lameta_completeScript.sml
+++ b/examples/lambda/barendregt/lameta_completeScript.sml
@@ -4453,6 +4453,16 @@ Definition equivalent_def :
            ~solvable M /\ ~solvable N
 End
 
+val _ = add_rule { block_style = (AroundEachPhrase, (PP.CONSISTENT, 0)),
+                   fixity = Infix (NONASSOC, 450),
+                   paren_style = OnlyIfNecessary,
+                   pp_elements = [HardSpace 1, TOK (UTF8.chr 0x2248),
+                                  BreakSpace (1,0)],
+                   term_name = "equivalent" };
+
+val _ = TeX_notation { hol = UTF8.chr 0x2248,
+                       TeX = ("\\HOLTokenWeakEQ", 1) };
+
 (* A more general definition (but many existing hard proofs are still
    using the above “equivalent”).
  *)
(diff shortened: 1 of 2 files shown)


mn200:
Hurray!  Please feel free to add to the release notes in the “new theories” section when you are ready to announce a finished product to the world.

PR #1991, opened 2026-05-19
vim-mode: hG opens a Resume sub-goal via markerLib.set_suspended_goal

tanyongkiam:
Previously `hG` only handled `Theorem name: ... Proof[...] ... QED` blocks (wrapping the region in `proofManagerLib.new_goalstack`). When the yanked region is a `Resume <thm>[<label>]:` header, set the suspended sub-goal on the proof manager via `markerLib.set_suspended_goal` instead — the same entry point Holmake uses, so the interactive goal matches the one the Resume body will be checked against.


xrchz (approved):

mn200:
Cool

PR #1994, opened 2026-05-20
Strong version of Böhm's separation theorem

binghe:
Hi,

This PR follows #1990 and improved the previous proved `[beta_separability_final]` from `==` (`lameq`) to `-b->*` (β-reduction):
```
[beta_separability_strong_final]
⊢ ∀M N.
    has_benf M ∧ has_benf N ∧ M =/== N ⇒
    ∀P Q. ∃pi. Boehm_transform pi ∧ apply pi M -β->* P ∧ apply pi N -β->* Q
```

Note that some versions of Böhm's Theorem indeed have β-reductions in the statements (Barendregt's book, however, doesn't use this form). This change is not hard but involves a few new lemmas about `-b->*`.

Now I mentioned those new `lambda` additions in next release notes.

--Chun


binghe:
In `semi_sensibleTheory`, previous there's the following definition of `separable` for a list of λ-terms:
```
[separable_def]
⊢ ∀Ms.
    separable Ms ⇔
    ∀Ns.
      LENGTH Ns = LENGTH Ms ⇒
      ∃c. ctxt c ∧ ∀i. i < LENGTH Ms ⇒ c (EL i Ms) == EL i Ns
```
Now I have proved another form of Böhm's separation theorem based on this `separable` definition:
```
[separable_thm]
⊢ ∀M N. has_benf M ∧ has_benf N ∧ M =/== N ⇒ separable [M; N]
```
Then, the question of separating a list of more than two λ-terms, is not trivial (a possible future work).

binghe:
Now I've removed the previous huge proof of `[subtree_equiv_lemma]` and re-proved it by the new `[vsubterm_equivalent_lemma]`. This removed over 4k lines (of similar proofs) from the code base, which is stable now. Now `lameta_complete` depends on the new `separability` theory (with `equivalent`-related stuff moved in), which depends on `boehm` theory.

Note that I didn't remove the old proof idea of the completeness result (in `lameta_completeScript.sml`, which is only 900+ lines now), so that all the theorems mentioned in our previous ITP paper are still there.

mn200:
Fantastic outcomes!

PR #2000, opened 2026-05-26
[extreal] contract and expand operations

binghe:
Hi,

I propose adding the following two operations (`contract` and `expand`) into `extreal_baseTheory`, which maps all extreals to/from reals in `[-1,1]`, preserving their order. It's learnt from [1].

```
[contract_def]
⊢ (∀r. contract (Normal r) = r / (1 + abs r)) ∧ contract +∞ = 1 ∧
  contract −∞ = -1

[expand_def]
⊢ ∀c. expand c =
      if 1 ≤ c then +∞ else if c ≤ -1 then −∞ else Normal (c / (1 − abs c))

[contract_11]
⊢ ∀x y. contract x = contract y ⇔ x = y

[contract_le_eq]
⊢ ∀x y. contract x ≤ contract y ⇔ x ≤ y

[expand_contract]
⊢ ∀x. expand (contract x) = x
```

For their usefulness, these operations may be used to store extreals into the `:real` slot of some data structures, preserving all information. I was trying to revise the metric definition for `extreals` using them, but then the plan didn't succeed. I consider these proofs may be useful in other (unknown) places...

--Chun

[1] Affeldt, R., Cohen, C.: Measure Construction by Extension in Dependent Type Theory with Application to Integration. Journal of Automated Reasoning. 67, 43 (2022).

### DIFF
--- a/src/real/extreal_baseScript.sml
+++ b/src/real/extreal_baseScript.sml
@@ -4092,3 +4092,196 @@ Proof
                 >> RW_TAC std_ss [] ))
   >> METIS_TAC [INSERT_SING_UNION,UNION_COMM]
 QED
+
+(* ------------------------------------------------------------------------- *)
+(*  Contract and expand (extreal to/from real, order-preserving)             *)
+(* ------------------------------------------------------------------------- *)
+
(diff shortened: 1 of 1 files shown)


mn200:
Thanks!

PR #2002, opened 2026-05-26
Make JSON lexer reject invalid numbers

### LINKED ISSUE #1998: parse_json lex function can lex invalid json and return unintended output on some valid json
I have been working a bit with the json lexer from [parse_jsonTheory](https://github.com/HOL-Theorem-Prover/HOL/blob/develop/examples/formal-languages/json/parse_jsonScript.sml) recently and I have noticed a few bugs specifically to do with number lexing.

Also, in my testing I abbreviated ``Positive`` to ``POS`` and ``Negative`` to ``NEG``:
```
Overload POS = “Positive”
Overload NEG = “Negative”
```

### Leading Zeros in Integer Component
By [ECMA-404](https://www.ecma-international.org/publications-and-standards/standards/ecma-404/) the integer component of the number should not be allowed to have leading zeros, but the lexer is able to lex numbers with this property as valid:
```
> EVAL ``lex "001 " []``;
val it = ⊢ lex "001 " [] = INL [NUM ((POS,1),NONE,NONE)]: thm
```
This is due to the main number lexing function, ``lex_sci``, calling ``lex_int`` to lex the integer part which then calls ``lex_num`` after lexing the negative sign if there is one. This is a problem because ``lex_num`` looks like this:
```
Definition lex_num_def:
  lex_num [] acc = (acc,[]) /\
  (lex_num (c::cs) acc =
    if isDigit c then
      lex_num cs (acc * 10 + (ORD c - ORD #"0"))
    else (acc, c::cs))
End
```
So it just end up ignoring any leading zeros and returns the numeric value. 
This is probably the behaviour wanted from ``lex_num`` but there might need to be a check in ``lex_int`` to prevent leading zeros, though that does mean that a different function would have to be written for lexing the integer component of the exponent, as that does allow for leading zeros.

### Lexing Empty Numbers
The ``lex_num`` function has the property that if it does not consume any characters (this happens if it immediately runs into a non-digit character) it returns the ``acc`` it entered with. This entry ``acc`` is always ``0`` in the code, but this causes some problems. Specifically, most functions that use ``lex_num`` are not programmed to tell whether ``lex_num`` returning ``0`` is due to the number actually being ``0`` or ``lex_num`` consuming no characters. There is a check specifically in ``lex`` to make sure the ``lex_sci`` consumes characters, but this check needs to be everywhere ``lex_num`` is used, because it is possible to write numbers that do consume characters but are still invalid json. Due to ``lex_frac`` and ``lex_exp`` not being able to tell whether characters are consumed, we can write fractions and exponents with no numbers, and the functions will just assume the numbers are ``0``:
```
> EVAL ``lex ". " []``;
val it = ⊢ lex ". " [] = INL [NUM ((POS,0),NONE,NONE)]: thm
```
And
```
> EVAL ``lex "E " []``;
val it = ⊢ lex "E " [] = INL [NUM ((POS,0),NONE,SOME (POS,0))]: thm
```
And
```
> EVAL ``lex ".E- " []``;
val it = ⊢ lex ".E- " [] = INL [NUM ((POS,0),NONE,SOME (NEG,0))]: thm
```
None of the above are valid json (the decimal component is also recorded as ``NONE``, due to ``lex_leading_zeroes`` not finding anything after the decimal. While this is correct, the decimal component is nonexistent, the lexer probably needs to exit in an error state once there is both a dot, but the fractional component is nonexistent).

### Decimals with Decimal Component Equal to 0 at the End of the String
I want to start by saying, these strings are valid and should be lexed, but there is inconsistency in how they are lexed depending on whether they are at the end of the string. When the numbers are not at the end of the string they lex as intended:
```
> EVAL ``lex "0.0 " []``;
val it = ⊢ lex "0.0 " [] = INL [NUM ((POS,0),SOME (0,0),NONE)]: thm
```
However, when they are at the end of the string:
```
> EVAL ``lex "0.0" []``;
val it = ⊢ lex "0.0" [] = INL [NUM ((POS,0),NONE,NONE)]: thm
```
The decimal component returns as ``NONE``. This seems to be due to ``lex_leading_zeroes``:
```
Definition lex_leading_zeroes_def:
  lex_leading_zeroes [] (acc:num) = (NONE, []) /\
  (lex_leading_zeroes (c::cs) acc =
    if c = #"0"
    then lex_leading_zeroes cs (acc+1)
    else
      (( \ (a, b). if a = 0 then (if acc > 0 then (SOME (acc-1, a), b) else (NONE, b)) else (SOME (acc, a), b)) (lex_num (c::cs) 0)))
End
```
If the number is not at the end of the string, it takes the ``c::cs`` case and returns properly, but if the number is at the end of the string then it takes the base case and exits early, despite the ``acc`` being non-zero.

I also found a similar problem with ending the string directly after ``E`` or ``e`` and the exponent returning ``NONE`` instead of ``SOME (POS,0)``, though this should not really be a problem if the previous issue is solved, as this only seems to happen when the exponent is empty.

didriklundberg:
This PR fixes HOL-Theorem-Prover/HOL#1998, containing a number of problems pointed out by @maxwell-zhou-smith. In short, these concern invalid JSON numbers which were not properly rejected by the lexer.

A few examples that can be tried out interactively:

```
(* GOOD *)
EVAL ``lex "0.0 " []``;
EVAL ``lex "0.00" []``;
EVAL ``lex "0.003" []``;
EVAL ``lex "10E+002" []``;
EVAL ``lex "0E+002" []``;
EVAL ``lex "0E-002" []``;

(* BAD *)
EVAL ``lex "0." []``;
EVAL ``lex "0.  " []``;
EVAL ``lex "." []``;
EVAL ``lex "10E+-002" []``;
EVAL ``lex "00E+002" []``;
EVAL ``lex "002" []``;
EVAL ``lex "E  " []``;
EVAL ``lex ".E- " []``;
```

In addition to what was pointed out by @maxwell-zhou-smith, this also disallows lexing of numbers that have both `+` and `-` after the `E` or `e` signifying exponentiation.

### DIFF
--- a/examples/formal-languages/json/parse_jsonScript.sml
+++ b/examples/formal-languages/json/parse_jsonScript.sml
@@ -372,7 +372,7 @@ End
 
 
 Definition lex_leading_zeroes_def:
-  lex_leading_zeroes [] (acc:num) = (NONE, []) /\
+  lex_leading_zeroes [] (acc:num) = (if acc > 0 then SOME (acc-1, 0) else NONE, []) /\
   (lex_leading_zeroes (c::cs) acc =
     if c = #"0"
     then lex_leading_zeroes cs (acc+1)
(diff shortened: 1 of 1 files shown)


maxwell-zhou-smith:
Correct me if I am wrong, but in ``lex_plus_def``:
```
Definition lex_plus_def:
  lex_plus [] = SOME (NONE, []) /\
  (lex_plus (c::cs) =
    if c = #"+" then
      if is_next_digit cs then
        let (num, cs') = lex_num cs 0 in
        SOME (SOME (Positive, num), cs')
      else NONE
    else
      if isDigit c \/ c = #"-" then
        (case lex_int (c::cs) of
         | SOME (int, cs') => SOME (SOME int, cs')
         | NONE => SOME (NONE, c::cs))
      else NONE)
End
```
Don't we want to be returning ``NONE`` if ``lex_int`` comes back as ``NONE``, because lex_int should only really come back as ``NONE`` there if we have a minus sign with no integer component after, and that results in an invalid json number.

I did do a quick test, though I am very bad at using github so my setup may be faulty, but I got:
```
> EVAL ``lex_sci "1E-"``;
val it = ⊢ lex_sci "1E-" = SOME (((Positive,1),NONE,NONE),"-"): thm
```

mn200:
It looks as if the CI failure is spurious, but perhaps there's a bit more to be done here.

didriklundberg:
> Correct me if I am wrong, but in `lex_plus_def`:
> 
> ```
> Definition lex_plus_def:
>   lex_plus [] = SOME (NONE, []) /\
>   (lex_plus (c::cs) =
>     if c = #"+" then
>       if is_next_digit cs then
>         let (num, cs') = lex_num cs 0 in
>         SOME (SOME (Positive, num), cs')
>       else NONE
>     else
>       if isDigit c \/ c = #"-" then
>         (case lex_int (c::cs) of
>          | SOME (int, cs') => SOME (SOME int, cs')
>          | NONE => SOME (NONE, c::cs))
>       else NONE)
> End
> ```
> 
> Don't we want to be returning `NONE` if `lex_int` comes back as `NONE`, because lex_int should only really come back as `NONE` there if we have a minus sign with no integer component after, and that results in an invalid json number.
> 
> I did do a quick test, though I am very bad at using github so my setup may be faulty, but I got:
> 
> ```
> > EVAL ``lex_sci "1E-"``;
> val it = ⊢ lex_sci "1E-" = SOME (((Positive,1),NONE,NONE),"-"): thm
> ```

Ultimately, we want to be using the top-level `lex` function. We get the following result:

```
lex "1E-" [] = INR "unexpected characters: -"
```

So "1E-" is indeed rejected, as it should be. However, I agree that ideally, the other functions should also make sense in isolation. Your suggested fix definitely makes for a better design overall, I pushed it in commit [05fdf21](https://github.com/didriklundberg/HOL/commit/05fdf21110f689e6bcb32cfd4bdd5a785f26a4cd).

maxwell-zhou-smith:
Ah, yes that does make sense, I failed to consider what would happen upon the lexer recursing on the minus sign.

The fix looks great though, thanks @didriklundberg!

mn200:
Thanks both!

PR #2003, opened 2026-05-28
Add JSON parser validation

didriklundberg:
I've added `validate.sh` and `validate.sml`, which make it easier to check whether the JSON parser properly accepts and rejects input using the [JSONTestSuite](https://github.com/nst/JSONTestSuite) (~300 test files). This could be useful for regression tests when modifying the parser.

Note that this does not validate whether the accepted input is actually parsed to a sensible format. It could be interesting to perform some cross-validation and look into the tests in `test_transform`.


examples/formal-languages/json/validation/validate.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -euo pipefail
+
+cd "$(dirname "$0")"
+
+HOLMAKE="../../../../bin/Holmake"
+
+if [ ! -d "JSONTestSuite/test_parsing" ]; then
```
mn200: This should be done with a `.hol_preexec` file; see the example of what's done in examples/l3-machine-code/arm8/asl-equiv/.hol_preexec

didriklundberg: Done!

mn200:
Thanks!

PR #2005, opened 2026-06-01
Added non-zero-mean version of Central Limit Theorem

KaiPhan:
This PR adds a non-zero-mean version of the Lyapunov Central Limit Theorem.

The existing zero-mean theorem is kept unchanged. The new theorem is proved by applying the existing result to the centred sequence

   ` Y i x = X i x - expectation p (X i)`

and then transferring convergence in distribution back to the original sequence.


mn200:
Thanks for this!  (The regression failures are due to unrelated mistakes in the generation of manuals.)

mn200:
In future, please do not make unrelated global edits turning Unicode into ASCII (or *vice versa*).

PR #2011, opened 2026-06-15
Remove stix package which globally changes TeX fonts

binghe:
Hi,

In HOL's provided `holtexbasic` TeX package, there's the use of `stix` font package. Its only purpose is to globally change the text (and math text) fonts, ranther than providing new TeX commands (therefore removing this package doesn't break anything.) Note that the `stix` font is more compact (and maybe also nicer) than the Computer Modern font (default of LaTeX `article` and many conference/journal templates), and as a side effect the paper writer may end up in putting more contents into limited page space...

Strictly speaking, changing fonts in LaTeX template when submitting papers, is a voilation of rules. The fact that in practice it didn't cause any problem, is (perhaps) because many people are so blind to see the font difference unless it's too obvious.

I suggest removing `stix` from `holtexbasic.sty`, and any user who still wants to "cheat" on fonts, should explicitly do it. (HOL official has nothing to do with it, after merging this PR.)

--Chun

### DIFF
--- a/src/TeX/holtexbasic.sty
+++ b/src/TeX/holtexbasic.sty
@@ -6,7 +6,7 @@
 % -------------------------------------------------------
 
 \NeedsTeXFormat{LaTeX2e}
-\usepackage{amssymb, underscore, fancyvrb, amsmath, stix}
+\usepackage{amssymb, underscore, fancyvrb, amsmath}
 
 \fvset{commandchars=\\\{\},xleftmargin=1mm,framesep=1.2mm,fontsize=\small}
 
--- a/src/TeX/holtexbasic.sty
+++ b/src/TeX/holtexbasic.sty
@@ -8,6 +8,10 @@
 \NeedsTeXFormat{LaTeX2e}
 \usepackage{amssymb, underscore, fancyvrb, amsmath}
 
+% Do not change the default text and math fonts (violating paper submission rules)
+% As of April 2018 the {stix} package is considered obsolete. Use {stix2} instead.
+\usepackage[notext,nomath]{stix2}
+
 \fvset{commandchars=\\\{\},xleftmargin=1mm,framesep=1.2mm,fontsize=\small}
 
 \newcommand{\HOLpagestyle}{}


mn200:
The `stix` package is there to support the tortoise shell brackets used for `EL`.  Dialogue with an AI suggests the following: 
```
% Borrow only the tortoise-shell brackets (U+2772/2773) from the STIX Type1
% extension font, without letting stix.sty take over the math fonts.
\DeclareFontEncoding{LS2}{}{}
\DeclareFontSubstitution{LS2}{stixex}{m}{n}
\DeclareSymbolFont{stixdelim}{LS2}{stixex}{m}{n}
\DeclareMathDelimiter{\lbrbrak}{\mathopen} {stixdelim}{"EE}{stixdelim}{"14}
\DeclareMathDelimiter{\rbrbrak}{\mathclose}{stixdelim}{"EF}{stixdelim}{"15}
```
should grab just those characters we need without contaminating everything else.

binghe:
Sorry, I should check `stix`'s documentation first to know its real functionality.  Then perhaps we should just not let the `stix` package to change default fonts (it changes not only math fonts but also text fonts) by adding some options, e.g. `[notext,nomath]`:
```
notext Do not change the default text fonts.
nomath Do not change the default math fonts.
not1 Do not change the default font encoding to T1.
notextcomp Do not load the textcomp package
...
```
On the other hand, CTAN [1] says "As of April 2018 this (stix) package is considered obsolete. See [stix2-otf](https://ctan.org/pkg/stix2-otf) and [stix2-type1](https://ctan.org/pkg/stix2-type1) instead."  `stix2` also provides the same symbols for tortoised brackets. Better to switch to it.

[1] https://ctan.org/pkg/stix

mn200:
Unfortunately, if you use `nomath` you then lose the ability to get `\lbrbrak`...

binghe:
Alright, I close this PR and leave the issue to you. (I don't use the new TeX representation of `EL` and thus can safely remove `stix` locally.)

PR #2012, opened 2026-06-23
CLT fixes (due to wrong definitions)

binghe:
Hi,

In current `central_limitTheory`, the definition of `absolute_third_moment` is (unfortunately) wrong. It should be a central moment (having `expectation p X` as the center, instead of 0):
```
[absolute_third_moment_def] (old)
    ⊢ ∀p X. absolute_third_moment p X = absolute_moment p X 0 3

[absolute_third_moment_def] (new)
    ⊢ ∀p X. absolute_third_moment p X = absolute_moment p X (expectation p X) 3
```
The old and new definitions are equivalent in the staged CLT results where all involved random variables (r.v.'s) have zero expectations, i.e. `expectation p (X n) = 0`. But in the final, more general version of CLT where the involved r.v.'s may have non-zero expectations, the wrong definition `absolute_third_moment` will lead to a much harder proof of CLT (it surprisingly can still be proved), whose overall statements are ...just wrong.

This PR fixes the wrong definition, and then greately simplified the recently added proof of `[CLT_Lyapunov]` (the final non-zero-mean version). Unused lemmas (and two definitions) are removed.  This fix is *urgent* because it seems that next HOL releases may come any time from now on.

P. S. Taking this chance, I also added a little new contents into `semi_sensibleTheory` of lambda examples (variants of the separability results)

--Chun

### DIFF
--- a/examples/lambda/barendregt/semi_sensibleScript.sml
+++ b/examples/lambda/barendregt/semi_sensibleScript.sml
@@ -136,7 +136,18 @@ Theorem separable_def =
         gen_separable_alt |> Q.SPEC ‘UNCURRY (==)’
                           |> SRULE [lambdathy_lameq]
 
-Theorem separable_incompatible :
+(* |- !Ms.
+        EVERY closed Ms ==>
+        (separable Ms <=>
+         !Ns.
+           LENGTH Ns = LENGTH Ms ==>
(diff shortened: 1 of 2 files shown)


mn200:
Thank you for your eagle eyes!

PR #2014, opened 2026-06-27
Add reduced build sequence up to bin/hol

xrchz:
Factor the prefix of tools/sequences/core-theories that builds the standard HOL executable into a new base-hol sequence.  This split occurs at the existing documented boundary before the Poly-only AI/tactictoe and holyhammer tail.

Add upto-hol as kernel plus base-hol, analogous to upto-parallel but stopping once bin/hol has been built.  This gives users and downstream tooling a supported reduced sequence for building enough of HOL to run the standard executable and load bossLib, without building later optional Poly-only components.

Motivated by https://github.com/HOL-Theorem-Prover/holbuild/issues/74


mn200:
Sure thing.  If you fancy it, you might add a flag (maybe `--core`) to `build` to alias the `--seq=...` option.

xrchz:
I'm fine just using seq

PR #2019, opened 2026-07-03
Changed lex_plus empty list case to return NONE

### LINKED ISSUE #2016: parse_json$lex can lex empty exponents at the end of strings
The `lex` function in [`parse_json`](https://github.com/HOL-Theorem-Prover/HOL/blob/develop/examples/formal-languages/json/parse_jsonScript.sml) can lex numbers with empty exponents only at the end of the string, e.g.

```
> EVAL ``lex "1E" []``;
val it = ⊢ lex "1E" [] = INL [NUM ((POS,1),NONE,NONE)]: thm
```
But
```
> EVAL ``lex "1E " []``;
val it = ⊢ lex "1E " [] = INR "unexpected characters: 1E ": thm
```
This seems to be because of the empty list case on the `lex_plus` function:
```
Definition lex_plus_def:
  lex_plus [] = SOME (NONE, []) /\
  (lex_plus (c::cs) =
    if c = #"+" then
      if is_next_digit cs then
        let (num, cs') = lex_num cs 0 in
        SOME (SOME (Positive, num), cs')
      else NONE
    else
      if isDigit c \/ c = #"-" then
        (case lex_int (c::cs) of
         | SOME (int, cs') => SOME (SOME int, cs')
         | NONE => NONE)
      else NONE)
End
```
Which returns `SOME (NONE, [])` causing the lexer to return a valid number with no exponent present, even though the `e` or `E` symbol was present.

maxwell-zhou-smith:
Fixes #2016.

Changing the base case of `lex_plus` to return `NONE` instead of `SOME (NONE, [])` should fix the issue, as we want to be returning `NONE` from `lex_plus` on the empty list as there is no exponent there to lex.
The example:
```
> EVAL ``lex "1E" []``;
val it = ⊢ lex "1E" [] = INL [NUM ((POS,1),NONE,NONE)]: thm
```
Now returns properly:
```
> EVAL ``lex "1E" []``;
val it = ⊢ lex "1E" [] = INR "unexpected characters: 1E": thm
```
As do other invalid json examples with the same issue:
```
EVAL ``lex "0.9230E" []``;
val it = ⊢ lex "0.9230E" [] = INR "unexpected characters: 0.9230E": thm
```
```
> EVAL ``lex "-1000.02E" []``;
val it = ⊢ lex "-1000.02E" [] = INR "unexpected characters: -1000.02E": thm
```
The change does not seem to affect any off the proofs in the file from my testing, so I do not think any other changes are necessary.

That being said, I would like a second opinion from @didriklundberg to see if he thinks these changes are good.

### DIFF
--- a/examples/formal-languages/json/parse_jsonScript.sml
+++ b/examples/formal-languages/json/parse_jsonScript.sml
@@ -407,7 +407,7 @@ Definition lex_frac_def:
 End
 
 Definition lex_plus_def:
-  lex_plus [] = SOME (NONE, []) /\
+  lex_plus [] = NONE /\
   (lex_plus (c::cs) =
     if c = #"+" then
       if is_next_digit cs then


didriklundberg:
This looks good at a glance, but I will take a closer look this evening (in 7-8 hours from now) when I can access a computer.

Did you re-do the validation according to the instructions in the README?

maxwell-zhou-smith:
I have, I got the following:
```
=== Summary ===
Expected:      318
Discrepancies: 0
Errors:        0
```

didriklundberg:
Yes, indeed this fix looks very good, and ready to merge!

maxwell-zhou-smith:
Great, thanks @didriklundberg!

mn200:
Thanks both!

PR #2035, opened 2026-08-10
Prove cv_compute selftest termination instead of cheating

lukaszcz:
The `exp_size_alt` example in `src/num/theories/cv_compute/automation/selftest.sml`
reached for `cheat` twice: once for the tactic argument of `tDefine`, and once for
the precondition goal handed to `cv_auto_trans_pre_rec`.

This PR discharges both obligations honestly:

  - **`tDefine`**: the definition is primitive recursive — as the comment sitting
    above it already remarks — so there is no termination argument to make, and
    `ALL_TAC` suffices.
  - **`cv_auto_trans_pre_rec`**: the precondition of the translated function is a
    termination argument over the sum type into which the two mutually recursive
    functions are packed, so a `measure` taking `cv_size` in either injection,
    followed by `cv_termination_tac`, proves it.

No change to any non-test source; the only effect is that the selftest now
actually exercises the translator's precondition machinery.


mn200:
Thanks!

Issue #2034, opened 2026-08-10
`computeLib`: a registered conversion returning `|- t = t` makes `CBV_CONV` loop forever — `REAL_RAT_REDUCE_CONV` hangs on rational literals via a `CHANGED_CONV` guard dropped in the HOL Light port


lukaszcz:
## Summary

Two related problems, one mechanism and one instance:

1. **Mechanism** (`computeLib`): a conversion registered with
   `computeLib.add_conv` counts as having made progress whenever it does not
   raise `HOL_ERR`. A conversion returning a reflexive theorem `|- t = t` is
   treated as a rewrite step, and `CBV_CONV` re-enters the same term forever.
   The failure mode is a hang with no diagnostic. The contract "a registered
   conversion must change the term or fail" is enforced by nothing and
   documented nowhere.

2. **Instance** (`RealField`): the public `REAL_RAT_REDUCE_CONV` hangs on any
   term containing a non-integral rational literal, e.g.
   `REAL_RAT_REDUCE_CONV ``&1 / &3 * x``` (verified). The cause is a
   `CHANGED_CONV` guard that upstream HOL Light has always had on this entry
   and that the HOL4 port dropped — see below.

## Minimal reproduction of the mechanism

`src/compute` only — no `real` theories involved. Save as `repro.sml` and run
`bin/hol run repro.sml` from the repo root. The counter exists only to turn the
hang into a printable message; remove it and the session hangs.

```sml
val calls = ref 0
fun refl_conv t =
  (calls := !calls + 1;
   if !calls > 500 then
     raise Fail ("DIVERGES: conv called " ^ Int.toString (!calls)
                 ^ " times on a 2-node term")
   else Thm.REFL t)

val cs = computeLib.add_conv (numSyntax.suc_tm, 1, refl_conv)
           (computeLib.new_compset [])

val r = Parse.thm_to_string (computeLib.CBV_CONV cs ``SUC (x:num)``)
        handle Fail m => m
val _ = print (r ^ "\n")

val _ = calls := 0
val cs' = computeLib.add_conv (numSyntax.suc_tm, 1,
                               Conv.CHANGED_CONV refl_conv)
            (computeLib.new_compset [])
val r' = Parse.thm_to_string (computeLib.CBV_CONV cs' ``SUC (x:num)``)
         handle Fail m => m
val _ = print ("with CHANGED_CONV guard: " ^ r' ^ "  [calls="
[...]

PR #2036, opened 2026-08-10
Stop REAL_RAT_REDUCE_CONV diverging on rational literals

lukaszcz:
## The bug

`REAL_RAT_DIV_CONV` rewrites `x / y` to `x * inv y` and back, so on a
literal already in lowest terms it returns `|- &1 / &3 = &1 / &3` —
success, no change.  `computeLib` re-evaluates whatever a registered
conversion hands back, so `CBV_CONV` went round that term forever:
`REAL_RAT_REDUCE_CONV` did not terminate on *any* term containing a
non-integral rational literal.

## The fix

`real_rat_compset` already guarded one entry — `REAL_RAT_NEG_CONV` —
with `CHANGED_CONV`.  Several of the others are no-ops on an
already-reduced argument in the same way, so the guard belongs on every
entry rather than on whichever operator's no-op was noticed first.  The
compset now adds all of them through a single `add_reduce_conv` that
applies it.

A registered conversion can report no progress in two ways: return a
reflexive theorem, or raise `UNCHANGED` (as `ALL_CONV`, hence
`TRY_CONV`, does).  Both are fatal — the first loops, the second escapes
`CBV_CONV` entirely, because `reduce_cst` catches only `HOL_ERR` when
deciding to move on to the next rule.  `Conv.CHANGED_CONV` covers both,
so it needs no `QCHANGED_CONV` around it; `QCHANGED_CONV` in its place
would be a regression, since it catches only the raise and lets the
diverging mode through.  A comment at the site records this so the
question is not reopened.

## Tests

`src/real/selftest.sml` gains a test that the conversion terminates on
`&1 / &3 * x`, plus a companion pinning that `&1 / &3 + &1 / &6` still
reduces to `&1 / &2` — so the termination guard cannot be satisfied by
simply disabling reduction.

## Scope

This is the instance half of #2034.  The underlying `computeLib`
mechanism — `reduce_cst` counting a reflexive result as progress — and
the `add_conv` documentation that never stated the contract are
deliberately left alone here.

### DIFF
--- a/src/real/RealField.sml
+++ b/src/real/RealField.sml
@@ -446,22 +446,8 @@ val sgn_tm = “real_sgn :real -> real”;
 val max_tm = “max :real -> real -> real”;
 val min_tm = “min :real -> real -> real”;
 
-(* computeLib re-evaluates whatever a registered conversion hands back, so
-   one that reports success without changing its term sends CBV_CONV round
-   the same term forever.  Several of these are no-ops on an argument that
-   is already reduced: REAL_RAT_DIV_CONV rewrites x / y to x * inv y and
-   back, so it takes &1 / &3 to &1 / &3 and REAL_RAT_REDUCE_CONV does not
-   terminate on any term containing a non-integral rational literal.  The
-   guard therefore belongs on every entry rather than on the one operator
-   (negation) whose no-op was noticed first.
-
-   A conversion can report no progress two ways: return a reflexive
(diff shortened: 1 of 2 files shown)


mn200:
Both new selftests should use `convtest`; only one does at the moment.

lukaszcz:
> Both new selftests should use `convtest`; only one does at the moment.

Fixed

mn200:
Thanks!

PR #2039, opened 2026-08-12
Add have, suff, contr to bossLib and documentation

myreen:
Why add more synonyms? Because these are more descriptive in my opinion. 

I find `have` more natural than `sg`, which is a bit cryptic for newcomers.

Why `contr`? I don't understand how people can keep writing `suppose_not_then strip_assume_tac` every time they want to prove by contradiction. I have always preferred `CCONTR_TAC`, but that seems shouty and old style.

Why `suff`? Well, it's nice and short. 

I have attempted to update the documentation and release notes.


mn200:
There's machinery for handling entry-point aliasing (as per the README in `help/src-sml`), but I'll do this after merging this.

PR #2047, opened 2026-09-02
Run proof-manager tactics outside the proofs slot lock

lukaszcz:
`proofManagerLib.expand*` applies tactics inside a `Context.Data.modify` callback, holding the `proofmanager.proofs` slot lock while the tactic runs.

If the tactic re-enters the proof manager—for example, when a failed `Define` sends its termination goal to `Defn.tgoal`—it attempts to acquire the same non-reentrant lock and deadlocks with interrupts masked.

Run tactics outside the slot lock by reading the current proof state, applying the proof transformation, and writing the resulting state back afterward.

This preserves the intended state behavior:

- Proof-manager operations performed by a failing tactic remain visible.
- Proof-manager operations performed by a successful tactic are replaced by the completed outer proof state.
- Tactics can safely invoke `e`, `set_goal`, and related proof-manager operations.

Regression tests cover:

- Reentrant tactics through `e`, `expandf`, `eall`, and `expand_listf`.
- Proof-state behavior when reentrant tactics succeed or fail.
- A failed `Define` inside a goal-stack tactic leaving its termination goal on the stack.
- A watchdog that reports a deadlock instead of allowing the test suite to hang.

### DIFF
--- a/src/num/termination/TotalDefn.sml
+++ b/src/num/termination/TotalDefn.sml
@@ -813,8 +813,7 @@ fun located_tDefine loc stem q tac =
              (def, NONE)
           end
         else
-        let val (def,ind) =
-              with_flag (proofManagerLib.chatting,false) Defn.tprove0(defn,tac)
+        let val (def,ind) = Defn.tprove0 (defn,tac)
            val def = def |> CONJUNCTS |> map GEN_ALL |> LIST_CONJ
         in
             delete_support defn (thy_consts (current_theory())) snap1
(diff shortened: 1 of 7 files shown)


mn200:
I'll hold off on this; the machinery around tactics and the global context is changing quite a bit on branch `new-tactics`.

mn200:
Yeah, merge conflicts are quite plausibly semantically significant here.

mn200:
Please fix this so that it merges and/or let me know if you think it's no longer required.

lukaszcz:
> Please fix this so that it merges and/or let me know if you think it's no longer required.

`tDefine` issue was fixed upstream, but the issue with the proof manager still remained and this PR fixes it

mn200:
Thanks!

PR #2048, opened 2026-09-02
ThmSetData: apply a theorem-set delta before recording it

lukaszcz:
## Issue

`ThmSetData.export_with_ancestry` registers a stored-attribute handler
(`store_attrfun`) for every theorem set created through it. That is the
path behind `Theorem foo[rule_induction]`, `store_thm` with an
attribute, and `ThmAttribute.store_at_attribute`. The handler did two
things in the wrong order:

```sml
#record_delta fullresult d;                            (* persist *)
#update_global_value fullresult (raw_apply_global d)   (* apply   *)
```

The delta was written into the current theory segment first, and only
then applied to the live global value. If the set type's
`apply_to_global` rejected the theorem, the exception escaped with the
delta already recorded. The bad delta was then exported with the theory,
and every descendant theory re-applied it on load, where it raised again
inside `apply_delta`. There is no user-level way to remove such a delta,
so one malformed attribute in one theory made every theory below it
unloadable.

Any set type whose `apply_to_global` can raise is affected. On `develop`
that includes `rule_induction`: `IndDefLib.ind_thm_to_consts` uses
`dest_imp` and `dest_thy_const` without a handler, so a theorem that is
not of the shape `!vs. premises ==> clauses`, or whose clause hypothesis
is headed by a variable, raises `HOL_ERR`.

Reproduction in a scratch theory:

```sml
val bad = GEN “P:num -> bool”
              (DISCH T (GEN “x:num” (DISCH “(P:num -> bool) x” TRUTH)))
val _ = save_thm ("bad", bad)
val _ = ThmAttribute.store_at_attribute
          {name = "bad", attrname = "rule_induction", args = [], thm = bad}
(* raises HOL_ERR, as it should *)
val _ = length (IndDefLib.thy_rule_inductions (current_theory ()))
(* one more than before: the rejected theorem was recorded anyway *)
```

`IndDefLib.export_rule_induction` had the same order, and so did a
handful of other exporters built directly on `AncestryData.fullmake`.
The exporters produced by `export_simple_dictionary` and friends were
already correct: they publish first, so a failed apply leaves nothing
recorded.

## Fix

Apply first, record only on success. `Context.Data.modify` leaves the
global value untouched when its callback raises, so nothing is
half-applied on the failure path either. The same swap is applied at
every remaining record-before-apply site so the rule is uniform:

- `src/1/ThmSetData.sml`: `store_attrfun` (the reported bug)
- `src/IndDef/IndDefLib.sml`: `export_rule_induction`
- `src/basicProof/BasicProvers.sml`: `export_rewrites`, `delsimps`
- `src/1/DefnBaseCore.sml`: `register_defn`
- `src/combin/combinpp.sml`: `new_form`
- `src/transfer/transferLib.sml`: `add_atomic_term`
- `src/marker/markerLib.sml`: `record_suspension_delta`,
  `record_resumption_delta`

Not addressed: publication and recording are still two steps, so an
asynchronous `Interrupt` between them can leave the live value and the
recorded delta out of step. Closing that window needs an
interrupt-masked critical section and a portable primitive that also
works under Moscow ML. It is independent of this fix and left as a
follow-up.

## Tests

Two rows added to `src/IndDef/selftest.sml`. A theorem that
`rule_induction` rejects is stored through the attribute path and
through `export_rule_induction`; in both cases the call must raise
`HOL_ERR` and the theory's recorded delta count must be unchanged. Each
row was checked against its own ablation: the attribute row fails with
the `ThmSetData` swap reverted, and the export row fails with only the
`IndDefLib` swap reverted.

### DIFF
--- a/src/1/DefnBaseCore.sml
+++ b/src/1/DefnBaseCore.sml
@@ -120,12 +120,12 @@ fun register_defn {tag, thmname} =
       val add = ThmSetData.mk_add thmname
       val p = case add of ThmSetData.ADD p => p | _ => raise Fail "Impossible"
     in
+      #update_global_value userdefs_db (register_defn_p tag p);
       if tag <> "user" then
         HOL_WARNING "DefnBase" "register_defn"
                     "Non-'user' definitions are only stored transiently"
       else
-        #record_delta userdefs_db add;
-      #update_global_value userdefs_db (register_defn_p tag p)
+        #record_delta userdefs_db add
     end
 
 
--- a/src/1/ThmSetData.sml
+++ b/src/1/ThmSetData.sml
@@ -281,8 +281,10 @@ fun export_with_ancestry
           if null args then
             let val d = rADD(lift name)
             in
-              #record_delta fullresult d;
-              #update_global_value fullresult (raw_apply_global d)
+              (* apply first: a delta the set type rejects must not be
+                 recorded, or every descendant theory fails to load *)
+              #update_global_value fullresult (raw_apply_global d);
+              #record_delta fullresult d
             end
           else raise ERR "store_attrfun"
                      ("Arguments not allowed for attribute " ^ attrname)
--- a/src/1/ThmSetData.sml
+++ b/src/1/ThmSetData.sml
@@ -281,8 +281,8 @@ fun export_with_ancestry
           if null args then
             let val d = rADD(lift name)
             in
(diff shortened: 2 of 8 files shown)


mn200:
Thanks; I suspect some of the delta-application functions will always succeed, but it's definitely a good idea to establish an idiom and use it consistently.

PR #2051, opened 2026-09-04
HolRefute: Quickcheck, narrowing and a Nitpick-style Kodkod model finder

lukaszcz:
## Summary

This PR adds `src/HolRefute`, a counterexample generator for HOL4 goals, built on Poly/ML only. It provides four diagnostic tactics (`REFUTE_TAC`, `QUICKCHECK_TAC`, `NARROWING_TAC`, `MODEL_REFUTE_TAC`), the SML entry points behind them, a user manual chapter (`Manual/Description/Refute.smd`), 33 help docfiles, a selftest with level-2 acceptance tables, theory-hygiene tests, and an executable example corpus adapted from Isabelle's Nitpick manual and example suites.

The library has four backends: exhaustive and random Quickcheck, native narrowing, and a Kodkod-based finite model finder. Substrates are untrusted accelerators; every counterexample is replayed and, where replay succeeds, turned into a HOL theorem. HolRefute leaves no type, constant, theorem or binding in the user's theory on any path (success, failure, timeout, interrupt); its only ambient effect is teaching the global `EVAL` compset closed `:rat` arithmetic and `:real` `inv`.

### Model finder: a port of Nitpick

The model finder is a module-for-module port of Nitpick (Isabelle2025-2, `src/HOL/Tools/Nitpick`). The pipeline, the intermediate languages and the trust model are Nitpick's; the adaptation is concentrated in the HOL-facing layer.

| Isabelle source | HolRefute module | Fidelity |
|---|---|---|
| `nitpick_preproc.ML` | `Refute_ModelFinder_Preproc` | Port: binarization, constructor destruction, specialisation, boxing, skolemisation. Traversal orders that fix variable numbering are preserved deliberately (rightmost-argument descent, `conjuncts_of` order, axiom-side skolemisation depth). |
| `nitpick_mono.ML` | `Refute_ModelFinder_Mono` | Port of the monotonicity calculus: mtypes, constraints, frames, SAT solving. `Id` and set products stay unfolded (no nut builtin), the Pure meta-connective cases, `bounteous_consts` and tracing are not ported, and `is_harmless_axiom` is narrowed. |
| `nitpick_scope.ML` | `Refute_ModelFinder_Scope` | Port of scope enumeration. |
| `nitpick_rep.ML` | `Refute_ModelFinder_Rep` | Port; relations are curried (HOL4 `relationTheory`) rather than sets of pairs, and the application boundaries are kept. |
| `nitpick_nut.ML` | `Refute_ModelFinder_Nut` | Port. HOL4 sets are functions and binders are named, so the environment records de Bruijn levels to keep Nitpick's `BoundName` convention. The `nut` datatype and its operator enumerations are unchanged; `cst` gains the word and char operations. One `NatToInt` totality bound is deliberately corrected against upstream and marked in the code. |
| `nitpick_kodkod.ML` | `Refute_ModelFinder_Kodkod` | Port of the FORL translation. Differences are confined to HOL4 semantics (`Num` of a negative integer, curried relations). |
| `nitpick_peephole.ML` | `Refute_ModelFinder_Peephole` | Port with three marked divergences (`s_all`/`s_exist` on an empty declaration, `rel_expr_intersects` on `AtomSeq`, `s_join`'s `Univ` clauses) plus the HOL4-side bit-width choice. |
| `nitpick_model.ML` | `Refute_ModelFinder_Model` | Port of model reconstruction and display, plus the HOL4-only certificate hook. |
| `nitpick_hol.ML` | `Refute_ModelFinder_HOL` | Port of the definition-harvesting and unfolding layer; the largest adaptation. DefnBase/TypeBase presentations replace Isabelle's code equations and `simps`; typedef, quotient and codatatype registries and the ersatz table are re-derived from HOL4 theory data. |
| `nitpick.ML` (`pick_them_nits_in_term`) | `Refute_ModelFinder` | Port of the control flow. Type-variable merging is the sortless specialisation of `merged_type_var_table_for_terms`, since HOL4 has no sorts. |
| `kodkod.ML` | `Refute_Forl` | Port of the Kodkodi process bridge; drives the same `kodkodi-1.5.7` component, without Isabelle's launcher. |
| `kodkod_sat.ML` | `Refute_ForlSat` | Port of SAT-solver selection, honouring the same `*_HOME` variables. |
| `prop_logic.ML`, cdclite part of `sat_solver.ML` | `Refute_PropSat` | Adaptation of Tjark Weber's BSD-licensed code, used by the monotonicity calculus. |
| `Nitpick.thy` | `refuteScript.sml` | Port of the ersatz definitions (`unknown`, `wf'`, `wfrec'`, `card'`, `sum'`, `safe_The`), with fmap rows added. |
| `nitpick_commands.ML` | no counterpart | Options are a `config` record with `upd_*` updaters and `REFUTE_TAC_WITH`, not Isar syntax. |

The verdict model is Nitpick's: `Genuine` / `QuasiGenuine` / `Potential` is decided by the encoding, `max_potential` is charged as in Nitpick (only a genuine model spends a `max_genuine` slot, where Nitpick's sound branch also charges quasi-genuine ones), and Refute trusts Kodkodi exactly as Nitpick trusts Kodkod (a `Genuine` verdict with no certificate is valid). Deliberate deviations, all documented in the README and manual:

- Hilbert choice is guarded per occurrence as `if ?x. P x then @P else unknown` where Isabelle leaves the occurrence unguarded and constrains it only through the witness-conditional `Eps_psimp`; a guard vetoes `NoCounterexample`.
- `NoCounterexample` is a total claim. Anything bounds-relative reports `Unknown` with a reason, and a value-position `unknown` in any harvested axiom vetoes it.
- Machine words of concrete width and `:char` are exact native carriers (`2^w` and 256 atoms), which turn smart binarization off. Nitpick has no counterpart.
- `real` is rational-valued, as in Nitpick. Polymorphic goals are tested at configured monomorphic instances and never earn `NoCounterexample`, where Nitpick varies the type variables' cardinalities like any other type.
- `sat_solver = "smart"` prefers a configured JNI or external solver; Nitpick's smart order never picks a JNI solver by itself.
- HOL4 completes non-exhaustive patterns with `ARB` in the raw definition; only the user's equations are harvested, so the same spurious-counterexample risk as Nitpick applies.

### Quickcheck: design ported, execution re-engineered

- **Exhaustive generators.** Plan compilation is ported from `exhaustive_generators.ML` (the `Refute_QC` comment cites the upstream lines). Smart generators follow Isabelle's predicate-compiler design: `Refute_SmartGen` mode-checks Horn SCCs for positive and negated first-order modes and flattens function equations into graph clauses; `upd_allow_function_inversion` defaults off like Isabelle's flag, and `upd_use_subtype` matches `use_subtype`.
- **Random generators.** Function values are drawn on demand as `random_fun_lift` does, but eagerly rather than lazily.
- **Execution differs.** Isabelle compiles generators through the code generator to ML. HolRefute compiles a goal to a test-plan IR (`Refute_Eval.plan`) that two substrates run: in-process SML extraction (`Refute_Extract`) and `computeLib`. Both consume one PRNG in the same order, so a seed reproduces the candidate stream on either.
- **`abstract_generators.ML`** becomes `abstract_generator`; **`find_unused_assms.ML`** becomes `Refute_Unused` (`check_/find_/print_unused_assms`), with the same maximal-droppable-set semantics.
- The driver in `quickcheck_common.ML` is not ported: budgets, iterative size deepening, the backend pool and certainty ceilings are `Refute_Core`'s own.

### Narrowing: generators ported, engine native

The type representation (`Narrowing_sum_of_products`), `finitize_functions` and the quantifier-pulling pass are ported from `narrowing_generators.ML`. The Haskell engines (`Narrowing_Engine.hs`, `PNF_Narrowing_Engine.hs`) are replaced by a native SML narrowing engine (`Refute_Narrow`, `Refute_QC_Narrow`), so no Haskell toolchain is needed. Finite function inputs are `ffun` update chains defined in `refuteScript.sml` instead of Isabelle's type-scoped `Constant` names.

### HOL4-only additions with no Isabelle counterpart

- **Certification** (`Refute_Cert`, `Refute_Cert_Narrow`, `Refute_Cert_Model`): QC hits are replayed with `computeLib`, narrowing replays its case tree, and model-finder values are replayed through Skolem provenance, bounded synthesis, one-layer cases, single-property induction, Presburger and `REAL_ARITH`. Replay is fail-closed and never weakens the encoding's certainty.
- **Concurrent backend pool** with one absolute deadline per call (`ParList`, below).
- **Theory hygiene**: all runtime definitions are bracketed in a theory snapshot/revert; `theory_tests/` checks that a descendant theory inherits nothing.
- Custom generator families (`register_generator_family`, fmap), `Refute_EvalFmap` for ground finite-map equality, and the rat/real compsets.

### `src/portableML/poly/concurrent/ParList.{sig,sml}`, `selftest.sml`, `Holmakefile` (new)

A small racing/mapping combinator library over Poly/ML threads: `map_with_workers`, `get_some_with_workers`, `get_some`, `get_first`, and `uninterruptible_wait`.

- **Why it is needed.** `Refute_Core` runs backend admission and the backend race through it (`map_with_workers` for admission, `get_some_with_workers` for the concurrent pool, `get_first` for `upd_sequential`). `Refute_QC` uses `uninterruptible_wait` so the theory-revert cleanup after an interrupted run still completes.
- **Why it lives here.** `Multithreading`, `Future`, `Timeout` and now `ParList` are built by the kernel band with `--poly_not_hol` and linked into sigobj. Naming this directory in HolRefute's `INCLUDES` would rebuild it with the overlay in scope and create an `Sref -> Overlay` dependency that breaks the next kernel bootstrap in `src/bool` (documented in HolRefute's Holmakefile).
- **What it fixes that nothing in the tree offered.** `Thread_Attributes.uninterruptible` clears the broadcast flag, and HOL's REPL delivers Ctrl-C as `Thread.broadcastInterrupt`, which Poly/ML drops rather than defers for a thread not accepting broadcasts. A masked join therefore swallowed Ctrl-C. `ParList` masks with `InterruptDefer` instead and keeps joins observant. Workers are interrupted, never `Thread.kill`ed, since a killed worker can leak the process-global theory mutex. Directed-interrupt tests cannot see any of this, so `selftest.sml` drives the masked windows through `ParList_Test` hooks. The Holmakefile change only wires that selftest in under `HOLSELFTESTLEVEL`.

### `src/IndDef`: `KeyedThmSet` and the `[coinduction]` attribute

- **`KeyedThmSet.{sig,sml}` (new).** The `ThmSetData`-backed "theorems keyed by the constants their clauses talk about" machinery that `IndDefLib` implemented inline for `rule_induction` is extracted into a factory parameterised by which clause part carries the key. `IndDefLib` now instantiates it with `Hypothesis`; the exported names, set type and map type are unchanged, so nothing downstream changes. What does change: malformed theorems get a diagnostic naming the set type and the part inspected, and non-implicational theorems are diagnosed instead of escaping as a bare `dest_imp` failure.
- **`CoIndDefLib`.** Instantiates the same factory with `Conclusion`, exporting `coinduction_map`, `thy_coinductions`, `add_coinduction`, `export_coinduction`, and registers the `coinduction` theorem attribute. `Hol_coreln` now exports its `_coind` theorem automatically, as `Hol_reln` already did for `_strongind`.
- **Why HolRefute needs it.** Nitpick classifies each inductive constant as an lfp or gfp to decide how to unfold and iterate it. `Refute_ModelFinder_HOL` answers that from `IndDefLib.rule_induction_map` and `CoIndDefLib.coinduction_map` (`is_registered_gfp`, `registered_stem`). Without a persistent coinduction registry, coinductive relations could only be recognised in the session that defined them.
- **`selftest.sml`, `Holmakefile`.** Tests for both key parts' diagnostics, the non-implicational case, the stored-attribute path, and a persisted well-formed export. The Holmakefile adds `KeyedThmSet.uo` to the selftest link line.
- **`Manual/Description/modern-syntax.smd`.** Documents `coinduction` in the theorem-attribute table.

### `src/coalgebras/pathScript.sml`, `selftest.sml`, `Holmakefile`

`pathTheory` had constructors, `path_cases`, injectivity and distinctness theorems but no case constant and no `TypeBase` entry. This adds `path_case` with its `compute`/`simp` equations, `case_cong`, `case_eq`, `case_elim`, a `case` overload so `case p of stopped_at x => ... | pcons x r q => ...` parses, and a `TypeBase` registration using `path_bisimulation` as the induction principle.

- **Why HolRefute needs it.** HolRefute's built-in codatatype registry (`llist`, `ltree`, `itree`, `itreeTau`, `lbtree`, `path`) validates each entry through its case constant and `TypeBase.constructors_of`; `path` could not be registered without both. The selftest's model-finder table exercises `path` injectivity, distinctness and bisimulation.
- **Side effect for other users.** `path` now behaves like the other coalgebraic types under `case` syntax, `EVAL` and the simplifier. The coalgebras selftest gains simp and `case`-syntax checks and a `TypeBase` registration check. `path_11` stays local since its two conjuncts are already exported.

### `src/parallel_builds/core/Holmakefile`

Adds `../../HolRefute` to `INCLUDES` under `POLY`, for every kernel. This is what pulls HolRefute into `bin/build`; no sequence file changes. HolRefute builds under `--otknl` as well as the standard kernel.

### `AGENTS.md` (new symlink to `CLAUDE.md`)

Lets agent tooling that reads `AGENTS.md` pick up the existing project notes. No content.

### `Manual/Description`, `help/Docfiles`

A new `Refute` chapter (wired into `chapters.txt`, `libraries.smd`, and the generated-file `.gitignore`), and 33 `Refute.*` docfiles for the tactics, entry points, registries, `config`, and the unused-assumption probes.

## Testing

- `HOLSELFTESTLEVEL=2 Holmake` in `src/HolRefute` is the quality gate: the selftest (level 2 adds cross-substrate conformance, the narrowing table, the corpus and the model-finder acceptance tables), then `theory_tests/`. Model-finder rows need `HOL4_KODKODI` pointing at an unpacked `kodkodi-1.5.7` and a Java runtime; without it they report inconclusive and the theory scripts still build.
- `Holmake examples` in `src/HolRefute` builds the example corpus.
- `src/IndDef`, `src/coalgebras` and `src/portableML/poly/concurrent` selftests cover the changes described above.


help/Docfiles/Refute.find_unused_assms.smd
```diff
@@ -0,0 +1,49 @@
...
+
+### See also
+
+[`Refute.check_unused_assms`](#Refute.check_unused_assms),
+[`Refute.print_unused_assms`](#Refute.print_unused_assms),
+[`Refute.config`](#Refute.config),
+[`Refute.try_refute`](#Refute.try_refute),
+[`DB.thms`](#DB.thms)
+
+The complete library guide is \ref{Description:the-holrefute-library}.
```
mn200, line 49: Please don't have text after the See-also list of idents

help/Docfiles/Refute.abstract_generator.smd
mn200, line 1: Doesn't conform to template (in `help/template`), which calls for an hrule after the type block and a one-line summary.

help/Docfiles/Refute.check_unused_assms.smd
mn200, line 1: Doesn't conform (as above)

help/Docfiles/Refute.config.smd
```diff
@@ -0,0 +1,326 @@
...
+`default_mf_config`.  `the_config` is the mutable interactive default
+read by `quickcheck`, `model_refute`, and the tactics when they are
+applied; `show_config ()` prints its complete value.  `apply_updates
+updates cfg` applies a list of updaters to `cfg` from left to right, and
+`current_config updates` does the same to `!the_config`;
+`REFUTE_TAC_WITH`, `refute_with`, and `refute_goal_with` are built from
+them.
+
+``` hol4
+val cfg = Refute.default_config
```
mn200, line 32: This should use scripting (with leading `>>`) so that the code actually runs and produces appropriate output in the generated file.

help/Docfiles/Refute.config.smd
```diff
@@ -0,0 +1,326 @@
...
+[`Refute.REFUTE_CONFIG_TAC`](#Refute.REFUTE_CONFIG_TAC),
+[`Refute.QUICKCHECK_TAC`](#Refute.QUICKCHECK_TAC),
+[`Refute.NARROWING_TAC`](#Refute.NARROWING_TAC),
+[`Refute.MODEL_REFUTE_TAC`](#Refute.MODEL_REFUTE_TAC),
+[`Refute.quickcheck`](#Refute.quickcheck),
+[`Refute.model_refute`](#Refute.model_refute),
+[`Refute.try_refute`](#Refute.try_refute),
+[`Refute.find_unused_assms`](#Refute.find_unused_assms)
+
+The complete library guide is \ref{Description:the-holrefute-library}.
```
mn200, line 326: Don't put material after the see-also section

help/Docfiles/Refute.config.smd
mn200, line 1: Suspect this entry is so long that it should just be in the DESCRIPTION manual (or (perhaps?) in a separate manual).

help/Docfiles/Refute.export_refute_psimp.smd
mn200, line 1: Doesn't conform

help/Docfiles/Refute.export_refute_simp.smd
mn200, line 1: Doesn't conform; are these small entry-points really user-facing?  If not, they probably don't need to be in the REFERENCE.

help/Docfiles/Refute.export_refute_unfold.smd
mn200, line 1: Doesn't conform

Manual/Description/Refute.smd
```diff
@@ -0,0 +1,1245 @@
...
+The library is available on Poly/ML builds.  In a theory script, name
+it in the `Libs` declaration:
+
+```sml
+Theory myTheory
+Libs
+  Refute
+```
+
+In an interactive session:
```
mn200, line 30: Don't document use of `load`

Manual/Description/Refute.smd
```diff
@@ -0,0 +1,1245 @@
...
+```
+
+The transcripts in this chapter are static.  Timings are
+machine-dependent, and reports are re-wrapped for the page from the one
+`<<HOL message: ...>>` block a session prints.  They select the
+portable `Compute` substrate and sequential execution, so that each
+report names a predictable backend.
+
+```{=latex}
+\begin{session}
```
mn200, line 80: This sort of material should (unless it's using `kodkod`) use the "scripting" interface to generate material dynamically.

Manual/Description/Refute.smd
```diff
@@ -0,0 +1,1245 @@
...
+universe index, which binary integers would replace, so a goal
+mentioning a word or a character disables binary integers; forcing
+them with `upd_binary_ints (SOME true)` refuses the word or character
+operation instead.  A typedef over `:num` or `:int` disables them for a
+different reason: binarized, its bound is a numeral needing a carrier
+atom of its own, and the scopes where the abstract type is large enough
+are exactly those where that numeral is missing.
+
+#### Types the executable backends can test
+
```
mn200, line 488: Does this mean that loading `Refute` adds a dependency on the theories of rationals and reals?

Manual/Description/Refute.smd
```diff
@@ -0,0 +1,1245 @@
...
+The theorem sets MF preprocessing uses are extended in a theory script
+with `export_refute_simp`, `export_refute_psimp`, and
+`export_refute_unfold`, each taking the name of an existing theorem;
+the sets persist with the theory.  Every conjunct must have a constant
+head in its conclusion -- `c t1 ... tn = u`, `c t1 ... tn`, or
+`~(c t1 ... tn)` -- because the clause is filed and looked up under
+that constant; when the model finder builds its tables, a clause with
+no constant head is dropped with a warning.
+
+Structural registrations require distinct type variables as parameters
```
mn200, line 1039: Would it be better to have this meta-data generated and stored by default.

src/HolRefute/Refute_EvalCompute.sml
mn200, line 1: This file should have a signature

src/HolRefute/Refute_Core.sml
mn200, line 1: This file should have a signature; and using `ref` for global values is extremely suspect.

src/HolRefute/Refute_Cert_Narrow.sml
mn200, line 1: This file should have a signature

mn200 (changes requested):
Documentation has multiple instances of similar problems:
- scripting should be used for the DESCRIPTION manual
- doc-files must conform to template
- some doc-files seem as if they can be omitted

Source-code:
- structures should almost always have signatures

I'd like to see the IndDefLib changes pulled out into a separate PR.

PR #2064, opened 2026-09-16
Misc fixes

ordinarymath:


src/metis/normalForms.sml
```diff
@@ -1369,18 +1369,19 @@ val REMOVE_ABBR_TAC =
 (*                                                                           *)
 (* Example:  f (if x then y else z)  =  (if x then f y else f z)             *)
 (* ------------------------------------------------------------------------- *)
-
+local
+val cond_rator_conv = REWR_CONV boolTheory.COND_RAND
```
mn200, line 1373: Given that there's a binding later in the same file that gives `cond_rator_conv` a definition involving `COND_RATOR`, I'm a bit suspicious about this.

src/metis/normalForms.sml
mn200, line 1: It would be good to have test-cases demonstrating how you've fixed a hitherto buggy behaviour.  (I am wondering if this code was ever exercised.)

mn200 (changes requested):

src/metis/normalForms.sml
ordinarymath: Oop. A more apt name would be cond_rand_conv

PR #2065, opened 2026-09-18
Make list compsets evaluate membership structurally

### LINKED ISSUE #2055: list_compset leaves closed ALL_DISTINCT singleton partially evaluated
On current `develop` (`a390cbabd3a4521bab4ee20281e3e42933a8a3ae`), both `listSimps.list_compset` and `wordsLib.words_compset` leave a closed singleton `ALL_DISTINCT` application partially evaluated, while `EVAL` closes it.

Minimal script:

```sml
Theory listCompsetExperiment
Ancestors
  list
Libs
  computeLib listSimps wordsLib

val input = ``ALL_DISTINCT [NONE : num option]``;

val list_result = computeLib.CBV_CONV listSimps.list_compset input;
val words_result = computeLib.CBV_CONV wordsLib.words_compset input;
val eval_result = EVAL input;

val _ = print ("list_compset: " ^ thm_to_string list_result ^ "\n");
val _ = print ("words_compset: " ^ thm_to_string words_result ^ "\n");
val _ = print ("EVAL: " ^ thm_to_string eval_result ^ "\n");
```

Output:

```text
list_compset: ⊢ ALL_DISTINCT [NONE] ⇔ NONE ∉ ∅
words_compset: ⊢ ALL_DISTINCT [NONE] ⇔ NONE ∉ ∅
EVAL: ⊢ ALL_DISTINCT [NONE] ⇔ T
```

The input is fully closed, so I expected the list-specific compset to reduce it to `T`.

This appears related to #182: list `MEM` is represented through `IN`/`LIST_TO_SET`, and the `ALL_DISTINCT` rule in `listSimps.list_rws` reaches predicate-set membership without `list_compset` containing enough support to finish the closed expression. Unlike the performance report in #182, this is a small closed term on which evaluation remains incomplete.

For context, the following specialization works around it with the current functional/sealed compset API:

```sml
val workaround_compset =
  listSimps.list_compset
  |> computeLib.copy
  |> pred_setLib.add_pred_set_compset
  |> (fn cs => computeLib.scrub_const cs ``bool$IN``)
  |> computeLib.add_thms [pred_setTheory.NOT_IN_EMPTY]
  |> (fn cs => computeLib.scrub_const cs ``ALL_DISTINCT``)
  |> computeLib.add_thms [listTheory.ALL_DISTINCT]
  |> computeLib.seal;

computeLib.CBV_CONV workaround_compset input;
(* ⊢ ALL_DISTINCT [NONE] ⇔ T *)
```

Is `list_compset` intended to close computations of this form? If so, it may need either structural `MEM` handling that takes precedence before `LIST_TO_SET` evaluation, or sufficient predicate-set support to finish the resulting membership test.

xrchz:
## Summary

- restore `listTheory.MEM` to `listSimps.list_rws` and remove the default `LIST_TO_SET_THM` rule
- keep explicit `LIST_TO_SET` computation available by opting into its theorem
- stop Patricia from persistently registering `LIST_TO_SET_THM` in the global compute theorem set
- add regressions for `ALL_DISTINCT`, closed `nub` computations, `words_compset`, explicit `LIST_TO_SET` evaluation, and the Patricia load-order case

This keeps list membership on its structural recursion equations instead of redirecting it through predicate-set membership. It fixes incomplete evaluation of closed terms and avoids the severe scaling problem for `nub` computations.

Fixes #2055.

## Testing

CI will provide validation.


mn200:
As per discussion on Zulip: I feel as if the better approach would be to continue to allow `LIST_TO_SET` to be used, but to remove `IN_CONV` and just use `IN_INSERT`, which embodies exactly the same logic as the `MEM` theorem (generates a disjunction of the equality check and the recursive "call").

xrchz:
Thanks — I prototyped this alternative by retaining `LIST_TO_SET_THM`, replacing finite-set use of `IN_CONV` with ordinary `IN_INSERT` and `NOT_IN_EMPTY` compute rules, and preserving `SET_SPEC_CONV` for set comprehensions.

I benchmarked it against structural `MEM` using the closed `(num option # string)` `nub` reproducer from #2055:

| Elements | Structural `MEM` | `LIST_TO_SET + IN_INSERT` |
|---:|---:|---:|
| 40 | 0.016s | 2.122s |
| 80 | 0.078s | 9.611s |

A preceding run gave essentially the same result:

| Elements | Structural `MEM` | `LIST_TO_SET + IN_INSERT` |
|---:|---:|---:|
| 40 | 0.019s | 2.287s |
| 80 | 0.082s | 9.609s |

Both routes fully close the results. Replacing `IN_CONV` is a substantial improvement over the previously observed 82–157 seconds at 80 elements, so it may be worthwhile independently. However, for this list computation it remains roughly 120× slower than structural membership and retains approximately quadratic scaling.

On that evidence, I think list compsets should still use structural `MEM`, as in this PR, rather than expanding list membership through `LIST_TO_SET`. Improving predicate-set membership separately would still be useful, but I would prefer the `IN_INSERT` change to be developed and reviewed in a separate PR.

mn200:
Interesting! I speculate that CBV is biting us here: we expand all of the `LIST_TO_SET` before doing any useful comparisons with `IN_INSERT`, generating a redundant copy of the list encoded as a set before the real work begins.  Anyway, those timings are pretty decisive.  So I'm happy to merge as is.

